PR #49, opened 2018-02-16
Connectable

### LINKED ISSUE #42: Types: Connection vs Connectable
The API and architecture conflate the concepts of a Connection and a Connectable. These are distinct, and the allowable operations on each are different. They should be represented by two different types in the API and architecture. This is a problem with BSD Sockets, and something we should avoid in the transport services architecture.

A `Connection` is something on which you can send data; something that has active protocol state. It supports operations such as `send()` and `close()`, and generates events such as `Received<>`. 

A `Connectable` is something that can be used to generate a `Connection`. It doesn't have an active transport protocol. It supports operations such as `setParameters()` or `connect()` (the latter being a method that consumes a `Connectable` and returns a `Connection` object). 

These are logically different types, and should be represented as such. It doesn't make sense to call `send()` on a `Connectable` or `setParemeters()` on a `Connection`, for example. 

In pre-establishment, we work with `Connectable` objects. Establishment then becomes the process of generating a `Connection` from a `Connectable`. Once we have a `Connection`, we're in the data transfer phase. 

Discuss.

csperkins:
Incorporate Preconnection into the architecture and API draft.

Closes #42


draft-pauly-taps-arch.md
```diff
@@ -205,103 +205,93 @@ Beyond the basic objects, there are several high-level groups of actions that an
...
+* Preconnection: A Preconnection object is a representation of a potential connection. It has state that describes parameters of a Connection that might exist in the future: the Local Endpoint from which that Connection will be established, the Remote Endpoint to which it will connect, and Path Selection Properties, Protocol Selection Properties, and Specific Protocol Properties that influence the choice of transport that Connection will use. A Preconnection can be fully specified and represent a single possible Connection, or it can be partially specified such that it represents a family of possible Connections.
```
tfpauly: May be good to note that the Local Endpoint is an optional part of the pre-connection properties

draft-pauly-taps-arch.md
```diff
@@ -205,103 +205,93 @@ Beyond the basic objects, there are several high-level groups of actions that an
...
+ +-- Specific Protocol     :      |                        :
+ |   Properties            :      |                        :
+ |                         :      |                        :
+ |   +----------+          :      |                        :
+ +-->| Listener |-----------------+                        :
+     +----------+ Connection Received                      :
+           ^               :                               :
+           |               :                               :
+        Listen()           :                               :
+~~~~~~~~~~
```
tfpauly: I like the diagram! Thanks for adding this.

draft-pauly-taps-arch.md
```diff
@@ -205,103 +205,93 @@ Beyond the basic objects, there are several high-level groups of actions that an
...
+* Preconnection: A Preconnection object is a representation of a potential connection. It has state that describes parameters of a Connection that might exist in the future: the Local Endpoint from which that Connection will be established, the Remote Endpoint to which it will connect, and Path Selection Properties, Protocol Selection Properties, and Specific Protocol Properties that influence the choice of transport that Connection will use. A Preconnection can be fully specified and represent a single possible Connection, or it can be partially specified such that it represents a family of possible Connections. 
 
-Listener
+* Connection: A Connection object represents an active transport protocol instance that can send and/or receive Content between a Local Endpoint and a Remote Endpoint. It holds state pertaining to the underlying transport protocol instance and any ongoing data transfer.
```
tfpauly: I'd like to not lose the original intent of the text before, that captured the idea that this "Connection" still works for stateless protocols like UDP, just by maintaining the five-tuple path. A note to say that the readiness of the connection to send and receive may or may not be based on protocol handshakes would be good.

draft-pauly-taps-arch.md
```diff
@@ -205,103 +205,93 @@ Beyond the basic objects, there are several high-level groups of actions that an
...
-
-Protocol Selection Properties
-
->> Protocol Selection Properties consist of the options that an application may set to influence the selection of transport protocol, configure the behavior of generic transport protocol features. These options come in the form of requirements, prohibitions, and preferences. Examples include reliability, service class, multipath support, and fast open support. 
+* Endpoint: An Endpoint represents one side of a transport connection. It is a concept that represents how an application views itself or a peer that may vary in levels of specificity. Examples include "IP address + port", "hostname + port", and "DNS service name".
 
-Specific Protocol Properties
+* Remote Endpoint: The Remote Endpoint in a properties \[MICHAEL: I can't parse "in a properties" here. Should this be removed?] represents the application's view of a peer that can participate in a transport connection.
```
tfpauly: Yeah, we should remove "in a properties"

draft-pauly-taps-arch.md
mwelzl: +1; but I find the "Content Received" arrow pointing at Content, which has a Send() arrow going back to Connection confusing.

draft-pauly-taps-arch.md
mirjak: Also i guess e.g. DNS resolution in done by the connection (not the preconception), right? In this case there a change that the connection fails even before an active transport protocol instance has been established. And the other question is, can a connection have more than one active protocol instance? I'd say yes, as you could emulate multipath by using e.g. two TCP connections.

draft-pauly-taps-arch.md
mwelzl: This is from long ago, when I once read the whole thing and inserted such comments throughout; you may want to do a cleanup, there are several such minor comments from me in the text.

draft-pauly-taps-arch.md
```diff
@@ -205,103 +205,93 @@ Beyond the basic objects, there are several high-level groups of actions that an
...
-
->> Path Selection Properties consist of the options that an application may set to influence the selection of path between itself and the Remote Endpoint. These options can come in the form of requirements, prohibitions, or preferences. Examples of options which may influence path selection include the interface type (such as a Wi-Fi Ethernet connection, or a Cellular LTE connection), characteristics of the path that are locally known like Maximum Transmission Unit (MTU), or expected throughput or latency.
-
-Protocol Selection Properties
-
->> Protocol Selection Properties consist of the options that an application may set to influence the selection of transport protocol, configure the behavior of generic transport protocol features. These options come in the form of requirements, prohibitions, and preferences. Examples include reliability, service class, multipath support, and fast open support. 
+* Endpoint: An Endpoint represents one side of a transport connection. It is a concept that represents how an application views itself or a peer that may vary in levels of specificity. Examples include "IP address + port", "hostname + port", and "DNS service name".
```
mirjak: Did we already decide if we want to keep this supperclass "Endpoint" or not?

draft-pauly-taps-arch.md
mirjak: Would ICE create new endpoints or add additional information to an existing endpoint? In both cases I think this definition is to limited by only relating a remote to the application's view...

draft-trammell-taps-interface.md
```diff
@@ -426,19 +429,21 @@ The following properties apply to Connections and Connection Groups:
...
-system optimize. However, a TAPS system will only actually set those protocol
-properties that are actually supported by the chosen transport protocol. These
-property all apply to Connections and Connection groups.
+Properties, depending on the available transport features. 
+As with Transport Preferences ({{transport-prefs}}), Protocol Properties are
+specified on the Preconnection object, and are using during initiation of a
+Connection to help the system choose an appropriate transport.
+The system will only actually set those protocol properties that are actually 
+supported by the chosen transport protocol. 
+These property all apply to Connections and Connection groups.
```
mwelzl: Should be "These properties all apply..."

draft-pauly-taps-arch.md
mirjak: Replying to myself.... I guess DNS resolution is done during Initiate() and if that fails no connection is established, right? I still wondering if there are cases where you have a connection but no active protocol instance (yet or anymore)...

draft-pauly-taps-arch.md
```diff
@@ -205,103 +205,93 @@ Beyond the basic objects, there are several high-level groups of actions that an
...
+* Path Properties Changed: A Path Properties Changed event notifies the application that some property of the Connection has changed that may influence how and where data is sent and/or received.
 
->> Close is the action an application may take on a Connection to indicate that it no longer intends to send data, is no longer willing to receive data, and that the protocol should signal this state to the remote endpoint if applicable.
+### Termination Actions {#termination}
 
-Abort
+* Close() is the action an application may take on a Connection to indicate that it no longer intends to send data, is no longer willing to receive data, and that the protocol should signal this state to the remote endpoint if applicable.
 
->> Abort is an action the application may take on a Connection to indicate a Close, but with the additional indication that the transport system should not attempt to deliver any outstanding data.
+* Abort() is an action the application may take on a Connection to indicate a Close, but with the additional indication that the transport system should not attempt to deliver any outstanding data.
```
mirjak: Should there also be an action to indicate that the application will not send any further data but is still willing to receive data?

mwelzl: No; this is explained in minset - exactly why I hesitate to move stuff to the appendix, people don't seem to read the appendix   :)    if you want to support a broad set of transport connections, you need these primitives to only expose the most limiting semantics of all of them. What you describe is a TCP specialty, it's the result of half-closed connections. If you expose this, and someone wants to use it, you limit the system to only support protocols that can do half-closed connections. Worse, you need to initially ask "will you want to use the primitive that still lets you receive after closing?" in order to choose the right set of protocols (or simply never use a protocol that can't do half-closed connections, which are by themselves not very useful anyway).

draft-pauly-taps-arch.md
mwelzl: About the "you could emulate multipath by using e.g. two TCP connections" bit: this is exactly the email conversation I just have / had with Tommy - I think that should be possible, but it would need us to add something to the API such that the application can say "my peer can handle this".

draft-pauly-taps-arch.md
csperkins: Yup - will fix

draft-pauly-taps-arch.md
csperkins: @tfpauly good point, will fix
@mirjak DNS resolution is done in Initiate(), and you'll get an error event if it doesn't resolve to something you can connect to. I haven't thought whether a Connection can support multiple underlying protocol instances, but the case where there is no longer an active protocol instance seems needed to represent a closed connection.

draft-pauly-taps-arch.md
csperkins: Not sure, but splitting Endpoint is coming in a separate pull request later (or, more likely, tomorrow)

draft-pauly-taps-arch.md
csperkins: For ICE I think we need: 1) `fn expand_candidates(p : Preconnection) -> List<Preconnection>` which can take a partially specified preconnection and expand it to a list of all possible candidates that match, so that list can be passed to the peer by the application; and 2) `fn initiate(peer_candidates : List<Preconnection>) ` which allows you to specify the remote candidates. I'll write this up, but it won't be today.

draft-pauly-taps-arch.md
csperkins: @mwelzl I see what you mean... will think whether I can improve

draft-pauly-taps-arch.md
britram: re resolution: I think whether this happens before or after Initiate() should be left up to an implementation. I could certainly see implementations that take action during Preconnect() in order to reduce Initiate() latency, and warming up the local name cache is one of those possible actions...

draft-pauly-taps-arch.md
britram: I agree with @mwelzl that this action is of limited utility for non-TCP stacks, and we could easily leave this out for London, but in general I don't think it's a problem to add actions which express application intent but which some/many stacks will treat as a no-op. I see `EndOfData()` (it's hard to name this, cf. socket's terrible `shutdown()`) as precisely this kind of thing -- it expresses an application-layer semantic, but some/many/most protocols might ignore it, and that's also okay as long as it doesn't cause incorrect operation of those protocols. (UDPLite's "checksum coverage" is a perfect example of this -- it's useless in TCP, and useless in UDP unless it's 0, but it doesn't express "don't checksum this", it expresses "the application is tolerant of corruption of this data". Same same.)

mwelzl: No sorry, this is different: if you write an application that sends data where it may be fine to not use a checksum, but you get a checksum, nothing bad happens. If you write an application where you issue close but you do expect to be able to read afterwards, but you can't, this is semantically different. Application 1 must ensure that communication is really over and it doesn't expect anything anymore from app 2 when the it closes - otherwise, the sent data from app2 is lost. So, at the end of the connection, you're in fact unreliable.

This isn't a big problem in practice, and that's why TCP's half-closed connections are somewhat obscure - but if you rely on them being there, your application can break. If you don't get "checksum disabled", nothing will ever break.

draft-pauly-taps-arch.md
csperkins: @britram this may be an optimisation hint we allow, but I'd like when resolution occurs to be deterministic by default

draft-pauly-taps-arch.md
britram: yep, agreed, that's why it's a different action: `EndOfData()`, not `Close()`. Also agree that `Close()` should mean the connection should be torn down and no further reads or writes are expected (with the caveat, of course, that in an asynchronous API the API cannot guarantee that no `Received<>` event will fire after `Close()` is called).

britram: (and I should point out that you still have to `Close()` after `EndOfData()`... it's all a bit obscure though, so I'd be happy to drop it for London)

mwelzl: Oh sure, that can happen. No problem though. The specific rules we had written up about this in NEAT for an active close were: "Triggered by calling neat_shutdown();
on_writable() will never be called again, calling neat_write() returns error
on_readable() can be called"

mwelzl: Thinking more about this - perhaps we should add this information (about the read-callback potentially still firing) to the description of the call?  I left it out on purpose to avoid making this text unnecessarily complex, but if you think it helps then we could include this, for "close" and "abort" ?

tfpauly:
Looks good. I pushed a commit to complete the change of the list style to include the implementation concepts section.

PR #50, opened 2018-02-16
cleaned up things a bit, removed some intents, inserted several comme…

mwelzl:
This ends my promised pass - I hope that in particular the app intents people can take a look at some of my comments (I wrote them inline as \[MICHAEL: blabla])

### DIFF
--- a/draft-trammell-taps-interface.md
+++ b/draft-trammell-taps-interface.md
@@ -245,7 +245,7 @@ Preconnection object during pre-establishment, forward-reference
 ## Specifying Endpoints {#endpointspec}
 
 The transport services API uses Endpoint objects to refer to local and remote endpoints.
-Endpoint objects can be configured using various representations of endpoint identifiers, 
+Endpoint objects can be configured using various representations of endpoint identifiers,
 including IP addresses, hostnames, or interface names as well as port
 numbers and service names:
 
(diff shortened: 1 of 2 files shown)


draft-trammell-taps-interface.md
```diff
@@ -1001,36 +1003,34 @@ application to receive the Content multiple times.
...
 
-###Protocol Properties
+#### ACK_immediately {#ack_immediately}
+
+This boolean property specifies, if true, that an application wants this
+Content to be immediately acknowledged by the receiver. In case of
+reliable transmission, this informs the transport protocol on the sender
+side faster that it can remove the Content from its buffer; therefore
+this property can be useful for latency-critical applications that maintain
+tight control over the send buffer (see {{#sending}}).
```
britram: I like this property, but I left it out because I wasn't sure I knew of any protocols that supported it...

draft-trammell-taps-interface.md
```diff
@@ -1306,6 +1309,8 @@ type Connection interface {
 }
 
 type Content interface {
+\[MICHAEL: We need something here for metadata, for a receiver -
+at least ECN must be a part of this metadata.]
```
britram: yeah, I thought about adding a simple Get(k string) interface{} metadata accessor, but we'd need to add metadata language to the receive section. I can do this after this PR and preconnection land...

draft-trammell-taps-interface.md
mwelzl: SCTP does. RFC 7053. It's really quite a special thing... I think you have to be quite an advanced application writer doing advanced juggling with the Sent event to make use of this (Sent will fire when it the transport buffer below drains, which will happen faster when ACKs come in faster). But let them have all the things that can reduce their latency, I'd say.

britram:
changed base to master since #49 landed

PR #52, opened 2018-02-18
Separate Endpoint into LocalEndpoint and RemoteEndpoint

### LINKED ISSUE #43: Types: local and remote endpoints
One of the things Post Sockets did was to separate `Local` and `Remote` into distinct types, rather than having a single `Endpoint` type. I think this is a useful distinction, that we should re-introduce it into the Transport Services API, since the two types are not interchangeable in the API calls. 

Examples:
* In Section 5, the `NewConnection()` call doesn't take two `Endpoints`, it takes a `Local` and a `Remote`.
* In Section 5.1, the allowable operations on `localSpecifier` are different to those on `remoteSpecifier` (the interface can be specified for the local, but not for the remote). 

I'm not much tied to the names, but I think it important that we distinguish logically separate types in the API and architecture. 

Discuss.

csperkins:
These commits separate Endpoint into LocalEndpoint and RemoteEndpoint types, and update the rest of the documents to match.

Closes #43


draft-pauly-taps-arch.md
```diff
@@ -213,9 +213,9 @@ Beyond the basic objects, there are several high-level groups of actions that an
  +-->| Preconnection |----------->| Connection |---------------> Closed
  |   +---------------+     :      +------------+ Connection:
  |                         :      ^   ^    |     Finished  :
- +-- Local Endpoint        :      |   |    |               :
+ +-- LocalEndpoint         :      |   |    |               :
```
britram: was this an overcorrection to CamelCase? Spaces (we should use "Local Endpoint", "Remote Endpoint" in the architecture draft; whether we camelcase in the API or just call things Local and Remote is a separate discussion

tfpauly: Agreed that this should be "Local Endpoint" and "Remote Endpoint"

draft-pauly-taps-arch.md
```diff
@@ -237,21 +237,19 @@ Beyond the basic objects, there are several high-level groups of actions that an
...
 
-* Listener: A Listener object accepts incoming transport protocol connections from Remote Endpoints  and generates corresponding Connection objects. It is created from a Preconnection object that specifies the type of incoming connections it will accept.
+* Listener: A Listener object accepts incoming transport protocol connections from RemoteEndpoints  and generates corresponding Connection objects. It is created from a Preconnection object that specifies the type of incoming connections it will accept.
 
 ### Pre-Establishment {#preestablishment}
 
-* Endpoint: An Endpoint represents one side of a transport connection. It is a concept that represents how an application views itself or a peer that may vary in levels of specificity. Examples include "IP address + port", "hostname + port", and "DNS service name".
+* RemoteEndpoint: The RemoteEndpoint represents the application's name for a peer that can participate in a transport connection. For example, a DNS name for the peer.
```
tfpauly: I don't think a DNS name is sufficient for an Endpoint, unless that's referring to the DNS Service name. It needs to be a DNS hostname + port. I think listing out the types as we had with endpoint is still necessary and useful.

My concern with removing the overall definition of an Endpoint is that we lose the common place to discuss the fact that an endpoint is not simply one type (hostname + port, or address + port, or anything else), but an abstract representation of something that may correspond via resolution or advertisement to one or more other endpoints.

What precisely is the advantage of removing the definition of the concept of an endpoint? It's still a term that is not really standardized, and without defining it, I think we're losing a fundamental type. It's essentially the equivalent of sockaddr for sockets—a type that indeed has meaning in multiple contexts, not just in remote/local dichotomies. If we don't specify this carefully, we risk interpretations of endpoints to be just one set of things (like how sockets used just addresses) that will pigeonhole us later on.

draft-pauly-taps-arch.md
```diff
@@ -321,7 +319,7 @@ The Transport System Implementation Concepts define the set of objects used inte
 
 * Path Racing: Path Racing is the act of attempting to establish, or scheduling attempts to establish, multiple Protocol Stacks that differ based on a selection from the available Paths.
 
-* Endpoint Racing: Endpoint Racing is the act of attempting to establish, or scheduling attempts to establish, multiple Protocol Stacks that differ based on the specific representation of the remote and local endpoints, such as IP addresses resolved from a DNS hostname.
+* Endpoint Racing: Endpoint Racing is the act of attempting to establish, or scheduling attempts to establish, multiple Protocol Stacks that differ based on the specific representation of the RemoteEndpoint and the LocalEndpoint, such as IP addresses resolved from a DNS hostname.
```
tfpauly: I don't really like this change either—when we're racing Endpoints (which now have no definition...), they're not using "the" original Remote or Local Endpoint, they're using new sets of derived local and remote Endpoints, which don't necessarily carry the full meaning of what is defined as the Local Endpoint and Remote Endpoint to the application. Specifically, I may have a "Local Endpoint" to hold my local state, and a "Remote Endpoint" to define the hostname, certificate policy, etc, for my peer; but the Endpoints representing what I resolved for my Remote Endpoint are sub-addresses that don't need their own state in the same way, but are still different entities, since I am trying different addresses.

draft-pauly-taps-arch.md
britram: on reflection -- i think one of the reasons I thought of this as an implementation detail is precisely because it's a `sockaddr`, but this...

>> What precisely is the advantage of removing the definition of the concept of an endpoint? It's still a term that is not really standardized, and without defining it, I think we're losing a fundamental type.

... is compelling to me.

tfpauly: Right, just as we want to live a post-sockets world (lower case post-sockets, here), we also want to live in a post-sockaddr world. 

For example, dealing with byte structures like sockaddrs in languages like Swift is really a pain. Using an Endpoint object with methods that replace the basic sockaddr fields is way easier.

draft-pauly-taps-arch.md
csperkins: Fixed.

draft-pauly-taps-arch.md
csperkins: Remote Endpoint should be defined by name and port, I agree. That was an oversight that I've fixed. 

For Endpoint, we need to be clear that RemoteEndpoint is an abstract type that can represent several different types of remote endpoint, and that name+port is just one example. Similarly for LocalEndpoint. And I'm certainly not advocating `struct sockaddr` style anything. But, it's still not clear to me what a generic Endpoint would include? 

I expect `LocalEndpoint` and `RemoteEndpoint` may well have some common methods (e.g., they'd both implement `Display` in a Rust implementation), but I don't see use cases for a generic `Endpoint`. All the cases that I can immediately think of, it matters whether it's a local or remote endpoint, so having a common super type doesn't give us anything.

This is not to say that I'm against `Endpoint` in principle, I'd just like a use case before adding it.

draft-pauly-taps-arch.md
csperkins: I think there are two issues here.

Firstly, should "Endpoint Racing" be a concept, or are we just describing Path Racing in a little more detail here? I suspect the latter.

Second, and independent of whether `Endpoint` exists as a super-type of `RemoteEndpoint` and `LocalEndpoint`, how does the racing relate to endpoints. I guess we pass in a list of RemoteEndpoint objects, each of which is resolved to one or more additional `RemoteEndpoint` objects, then we race connections to those remote endpoint? I'm note sure I understand the comment about "don't necessarily carry the full meaning"?

draft-pauly-taps-arch.md
tfpauly: I agree that a raw Endpoint that isn't Remote or Local may not be used in any of the externally-facing APIs to create a Connection. From an API standpoint, it's fine to require a specific type of the general class.

All I'm arguing is that in our list of concepts and definitions, we need to define that there is a basic concept of an Endpoint, that isn't necessarily local or remote, since local and remote only become relevant in the context of a connection, or some communication. Having a definition of Remote and Local Endpoints without ever saying what an endpoint is (or duplicating the definitions, saying Local: See the generic parts of Remote) seems unnecessary.

Also, from an implementation standpoint, the Endpoint really isn't as trivial as you're saying. It's not just a matter of displaying. Since we don't use sockaddrs in the exposure of the address form of endpoint, the Endpoint does the following:
- Can create hostname/address from a string, which automatically parses into a hostname, IPv4, or IPv6 address. If the IPv6 address is a mapped v4 address, it parses that out.
- Create port with a string or number ("http" or "80")
- Extract address byte encoding for use in forming packets
- Extract port in either network or host byte order
- Extract IP address family
- Do subnet or DNS name segment matching and comparisons between endpoints (are these two in the same slash #)
- Allow definition of custom endpoint subclasses that can resolve into standard types

All of these operations are useful for both local and remote endpoints, such as when you are filling out or parsing headers within the protocol stack. If as an implementation, I was told to make the LocalEndpoint and RemoteEndpoint completely different classes in a strongly typed language, I would need to redefine all of these? If they're not subclasses of the same Endpoint generic class, at least they both implement the template/protocol of an Endpoint so we can re-use some logic.

To that end, defining the concept of Endpoint first, then explaining Remote and Local Endpoint, still makes sense in the architecture.

tfpauly:
Since the previous comment was on a now-hidden commit:

I agree that a raw Endpoint that isn't Remote or Local may not be used in any of the externally-facing APIs to create a Connection. From an API standpoint, it's fine to require a specific type of the general class.

All I'm arguing is that in our list of concepts and definitions, we need to define that there is a basic concept of an Endpoint, that isn't necessarily local or remote, since local and remote only become relevant in the context of a connection, or some communication. Having a definition of Remote and Local Endpoints without ever saying what an endpoint is (or duplicating the definitions, saying Local: See the generic parts of Remote) seems unnecessary.

Also, from an implementation standpoint, the Endpoint really isn't as trivial as you're saying. It's not just a matter of displaying. Since we don't use sockaddrs in the exposure of the address form of endpoint, the Endpoint does the following:

Can create hostname/address from a string, which automatically parses into a hostname, IPv4, or IPv6 address. If the IPv6 address is a mapped v4 address, it parses that out.
Create port with a string or number ("http" or "80")
Extract address byte encoding for use in forming packets
Extract port in either network or host byte order
Extract IP address family
Do subnet or DNS name segment matching and comparisons between endpoints (are these two in the same slash #)
Allow definition of custom endpoint subclasses that can resolve into standard types
All of these operations are useful for both local and remote endpoints, such as when you are filling out or parsing headers within the protocol stack. If as an implementation, I was told to make the LocalEndpoint and RemoteEndpoint completely different classes in a strongly typed language, I would need to redefine all of these? If they're not subclasses of the same Endpoint generic class, at least they both implement the template/protocol of an Endpoint so we can re-use some logic.

To that end, defining the concept of Endpoint first, then explaining Remote and Local Endpoint, still makes sense in the architecture.

mirjak:
@tfpauly I'd say that what you described about the use of endpoint should probably be noted in the implementation draft. I'm still not sure if Endpoint should be noted in the architecture because it's not an architectural concept but rather an implementation issue, but I think I'm okay to define it for sake of having a good definition we can refer to.

@csperkins I'm still wondering if we generate new remotes in the resolution step or if we store all we know about one actual endpoint in the same remote. If we get multiple IP addresses for the same URL, is that multiple hosts serving the same content, or one entity that is presumably distributed over multiple physical (frontend) host, or one host having multiple addresses. My answer would be, as we don't know the actually setup, we should keep all information we learned based on the initial input provided for the Remote within that Remote object. I guess this goes all back to the (StackEvo) question: what is an endpoint? So maybe good to have a definition...

britram:
...fwiw i'm holding off on my big editorial pass until this lands...

tfpauly:
@mirjak I think that while the concept is most critical for the implementation, it still affects the application developer that uses the API. If, as a developer, I want to create a Remote or Local Endpoint, I need to know how to specify my address/hostname/port (which needs to be common types between the two). We also have application level tests that do things like create a listener, extract the local endpoint, and then create a connection back to ourselves over loopback, using the the local as a remote. It would be strange if that wasn't possible.

Mainly, I'm just arguing we should define the existence of a generic "address/hostname/service" blob from the start for clarity. Removing the definition of the base endpoint concept didn't make the remote or local endpoints stronger definition; it just removed clarifying text that's general to both.

csperkins:
@tfpauly on reflection, putting endpoint into the architecture seems okay, but I'm still unconvinced we need it in the API. Or, at least, before we add Endpoint to the API, we need to more clearly understand the range of different endpoint types to really see what's common between them. I wonder if things like, e.g., LocalEndpoints obtained via a TURN server, break some of the assumptions of commonality.

@mirjak I was thinking that `resolve(RemoteEndpoint)` would return a list of `RemoteEndpoint` objects. Maybe they should reference what they were resolved from, but I don't think we have any way of knowing whether they refer to the same host or not.

tfpauly:
@csperkins I agree on the point for the API document. I don't think we need to have the generic endpoint there for now, as long as the architecture explains the relationship.

With regards to resolution, our implementation takes an endpoint, and resolves it into a set of endpoints (of a different type, generally; used for bonjour service -> hostname, hostname -> address, v4 address -> NAT64 translated v6 address, etc). We've found it useful for the endpoint to track the "parent" endpoint it was derived from, mainly so that when protocols that care about the original hostnames need to access them (like TLS for cert validation, or SOCKS for figuring out the original hostname to request to the proxy), it's available. Also useful for doing per-domain policy evaluation for flows.

csperkins:
@tfpauly ack - will try to write this up tomorrow (today was a write-off due to lecturing...)

tfpauly:
Looks good!

PR #71, opened 2018-02-22
Add text for UDP Send

adventureloop:


tfpauly:
The main point of this new text seems to be around how UDP socket API allows sending and receiving on multiple addresses; however, that same effect can be handled by UDP Connections with fully bound five-tuples (and UDP Listeners). Should this instead add text on how to map the abstraction of UDP connections onto traditional unconnected socket implementations?

britram:
@adventureloop @tfpauly is there any reason not to just deconflict and land this one?

britram:
Timing out, pushing to ietf-01.

draft-brunstrom-taps-impl.md
```diff
@@ -418,6 +418,12 @@ How to handle sending data onto examples like TCP, UDP, and a basic Length-Value
 
 How to handle and notify errors when sending.
 
+#### Sending content with UDP
+
+The UDP Send Primitives described in section 3.1 of {{RFC8304}} allow applications to send data with either an explicit destination for each packet or the option to bind the destination address for each send call to a single host. The UDP send primitive also offers applications the ability to set a range of IP options which are required to build applications that can use the network effectivly. Implementation MUST expose these mechanisms or provide services to implement their transport features. 
+
+Implementations should consider the advice in the UDP Usage Guidlines (RFC8085) and create a 'connect' style for each UDP Connection object.
```
tfpauly: Typo on Guidlines -> Guidelines

draft-brunstrom-taps-impl.md
```diff
@@ -418,6 +418,12 @@ How to handle sending data onto examples like TCP, UDP, and a basic Length-Value
 
 How to handle and notify errors when sending.
 
+#### Sending content with UDP
+
+The UDP Send Primitives described in section 3.1 of {{RFC8304}} allow applications to send data with either an explicit destination for each packet or the option to bind the destination address for each send call to a single host. The UDP send primitive also offers applications the ability to set a range of IP options which are required to build applications that can use the network effectivly. Implementation MUST expose these mechanisms or provide services to implement their transport features.
```
tfpauly: Exposing per-send addresses is specifically not exposed right now in the API, and I think this is intentional. We can treat all UDP flows as "connected" so that sends and receives stay simple, and we can have back pressure from per-flow interface queues. I believe we can make an argument in this text that everything that can be expressed with an unconnected UDP socket is expressible with Listeners and Connections.

britram:
This seems OBE; @adventureloop, can you send an updated PR if this is still current?

Issue #90, opened 2018-02-25
Semantics of "low latency"


mwelzl:
In issue #65, we discussed that "Immediate Acknowledgement" (from SCTP) is a bit too protocol-specific and could maybe be called "low latency", allowing the stack to automatically turn this on "under the hood".

I think the only use case for this is when an application executes tight control over the send buffer, and wants to be able to change what it hands over to the transport system at the last minute - then, it is probably good to know ASAP when a particular Message has really arrived at the receiver.
This would be an application that requires low latency.

The API currently has another send parameter called "immediate". This would turn off the Nagle algorithm, for example - it represents a wish to send stuff fast, possibly at the cost of wasting capacity  (more header overhead, or other things).

I don't think that an application that needs "immediate" would always also need "immediate acknowledgement" or vice versa - these seem to be two different things.

Should we have two boolean send parameters for these two types of requests?
"Prefer low latency" (to trade against capacity) and "ACK ASAP" ?

gorryfair:
I'd suggest that when you say "the only use case for this is", you're starying to make me think this is a very specific application use.  The immediate stuff in SCTP was a very specific use. I'm quite OK with believing other use-cases exist, what I'd question is whether this specific feature needs to be in a generic API?

mwelzl:
I think so, yes. In SCTP, there were SCTP-internal reasons AND there was the app reason: https://tools.ietf.org/html/rfc7053#section-4.1    I now wonder whether this is something that COULD somehow be handled internally in the transport system, or if we need to expose a flag to enable this kind of thing. I'll follow up with emails.

mwelzl:
TL;DR: I thought more about this. I started to write an email, and writing made me think deeper and understand better. I conclude that, whether to get ACKs back immediately or not is also just a matter of trading latency against overhead, and indeed one "low latency" bit is enough.

Below, I describe the background for my conclusion. It would be good if someone could check it for potential logical errors.

Tommy once described Apple's model of buffer control to me as follows:

***
I consider TCP_NOTSENT_LOWAT to be an implementation detail of the sockets API, and not really the right model here. 
The backpressure we have is based on the asynchronous callback. Nothing is blocking.
If the application wants to slow down the generation of its content based on back pressure from the stack, it should wait to generate another send based on the callback to the previous send being fired. However, you may enqueue multiple sends back to back, since the data may already be generated. This gives all of the capabilities of back pressure, without blocking, or forcing it.
***

[...]

Issue #69, opened 2018-02-21
Sweep properties/features without group consensus to a new Appendix A


britram:
During my final editorial review pre-00, I'll take everything we're still discussing into a new Appendix A, with some text along the lines "an advantage of this framework is it has extension points for properties and preferences beyond those in the proposed standard interface; here are some potential properties and preferences that we have some design and implementation experience with that may prove useful, and may be added to the core interface in future revisions of this document."

britram:
So far, tracking the discussion in #60, I propose the following:

- Everything in section 5.2.3 goes into Appendix A, except:
  - Traffic Category which gets lifted up into Transport Preferences (and possibly renamed Capacity Profile by reference to Minset), with additional language that (1) the only useful preference level is Prefer and (2) in addition to being used to select transports, it can also be used to configure selected transports for e.g. DSCP marking.

- The Immediate boolean send property gets removed to the appendix in favor of disabling Nagle via Traffic Category.

britram:
(On rewriting Traffic Category as a path/protocol preference, it's not clear to me what the stack is supposed do with any of these other than the interactive/default/bulk classes. I'm leaving Stream/Constant Rate but I'm not sure what to say about it. We can fix this post-00)

philsbln:
We are currently writing the implementation part for all Intents – text for that will appear in the intents-implementation branch in the next days

mwelzl:
I think we ended up with a confusion due to renaming: I thought that the thing that is based on the minset "capacity profile" is "timeliness", not "traffic category". And why remove Immediate?

Actually:

- an app needs a way to say that we require low latency, even at the cost of more overhead. That can be Nagle-like Message bundling, or other things (e.g. see issue #90). Because this request may (as in the case of Message bundling) be necessary for a message, I think it would be good to keep "Immediate".

- removing this from the minset capacity profile, it only covers table 1 in draft-ietf-tsvwg-rtcweb-qos-18. I think we want applications to be able to choose cells in this table, *somehow*. I don't know how to do this most cleanly; for version -00, maybe we could simply point at this draft and say that the "Traffic Category" would choose cells in this table?  Or some such solution...

britram:
> I think we ended up with a confusion due to renaming: I thought that the thing that is based on the minset "capacity profile" is "timeliness", not "traffic category

aaaah right. yep that makes more sense.

>  And why remove Immediate?

because capacity profile "Interactive" does this for you on a per connection basis?

[...]

PR #96, opened 2018-02-26
Brian's Monday editorial pass on API

britram:
Fixes all outstanding London API issues except #92 and #93

### DIFF
--- a/draft-trammell-taps-interface.md
+++ b/draft-trammell-taps-interface.md
@@ -1021,12 +1021,27 @@ underlying Protocol Stack's framing.  See {{receive-framing}} for handling
 framing in situations where the Protocol Stack provides octet-stream transport
 only.
 
-Note that as with sending, the Message object passed to Received may represent
-a partial message, larger than the maximum message size than can be received
-atomically. In this case, the Message object passed contains an indication
+The Message object passed to Received is complete and atomic, unless one of the following
+conditions holds:
+
+* the underlying protocol stack supports message boundary preservation, and
+  the size of the Message is larger than the buffers available for a single
+  message;
+* the underlying protocol stack does not support message boundary
+  preservation, and the deframer (see {{receive-framing}}) cannot determine
+  the end of the message using the buffer space it has available; or
+* the underlying protocol stack does not support message boundary
+  preservation, and no deframer was supplied by the application
+
+In this case, the Message object passed to Received contains an indication
 that the object received is partial, a reference to the full Message it
-belongs to, as well as the byte range of the data content within the partial
-Message.
+belongs to, the byte offset of the data in the partial Message within the full
+Message, and an indication whether this is the last (highest-offset) partial
+Message in the full Message.
+
+Note that in the degenerate case -- no message boundary preservation and no
+deframing -- the entire connection is represented as one large message of
+indeterminate length.
 
 ~~~
 Connection -> ReceiveError<>
--- a/draft-trammell-taps-interface.md
+++ b/draft-trammell-taps-interface.md
@@ -157,19 +157,27 @@ with these Actions and Events.
 
 The following notations, which can be combined, are used in this document:
 
-- Object := Action()
+~~~
+Object := Action()
+~~~
 
 >> An Action creates an Object.
 
-- Object.Action()
+~~~
+Object.Action()
+~~~
 
 >> An Action is performed on an Object.
 
-- Object -> Event&lt;>
+~~~
+Object -> Event<>
+~~~
 
 >> An Object sends an Event.
 
-- Action(parameter, parameter, ...) / Event&lt;parameter, parameter, ...>
+~~~
+Action(parameter, parameter, ...) / Event<parameter, parameter, ...>
+~~~
 
 >> An Action takes a set of Parameters; an Event contains a set of Parameters.
 
@@ -616,7 +624,7 @@ multiple candidates.
 The following events may be sent by the Connection after Initiate() is called:
 
 ~~~
-Connection -> Ready&lt;>
+Connection -> Ready<>
 ~~~
 
 The Ready event occurs after Initiate has established a transport-layer
@@ -625,7 +633,7 @@ candidate Path. No Receive events (see {{receiving}}) will occur before
 the Ready event for connections established using Initiate.
 
 ~~~
-Connection -> InitiateError&lt;>
+Connection -> InitiateError<>
 ~~~
 
 An InitiateError occurs either when the set of transport and cryptographic
@@ -656,7 +664,7 @@ called, no further parameters may be bound to the Preconnection, and no
 subsequent establishment call may be made on the Preconnection.
 
 ~~~
-Preconnection -> ConnectionReceived&lt;Connection>
+Preconnection -> ConnectionReceived<Connection>
 ~~~
 
 The ConnectionReceived event occurs when a Remote Endpoint has established a
@@ -668,7 +676,7 @@ event, and is ready to use as soon as it is passed to the application via the
 event.
 
 ~~~
-Preconnection -> ListenError&lt;>
+Preconnection -> ListenError<>
 ~~~
 
 A ListenError occurs either when the Preconnection cannot be fulfilled for
@@ -697,7 +705,7 @@ been called, no further parameters may be bound to the Preconnection, and
 no subsequent establishment call may be made on the Preconnection.
 
 ~~~
-Preconnection -> RendezvousDone&lt;Connection>
+Preconnection -> RendezvousDone<Connection>
 ~~~
 
 The RendezvousDone<> event occurs when a connection is established with the
@@ -708,7 +716,7 @@ resulting Connection is contained within the RendezvousDone<> event, and is
 ready to use as soon as it is passed to the application via the event.
 
 ~~~
-Preconnection -> RendezvousError&lt;msgRef, error>
+Preconnection -> RendezvousError<msgRef, error>
 ~~~
 
 An RendezvousError occurs either when the Preconnection cannot be fulfilled
@@ -798,7 +806,7 @@ Protocol Stacks that support it.
 Like all Actions in this interface, the Send action is asynchronous.
 
 ~~~
-Connection -> Sent&lt;msgRef>
+Connection -> Sent<msgRef>
 ~~~
 
 The Sent event occurs when a previous Send action has completed, i.e., when the
@@ -815,7 +823,7 @@ times without waiting for a Sent event, it has created more buffer inside the
 transport system than an application that only issues a Send after this event fires.
 
 ~~~
-Connection -> Expired&lt;msgRef>
+Connection -> Expired<msgRef>
 ~~~
 
 The Expired event occurs when a previous Send action expired before completion;
@@ -825,7 +833,7 @@ partially reliable transports. The Expired event contains an
 implementation-specific reference to the Message to which it applies.
 
 ~~~
-Connection -> SendError&lt;msgRef>
+Connection -> SendError<msgRef>
 ~~~
 
 A SendError occurs when a Message could not be sent due to an error condition:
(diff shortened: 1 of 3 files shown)


draft-trammell-taps-interface.md
```diff
@@ -285,94 +288,69 @@ localSpecifier.withStunServer(address, port, credentials)
...
-Default preference levels and possible combinations of Transport Parameters and
-preference levels are specified in {{appendix-preferences}}.
-
-
-
-### Transport Preferences {#transport-prefs}
-
-Transport Preferences drive protocol selection and path selection on
-connection establishment. Since there could be paths over which some transport
+requirements and preferences for the transport. These include protocol and path
```
tfpauly: Capitalize Protocol and Path Selection here, since they are terms from the architecture?

draft-trammell-taps-interface.md
```diff
@@ -1073,117 +859,117 @@ decrease.
...
 
-#### Send Bitrate {#send-bitrate}
+#### Instantaneous Capacity Profile
 
-This numeric property in Bytes per second specifies at what bitrate the
-application wishes the content to be sent. A transport supporting this
-feature will not exceed the requested Send Bitrate even if flow-control
-and congestion control allow higher bitrates. This helps to avid bursty
-traffic pattern on busy video streaming servers.
+This enumerated property specifies the application's preferred tradeoffs for sending this Message; it is a per-Message override of the Capacity Profile protocol and path selection property (see {{transport-params}}). Th
```
tfpauly: Is there a dangling "Th" here?

draft-trammell-taps-interface.md
```diff
@@ -1073,117 +859,117 @@ decrease.
...
-Content to be acknowledged immediately by the receiver. In case of reliable
+Message to be acknowledged immediately by the receiver. In case of reliable
 transmission, this informs the transport protocol on the sender side faster
-that it can remove the Content from its buffer; therefore this property can be
+that it can remove the Message from its buffer; therefore this property can be
 useful for latency-critical applications that maintain tight control over the
 send buffer (see {{sending}}).
 
-#### Send Bitrate {#send-bitrate}
+#### Instantaneous Capacity Profile
```
mwelzl: Interesting idea to make this whole choice per-message. I'm not 100% sure about everything here - e.g. the value choices ("constant rate" in particular). But we should make progress, and I think this is a fine discussion to postpone until after London.

draft-trammell-taps-interface.md
```diff
@@ -181,33 +182,39 @@ language and the platform. Actions could be implemented as functions or method
...
-  and for long-term caching of cryptographic identities and parameters for
-  associations among endpoints.
+- The interface provides asynchronous connection establishment,
+  transmission, and reception, allowing most application interactions with the
+  transport layer to be event-driven, in line with developments in modern
+  platforms and programming languages
 
-- Atomic transmission of content, using application-assisted framing and
-  deframing where the underlying transport does not provide these.
+- Explicit support for multistreaming and multipath transport protocols, and the
```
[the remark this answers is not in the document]
britram: will do...

draft-trammell-taps-interface.md
```diff
@@ -1073,117 +859,117 @@ decrease.
...
 underlying Protocol Stack's framing.  See {{receive-framing}} for handling
 framing in situations where the Protocol Stack provides octet-stream transport
 only.
 
+Note that as with sending, the Message object may represent a partial message,
+larger than the maximum message size than can be received atomically. In this
+case, the Message object passed contains some reference to the full Message it
+belongs to, as well as the byte range of the data content within the partial
+Message.
+
```
mwelzl: What the receiver gets with this call may or may not coincide with what the sender has handed over as a message. The application must be able to figure out where a message begins and ends. No problem to do this generically, via framers and de-framers, but simply receiving a "message" doesn't tell the receiver anything about the sender's message size, and hence this paragraph makes no sense to me.

draft-trammell-taps-interface.md
```diff
@@ -1243,28 +1024,93 @@ connectionProperties := Connection.getProperties()
...
+Note that Protocol Properties are also set during pre-establishment, as
+transport parameters, to preconfigure Protocol Stacks during establishment.
+
+Generic Protocol Properties include:
+
+* Timeout for aborting Connection: 
+  This numeric property specifies how long
+  to wait before aborting a Connection during establishment, or after a
+  connection has failed after establishment. It is given in seconds.
+
```
mwelzl: "after a connection has failed after establishment" seems recursive, as such failure would be the result of the timeout. Suggestion: "...how long to wait before aborting a Connection during establishment, or before deciding that a connection has failed after establishment."

draft-trammell-taps-interface.md
```diff
@@ -1243,28 +1024,93 @@ connectionProperties := Connection.getProperties()
...
+  This numeric property, if applicable, represents the maximum Message size 
+  that can be sent without incurring network-layer fragmentation and/or transport layer segmentation at the sender. This property is read-only.
+
+* Maximum non-partial Message size on send:
+  This numeric property represents the maximum Message size that can be sent as
+  a non-partial Message. This property may be read-only.
+
+* Maximum non-partial Message size on receive: 
+  This numeric property represents
+  the maximum Message size that can be received as a non-partial Message. This property may be read-only.
```
mwelzl: Why "may be"?  The ones above just read "...is read-only", which seems better.

draft-trammell-taps-interface.md
```diff
@@ -1243,28 +1024,93 @@ connectionProperties := Connection.getProperties()
...
+  before or during Connection establishment, see also {{send-idempotent}}.
+  It is given in Bytes. This property is read-only.
+
+* Maximum Message size before fragmentation or segmentation:
+  This numeric property, if applicable, represents the maximum Message size 
+  that can be sent without incurring network-layer fragmentation and/or transport layer segmentation at the sender. This property is read-only.
+
+* Maximum non-partial Message size on send:
+  This numeric property represents the maximum Message size that can be sent as
+  a non-partial Message. This property may be read-only.
```
mwelzl: Why "may be"?  The ones above just read "...is read-only", which seems better.

draft-trammell-taps-interface.md
```diff
@@ -1243,28 +1024,93 @@ connectionProperties := Connection.getProperties()
...
+
+Specific Protocol Properties can additionally be associated with a specific
+protocol. For example, the application can specify a set of TCP Options to use
+if and only if TCP is selected by the system, including options such as the
+Maximum Segment Size (MSS), and options around Acknowledgement Stretching. Such
+properties should not be assumed to apply across different protocols, but must
+be possible to specify if required by the application. Attempts to set specific
+protocol properties on a protocol stack not containing that specific protocol
+are simply ignored, and do not raise an error.
+
```
mwelzl: In issue #69, I proposed the following text as a replacement for this:

***
In order to specify Specific Protocol Properties, Transport System implementations MAY offer applications to attach a set of options to the Preconnection object, associated with a specific protocol. For example, an application could specify a set of TCP Options to use if and only if TCP is selected by the system. Such properties MUST NOT be assumed to apply across different protocols.
***

If you don't agree with the RFC21119 style keywords, don't capitalize them. If you generally don't agree with this text replacement, at least remove ", and options around Acknowledgement Stretching", please.

draft-trammell-taps-interface.md
britram: This one seems like `setsockopt(SO_SNDBUF)` to me... i.e., since the maximum non-partial size is a function of stack and system buffers, you might be able to change it. Happy to "is read only" this for London though.

draft-trammell-taps-interface.md
britram: ACK. I'd like to keep the part about not raising errors though...

draft-trammell-taps-interface.md
britram: Partial messages are different than deframed messages. However, the two play nicely together: if you have a non-framing Protocol Stack with vanilla TCP on top, and you have a deframer that knows how to split the stream into application layer messages, you might still end up in a situation where the framer says "hey, I got a length-prefixed thing that says it's 24MB" and the system only supports buffering messages 512kB at a time. So the deframer would chop the stream up into buffer-sized Partials and pass them up.

(Indeed, in the degenerate case where you're running on vanilla TCP and you don't pass a deframer down, the system default deframer would treat the entire stream as a single giant message, and pass partials down according to its own buffering and scheduling requirements)...

Not really clear what the problem is. Suggest we defer to London.

mwelzl: I'm fine with deferring the discussion to London, but until then I consider this paragraph as wrong and don't think it should be added. We're trying to kick out stuff for which we don't have consensus - so: first convince me, then add it, not vice versa.

draft-trammell-taps-interface.md
mwelzl: Ah, sure, yes, I missed this in my suggested text update.

draft-trammell-taps-interface.md
tfpauly: I'd really like to see this left in, and I thought it had been discussed and agreed in Cambridge? Either way, let's discuss more for sure, but I think is necessary to have a document that actually works for these protocol use cases. If we decide some other way, we can change it, but I think we need this text in here for London.

@mwelzl one point to re-iterate is that the notion of Message does not need to match between the two sides—it's just a matter of how a local side parses bytes on a connection into chunks.

draft-trammell-taps-interface.md
britram: yeah i kind of hate "instantaneous capacity profile" as a name ("timeliness" isn't great either, though"); inclined to leave this as is and revisit after we decide on #90.

draft-trammell-taps-interface.md
britram: so it's clear that we need partial message support on both send and receive, simply because there are buffers in the world. we handwaved on this wildly in previous revisions of this document (and allowed ourselves to do so because "content"). this corrects that.

I... really don't understand your objection here, Michael ("this paragraph is bad" is not extremely helpful). Can you help me out a bit? Is the problem the interplay between deframers and partial messages? Is the text just bad?

draft-trammell-taps-interface.md
mwelzl: Sure, I agree

draft-trammell-taps-interface.md
britram: Oops, missed this one. Fixed.

draft-trammell-taps-interface.md
mwelzl: Okay. As it's written it doesn't make sense to me - I absolutely agree with having stuff that actually works, but so does NEAT, and NEAT doesn't need this...  maybe I'm missing out on one of your use cases.

Brian's "indeed, ..." bracket above describes exactly what's on my mind: in this case, why would you need to consider this a partial message, and give a "reference to the full message it belongs to"?  This seems unnecessary to me.  Can we agree on phrasing this as optional? Phrasing it as factual (as it is now) indicates that there is no other way, which would IMO mean "no vanilla TCP". I would be fine with "In this case, the Message object passed may contain some reference to the full Message it belongs to, as well as the byte range of the data content within the partial Message."

draft-trammell-taps-interface.md
mwelzl: Let's go for "is read only". JFYI, the background for this (see minset) is indeed an unchangeable value that comes from the transport protocol. E.g., even if you allow fragmentation, UDP has a fixed limit for the size of the message that you can give it.

draft-trammell-taps-interface.md
mwelzl: Same rationale as above - a function of the transport protocol, imposing a max value.

draft-trammell-taps-interface.md
britram: 735da31582ec00bbebfd9cd4fa5d04f70a6bad61 contains an attempt to make this more explicit (and to note that vanilla TCP is of course supported). If making "full message reference" optional is better, happy to do that.

PR #99, opened 2018-02-26
fix #92 (in Notation); add note on Receive callback.

### LINKED ISSUE #92: Write §11 on error handling

britram:


britram:
note this one has an edit about Receive callbacks that only makes sense once #96 lands...

PR #106, opened 2018-02-27
Clone preconnections too (fix #95)

### LINKED ISSUE #95: Work out the relationship between Connection Groups and Preconnection
We create Connection Groups via `Clone().`

It would be nice to be able to create groups during pre-establishment.

But you need a Connection to `Clone()` a connection.

There are a few possible syntaxes to use here, we should pick one. For now, will drop mention of being able to clone in pre-establishment.

britram:


draft-trammell-taps-interface.md
```diff
@@ -758,31 +761,37 @@ configure the remote.
...
+supporting efficient connection group usage are preferred by the application.
+
+There is only one Protocol Property that is not entangled, i.e., it is a
+separate per-Connection Property for individual Connections in the group:
+niceness. Niceness works as in {{send-niceness}}: when allocating available
+network capacity among Connections in a connection group, sends on connections
+with higher Niceness values will tend to yield to connections with lower
+Niceness values. The priority setting is purely advisory. No guarantees are
+given, and each implementation is free to implement exact capacity allocation as
+it sees fit.
```
mwelzl: "tend to yield to connections" is IMO too weak - it doesn't say anything about how the available capacity is really divided; however, when we DO get a multi-streaming protocol, the capacity division can become quite precise. That's why I included the equation for an "ideal transport system implementation". Why not do the same? Let me try:

An ideal transport system implementation would assign the Connection the capacity share (M-N) x C / M, where N is the Connection's Niceness value, M is the maximum Niceness value used by all Connections in the group and C is the total available capacity (this assumes that positive integer or float values are used to represent Niceness).

draft-trammell-taps-interface.md
```diff
@@ -758,31 +761,37 @@ configure the remote.
 
 ## Connection Groups {#groups}
 
-Groups of Connections can be created using Clone action:
+Groups of Preconnections or Connections can be created using the Clone action:
 
 ~~~
+Preconnection : Preconnection.Clone()
+
```
csperkins: := presumably?

draft-trammell-taps-interface.md
csperkins: Should be explicit that calling `Initiate()`, `Listen()`, or `Rendezvous()` on a cloned `Preconnection` does not cause the other clones to be established, but if they are established then they'll be entangled with the existing connection.

Also suggest adding: "If `Clone()` is called on a `Preconnection` that is subsequently established into an active `Connection`, then attempts to change the parameters of the cloned `Preconnection` will fail."

britram: Niceness in send properties is (was originally) specified as a set of discrete classes, as opposed to an inverse numeric priority, but yeah, this is a better way to do it (as long as the "as it sees fit" language stays in). will change.

draft-trammell-taps-interface.md
britram: yep will change

draft-trammell-taps-interface.md
britram: agreed, will change.

Issue #79, opened 2018-02-23
Replace Content with Message


gorryfair:
I am not hugely sold on the definition of /Content: A Content object/. I struggled in NEAT with the best word, and we eventually decided on /data block/. That's not perfect, but it at least keeps away from the notion of content delivery networks, and various other uses of the word "Content". 

Can we *please* reconsider this?

tfpauly:
Sure, we can reconsider. Data Block is one option; we could also just do Data, but then the singular chunk is more awkward.

Some test phrases:
- "a given Connection is ready to send and/or receive Content"
- "a given Connection is ready to send and/or receive Data Blocks"
- "a given Connection is ready to send and/or receive Data"

Yeah, I don't like making this seem like a Content Delivery Network, but when this does get applied to protocols like QUIC and HTTP/2 as a transport, then there are instances in which this is the Content portion of the Header/Content divide.

Any other votes from the group?

britram:
<whispers> message

tfpauly:
I mean, message works nicely—other than the fact that it's a bit awkward when describing incomplete message content.

abrunstrom:
Hard to come up with a good term for this one. In the test phrases above I think Data works quite well, but I am not sure it will in all situations.

gorryfair:
I suggested "data block" or "data", but I really would not have questioned "message".

philsbln:
With our Socket Intents work, we started at "object" and finally came to "message". A message can easily be spread across several PUDs for transmission. If re-assembly fails, we end up with partial messages.

tfpauly:
One comment on the relationship between Message and Content: if we do Message as the main architecture concept (which I'm fine with to be clear), I can imagine that implementations or APIs can still refer to the "content of the message" as opposed to metadata about the message—for example, when I receive a UDP packet as a Message, the content is the interior payload of the datagram, but my metadata includes the ECN flags received on the packet. Does that sound good?

gorryfair:
Foe me trhat would be fine ((... we could people thinking about the content of messages, or content messages, etc - but at least we don't have "content of content" etc ;-) ))

britram:
#85 does this for Architecture; we still need to make this change in API (and make sure it carries forward into Implementation as necessary)

tfpauly:
Yes, needed for API as well. I already modified Implementation for this.

tfpauly:
[...]

PR #107, opened 2018-02-27
Fixes to sending and receiving messages to support vanilla TCP

mwelzl:
The previous text didn't support sending multiple messages across vanilla TCP without framers/deframers or a top-level protocol in the stack that would de-frame. This change allows sending multiple messages across vanilla TCP and handing them over as bytes to an application that, by itself, detects message boundaries from the chunks of bytes that it gets.

### DIFF
--- a/draft-brunstrom-taps-impl.md
+++ b/draft-brunstrom-taps-impl.md
@@ -427,7 +427,7 @@ Protocols that provide the framing (such as length-value protocols, or protocols
 
 ### Sending Messages
 
-The effect of the application sending a Message is determined by the top-level protocol in the established Protocol Stack. That is, if the top-level protocol provides an abstraction of framed messages over a connection, the application will be able to send multiple Messages on that connection, even if the framing protocol is built on a byte-stream protocol like TCP.
+The effect of the application sending a Message is determined by the top-level protocol in the established Protocol Stack. That is, if the top-level protocol provides an abstraction of framed messages over a connection, the receiving application will be able to obtain multiple Messages on that connection without supplying a deframer, even if the framing protocol is built on a byte-stream protocol like TCP.
 
 #### Send Parameters
 
@@ -458,7 +458,7 @@ Since sending a Message may involve a context switch between the application and
 
 ### Receiving Messages
 
-Similar to sending, Receiving a Message is determined by the top-level protocol in the established Protocol Stack. The main difference with Receiving is that the size and boundaries of the Message are not known beforehand. The application can communicate in its Receive action the parameters for the Message, which can help the implementation know how much data to deliver and when. For example, if the application only wants to receive a complete Message, the implementation should wait until an entire Message (datagram, stream, or frame) is read before delivering any Message content to the application. Alternatively, the application can specify the minimum number of bytes of Message content it wants to receive (which may be just a single byte) to control the flow of received data.
+Similar to sending, Receiving a Message is determined by the top-level protocol in the established Protocol Stack. The main difference with Receiving is that the size and boundaries of the Message are not known beforehand. The application can communicate in its Receive action the parameters for the Message, which can help the implementation know how much data to deliver and when. For example, if the application only wants to receive a complete Message, the implementation should wait until an entire Message (datagram, stream, or frame) is read before delivering any Message content to the application. This requires the implementation to understand where messages end, either via a supplied deframer or because the top-level protocol in the established Protocol Stack preserves message boundaries; if, on the other hand, the top-level protocol only supports a byte-stream and no deframers were supported, the application must specify the minimum number of bytes of Message content it wants to receive (which may be just a single byte) to control the flow of received data.
 
 If a Connection becomes finished before a requested Receive action can be satisfied, the implementation should deliver any partial Message content outstanding, or if none is available, an indication that there will be no more received Messages.
 
@@ -596,7 +596,7 @@ Connection lifetime for TCP translates fairly simply into the the abstraction pr
 
 If the application sends a Close, that can translate to a graceful termination of the TCP connection, which is performed by sending a FIN to the remote endpoint. If the application sends an Abort, then the TCP state can be closed abruptly, leading to a RST being sent to the peer.
 
-Without a layer of framing above TCP, the cleanest abstraction for sending and receiving Messages over TCP is to treat each direction of the stream of bytes as a single Message, terminated by a FIN. That means that if the application can make several Send calls to enqueue subsequent data chunks for the same Message to continue writing, but marking the Message complete corresponds to closing the sending stream. Similarly, when the application receives the final portion of a Message, it knows that the receiving stream has been closed.
+Without a layer of framing above TCP, the receiver side of the implementation has to rely on a supplied deframer to determine message boundaries. In the absence of such a deframer, a request by the application to only receive complete messages fails.
 
 ## UDP
 
--- a/draft-brunstrom-taps-impl.md
+++ b/draft-brunstrom-taps-impl.md
@@ -427,7 +427,7 @@ Protocols that provide the framing (such as length-value protocols, or protocols
 
 ### Sending Messages
 
-The effect of the application sending a Message is determined by the top-level protocol in the established Protocol Stack. That is, if the top-level protocol provides an abstraction of framed messages over a connection, the receiving application will be able to obtain multiple Messages on that connection without supplying a deframer, even if the framing protocol is built on a byte-stream protocol like TCP.
+The effect of the application sending a Message is determined by the top-level protocol in the established Protocol Stack. That is, if the top-level protocol provides an abstraction of framed messages over a connection, the receiving application will be able to obtain multiple Messages on that connection, even if the framing protocol is built on a byte-stream protocol like TCP.
 
 #### Send Parameters
 
@@ -596,7 +596,7 @@ Connection lifetime for TCP translates fairly simply into the the abstraction pr
 
 If the application sends a Close, that can translate to a graceful termination of the TCP connection, which is performed by sending a FIN to the remote endpoint. If the application sends an Abort, then the TCP state can be closed abruptly, leading to a RST being sent to the peer.
 
-Without a layer of framing above TCP, the receiver side of the implementation has to rely on a supplied deframer to determine message boundaries. In the absence of such a deframer, a request by the application to only receive complete messages fails.
+Without a layer of framing (a top-level protocol in the established Protocol Stack that preserves message boundaries, or an application-supplied deframer) on top of TCP, the receiver side of the transport system implementation can only treat the incoming stream of bytes as a single Message, terminated by a FIN when the Remote Endpoint closes the Connection.
 
 ## UDP
 
--- a/draft-trammell-taps-interface.md
+++ b/draft-trammell-taps-interface.md
@@ -1033,11 +1033,17 @@ conditions holds:
 * the underlying protocol stack does not support message boundary
   preservation, and no deframer was supplied by the application
 
-In this case, the Message object passed to Received may contain an indication
-that the object received is partial, the byte offset of the data in the
-partial Message within the full Message, an indication whether this is the
+The Message object passed to Received will indicate one of the following:
+
+1. this is a complete message;
+2. this is a partial message containing a section of a message with a known message boundary (made partial for local buffering reasons, either by the underlying protocol stack or the deframer). In this case, the Message object passed to Received may contain the byte offset of the data in the partial Message within the full Message, an indication whether this is the
 last (highest-offset) partial Message in the full Message, and an optional
-reference to the full Message it belongs to.
+reference to the full Message it belongs to; or
+3. this is a partial message containing data with no definite message boundary, i.e. the only known message boundary is given by termination of the Connection
+
+Note that in the absence of message boundary preservation and without
+deframing, the entire connection is represented as one large message of
+indeterminate length.
 
 ~~~
 Connection -> ReceiveError<>
--- a/draft-trammell-taps-interface.md
+++ b/draft-trammell-taps-interface.md
@@ -1033,15 +1033,11 @@ conditions holds:
 * the underlying protocol stack does not support message boundary
   preservation, and no deframer was supplied by the application
 
-In this case, the Message object passed to Received contains an indication
+In this case, the Message object passed to Received may contain an indication
 that the object received is partial, the byte offset of the data in the
 partial Message within the full Message, an indication whether this is the
 last (highest-offset) partial Message in the full Message, and an optional
 reference to the full Message it belongs to.
-
-Note that in the degenerate case -- no message boundary preservation and no
-deframing -- the entire connection is represented as one large message of
-indeterminate length.
 
 ~~~
 Connection -> ReceiveError<>


draft-brunstrom-taps-impl.md
```diff
@@ -427,7 +427,7 @@ Protocols that provide the framing (such as length-value protocols, or protocols
 
 ### Sending Messages
 
-The effect of the application sending a Message is determined by the top-level protocol in the established Protocol Stack. That is, if the top-level protocol provides an abstraction of framed messages over a connection, the application will be able to send multiple Messages on that connection, even if the framing protocol is built on a byte-stream protocol like TCP.
+The effect of the application sending a Message is determined by the top-level protocol in the established Protocol Stack. That is, if the top-level protocol provides an abstraction of framed messages over a connection, the receiving application will be able to obtain multiple Messages on that connection without supplying a deframer, even if the framing protocol is built on a byte-stream protocol like TCP.
```
csperkins: I don't get it: if you don't supply a deframer for a byte-stream protocol, how does the top-level protocol provide the abstraction of framed messages over the connection? Surely the point of the deframer is to provide this.

draft-trammell-taps-interface.md
```diff
@@ -1033,16 +1033,12 @@ conditions holds:
 * the underlying protocol stack does not support message boundary
   preservation, and no deframer was supplied by the application
 
-In this case, the Message object passed to Received contains an indication
+In this case, the Message object passed to Received may contain an indication
```
britram: I don't get this. Handing a message to a receiver without telling it that it's partial risks incorrect functionality. If the API is going to break its contract that messages are meaningfully atomic (albeit because the application told it to), then it has a responsibility to indicate this (even if the application ignores it)

draft-trammell-taps-interface.md
```diff
@@ -1033,16 +1033,12 @@ conditions holds:
...
 that the object received is partial, the byte offset of the data in the
 partial Message within the full Message, an indication whether this is the
 last (highest-offset) partial Message in the full Message, and an optional
 reference to the full Message it belongs to.
 
-Note that in the degenerate case -- no message boundary preservation and no
-deframing -- the entire connection is represented as one large message of
-indeterminate length.
-
```
britram: It seems important to leave this in (though on reading I can see that it might be useful to use less potentially-but-unintentionally-pejorative word than "degenerate" as an adjective for this case)

draft-brunstrom-taps-impl.md
mwelzl: Surely that's the point of a deframer, but it's also possible without providing one to the transport system: the receiving application can determine message boundaries on its own.

mwelzl:
It seems we never came to consensus over the main issue here - we did have consensus that it's fine to have framers and deframers. It seems that the crux of the matter is the following:

Tommy and you want: a sender saying "the frame ends here" will terminate a TCP connection; so if you really really want to hand over multiple chunks of data from a stream, then you should use one huge message and only terminate when finished with everything.

I don't want that, for the reason I explained so often under the heading "Application-Framed Bytestream": the sending application doesn't know whether it gets vanilla TCP or e.g. SCTP. If it gets SCTP, messages can be sent as messages, with all their properties: lifetime, unordered delivery ...   but this can only work if the sender knows that messages are messages. If the transport system gets TCP, however, it must be able to send the messages directly on the bytestream. Either way this is fine as long as the receiving application is able to determine boundaries in the chunk of data that it gets: in case of SCTP, it really does get messages, in case of TCP, it gets a bytestream.

By requiring "atomic" message transmission from one end to the other, and not being able to send multiple messages on a single TCP connection, this functionality becomes unavailable.

If this seems confusing, please consider taking another look at the animations, e.g. here:
http://heim.ifi.uio.no/michawe/research/publications/ietf98-taps-discussion.pptx

If you want a longer, more detailed explanation, please read:
https://tools.ietf.org/html/draft-ietf-taps-minset-01#appendix-A.3.1

If you don't believe this can work, please download and test NEAT.

I don't understand all this "a TCP connection represents a large message business". It's a strange limit that gets in the way of efficiently falling back to vanilla TCP.

draft-trammell-taps-interface.md
mwelzl: Ah, right. We can fix this by flipping the logic. I suggest to replace:

In this case, the Message object passed to Received may contain an indication
that the object received is partial, the byte offset of the data in the
partial Message within the full Message, an indication whether this is the
last (highest-offset) partial Message in the full Message, and an optional
reference to the full Message it belongs to.

with:

Complete Messages passed to Received are marked as such.
When a message is incomplete, it may contain an indication
that the object received is partial, the byte offset of the data in the
partial Message within the full Message, an indication whether this is the
last (highest-offset) partial Message in the full Message, and an optional
reference to the full Message it belongs to.


... now, no more contract broken. You know what you get.

draft-trammell-taps-interface.md
mwelzl: It's not about the word "degenerate": it hints at "the FIN indicates the end of a message", which is a bad idea.

britram:
our lack of consensus seems to me to come from a giant pile of subtle misunderstandings:

> Tommy and you want: a sender saying "the frame ends here" will terminate a TCP connection; 

nope. Well, at least I don't want this. This would be silly. 

There's no guarantee that one side of a connection using this interface will be connected to another side of a connection also using the interface. Message sends are passed down to the underlying protocol stack atomically, so that stacks that do indeed support boundary preservation have boundaries. If you get TCP, it'll ignore the boundaries, and hope the other side provides deframing. (If a sender absolutely requires atomicity without relying on receiver-side deframing, it could prohibit the selection of protocol stacks that don't support it, of course.) 

> By requiring "atomic" message transmission from one end to the other, and not being able to send multiple messages on a single TCP connection, this functionality becomes unavailable.

But we don't require atomic message transmission. We require an API which presents atomic message transmission to the application when it is available, and allows (asks) for the application to help provide this when it is not.

Here's what this looks like when a sender writes multiple messages on a stack not supporting boundary preservation to a receiver not running this interface:

```
  sender [msg 1] [msg 2] [msg 3] -> TCP [bytes 1] [bytes 2] [bytes 3] -> receiver [pile of bytes] -> application layer parser
```

The receiver gets a pile or bytes in order and hands them to an app layer parser, just like in TCP.

Here's what this looks like when a TCP only-sender writes three app-layer-delimited messages to a stream to a receiver running this interface with deframers:

```
TCP [bytes 1] [bytes 2] [bytes 3] -> interface deframer [msg 1] [msg 2] [msg 3] -> receiver
```

The receiver pushed the deframe logic down into the interface, and got to simplify its own workflow -- every callback has an atomic message.

Here's what this looks like in the case this PR seems to care about: TCP, no deframer:

```
TCP [bytes 1] [bytes 2] [bytes 3] -> receiver [pile of partial messages containing bytes 1-3 in order] -> application layer parser
```

Just like the TCP case above. Each `Received<>` event is just like a stream `read()` call in TCP. (I can, indeed, see why you'd want to provide length hints to `Receive()` to support this, and I think we should add that.)

> If you don't believe this can work, please download and test NEAT.

I believe it can work, that's not the question at all. I just don't think it's a very useful abstraction if a goal of building a new interface is moving application development away from stream-centric assumptions in order to allow it to take advantage of new transports that provide better-than-single-stream abstractions. 

> I don't understand all this "a TCP connection represents a large message business". It's a strange limit that gets in the way of efficiently falling back to vanilla TCP.

There's no limit here. To translate this interface into bytestream thinking: a partial message is simply a section of a bytestream with a running offset (which can itself be calculated trivially by the stack or the system). When a receiver doesn't have enough information to atomize messages, it presents bytestream(s) as partials.

mwelzl:
So in this comment, I agree with absolutely everything. Just two comments:


> > Tommy and you want: a sender saying "the frame ends here" will terminate a TCP connection;
> nope. Well, at least I don't want this. This would be silly.

Great!!   But I was under the strong impression that this is what Tommy wants because of things that he wrote, both in the text and in issue #79. Fantastic if I misunderstood all that, but then my suggested text changes should be fine. All I want is to allow:
1) sending many messages (multiple "message over" markers from the application) over a single TCP connection
2) receiving blocks of data (similar to normal TCP data reception today) from which an application determines where messages begin and end.


> (I can, indeed, see why you'd want to provide length hints to Receive() to support this, and I think we should add that.)

Just for the record, that didn't come from me - I suspect it was from Tommy. Anyway, it makes sense to me as well.

draft-trammell-taps-interface.md
mwelzl: If you insist on keeping this, please give better arguments than "it seems important" - because to me this text is confusing and potentially misleading.

draft-trammell-taps-interface.md
mwelzl: Do we agree on this replacement?

britram:
> > (I can, indeed, see why you'd want to provide length hints to Receive() to support this, and I think we should add that.)

> Just for the record, that didn't come from me - I suspect it was from Tommy. Anyway, it makes sense to me as well.

hm i think that might have been me. such is the nature of the pile of confusions...

draft-brunstrom-taps-impl.md
csperkins: If you have HTTP over TCP, then you have an HTTP deframer, surely?

britram:
> All I want is to allow:
> 1. sending many messages (multiple "message over" markers from the application) over a single TCP connection

with the caveat that these might be completely lost on send, yep, absolutely we need to allow this.

> 2. receiving blocks of data (similar to normal TCP data reception today) from which an application determines where messages begin and end.

yep, with the caveat that we do not yet have a way to pass the max length the application wants to receive (but can easily add this).

> but then my suggested text changes should be fine

Hm, I think you still need the "this is a partial message" indicator on the receive (which of course the application knows, but allowing the "partial message" marker to be omitted when the underlying transport / deframe stack doesn't have any boundary information seems like a contract violation to me.

draft-trammell-taps-interface.md
csperkins: If you have a raw TCP connection with no framing protocol specified, how else do you know when the Message ends other than the end of the Connection? (i.e., the FIN indicates the end of message)

Note, though, that this is an extremely rare case, since you almost always have a framing protocol.

draft-brunstrom-taps-impl.md
mwelzl: So your point is that the receiving application MUST tell the transport system about its deframer?

csperkins:
> 1. sending many messages (multiple "message over" markers from the application) over 
> a single TCP connection

Yes, of course. Same as happens today.

> 2. receiving blocks of data (similar to normal TCP data reception today) from which an 
> application determines where messages begin and end.

Yes, although the way I was envisaging this working was that the application provides a de-framing function to the stack, and that de-framing function determines where the messages begin and end, so that what's returned are complete messages. That way, you get a message based abstraction even if running over a byte-stream transport like TCP.

mwelzl:
I agree with the caveats you list above, that's all fine. About this:

> Hm, I think you still need the "this is a partial message" indicator on the receive (which of course the application knows, but allowing the "partial message" marker to be omitted when the underlying transport / deframe stack doesn't have any boundary information seems like a contract violation to me.

The point is that the receiving application must be able to know if it did get a complete message. When it does, the transport system can say so, so there you get your contract. When it does not, the receiver-side transport system may not know about it because it doesn't know about sender-supported message boundaries in the first place (the first of your two mentioned caveats). If it got the data as a TCP stream, then at best, it can say "perhaps this isn't a complete message, I'm not sure", but it certainly can't provide the other things your text requires in this case (whether it's a partial message, a reference to the complete message, ..).

If we phrase it as I suggest, I think we break no contract and we promise as much as we can promise.

draft-brunstrom-taps-impl.md
csperkins: Yes. This doesn't change anything on the wire, or affect interoperability with existing endpoints or protocols, it just means the application gives the de-framing function to the stack to run, rather than doing it itself.

britram:
AHHHH okay.

Actually, I think this should be a three-state indicator:

(1) This is a complete message (known because the deframer said so, or the transport stack did).
(2) This is a partial message containing a section of a message with a known message boundary (made partial for local buffering reasons, either by the underlying protocol stack or the deframer)
(3) This is a partial message containing a data with no definite message boundary (i.e. stream protocol and no deframer)...

draft-brunstrom-taps-impl.md
mwelzl: I see. Sure, this way, it can work. It wasn't clear to me that, in order to obtain messages, a receiver MUST provide a de-framer. I'm not sure this is very explicit in the text?

draft-trammell-taps-interface.md
mwelzl: This is connected to the discussion above: I now understand that you interpret this statement as "if you want anything better than the entire connection to be one huge message, you MUST provide a deframer".  So this is a way of saying "you really MUST provide a deframer." Now I think I also understand why Brian thought this is important. So, I think that's okay... I understand that this is a good idea because the transport system could then even apply this on the TCP byte stream (provided it has access to it) in order to cut out messages that arrive out of order. But then (for -01 ?) we should add such a flag to the receiver side to optionally offer, to do these Minion / Hollywood - like things.

Anyway. If my interpretation above is right, then I agree to keep this sentence.

mwelzl:
CONVERGED!  I agree with the 3-state indicator proposal  :)

tfpauly:
@britram I like your last list of three, but I will still point out that over any stream protocol with no deframer, there will be a definite end to received data (the equivalent of a FIN).

My previous comments about FINs, etc, were just to point out that it is possible to map a stream this way. If we're doing HTTP/2 or QUIC streams, we may indeed want the end of the message to correspond to the end of the stream, since HTTP over HTTP/2 generally will do one request per stream identifier. However, I had also mentioned that it is possible to have the stream protocol ignore any claimed message boundaries, but this means that the message boundaries are understood to be lost when mapped into the bytes on the wire.

csperkins:
@britram yes..... I _think_ that makes sense

mwelzl:
@tfpauly - this sounds fine!  (phew!)    So we agree that a sender can send 50 small messages over a single connection (each marked with "end of message"), right? And when the transport system ends up using vanilla TCP, and the receiver-side application then wants to obtain the messages, it must provide a deframer (note that the "must" here is new to me, but I think it's fine and probably even a good idea in order to do cooler things later).

draft-brunstrom-taps-impl.md
```diff
@@ -427,7 +427,7 @@ Protocols that provide the framing (such as length-value protocols, or protocols
...
```
tfpauly: The point of this paragraph was to describe how you can send multiple message using a framer. I don't think we should add "without a deframer", since the point was to describe the framer.

draft-brunstrom-taps-impl.md
```diff
@@ -458,7 +458,7 @@ Since sending a Message may involve a context switch between the application and
...
+Similar to sending, Receiving a Message is determined by the top-level protocol in the established Protocol Stack. The main difference with Receiving is that the size and boundaries of the Message are not known beforehand. The application can communicate in its Receive action the parameters for the Message, which can help the implementation know how much data to deliver and when. For example, if the application only wants to receive a complete Message, the implementation should wait until an entire Message (datagram, stream, or frame) is read before delivering any Message content to the application. This requires the implementation to understand where messages end, either via a supplied deframer or because the top-level protocol in the established Protocol Stack preserves message boundaries; if, on the other hand, the top-level protocol only supports a byte-stream and no deframers were supported, the application must specify the minimum number of bytes of Message content it wants to receive (which may be just a single byte) to control the flow of received data.
```
tfpauly: This part looks fine to me.

britram:
> but I will still point out that over any stream protocol with no deframer, there will be a definite end to received data (the equivalent of a FIN).

Well, yes... but the application won't know what the end is *until* it sees the FIN. so s/no definite message boundary/no known message boundary/ in state 3?

(git logistics question: @mwelzl, do you want to update this branch (which will automatically update the PR) with receive length and three-state partials, or shall we land this PR then do those in another?)

draft-brunstrom-taps-impl.md
```diff
@@ -596,7 +596,7 @@ Connection lifetime for TCP translates fairly simply into the the abstraction pr
 
 If the application sends a Close, that can translate to a graceful termination of the TCP connection, which is performed by sending a FIN to the remote endpoint. If the application sends an Abort, then the TCP state can be closed abruptly, leading to a RST being sent to the peer.
 
-Without a layer of framing above TCP, the cleanest abstraction for sending and receiving Messages over TCP is to treat each direction of the stream of bytes as a single Message, terminated by a FIN. That means that if the application can make several Send calls to enqueue subsequent data chunks for the same Message to continue writing, but marking the Message complete corresponds to closing the sending stream. Similarly, when the application receives the final portion of a Message, it knows that the receiving stream has been closed.
+Without a layer of framing above TCP, the receiver side of the implementation has to rely on a supplied deframer to determine message boundaries. In the absence of such a deframer, a request by the application to only receive complete messages fails.
```
tfpauly: A deframer *is* a layer of framing above TCP. So this is saying "if you don't have a deframer, use a deframer". I think that if you choose not to use a deframer, you should at the very least get an indication that a FIN has been received by the received content being marked as complete. That was actually a big problem for stream clients we had when we didn't give them this bit.

draft-brunstrom-taps-impl.md
mwelzl: I get it now, agreed

mwelzl:
I'll update the PR next  (first time I'm doing this - sorry for being clumsy)

draft-brunstrom-taps-impl.md
mwelzl: Sorry about the language silliness, which I'll fix. But: a "layer of framing" could also be taken to mean a protocol - something that the sender and receiver both know about, e.g. HTTP/TCP inside the transport system.

The part I dislike about the original text is: "That means that if the application can make several Send calls to enqueue subsequent data chunks for the same Message to continue writing, but marking the Message complete corresponds to closing the sending stream."   Besides the sentence being broken, this again sounds like prohibiting my wish of "sending 50 short messages across a single connection".

I also generally find it strange to conflate "end of message" with "FIN" because the latter means "connection closed". You can always tell your applications that the remote endpoint has closed.

tfpauly: There's a very important difference between "connection closed/finished" and one side sending a FIN. I can send a write-close after my initial request, and still keep the TCP connection up for a long time. If an application needs to know when the other send send the end-of-message/file, then that should be communicated in the Receive path, not in an event that "Connection Finished", which generally implies no more traffic at all.

abrunstrom: But we do not have a write-close so in the current API I guess the Close call is what will generate the FIN, correct? I am fine this being communicated in the receive path, but I think the "Connection Finished" will also follow from the FIN.

britram:
This looks good to me now. I'll do a separate PR to add receive size to the API.

mwelzl:
Ah, sorry, I forgot this to do this (receive size); thanks!

britram:
...see #121

britram:
Will land both of these; open subsequent issues if we need to tweak the text, but IMO this is certainly -00 quality...

PR #122, opened 2018-02-28
Intents implementation

theri:
In the Implementation draft, we add text to explain how Application Intents can be implemented in a transport system.
The two main functions are Path Selection, which can be implemented by ranking branches in the tree of connection options that will be raced, and Protocol Configuration, which happens after the racing is complete and a protocol stack has been chosen.


draft-brunstrom-taps-impl.md
```diff
@@ -330,6 +343,32 @@ Protocol options are checked next in order. Whether or not a set of protocol, or
...
+
+* Capacity Profile:
+An implementation may use the Capacity Profile to prefer paths optimized for the application's expected traffic pattern:
+   * Interactive/Low Latency:
+     Prefer paths with the lowest expected latency
+   * Constant Rate:
+     Prefer paths that can satisfy the requested Stream Send or Stream Receive Bitrate
+   * Scavenger/Bulk:
+     Prefer paths with the highest expected available bandwidth
+
```
mwelzl: For each of these path properties, I'd ask: "how can you know?". Please give an example.

draft-brunstrom-taps-impl.md
```diff
@@ -330,6 +343,32 @@ Protocol options are checked next in order. Whether or not a set of protocol, or
...
+* Size to be Sent or Received:
+An implementation may use the Size to be Sent or Received to compute an estimate of the completion time of a transfer over different available paths and then prefer the path with the shorter expected completion time. This Intent may be used instead of the Capacity profile, as the application does not always know whether its transfer will be latency-bound or bandwidth-bound, and thus may not be able to specify a Capacity Profile. A related paper is currently under submission.
```
mwelzl: But again I'd like to know how you know about the path property - in this case, the expected completion time per path. Is this from previous measurements?

draft-brunstrom-taps-impl.md
```diff
@@ -330,6 +343,32 @@ Protocol options are checked next in order. Whether or not a set of protocol, or
...
+
+* Send / Receive Bitrate:
+If the application indicates an expected send or receive bitrate, an implementation may prefer a path that can likely provide the desired bandwidth.
```
mwelzl: And again, how do you know about this path property?

draft-brunstrom-taps-impl.md
```diff
@@ -386,6 +425,17 @@ Implementations may select the criteria by which a leaf node is considered to be
...
 
+### Protocol Configuration
+
+Once a connection is established and a protocol stack has been chosen, an implementation SHOULD configure any Protocol Properties given by the application which apply to the chosen protocol stack.
+
+Additionally, an implementation MAY configure a transport protocol according to following Application Intents:
+
+* Traffic Category, Capacity Profile:
+The implementation may map information about the anticipated traffic pattern or priority to a DSCP code point value, e.g., see [I-D.ietf-tsvwg-rtcweb-qos]. Furthermore, if an application indicates a preference for low latency, the implementation may disable the Nagle algorithm on a TCP connection.
```
mwelzl: But this is now already written for the "instantaneous capacity profile". Maybe remove "traffic category, capacity profile" here, or write TBD in case the intention is to be broader than what the "instantaneous capacity profile" offers?

draft-brunstrom-taps-impl.md
```diff
@@ -386,6 +425,17 @@ Implementations may select the criteria by which a leaf node is considered to be
...
+
+* Send Bitrate: If an application indicates a certain bitrate it wants to send on the connection, the implementation may limit the bitrate of the outgoing communication to that rate, for example by setting an upper bound for the TCP congestion window of a connection calculated from the Send Bitrate and the Round Trip Time. This helps to avoid bursty traffic patterns on video streaming servers, see [Trickle].
```
mwelzl: My immediate reaction to this is this max cwnd would limit the "max. bitrate", not the bitrate - which goes back to my prior comment that a bitrate is an average over a certain time interval, and so this is a pretty vague parameter. But I haven't had time to read the Trickle paper yet (btw thanks for the reference). So I think this can be left standing as it is, for now.

draft-brunstrom-taps-impl.md
theri: We can observe the maximum throughput over time in a performance cache, which we use as an estimate of the available bandwidth. I added some clarification to the parts you marked as well as one sentence to Section 8.2 (Performance Caches).

draft-brunstrom-taps-impl.md
theri: Yes, we use cached performance estimates of Round Trip Time and expected maximum bandwidth to compute an estimated completion time on the paths. I modified the text to clarify this.

draft-brunstrom-taps-impl.md
theri: Indeed the Capacity Profile in the current API draft already reflects most of what we specify here, although we may get broader in the future as we figure out the relation between Capacity Profile, our Timeliness Intent, and our Category Intent, which I think is a post-London discussion.
Nevertheless, I think it makes sense to mention Protocol Configuration aspects from the API draft in the Implementation draft, too, also for including the Capacity Profile. Do you agree with this in general?

mwelzl: In general yes, but my point is that this text repeats pretty much what I've already written in the implementation draft - if you search for Nagle or DSCP you'll see it

theri: Oh, now I see, that was an oversight on our part, sorry.
I removed the part for now, broadening the concept is TBD post-London.

draft-brunstrom-taps-impl.md
theri: If we call this Maximum Bitrate to be Sent instead, does it fix the problem?

Also, as the Send Bitrate was now the only "Protocol Configuration" and we think that it also fits as a Send Parameter (e.g., if the message is an entire video being streamed), we moved the Send Bitrate to the "Send Parameters" (Section 5.1.1.1).

draft-brunstrom-taps-impl.md
```diff
@@ -330,6 +343,35 @@ Protocol options are checked next in order. Whether or not a set of protocol, or
 
 Branching for derived endpoints is the final step, and may have multiple layers of derivation or resolution, such as DNS service resolution and DNS hostname resolution.
 
+
+## Ranking Branches
```
tfpauly: Can we call this step "sorting" rather than "ranking"? This would be more in line with Happy Eyeballs (https://tools.ietf.org/html/rfc8305#section-4)

draft-brunstrom-taps-impl.md
```diff
@@ -330,6 +343,35 @@ Protocol options are checked next in order. Whether or not a set of protocol, or
...
+
+Implementations SHOULD rank the branches of the tree of connection options in order to prioritize them for candidate racing. Leaf nodes on branches with higher rankings represent connection attempts that will be raced first. Implementations SHOULD order the branches to reflect the preferences expressed by the application for its new connection, including Protocol and Path Selection Properties as well as Application Intents, which are specified in {{draft-trammell-taps-interface}}. In addition to the properties provided by the application, an implementation MAY include additional criteria such as cached performance estimates, see {{performance-caches}}, or system policy, see {{role-of-system-policy}}, in the ranking.
```
tfpauly: Let's leave off Application Intents here, since that's still just in the appendix of the API for now.

draft-brunstrom-taps-impl.md
```diff
@@ -330,6 +343,35 @@ Protocol options are checked next in order. Whether or not a set of protocol, or
...
+
+The Transport Parameters specified in {{draft-trammell-taps-interface}} may be used to rank branches in the following ways:
+
+* Interface Type:
+If the application specifies an interface type to be required, an implementation MUST purge all branches that correspond to paths over interfaces that do not match this type. Similarly, it MUST purge paths that correspond to interface types that the application prohibits. If the application specifies an interface type to be preferred or avoided, implementations SHOULD rank paths accordingly.
```
tfpauly: I disagree with the approach here: we shouldn't:

1. Gather all possible paths into a tree
2. During sorting, prune out the ones that aren't allowed

Instead, it's simpler to:

1. Gather all permissible paths
2. Sort them for preference

Essentially, the pruning is already covered by Candidate Gathering. We can add more text there to emphasize that (it already has "This list is determined by the requirements, prohibitions, and preferences of the application as specified in the Path Selection Properties and Protocol Selection Properties."), but that's where that operation belongs.

draft-brunstrom-taps-impl.md
```diff
@@ -330,6 +343,35 @@ Protocol options are checked next in order. Whether or not a set of protocol, or
...
+* Capacity Profile:
+An implementation may use the Capacity Profile to prefer paths optimized for the application's expected traffic pattern according to cached performance estimates, see {{performance-caches}}:
+   * Interactive/Low Latency:
+     Prefer paths with the lowest expected Round Trip Time
+   * Constant Rate:
+     Prefer paths that can satisfy the requested Stream Send or Stream Receive Bitrate, based on observed maximum throughput
+   * Scavenger/Bulk:
+     Prefer paths with the highest expected available bandwidth, based on observed maximum throughput
+
+* Size to be Sent or Received:
```
tfpauly: Please remove items that are still just in the appendix and not agreed upon for now. (that's Size, Bitrate, and Cost)

draft-brunstrom-taps-impl.md
```diff
@@ -445,6 +487,9 @@ The effect of the application sending a Message is determined by the top-level p
 
 - Instantaneous Capacity Profile: when this is set to "Interactive/Low Latency", the Message should be sent immediately, even when this comes at the cost of using the network capacity less efficiently. For example, small messages can sometimes be bundled to fit into a single data packet for the sake of reducing header overhead; such bundling SHOULD not be used in case this parameter is true. For example, in case of TCP, the Nagle algorithm SHOULD be disabled when Immediate=true. Scavenger/Bulk can translate into usage of a congestion control mechanism such as LEDBAT, and/or the capacity profile can lead to a choice of a DSCP value as described in {{I-D.ietf-taps-minset}}).
 
+- Send Bitrate: If an application indicates a certain bitrate it wants to send on the connection, the implementation may limit the bitrate of the outgoing communication to that rate, for example by setting an upper bound for the TCP congestion window of a connection calculated from the Send Bitrate and the Round Trip Time. This helps to avoid bursty traffic patterns on video streaming servers, see [Trickle].
```
tfpauly: Let's remove this until it gets out of the appendix

draft-brunstrom-taps-impl.md
```diff
@@ -581,7 +626,7 @@ In addition to protocol state, Protocol Instances SHOULD provide data into a per
...
+These items can be cached on a per-address and per-subnet granularity, and averaged between different values. The information SHOULD be cached on a per-network basis, since it is expected that different network attachments will have different performance characteristics. Besides Protocol Instances, other system entities may also provide data into performance-oriented caches. This could for instance be signal strength information reported by radio modems like Wi-Fi and mobile broadband or information about the battery-level of the device. Furthermore, the system may cache the observed maximum throughput on a path as an estimate of the available bandwidth.
```
tfpauly: Great!

draft-brunstrom-taps-impl.md
philsbln: I am fine with renaming it – needs a little more explanation then, but we can do that

draft-brunstrom-taps-impl.md
philsbln: I guess we can move them to an appendix for now - completely removing them makes this pull-request pointless.

draft-brunstrom-taps-impl.md
philsbln: Same as above – moving it to an appendix for now

draft-brunstrom-taps-impl.md
philsbln: Should be addressed now by #8ecb7df

draft-brunstrom-taps-impl.md
philsbln: Fixed with commit #9d3bf58

philsbln:
Commit #9d3bf58 should address all changes requested - please reconsider your reviews

draft-brunstrom-taps-impl.md
theri: Yes, I think this relates to the cached performance parameters, how the exact calculation of such parameters is done, their cache lifetime, etc. 
I'm currently working on another paper that deals with how to get good performance estimates, and I'm looking forward to discussing this more and perhaps contributing some insights to the draft... perhaps post-London. :)

draft-brunstrom-taps-impl.md
theri: This probably relates to the point above of how to get good metrics. Otherwise I think it would be bitrate in (M)Bits per Second, as given e.g. in an MPEG-DASH manifest file?

mwelzl: As long as it's related to a time interval or a certain amount of data, I guess it's fine...

Issue #112, opened 2018-02-27
API: How to specify idempotent data?


csperkins:
How is idempotent data, to be send in the zero-RTT connection establishment, specified?

The Use 0-RTT session establishment with an idempotent Message property "specifies whether an application would like to supply a Message to the transport protocol before Connection establishment" which suggests the idempotent message is set on the `Preconnection` and used later when the `Connection` is established, but it might be more natural to specify the idempotent data as a parameter to `Initiate()`/`Rendezvous()`?

britram:
paging @tfpauly to discuss the pros and cons of the Apple and Linux approaches to TFO...

tfpauly:
There are three top-level approaches that we’ve used/tried:
1. Mark fast open in parameters (Preconnection) such that connect() (Initiate) becomes a fake action, and you then send after that, and the first send kicks off the real Initiate. 
2. Add your initial fast-open data to the parameters (Preconnection), such that it’s already known by the time Initiate is called. 
3. Allow marking sends as idempotent, and allow enqueuing sends after create and before start. 

(1) gets tricky if you want to monitor when connections actually complete, because they’re marked as Connected/Ready too early. This is a bad application contract, especially when there’s racing involved. 

(2) is not great since it requires an entirely new way of specifying data that isn’t Send(). 

Neither (1) nor (2) clearly support sending multiple chunks of idempotent data into the stack as fast open data. 

Thus, we chose (3). 

However, Preconnection has certainly messed this up a bit. My suggestion is to no longer take Preconnection as an argument to Initiate(), but instead use a NewConnection(Preconnection) -> Initiate() pattern, which you can insert Sends in between. There are many other reasons to do this as well—if I want to set my callbacks for events for this connection, etc, in languages like C that’s a lot easier if I can call create() and then set my event handlers, and then start once those are set. 

Note: I already removed the text in the architecture that referred to Preconnection being passed to calls like Initiate and Listen, since that was API, not architecture, and since I didn’t quite agree :)

csperkins:
So you create a `Connection` from a `Preconnection`, then call `Initiate()` on the `Connection`? That would work, provided we're clear when the parameters of that connection lock in (when it's created or when it's initiated)

tfpauly:
Yes, the parameters from the Preconnection are immutable once the Connection is created.

csperkins:
[...]

PR #124, opened 2018-02-28
Handling of idempotent send on initiate

britram:


britram:
Will fix #112 when it lands.

britram:
so this update is definitely better than the version before it, but it seems like there should be some way to make the `Start()` phase optional. Otherwise, "I `Initiate()`d, but forgot to `Start()`" seems like the most common programmer error that this interface will elicit.

tfpauly:
I think the concepts outlined here work well; semantically, language implementations that work well for the model can certainly do a InitiateAndStart() equivalent. We *could* imagine changing some of the naming to make the programmer error less easy to fall into, like replacing Initiate() with Allocate/Create|Outbound/Initiator(), and replacing Listen() with Allocate/Create|Inbound/Listener().

britram:
First, please don't land this yet; I wanted this in PR form to be able to discuss concrete text, not because I'm at all convinced it's not terrible.

Second follows.

britram:
Second:

>  ...can certainly do a InitiateAndStart() equivalent...

ugh ugh ugh ick ack thpbbbt no.

Splitting this into three steps (put together some parameters, which might already be doing some preracing stuff in the background; instantiate that set of parameters i.e. while telling it what kind of thing it's going to be, then actually telling that thing to go) is very easy to implement but IMO exposes too much of the internals of the process we envision to the application. To try and elaborate on my comment last night... we have a paradox here:

1. We would like the system to be able to do things that are Good and Cheap and Fast for the application on its own, without the application specifically asking for them.

2. This particular thing, Good and Cheap and Fast via 0RTT, requires that the application give the data it would like to initiate with earlier than it otherwise would; indeed, if latency is important, it has to hand over the data it would like to initiate with *before the system might conceivably make a decision to start racing*.  This is *only* useful for initiation (I still need to think about rendezvous) -- listeners only have to be willing to handle 0RTT initiation. 

I can't think of any other thing the application might want the system to do on its behalf that reverses the order of operations like this, so 0RTT really is special. The question is how much do we bend the API for this specialness, given that it's actually kind of important for performance.

There seem to me to be four (or two and two halves) approaches for moving forward:

1. This PR or something like it: split all the steps out, make the application go through all of them in order, provide syntactic sugar for the "I really don't care about 0RTT" case. The downside of this is that applications will then use that syntactic sugar, and will have to be mildly tweaked to use 0RTT in case they decide they want to use it in the future. (IMO this will reduce the deployment of 0RTT to applications built to APIs like this one).

2. (or really, 1.5): This PR, with the operation that the application will perform on a Preconnection embedded into that Preconnection at instantiation time; i.e. you have Preinitiations and Prelisteners and Prerendezvous; or, with less terrible naming, a single Preconnection type that can be bound to a later operation by `PreconnectionForInitiation()`, `NewPreconnectionForListening()`, and so on. I still don't like this much but it seems to reduce the three-step thing to a two-step thing (i.e., `NewPreconnectionForInitiation()` does basically what `Preconnect()` and `Initiate()` do in this PR, and `Initiate()` does what `Start()` does.)

3. Handle this via a configuration directive. This would work something like a new semantic attached to the "I want 0RTT" protocol selection property, stating that when 0RTT is requested and available, an Initiate() call actually *delays* `Initiate()` until the first `Send()`. This complicates the semantics of the API a little (and can complicate implementation quite a bit; i.e. I can't really see how to race a 0RTT-capable transport against a non-0RTT capable transport in this case). It also has the problem that you can't override the delayed initiation without playing protocol selection games (which we'd like to discourage) and you can't express that you really, really want five messages in the first 0RTT flight.

4. (3.5) as 3, but with two initiate calls: `InitiateNow()` which will override any delay (i.e., I might have 0RTT but know I'm connecting to a server which will send me a banner first, so please don't wait for my first `Send()`), and `Initiate()` with the potential delay-to-first-Send; and a new state on the Immediate send parameter which basically says "please please please coalesce this Message with the next one" and, in the case of 0RTT, extends the send delay until it sees the first Send without that state.

I'm not actually sure what the tradeoffs are here yet. They're all a little hacky, but I'm leaning toward 4 for purely external-aesthetic reasons.

In any case, I think we should ship -00 before this discussion converges. If we're fast (and don't have A Lot Of Other Stuff To Do in the two weeks to London) then we can always drop a -01 on IETF Monday. If not, we get to continue this discussion over a beer or three, which seems like it might be productive.

britram:
@csperkins was bba9a56 meant for this branch, or should I cherry-pick it over to master?

csperkins:
@britram bba9a56 was intended for here, since it assumes a `Rendezvous` object that doesn't exist in the API in master. It might make sense to cherry-pick it for the general concept though.

philsbln:
Now, that I have time to go through the issues, I want to drop an Idea on that one:

 5. Allow send() on a Pre-Connection. If 0-RTT is requested and send is idem-potent, do the send as 0-RTT if possible, otherwise buffer the send and send directly after the connection is fully established. As we do not allow to clone pre-connections anymore, it is safe to enqueue sends here.

mwelzl:
wasn’t that always the plan??   if we have moved away from it, we did it for a reason - but what was that reason?

or is this *even* earlier than what we currently have?  good, then!

britram:
For in-person discussion in Montreal; we should turn the suggestions in this PR and in #112 into a message to the list.

britram:
Can we consider this PR moot by #214?

britram:
closing because #214 landed

PR #128, opened 2018-03-01
Add text about framing

csperkins:
This is a reworking of the text from Section 4.2 of draft-trammell-taps-post-sockets-03, intended to help motivate the choice of Message-based APIs with framing for the transport services architecture. The level of detail here seems appropriate for the architecture document, rather than the API.

### DIFF
--- a/draft-pauly-taps-arch.md
+++ b/draft-pauly-taps-arch.md
@@ -339,6 +339,56 @@ The Transport System Implementation Concepts define the set of objects used inte
 
 * Endpoint Racing: Endpoint Racing is the act of attempting to establish, or scheduling attempts to establish, multiple Protocol Stacks that differ based on the specific representation of the Remote Endpoint and the Local Endpoint, such as IP addresses resolved from a DNS hostname.
 
+### Message Framing, Parsing, and Serialization {#framing}
+
+While some transports expose a byte stream abstraction, most higher level
+protocols impose some structure onto that byte stream. That is, the higher
+level protocol operates in terms of messages, protocol data units (PDUs),
+rather than using unstructured sequences of bytes, with each message being
+processed in turn.  Protocols are specified in terms of state machines
+acting on semantic messages, with parsing the byte stream into messages
+being a necessary annoyance, rather than a semantic concern.  Accordingly,
+the Transport Services architecture exposes messages as the primary 
+abstraction.  Protocols that deal only in byte streams, such as TCP,
+represent their data in each direction as a single, long message.  When
+framing protocols are placed on top of byte streams, the messages used in
+the API represent the framed messages within the stream.
+
+Providing a message-based abstraction also provides:
+
+* the ability to associate deadlines with messages, for transports that
+  care about timing;
+
+*  the ability to provide control of reliability, choosing what messages to
+   retransmit in the event of packet loss, and how best to make use of the
+   data that arrived;
+
+* the ability to manage dependencies between messages, when some messages
+  may not be delivered due to either packet loss or missing a deadline, in
+  particular the ability to avoid (re-)sending data that relies on a previous
+  transmission that was never received.
+
+All require explicit message boundaries, and application-level framing of
+messages, to be effective.  Once a message is passed to the transport, it
+can not be cancelled or paused, but prioritization as well as lifetime and
+retransmission management will provide the protocol stack with all needed
+information to send the messages as quickly as possible without blocking
+transmission unnecessarily. The transport services architecture facilitates
+this by handling messages, with known identity (sequence numbers, in the
+simple case), lifetimes, niceness, and antecedents.
+
+Transport protocols such as SCTP provide a message-oriented API that has
+similar features to those we describe.  Other transports, such as TCP, do
+not.  To support a message oriented API, while still being compatible with
+stream-based transport protocols, implementations of the transport services
+architecture must provide APIs for framing and de-framing messages.  That
+is, we push message framing down into the transport services API, allowing
+applications to send and receive complete messages.  This is backwards
+compatible with existing protocols and APIs, since the wire format of
+messages does not change, but gives the protocol stack additional
+information to allow it to make better use of modern transport services.
+
+
 # IANA Considerations
 
 RFC-EDITOR: Please remove this section before publication.
--- a/draft-pauly-taps-arch.md
+++ b/draft-pauly-taps-arch.md
@@ -381,7 +381,7 @@ Transport protocols such as SCTP provide a message-oriented API that has
 similar features to those we describe.  Other transports, such as TCP, do
 not.  To support a message oriented API, while still being compatible with
 stream-based transport protocols, implementations of the transport services
-architecture must provide APIs for framing and de-framing messages.  That
+architecture should provide APIs for framing and de-framing messages.  That
 is, we push message framing down into the transport services API, allowing
 applications to send and receive complete messages.  This is backwards
 compatible with existing protocols and APIs, since the wire format of


draft-pauly-taps-arch.md
```diff
@@ -339,6 +339,56 @@ The Transport System Implementation Concepts define the set of objects used inte
...
+While some transports expose a byte stream abstraction, most higher level
+protocols impose some structure onto that byte stream. That is, the higher
+level protocol operates in terms of messages, protocol data units (PDUs),
+rather than using unstructured sequences of bytes, with each message being
+processed in turn.  Protocols are specified in terms of state machines
+acting on semantic messages, with parsing the byte stream into messages
+being a necessary annoyance, rather than a semantic concern.  Accordingly,
+the Transport Services architecture exposes messages as the primary 
+abstraction.  Protocols that deal only in byte streams, such as TCP,
+represent their data in each direction as a single, long message.  When
```
mwelzl: Just for the record, I don't have a problem with the "TCP represents their data as a single long message" statement per se (although I don't think it adds much value to say this) - my issue is when this translates into "thus, sending 50 small messages may mean 50 small TCP connections".  Obviously we don't want that.

Anyway - for this PR, this is fine by me!

draft-pauly-taps-arch.md
```diff
@@ -339,6 +339,56 @@ The Transport System Implementation Concepts define the set of objects used inte
...
+   data that arrived;
+
+* the ability to manage dependencies between messages, when some messages
+  may not be delivered due to either packet loss or missing a deadline, in
+  particular the ability to avoid (re-)sending data that relies on a previous
+  transmission that was never received.
+
+All require explicit message boundaries, and application-level framing of
+messages, to be effective.  Once a message is passed to the transport, it
+can not be cancelled or paused, but prioritization as well as lifetime and
```
mwelzl: "Once it is passed, it can not be cancelled or paused" - @tfpauly , is this really true in your implementation too? Else we probably shouldn't say that...

draft-pauly-taps-arch.md
```diff
@@ -339,6 +339,56 @@ The Transport System Implementation Concepts define the set of objects used inte
...
+information to send the messages as quickly as possible without blocking
+transmission unnecessarily. The transport services architecture facilitates
+this by handling messages, with known identity (sequence numbers, in the
+simple case), lifetimes, niceness, and antecedents.
+
+Transport protocols such as SCTP provide a message-oriented API that has
+similar features to those we describe.  Other transports, such as TCP, do
+not.  To support a message oriented API, while still being compatible with
+stream-based transport protocols, implementations of the transport services
+architecture must provide APIs for framing and de-framing messages.  That
```
mwelzl: "must" is pretty strong here. I can agree to having this as a "must" on the receiver side for the added value that it can give there, but I'm less sure about the sender side being a "must"

draft-pauly-taps-arch.md
tfpauly: 50 small messages *does not* mean 50 small TCP connections, and I am not advocating that anyone does that. =)

draft-pauly-taps-arch.md
tfpauly: Yeah, you can't cancel a message once you send it.

draft-pauly-taps-arch.md
tfpauly: I agree that this should be a "should", since a perfectly reasonable use case may not need messages on TCP. There *are* some long stream use cases, like a big file transfer.

tfpauly:
Well written, @csperkins!

draft-pauly-taps-arch.md
mwelzl: Well this is getting into the conversation we wanted to have in London, not before, but my concern is "must use a framer if I don't want 50 connections for 50 short messages over TCP".   What's the point of a sender-side framer anyway?  Only that you can offer common functionality in a library - but when I don't care about that, why can't it be good enough to just use "send"?

THAT is my concern about the "must"

draft-pauly-taps-arch.md
mwelzl: I thought it does, in your system, if I don't care about using framers and just use "send" ?   Anyway sure we all agree that nobody wants 50 short TCP connections  :)

draft-pauly-taps-arch.md
mwelzl: Then that text is fine I think  (and well written!)

draft-pauly-taps-arch.md
mwelzl: Either way: this text is also fine to keep. So there's only one small item left below that is a change request - the "must" / "should"

draft-pauly-taps-arch.md
britram: +1 to should here, though I still think that the "every message starts a new TCP connection" thing is so silly on its face that if someone really, really wants to implement that, it's their own damn fault and we don't have any responsibility at all to save them from themselves.

Sender-side framers are really just a convenience to allow applications to pass down an application-layer object and have it turn into a bytestream with appropriate framing that a receiver-side deframer can deframe (except in the case that the application (1) requires message semantics and (2) provides no framing, which I'd argue means it absolutely can't run over TCP anyway, TAPS or not -- i.e. this is another one of those silly things we have no duty to mention.)

mwelzl:
What's convenient about the framer? If I have messages of 100 bytes and the length is what defines my message, what's the point of requiring me to define a framer function that I give to the system? I just send, send, send, off go 3 100-byte messages. Hard to get more convenient than that  :)

I understand that the hook to a sender-side and receiver-side framer can allow you to offer a library that does framing when needed; I don't personally see this as a big win, but okay...    and I understand that forcing apps to use a receiver side de-framer can, in principle, let us do cool things on the receiver side (accessing messages out of the bytestream). But for sending this just seems to be an uncomfortable add-on.

I guess I shouldn't be arguing it as somehow it's again just syntactical sugar, and I said before that I don't care too much about syntax here. I get it, for the 100-byte messages the framer just won't do anything, and for sending I have to use Connection.Send(Framer.Frame(Message)) instead of Connection.Send(Message)... these calls could basically be renamed into SendMessage vs. SendSomethingHuge. I said before that I don't care too much about these syntactical bits.

I guess it's just that tying "Connection.Send(Message)" to "we assume your Message is big and terminates a TCP connection" is really counterintuitive to me.

mwelzl:
... and I think I answered to @britram in the wrong place  :(    sorry github

draft-pauly-taps-arch.md
csperkins: Saying “must provide APIs” is not “must use APIs”...

That said, I’m surprised by the resistance to this, since the overwhelming majority of applications have the equivalent of a framer/de-framer in them today as part of their protocol implementation. This isn’t a new feature: it’s breaking out an existing feature to make it reusable.

csperkins:
@mwelzl what’s convenient about the framer? In a C based implementation, very little. In a high level language implementation, however, it means I can `Connection.send(HttpRequestObject)` and have it convert the high-level structured data into the on the wire format automatically, using a reusable library.

britram:
no, no, that was right.

Framers can take an object by reference or value of any type, and turn them into an appropriately framed octet array. Syntactic sugar, yes, and pointless in some languages. But have a look at https://godoc.org/github.com/mami-project/postsocket -- here, you're pushing down the thing that converts application structures into wire structures into the API, so the API speaks to you in terms of your own structures (in Go, this is generally called a `Marshaler`; IIRC from Java `Serializer` is the terminology there).

If your application-internal structures are already serialized byte arrays, you're right, there is zero win here.

tfpauly:
@csperkins Framers are still very very useful in C. You can use callbacks, function pointers, etc, to help lay our your data, and doing that in the context of the networking thread improves efficiency (beyond the simpler API).

csperkins:
@tfpauly even better, then!

britram:
also:

> I guess it's just that tying "Connection.Send(Message)" to "we assume your Message is big and terminates a TCP connection" is really counterintuitive to me.

so this whole line of argument is extremely perplexing to me, and indicative of some sort of fundamental failure to communicate on this particular point. absolutely nobody is saying that this interface implies "your Message... terminates a TCP connection" *on the sender side*. I can't see where you're getting the impression otherwise in the text. 

(When receiving, yes, if the receiver has no way to deframe the stream, you get a message that never ends, as a series of partial reads that looks just like a bunch of `read(2)` calls on a stream; i.e. when it fails, it fails to something that basically works exactly like sockets with callbacks.

 I *really* suggest we table this whole discussion until London, because we don't seem to be making much progress in here, and ask you to trust me until then when I say that there is never, ever, ever a situation in which an implementation of this interface that was actually working would ever look at a single message on the sender side "oh, cool, that's done, guess I'll send a FIN then!", ever, because that would be an incredibly silly thing to do.

(where's the blink tag in github markdown? i really need the blink tag for this)

mwelzl:
On this:

> absolutely nobody is saying that this interface implies "your Message... terminates a TCP connection" on the sender side. I can't see where you're getting the impression otherwise in the text.

Sorry, there's a lack of context - I interpreted this out of private emails from Tommy, which seemed to  pretty clearly say that, if you just call "send" without providing a framer, you do send a FIN - and I just found that awkward. But heck! Clearly we all don't want silliness to happen and we want a nice clean interface, and clearly we can quickly agree on this in London. So no worries, and see y'all soon!

PR #163, opened 2018-04-11
clarify soft errors, fix #141

### LINKED ISSUE #141: informing an app about ICMP
TEXT:
      informed when an ICMP error message arrives.
- I think this is OK. but is this intended to be for soft-errors, since hard-errors break the connection? I think I get this, although some transports obvioulsy "eat" the ICMP packets they receive, and many ICMP messages have no meaning to an app, but some can be important. Tell me more please...

britram:


draft-trammell-taps-interface.md
```diff
@@ -426,11 +426,15 @@ The following properties can be used during Protocol and Path selection:
...
-  useful to be informed when an ICMP error message arrives. This property
-  applies to Connections and Connection Groups. This is not a strict
-  requirement. The default is to have this option.
+* Notification of ICMP soft error message arrival:
+  This boolean property specifies whether an application considers it useful
+  to be informed when an ICMP error message arrives that does not force
+  termination of a connection. This property applies to Connections and
+  Connection Groups. Received ICMP errors will be available as SoftErrors.
+  Note that even if a protocol supporting this property is selected, not all
+  ICMP errors will necessarily be properly routed, so applications cannot rely
```
mwelzl: "properly routed" seems strange to me; suggest to replace with "delivered"

PR #167, opened 2018-04-17
fix #138 fix #137 add refs

### LINKED ISSUE #137: TAPS Architecture - second para of section 3
The sentence reads:
   Beyond these, if the API
   for the system remains the same over time, new protocols and features
   may be added to the system's implementation without requiring
   significant changes in applications for adoption.
However, this kind of suggests that some small change is needed to apps. That's not necessarilly true, and I think the key point is the converse (i.e, 3.1 below). Can I suggest:
   Beyond these, if the API
   for the system remains the same over time, new protocols and features
   may be added to the system's implementation without requiring
   changes in applications for adoption.

### LINKED ISSUE #138: Architecture: Security Considerations - Expand acconyms
The Sec Con section contains a large number of unexpanded abbreviations. Compared to the remainder of the document this requires much more knowledge of these terms, suggest we expand them here.

britram:


britram:
seeing no comment... :)

tfpauly:
Sorry, looks good =) Busy week over here!

Issue #165, opened 2018-04-16
Fix Transport Parameters inconsistencies


philsbln:
This is a fork of Issue #153 – The part of it that does not need discussion.

 - Annotate "data type" for each Transport Parameter
 - Change "boolean" Protocol and Path Selection Parameters to type "Preference"
 - Remove editing artifacts

Issue #153, opened 2018-03-21
Restructure Transport Parameters Section


theri:
We have Protocol Selection Properties and Path Selection Properties (§5.2 of -00), General Protocol Properties and Specific Protocol Properties (§9.1), Transport Features and Path Properties (mentioned in §9), Security Properties (§5.3), Message Properties (called Send Properties in §7.1), unless I'm forgetting some. 

- We definitely need to **restructure** the API draft.
I propose that these Properties all go in the same Section (former §9, maybe move it up in the draft more).
- Perhaps we may need to **rename** some of them, to reduce confusion, and make sure our descriptions are clear enough.
- We need to make clear **when** which of these properties can be specified and when they become immutable.

abrunstrom:
We should also consider if Transport Parameters and Security Parameters are good names when we look over the terminology/structure.

philsbln:
In addition to the todos raised by @theri, we also need to fix and define data types taken by the properties.
 - Most of the `boolean` typed properties are actually preferences
 - Moving the somehow-displaced definition of preferences and making this a type definition should help

Getting types/categories of parameters right might also be a first step to an IANA registry

philsbln:
The front matter also has become tediously long as it defines a lot of things inclunding 
 - data types (cross check with #173) 
 - how preferences work
 - interaction between path- and protocol selections

We need to better separate it

philsbln:
Rough agenda: 
1. first propose a categorization for the different kinds of parameters to the list
2. Make a PR based on the outcome

theri:
Philipp and I just discussed the following regarding the conceptual organisation of "Properties" or “Parameters”. The goal to reduce complexity and to avoid confusion:

- Everything should be named either "Properties" or "Parameters". We should agree on one of these synonyms, not mix them. We prefer "Properties".

- Properties apply on a Connection _logical or_ a Message. So we have Connection Properties and Message Properties (see #146 for a discussion - The "Send Parameters” in the current API would then be renamed to Message Properties (and some may apply on partial messages).

- Open question: As some Properties might applicable to both a Connection and a Message, @philsbln sees the distinction between (Selection Properties/Intents/Protocol Properties) and between (Message Properties/Connection Properties) as orthogonal . @theri thinks that Selection Properties, Intents, and Protocol Properties are Connection Properties, and that Message Properties should be a category separate from them, as per-Message Path Selection is out of scope for this draft.

[...]

Issue #35, opened 2018-02-12
Rewrite Interface Types to be useful and implementable


britram:
We should be a bit more specific about how an application is supposed to interact with path selection properties. Do we want to bind this to a world in which the system specifies a set of interface type labels and the application has to choose, or is there some other approach here?

tfpauly:
So, one approach is to indeed have enumerations of path types and services that are allowed. To give the example of how we do this:

- Require/Prohibit/etc "interface name" like "en0". This sucks, but is on occasion necessary if you need to be really specific as an application.

- Require/Prohibit/etc "interface type", which is "Wi-Fi", "Wired Ethernet", "Cellular", "Loopback"; with subtypes for "Peer-to-Peer Wi-Fi", etc.

- Require/Prohibit/etc "service", "PvD" or what we call a "network agent". This can be a class of a network service like "VPN", "Proxy",  etc, or a specific instance of a VPN, etc.

That last category allows a fairly extensible set of options. I think in general, this area should take lessons from the PvD work. See: https://tools.ietf.org/html/draft-kline-mif-mpvd-api-reqs-00

britram:
do we need to assign someone to write text for this for the -00 milestone, or can we come back to this after London?

philsbln:
I am fine with moving this to post-London

gorryfair:
I think we need to address what I think is a basic issue in the API draft. The current text seems to suggest the App signals specific characteristics of the App across the API. I have problems with this - the experience from RSVP, etc - is that Apps seldom understand how to express themselves regarding their own behaviour. Instead, I would advocate that we explicitly focus on what they want from the network path: 

Examples I suggest are helpful include: an abstract treatment based on a PHB; a minimum PMTU; a preference for lower latency rather than throughput; a need for no cost... etc.
Examples I would argue against: A need for a specific transmission rate; understanding of the size of a transfer at the time the connection is made; ... etc.

tfpauly:
+1 for specifying requirements of paths and protocols, rather than properties of an application. Any time we've attempted to implement the items you mention as the "argue against", they've been dead ends that have either been misused or deprecated.

philsbln:
Lets look how many use cases for this stay when you add support for requesting explicit pvds

britram:
#181 closed this

PR #168, opened 2018-04-17
Transport Parameter Fixes for -00

philsbln:
This commit fixes bug #165 by making transport parameter types explicit.

It also fixes  Path- and Protocol Selection Properties from "binary + preference" to preferences type.
With the old boolean properties, one could do things like "prohibit Reliable Transport false" - is that the same as "prefer Reliable Transport true"?

### DIFF
--- a/draft-trammell-taps-interface.md
+++ b/draft-trammell-taps-interface.md
@@ -338,14 +338,11 @@ selection parameters, as well as Generic and Specific Protocol Properties for
 configuration of the detailed operation of the selected Protocol Stacks.
 
 All Transport Parameters are organized within a single namespace shared with
-Send Parameters (see {{send-params}}). All transport parameters take
-paremeter-specific values. Protocol and Path Selection properties additionally
-take one of five preference levels, though not all preference levels make
-sense with all such properties.
+Send Parameters (see {{send-params}}). These transport parameters take
+values of paremeter-specific types.
 
 Note that it is possible for a set of specified transport parameters to be
-internally inconsistent, or for preferences to be inconsistent with the later
-use of the API by the application. Application developers can reduce
+internally inconsistent, or to be inconsistent with the later use of the API by the application. Application developers can reduce
 inconsistency by only using the most stringent preference levels when failure
 to meet a preference would break the application's functionality (e.g. the
 Reliable Data Transfer preference, which is a core assumption of many
@@ -360,9 +357,9 @@ endpoints that support only specific network addresses or transports,
 transport protocol selection is necessarily tied to path selection. This may
 involve choosing between multiple local interfaces that are connected to
 different access networks.
-
-To reflect the needs of an individual Connection, they can be
-specified with five different preference levels:
+Conflicts between protocol and path selection properties should be resolved if possible in a way  that path selection properties take precedence.
+The type of most  Protocol and Path Selection properties is "preference" and has
+with five different preference levels:
 
    | Preference | Effect                                                             |
    |------------|--------------------------------------------------------------------|
@@ -377,8 +374,6 @@ and path selection properties. The defaults given for each property below
 represent a configuration that can be implemented over TCP. An alternate set of
 default Protocol Selection Properties would represent a configuration that can
 be implemented over UDP.
-
-
 
 All transport parameters used in the pre-establishment phase are collected
 in a TransportParameters Object that is passed to the Preconnection Object.
@@ -425,7 +420,9 @@ The following properties can be used during Protocol and Path selection:
 
 ### Reliable Data Transfer {#prop-reliable}
 
-This boolean property specifies whether the application wishes to use a
+Type: Preference
+
+This property specifies whether the application wishes to use a
 transport protocol that that provides mechanisms to help ensure that all data
 is received and without corruption on the other side. This also entails being
 notified when a Connection is closed or aborted. This property applies to
@@ -434,7 +431,9 @@ is to enable Reliable Data Transfer.
 
 ### Preservation of data ordering {#prop-ordering}
 
-This boolean property specifies whether the application wishes to use a
+Type: Preference
+
+This property specifies whether the application wishes to use a
 transport protocol that provides mechanisms to ensure that data is received
 by the application on the other end in the same order as it was sent. This
 property applies to Connections and Connection Groups. This is a strict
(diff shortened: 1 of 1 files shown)


draft-trammell-taps-interface.md
```diff
@@ -338,14 +338,11 @@ selection parameters, as well as Generic and Specific Protocol Properties for
 configuration of the detailed operation of the selected Protocol Stacks.
 
 All Transport Parameters are organized within a single namespace shared with
-Send Parameters (see {{send-params}}). All transport parameters take
-paremeter-specific values. Protocol and Path Selection properties additionally
-take one of five preference levels, though not all preference levels make
-sense with all such properties.
+Send Parameters (see {{send-params}}). These transport parameters take
+values of paremeter-specific types.
```
britram: s/paremeter/parameter/

draft-trammell-taps-interface.md
```diff
@@ -458,22 +461,28 @@ is to not have this option.
...
+
+This property specifies that the application would prefer multiple
 Connections within a Connection Group to be provided by streams of a single
 underlying transport connection where possible. This is not a strict
 requirement. The default is to not have this option.
 
 ### Notification of excessive retransmissions {#prop-retrans-notify}
 
-This boolean property specifies whether an application considers it
+Type: Boolean
```
britram: I'm not sure how this is supposed to work. Is true == Require or Prefer; is false == Avoid or Prohibit? How do I ignore?

I think this might be a Preference, though not one that most apps will use.

draft-trammell-taps-interface.md
```diff
@@ -458,22 +461,28 @@ is to not have this option.
...
+This property specifies whether an application considers it
 useful to be informed in case sent data was retransmitted more often than
 a certain threshold. This property
 applies to Connections and Connection Groups. This is not a strict
 requirement. The default is to have this option.
 
 ### Notification of ICMP soft error message arrival {#prop-soft-error}
 
-This boolean property specifies whether an application considers it useful
+Type: Boolean
```
britram: As above.

draft-trammell-taps-interface.md
```diff
@@ -494,7 +505,9 @@ requiring a checksum when receiving.
 
 ### Interface Type {#prop-intf-type}
 
-This enumerated property specifies which kind of access network interface,
+Type: Enumeration
```
britram: An Enumeration is a set of (Type, Preference) tuples, correct? We might want to say that somewhere.

draft-trammell-taps-interface.md
```diff
@@ -504,7 +517,9 @@ interface configured in the system policy.
 
 ### Capacity Profile {#prop-cap-profile}
 
-This enumerated property specifies the application's expectation of the
+Type: Enumeration
```
britram: As above.

draft-trammell-taps-interface.md
philsbln: Oh - that is not what I wanted to state here – my idea was just to use this as "prefer interface type", e.g., Wifi.

However, I like your Idea of making that a tuple much better.

draft-trammell-taps-interface.md
philsbln: Good question – I thought of it only as boolean whether to fire an event when available these notifications are available.
If we assume there are applications that do not work without this kind of notification, we have to make this a Preference, otherwise, I think it is a Boolean.

draft-trammell-taps-interface.md
philsbln: Same Comment as above ;-)

draft-trammell-taps-interface.md
britram: IMO we should ask a slightly different question here, not "can we think of an application that would really want to prohibit selection of a transport based on the inability to see RTX notifications or ICMP soft errors", but "do we have a good reason for creating a third class of selection parameter when two suffice"? If True = "prefer" and False = "ignore", then we can just say that in the docs rather than specifying additional codepaths (every codepath, remember, is another way you can accidentally screw up :) )

draft-trammell-taps-interface.md
britram: > The application tells the Transport system the traffic is Constant Rate, not that it prefers / wants to avoid that the traffic is treated as constant rate.

How is the TAPS system's action based on this information different, though?

draft-trammell-taps-interface.md
philsbln: I don't see a functional difference between preference with two valid levels and boolean, just, from an application programmers perspective, I still think that "Boolean" is less confusing here.

I would like to get at least a second opinion from @tfpauly on that -- after that, I change it to the one that makes this change land for -00.

draft-trammell-taps-interface.md
philsbln: This case is different from the Interface Type case:
The Interface type case needs a parameter in addition to the preference.
For the Capacity Profile, the preference makes no difference here, adding it is redundant and mildly confusing. So I prefer it being just a Enum of the envisioned capaity profiles.

philsbln:
General clarification:
 - This pull request changed "preference" to a data type a value can have rather than being something weakly defined that can be specified along with every Transport Parameter.
   So we go from a Key / Preference / Value tripel where Preference or Value is pointless for (n-1) cases to a Key / Value pair with typed values. 
 - The only Transport Parameter (including the Intents in the Appendix) that uses a parameter and a preference is the "Interface Type".
 - I would like to land this for -00 as this is good enough for discussion while the old state is confusing.
 - ToDos for -01:
   - Explain types clearly
   - make  "Interface Type" easier in -01
   - discuss extendability

draft-trammell-taps-interface.md
britram: okay, that makes sense. There should be some language, either here or in Implementation, that a selection of a capacity profile may be used by the TAPS system as a "preference" for a transport that is "good" for the given capacity profile.

britram:
> I don't see a functional difference between preference with two valid levels and boolean...

If we're doing booleans, the document needs to define a concise and generally applicable rule for how booleans interact with preferences. ISTM that boolean True == preference Prefer and boolean False == preference Avoid (in which case, why not just say Prefer and Avoid instead of True and False?)

britram:
also @philsbln: Please file everything in the comments on this PR that should be addressed in -01 as an issue (if you haven't already)... thanks!

philsbln:
This should be ready for landing. – @britram please merge if you agree

Open Issues for Milestone itef-01:
 - Restructure Transport Parameters Section #153 
 - Explain data types explicitly #173
 - Rewrite Interface Types to be useful and implementable #35

Issue #149, opened 2018-03-21
Handling conflicts among path/protocol selection properties


britram:
See thread at https://mailarchive.ietf.org/arch/msg/taps/i8Ax-WqOoOg4obYU8MgBzb0F3oc

Suggested resolution: implementation must have a deterministic tiebreaker, current recommendation is that path properties take precedence. API should note expected behavior. There should be additional text on candidate selection and ordering.

tfpauly:
Give example for path candidate gathering and sorting (get selection properties, get available paths in system preferred order, walk through them cancelling out the prohibited ones, and ranking preferences; get ultimate list). Do the same for protocols likely.

theri:
Added text to API and Implementation drafts on branch "conflicting-policies".
API has an example, implementation still needs one. Feel free to add, otherwise I'll come back to it later today or tomorrow.

philsbln:
I guess that works for -00

For -01, we might want to integrate the resolution a little bit more into the decision process.

philsbln:
Resolved by PR #167

PR #169, opened 2018-04-18
Conflicting policies

theri:
Added text about what happens when an application specifies both protocol and path selection properties. Path selection properties have precedence. I added examples to both API and implementation draft to illustrate this. Relates to #149

### DIFF
--- a/draft-brunstrom-taps-impl.md
+++ b/draft-brunstrom-taps-impl.md
@@ -349,6 +349,19 @@ An implementation may use the Capacity Profile to prefer paths optimized for the
 
 \[Note: See {{branch-sorting-non-consensus}} for additional examples related to Properties under discussion.]
 
+If both WiFi and LTE are available, but the application has indicated a preference for WiFi, branches will be sorted accordly, with WiFi at the top. If the application has also indicated a preference for a feature only available in SCTP, branches with SCTP will be sorted to the top within their subtree. However, if the implementation has cached the information that SCTP is not available on the path over WiFi, there is no SCTP node in the WiFi subtree.
+Here, the path over WiFi will be tried first, and, if connection establishment succeeds, TCP will be used. So the path selection property of preferring WiFi takes precedence over the protocol selection property of preferring SCTP.
+
+~~~~~~~~~~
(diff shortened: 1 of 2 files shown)


draft-brunstrom-taps-impl.md
```diff
@@ -328,6 +328,9 @@ Branching for derived endpoints is the final step, and may have multiple layers
 Implementations should sort the branches of the tree of connection options in order of their preference rank. 
 Leaf nodes on branches with higher rankings represent connection attempts that will be raced first.
 Implementations should order the branches to reflect the preferences expressed by the application for its new connection, including Protocol and Path Selection Properties, which are specified in {{I-D.trammell-taps-interface}}. 
+Path Selection Properties take precedence over Protocol Selection Properties.
+
+If Protocol and Path Selection Properties contain any prohibited properties, the implementation should purge branches containing nodes with these properties.
```
britram: Does this mean that Prohibit takes precedence over all others? (I think it does, and I think it's correct). I think what I'd like to see here (and maybe I'm just a languages geek) is a strict ordering, e.g. something like:

`Prohibit` > `Require` > `Prefer` > `Avoid`

draft-trammell-taps-interface.md
```diff
@@ -361,6 +361,13 @@ transport protocol selection is necessarily tied to path selection. This may
 involve choosing between multiple local interfaces that are connected to
 different access networks.
 
+In case of conflicts between protocol and path selection properties,
+path selection properties take precedence.
+For example, if an application indicates a preference for a specific path, but
+also a preference for a protocol not available on this path, the transport
+system will try the path first, so the protocol selection property might not
+have an effect.
+
```
britram: Precedence order (as above) should IMO appear here too...

draft-brunstrom-taps-impl.md
theri: Makes sense. I added more text to the implementation that details how this should work.

draft-trammell-taps-interface.md
theri: Agreed, I added some text here too, and moved it below the table. This section should perhaps be restructured in the future.

Issue #136, opened 2018-03-05
Re-add batching Send operations to API and Implementation


tfpauly:
In aligning the implementation with the API for the initial draft, we removed this text since we still need text in the API:

#### Batching Sends

Since sending a Message may involve a context switch between the application and the transport system, sending patterns that involve multiple small Messages can incur high overhead if each needs to be enqueued separately. To avoid this, the application should have a way to indicate a batch of Send actions, during which time the implementation will hold off on processing Messages until the batch is complete. This can also help context switches when enqueuing data in the interface driver if the operation can be batched.

gorryfair:
This seems equivalent to the idea of PUSH in existing systems and would make sense for lots of short write calls.

britram:
+1.

The higher-level question here is how to support high-performance applications. Send batching is one thing you can do here. Receive batching would be another obvious one. Zero-copy is another, though that's probably so tied to implementation and platform that it's not useful to standardize. But there should be some way for this interface to coexist with it...

mwelzl:
+1 from me too

philsbln:
+1 

In my opinion, support for high-performance applications is nothing we can add to the API, but rather something we must sure to not prevent or make hard to implement.
So we should ask ourselves for every point in the API "can this be implemented zero-copy" or "can an implementation batch this" – if the answer is no, we have to fix that.

britram:
this one seems to have consensus and does not seem to be too much work. @tfpauly, if you can get a PR in in the next week or so let's ship it with ietf-00...

tfpauly:
I'll try to get this in soon!

britram:
fixed with merge of #171

PR #171, opened 2018-04-18
Add back batching sends

tfpauly:
See issue #136 

The one question I have here is the semantics around batching. In our implementation, we use block semantics (from clang), to batch looks like:

connection_batch(^{
    connection_send(...);
    connection_send(...);
});

where the block is called inline, and the batch is over when the block returns. I'm not sure what the right style for the pseudo-code is for the draft.


draft-brunstrom-taps-impl.md
```diff
@@ -479,6 +479,10 @@ The effect of the application sending a Message is determined by the top-level p
 
 The application should be notified whenever a Message or partial Message has been consumed by the Protocol Stack, or has failed to send. The meaning of the Message being consumed by the stack may vary depending on the protocol. For a basic datagram protocol like UDP, this may correspond to the time when the packet is sent into the interface driver. For a protocol that buffers data in queues, like TCP, this may correspond to when the data has entered the send buffer.
 
+#### Batching Sends
+
+Since sending a Message may involve a context switch between the application and the transport system, sending patterns that involve multiple small Messages can incur high overhead if each needs to be enqueued separately. To avoid this, the application should have a way to indicate a batch of Send actions, during which time the implementation will hold off on processing Messages until the batch is complete. This can also help context switches when enqueuing data in the interface driver if the operation can be batched.
+
```
britram: Yep, this is good.

draft-trammell-taps-interface.md
```diff
@@ -984,6 +984,21 @@ The following values are valid for Transmission Profile:
...
+application may want to batch several Send actions together. This provides a hint to
+the system that the sending of these Messages should be coalesced when possible,
+and that sending any of the batched Messages may be delayed until the last Message
+in the batch is enqueued.
+
+~~~
+Connection.Batch(
+	Connection.Send(Message, sendParameters)
+	Connection.Send(Message, sendParameters)
+)
```
britram: I'd almost leave out the pseudocode, but this is fine for now. Like events, the implementation platform is going to almost completely determine the syntax. In Go, I'd probably do something like:

```go
func (c Connection) Batch (sendfunc func());

var c Connection
c.Batch(func() {
    c.Send(...)
    c.Send(...)
})
```

which is almost the same thing but makes the closure explicit (I presume the Swift thing has a closure behind it anyway).

Another pattern I've seen (usually with RDBMS drivers and such) is to have a `Batch` object that has a `Send` method and a `Commit` method.

In C you'd probably end up with something like `sendv(Connection, messages...)`.

PR #181, opened 2018-05-27
Rework text for interface types

tfpauly:
Implements #35

### DIFF
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -91,6 +91,7 @@ normative:
 
 informative:
   I-D.ietf-taps-transport-security:
+  RFC7556:
 
 --- abstract
 
@@ -521,18 +522,54 @@ strict requirement, as it signifies a reduction in reliability. The
 default is full checksum coverage without being able to change it, and
 requiring a checksum when receiving.
 
-### Interface Type {#prop-intf-type}
+### Interface Instance or Type {#prop-interface}
 
 Type: Tuple (Enumeration, Preference)
 
-This property specifies which kind of access network interface,
-e.g., WiFi, Ethernet, or LTE, to prefer over others for this Connection, in
-case they are available. In general, Interface Types should be used only with
-the `Prefer` and `Prohibit` preference level. Specifically, using the
-`Require` preference level for Interface Type may limit path selection in a
-way that is detrimental to connectivity. The default is to use the default
-interface configured in the system policy.
-The valid values for the access network interface kinds are implementation specific.
+This property allows the application to select which specific network interfaces
+or categories of interfaces it wants to `Require`, `Prohibit`, or `Prefer`.
+
+If a system supports discovery of specific interface identifiers, such as `en0` or `eth0`
+on Unix-style systems, an implemention should allow using these identifiers to
+define path preferences. Note that marking a specific interface as `Required`
+strictly limits path selection to a single interface, and leads to less flexible and
+resilient connection establishment.
+
+The set of valid interface types is implementation- and system-specific. For example,
+on a mobile device, there may be `Wi-Fi` and `Cellular` interface types available;
+whereas on a desktop computer, there may be `Wi-Fi` and `Wired Ethernet`
+interface types available. Implementations should provide all types that are supported
+on some system to all systems, in order to allow applications to write generic code.
+For example, if a single implementation is used on both mobile devices and desktop
+devices, it should define the `Cellular` interface type for both systems, since an
+application may want to always `Prohibit Cellular`. Note that marking a specific 
+interface type as `Required` limits path selection to a small set of interfaces, and leads
+to less flexible and resilient connection establishment.
+
+The set of interface types is expected to change over time as new access technologies
+become available.
+
+### Provisioning Domain Instance or Type {#prop-pvd}
+
+Type: Tuple (Enumeration, Preference)
+
+Similar to interface instances and types {{prop-interface}}, this property allows
+the application to control path selection by selecting which specific Provisioning Domains
+or categories of Provisioning Domains it wants to  `Require`, `Prohibit`, or `Prefer`.
+Provisioning Domains define consistent sets of network properties that may be more
+specific than network interfaces {{RFC7556}}.
+
+The indentification of a specific Provisioning Domain (PvD) is defined to be implementation-
+and system-specific. For example, this may be a string name or an integer. As with 
+requiring specific interfaces, requiring a specific PvD strictly limits path selection.
+
+Categories or types of PvDs are also defined to be implementation- and system-specific.
+These may be useful to identify a service that is provided by a PvD. For example, if an application
+wants to use a PvD that provides a Voice-Over-IP service on a Cellular network, it can use
+the relevant PvD type to require some PvD that provides this service, without needing to
+look up a particular instance. While this does restrict path selection, it is more broad than
+requiring specific PvD instances or interface instances, and should be preferred over those
+options.
 
 ### Capacity Profile {#prop-cap-profile}
 
@@ -1204,7 +1241,7 @@ Connection properties include:
 
 * Path Properties of the path(s) in use, once the Connection has been
   established. These properties can be derived from the local provisioning
-  domain, measurements by the Protocol Stack, or other sources. They can only
+  domain {{RFC7556}}, measurements by the Protocol Stack, or other sources. They can only
   be queried.
 
 ## Protocol Properties {#protocol-props}
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -527,7 +527,7 @@ requiring a checksum when receiving.
 Type: Tuple (Enumeration, Preference)
 
 This property allows the application to select which specific network interfaces
-or categories of interfaces it wants to `Require`, `Prohibit`, or `Prefer`.
+or categories of interfaces it wants to `Require`, `Prohibit`, `Prefer`, or `Avoid`.
 
 If a system supports discovery of specific interface identifiers, such as `en0` or `eth0`
 on Unix-style systems, an implemention should allow using these identifiers to
@@ -549,18 +549,24 @@ to less flexible and resilient connection establishment.
 The set of interface types is expected to change over time as new access technologies
 become available.
 
+Interface types should not be treated as a proxy for properties of interfaces such as
+metered or unmetered network access. If an application needs to prohibit metered
+interfaces, this should be specified via Provisioning Domain attributes {{prop-pvd}}
+or another specific property.
+
 ### Provisioning Domain Instance or Type {#prop-pvd}
 
 Type: Tuple (Enumeration, Preference)
 
 Similar to interface instances and types {{prop-interface}}, this property allows
 the application to control path selection by selecting which specific Provisioning Domains
-or categories of Provisioning Domains it wants to  `Require`, `Prohibit`, or `Prefer`.
+or categories of Provisioning Domains it wants to  `Require`, `Prohibit`, `Prefer`, or `Avoid`.
 Provisioning Domains define consistent sets of network properties that may be more
 specific than network interfaces {{RFC7556}}.
 
 The indentification of a specific Provisioning Domain (PvD) is defined to be implementation-
-and system-specific. For example, this may be a string name or an integer. As with 
+and system-specific, since there is not a portable standard format for a PvD identitfier.
+For example, this identifier may be a string name or an integer. As with 
 requiring specific interfaces, requiring a specific PvD strictly limits path selection.
 
 Categories or types of PvDs are also defined to be implementation- and system-specific.


draft-ietf-taps-interface.md
```diff
@@ -521,18 +522,54 @@ strict requirement, as it signifies a reduction in reliability. The
...
+on Unix-style systems, an implemention should allow using these identifiers to
+define path preferences. Note that marking a specific interface as `Required`
+strictly limits path selection to a single interface, and leads to less flexible and
+resilient connection establishment.
+
+The set of valid interface types is implementation- and system-specific. For example,
+on a mobile device, there may be `Wi-Fi` and `Cellular` interface types available;
+whereas on a desktop computer, there may be `Wi-Fi` and `Wired Ethernet`
+interface types available. Implementations should provide all types that are supported
+on some system to all systems, in order to allow applications to write generic code.
```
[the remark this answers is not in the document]
tfpauly: No, we decided on the call that a registry is definitely not the way to go—I'm saying here that any given implementation likely has a list of types it supports, and it should make all types it knows about available across various platforms to make the code as reusable as possible.

draft-ietf-taps-interface.md
```diff
@@ -521,18 +522,54 @@ strict requirement, as it signifies a reduction in reliability. The
...
+For example, if a single implementation is used on both mobile devices and desktop
+devices, it should define the `Cellular` interface type for both systems, since an
+application may want to always `Prohibit Cellular`. Note that marking a specific
```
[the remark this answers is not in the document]
tfpauly: Mmm, it's *almost* the same, but mentions specific interface "type", and notes that it's not quite as strict as a specific interface.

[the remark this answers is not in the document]
chris-wood: Sorry, right, my point was that maybe we could factor this out into a single paragraph that covers interfaces and types at the end of this section... or not. Either way works.

draft-ietf-taps-interface.md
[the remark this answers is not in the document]
chris-wood: Right, that's what I thought, but the text makes it seem like *different* implementations might share these same types. You do mention that these are implementation specific, yet there's no accompanying example like there is for the system-specific types. Maybe we could just add one such example?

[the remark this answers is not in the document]
mwelzl: I think we should be more concrete and more prescriptive for the sake of portability. I'd replace "Implementations should provide all types that are supported on some system to all systems, in order to allow applications to write generic code." with 'For the sake of application portability, all implementations should provide at least the types "Wi-Fi", "Cellular", "Wired Ethernet".'   - these types are pretty generic anyway, what's the harm in always offering them?  I'm afraid that a relaxed statement like the one that's now here is not going to lead to a coherent set across systems (because the point of reference are only other implementations, not this draft!), and so the idea of application portability is lost.

mwelzl:
@britram - I don't think I'll fight for this one, but to begin with, tell me: what do you think is wrong about my view?

draft-ietf-taps-interface.md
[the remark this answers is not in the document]
mwelzl: I can see that. But what about my counter-argument that saying "Implementations should provide all types that are supported on some system to all systems" isn't ever going to give us enough coherence across implementations that application code is going to really be portable?  What's a better alternative?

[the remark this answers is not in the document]
britram: I'm not sure we need a common interface type system for portability here, because in most case the specific semantics of whether a given application should or should not use a given transport/interface are administrative- or user-level choices, not application developer choices.

In other words, a system would provide a way for an application to ask a user (or, as in Apple's case, provide a system UI to ask the user) which apps should work with which interfaces. This is probably an implementation-level thing though, as it might also tie into other specifics of the platform...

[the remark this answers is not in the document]
mwelzl: I don't understand what this has to do with hard-coding L2-style names or not. Indeed, admin choices will tell us which interfaces can be used, but how can you write an application and port it between systems if you have no common name to use for the WiFi interface?  But as I said, of course I also understand your other argument, that hardcoding L2 names isn't going to be great for all time in the future. I just wanted to raise the concern, I think it would the better compromise to hard-code these names after all, but I'm not going to fight for a paricular choice here. My view is that Tommy should read our discussion and then decide, and let's go for what he picks.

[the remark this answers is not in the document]
britram: > but how can you write an application and port it between systems if you have no common name to use for the WiFi interface?

I guess I'm having a hard time understanding why a (non-network-admin-tool) would want, as part of its application logic, as opposed to in response to some user configuration demand, to define a constraint on interface types. I can think of one such constraint: having two connections, one "essential" and one "higher-bandwidth", and constraining the latter to cheaper connections, but that's really a higher-level policy which (1) requires user input and (2) probably should be expressed in higher-level goals than "wifi or not"...

I think this is a larger discussion (and indeed is kind of the problem PvD is meant to solve)... for this particular issue, though, I think what's here is good. I'm not going to fight awfully hard either, so +1 to making this @tfpauly's problem :)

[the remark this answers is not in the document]
mwelzl: Yes, +1 on this being difficult to understand - I was also just beginning to think along the same lines, that we should abstract from "WiFi" into "cheap" or whatever. Anyway, we agree: this is in @tfpauly 's hands - case closed, problem solved   :-D

tfpauly:
Thanks for the hearty discussion!

On this point, I do agree with @britram that we're not going to achieve full portability across systems for interfaces and PvDs, and it likely doesn't need to be a goal.

If I'm writing an implementation in ten years for some new system that only uses Wi-Fi and [INSERT FANCY NEW RADIO TYPE HERE], why would I need to add support for this wired ethernet thing that I don't have any devices for?

Since moving between different implementations isn't likely going to be 100% portable anyhow due to language differences and symbol differences, applications will need to change. However, a specific implementation that is portable (like NEAT, or like Apple's if we make it cross-platform) should endeavor to offer the same types across systems.

mwelzl:
Fine by me, go ahead

draft-ietf-taps-interface.md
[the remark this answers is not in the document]
theri: I read this part similar to @chris-wood, that different implementations would have to provide the same types, or even provide *all* types that any other implementation out there provides. If it means "an implementation that can run on multiple platforms should provide the same types on all these platforms", that's fine with me, but it should be made a little more clear in the text.

I think generally it's good to offer more types and live with the fact that not all types are going to be available on every system an implementation is running on. An implementation will still know what types are actually available on this particular system at run time. If the application asks to "Require" something that isn't there, the implementation will error out early according to Section 3.1 in the Implementation Draft. Maybe this is worth another sentence in the "types" paragraph - a "Require" on a particular interface type will also restrict the selection, maybe unnecessarily.

Imo this kind of interface selection is often about assumptions that an application developer has about a particular type of access network - like "WiFi = cheap" as @mwelzl was saying, and as @britram probably meant by "higher-level goal". I agree with them that for such use-cases we should abstract from specific interface choice and encourage applications to request what they actually want from the network (e.g., cheap), instead of relying on the developer's assumptions, which might be wrong. This is what we are trying to achieve with Intents, too.

draft-ietf-taps-interface.md
```diff
@@ -521,18 +522,54 @@ strict requirement, as it signifies a reduction in reliability. The
...
+The set of interface types is expected to change over time as new access technologies
+become available.
+
+### Provisioning Domain Instance or Type {#prop-pvd}
+
+Type: Tuple (Enumeration, Preference)
+
+Similar to interface instances and types {{prop-interface}}, this property allows
+the application to control path selection by selecting which specific Provisioning Domains
+or categories of Provisioning Domains it wants to  `Require`, `Prohibit`, or `Prefer`.
```
theri: Why not "Avoid"? I would add that one here.

philsbln: I agree with @theri – maybe one might want to avoid using the infrastructure of a competitor, but not prohibit it.

draft-ietf-taps-interface.md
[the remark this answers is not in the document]
philsbln: I can live with a spirit of the text @tfpauly wrote  – But I guess the Interface Type or Instance should only be a corner case anyway. It is implementation specific and I guess it will have surprising results.
What should the system do, if one specifies "prohibit cellular" but it knows that the uplink for the Wifi interface is also cellular?

We should add a warning against using "Wifi" as a representant for "cheap" or "Cellular" as a representant for "expensive" – for this use case, we should use something like the cost preference Intent.

draft-ietf-taps-interface.md
```diff
@@ -521,18 +522,54 @@ strict requirement, as it signifies a reduction in reliability. The
...
+Provisioning Domains define consistent sets of network properties that may be more
+specific than network interfaces {{RFC7556}}.
+
+The indentification of a specific Provisioning Domain (PvD) is defined to be implementation-
+and system-specific. For example, this may be a string name or an integer. As with 
+requiring specific interfaces, requiring a specific PvD strictly limits path selection.
+
+Categories or types of PvDs are also defined to be implementation- and system-specific.
```
philsbln: - We should have at least a portable way to specify PvD strings. 
- Maybe we should also have a portable way to map 3GPP APNs tp PvDs (but most probably out of scope for this document)
Other stuff can indeed be implementation specific.

draft-ietf-taps-interface.md
```diff
@@ -521,18 +522,54 @@ strict requirement, as it signifies a reduction in reliability. The
...
+These may be useful to identify a service that is provided by a PvD. For example, if an application
+wants to use a PvD that provides a Voice-Over-IP service on a Cellular network, it can use
+the relevant PvD type to require some PvD that provides this service, without needing to
+look up a particular instance. While this does restrict path selection, it is more broad than
+requiring specific PvD instances or interface instances, and should be preferred over those
+options.
```
philsbln: Is this the right place to request such a service? I have no idea how such a PvD Type specifier could look like…

draft-ietf-taps-interface.md
tfpauly: Sure, we can add avoid. Many use cases for that.

draft-ietf-taps-interface.md
tfpauly: There is no portable definition of a PvD string, and we certainly aren't going to go into the business of creating APN registries...

draft-ietf-taps-interface.md
tfpauly: Yup, we use this a lot.

In your APN example, the PvD type is like "Cellular : VVM" for visual voicemail. The Instance is a specific instance of a PvD, which there may be multiple of a type on a device.

draft-ietf-taps-interface.md
philsbln: Oh snap… RFC7558 really does not define any format… If we take the assumption that PvDs are FQDNs from draft-bruneau-pvd, it would be so much easier, as we could be sure the PvD does not contain a colon.
This would enable clearly sepatating explicit PvDs (without colon) from PvD classes (with colon)

Reverring to a specific APN could be realized with a APN class: `APN:[apn name]`

draft-ietf-taps-interface.md
philsbln: Ok - not that makes sense to me.

I guess we should still describe this in more detail and at least give a structure how these classes look like (see comment above)

draft-ietf-taps-interface.md
tfpauly: This document is certainly not the place to go into the debate of how to identify PvDs. This is a very long and complicated discussion, and we already have two platforms (Android and iOS) that handle them with quite different identifiers. Thus, the text here represents the reality that these will indeed need to be system-specific.

draft-ietf-taps-interface.md
tfpauly: Added text to explain why the PvD identifier isn't strongly defined.

draft-ietf-taps-interface.md
tfpauly: Added text to explain why the PvD identifier isn't strongly defined.

draft-ietf-taps-interface.md
tfpauly: Added avoid

draft-ietf-taps-interface.md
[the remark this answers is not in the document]
tfpauly: Added a comment warning against using interface types as a proxy for metered and unmetered networks.

tfpauly:
@theri @philsbln please take a look at the update!

@mwelzl, I assume based on your last comment that you're okay with the change going in now?

mwelzl:
Yes, I agree with it!

draft-ietf-taps-interface.md
```diff
@@ -521,18 +522,60 @@ strict requirement, as it signifies a reduction in reliability. The
...
-This property specifies which kind of access network interface,
-e.g., WiFi, Ethernet, or LTE, to prefer over others for this Connection, in
-case they are available. In general, Interface Types should be used only with
-the `Prefer` and `Prohibit` preference level. Specifically, using the
-`Require` preference level for Interface Type may limit path selection in a
-way that is detrimental to connectivity. The default is to use the default
-interface configured in the system policy.
-The valid values for the access network interface kinds are implementation specific.
+This property allows the application to select which specific network interfaces
+or categories of interfaces it wants to `Require`, `Prohibit`, `Prefer`, or `Avoid`.
```
philsbln: nit: according to the definition of a preference, everything after the OR is redundant

draft-ietf-taps-interface.md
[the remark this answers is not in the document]
philsbln: I can not parse `on some system to all systems` in the last paragraph.

I think the whole last paragraph still hard to understand as it introduces a new concept without saying so – how about:

   The set of valid interface types is implementation- and system-specific and can include mnemonic types.
   On a mobile device, for example, there may be the interface types`Wi-Fi` and `Cellular` available; on a desktop computer, there may be `Wi-Fi` and `Wired Ethernet` available.
   Implementations should provide a way to enumerate interface mnemonic types order to allow applications to write generic code.

draft-ietf-taps-interface.md
```diff
@@ -521,18 +522,60 @@ strict requirement, as it signifies a reduction in reliability. The
...
+to less flexible and resilient connection establishment.
+
+The set of interface types is expected to change over time as new access technologies
+become available.
+
+Interface types should not be treated as a proxy for properties of interfaces such as
+metered or unmetered network access. If an application needs to prohibit metered
+interfaces, this should be specified via Provisioning Domain attributes {{prop-pvd}}
+or another specific property.
+
```
philsbln: I like that, but the text above still leads in another direction (how one can use interface types for that)

Maybe move that paragraph up directly after the paragraph defining the interface types and shorten the explanation in between.

draft-ietf-taps-interface.md
philsbln: I guess that works for now.

nit: maybe double-check whether there is a need to talk about `Categories` and  `Types` of PvDs

britram:
let's land this one then...

Issue #53, opened 2018-02-21
Do we need an indication that the application will not send any further data but is still willing to receive data (aka EndOfData())?


mirjak:
As discussed with PR #49 we need to decide (post-London) if we want an EndOfData() call. It was noted that this is not supported by all protocols and that TCP's close() mixes things up wrongly. Therefore EndOfData() must be separate from Close() and might not have any effect if not supported by a certain protocol.

gorryfair:
My first reaction is that close and abort are the correct primitives here. 

Drilling into to details (TCP, SCTP/UDP, etc) is unlikely to be a useful thing in this spec. If we were to specify the semantics correctly, we do not need an additional primative. My fealing is probably flavoured by the belief that applications should anyway decide at a higher level that they have finished communication before they issue close, and not rely on a callback to do this  - anything left in the transport on calling close is unlikely to make it to the endpoint.

tfpauly:
I agree with Gorry that we can just use Close and Abort as top level Connection API calls.

However, for Send(), I currently pass { Content, Metadata, isComplete, Callback } as parameters. The isComplete indicates if the Content is a full chunk for a message. When used over a streaming protocol like TCP directly, TCP can choose to interpret this as an indication to send a FIN. That replaces the notion of write-close for us.

mwelzl:
But the description "open, send, close" doesn't capture what this issue is about: "... will not send any further data but is still willing to receive data". write-close per se isn't a problem, but the "still receive" part is problematic. I also don't understand Mirja's statement that TCP's close "mixes things up wrongly" - I think TCP, and perhaps only TCP, gives you exactly what you want ("...will not send any further data but is still willing to receive"). SCTP is a counter-example, as it doesn't support half-closed connections, and the same is probably true for streams that are used as connections.

Anyway: if we want to support "will not send any further data but is still willing to receive", we need to understand how this should work. I can imagine that, with a protocol that doesn't support half-closed connections, the transport system sends, remembers (above the transport) that the application is done sending, and the connection is kept open so the application can still receive, until .... when?  Who truly closes the connection, when? The other side?  I think that makes it application specific. I think it would work like that with web servers, though.

tfpauly:
I'd like to resurrect this for discussion in -01.

tfpauly:
The place I'm currently at with how to express sending a half-close is this:
- Mark a property on the Message that it is the "final" message
- Once that Message has been completely sent, the application has indicated it is done sending on that connection
[...]

PR #183, opened 2018-05-27
Add text for Final Messages to support half-close

tfpauly:
Satisfies #53

### DIFF
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -980,7 +980,7 @@ application to receive the Message multiple times.
 
 ### Final {#send-final}
 
-Final is a boolean property. If true, this Message is the final message that
+Final is a boolean property. If true, this Message is the last one that
 the application will send on a Connection. This allows underlying protocols
 to indicate to the Remote Endpoint that the Connection has been effectively
 closed in the sending direction. For example, TCP-based Connections can
@@ -988,9 +988,9 @@ send a FIN once a Message marked as Final has been completely sent.
 Protocols that do not support signalling the end of a Connection in a given
 direction will ignore this property.
 
-Note that a final Message must always be sorted to the end of a list of Messages.
-The final property overrides Niceness and any other property that would re-order
-Messages. If another Message is sent after a Message marked as final has already
+Note that a Final Message must always be sorted to the end of a list of Messages.
+The Final property overrides Niceness and any other property that would re-order
+Messages. If another Message is sent after a Message marked as Final has already
 been sent on a Connection, the new Message will report an error.
 
 ### Corruption Protection Length {#send-checksum}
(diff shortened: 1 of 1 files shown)


draft-ietf-taps-interface.md
```diff
@@ -978,6 +978,21 @@ Send Action. It is used to mark data safe for certain 0-RTT establishment
...
+closed in the sending direction. For example, TCP-based Connections can
+send a FIN once a Message marked as Final has been completely sent. 
+Protocols that do not support signalling the end of a Connection in a given
+direction will ignore this property.
+
+Note that a final Message must always be sorted to the end of a list of Messages.
+The final property overrides Niceness and any other property that would re-order
+Messages. If another Message is sent after a Message marked as final has already
+been sent on a Connection, the new Message will report an error.
+
```
mwelzl: Capitalization of "Final" vs. "final" is inconsistent above - I suggest to write "The Final property overrides..." and "...marked as Final ..."  in the last two sentences above.

draft-ietf-taps-interface.md
```diff
@@ -1130,6 +1145,21 @@ received that reception has failed. Such conditions that irrevocably lead the
 the termination of the Connection are signaled using ConnectionError instead
 (see {{termination}}).
 
+## Receiving Final Messages
+
+The application may check a property to determine if a received Message is
+the final Message on a Connection. For any Message that is marked as final,
```
mwelzl: Suggest "marked as Final"

draft-ietf-taps-interface.md
```diff
@@ -1130,6 +1145,21 @@ received that reception has failed. Such conditions that irrevocably lead the
...
+the application can assume that there will be no more Messages received on the
+Connection once the Message has been completely delivered. This corresponds
+to the Final property that may be marked on a sent Message {{send-final}}.
+
+Final Messages are delivered by protocols that support closing one side of a Connection.
+For example, if a Connection using TCP receives a FIN, the application can determine
+that the stream is complete once the final Message has been marked complete.
+Protocols that do not support this functionality, such as UDP, may never deliver a Message
+marked as final.
+
```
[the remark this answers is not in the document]
britram: I also don't think we need to go into this level of detail *here*. You might get a Final message, but you, the application, should not rely on it. (Indeed, if the BSD API for TCP had this functionality, any peer that did the send-last-drop-state-start-RSTing-ACKS trick some CDNs use (used?) as a performance optimization would never send Final messages, even though the underlying transport supports them).

First para is okay. Suggest replacing second para with something like "Some transport protocols and peers may not support signaling of a Final message. Applications therefore must not rely on receiving a message marked Final to know that the other endpoint is done sending on a connection"

tfpauly:
Okay, addressed capitalization and took Brian's suggested paragraph (with one tiny change).

philsbln:
Fine with me - go ahead

tfpauly:
@mwelzl, @britram any further comments on this one?

britram:
ship it

mwelzl:
Sorry, seems I should have clicked "approve" earlier. Still being clumsy in github  :(

Issue #184, opened 2018-06-01
Replace long author lists with editors


tfpauly:
The RFC Editor has become rather strict in implementing the 5 author limit for drafts:

This document has more than five authors listed, which is considered excessive under normal circumstances. If you plan to request publication as an RFC, this will require additional consideration by the stream manager (for example, the IESG), and publication may be declined unless sufficient justification is provided. See RFC 7322, section 4.1.1 for details.

We should update these for -01

gorryfair:
In cases where one or two authors are collating stuff, then this is certainly correct. I'd need convincing this is true for at least some of the author list - in cases where more than 5 authors need to be in RFC-Ed review, I was always told the process could be over-ridden, where the author team needed that level of interaction with the process.

mwelzl:
The phrasing in RFC 7322 is "If there is a request for more than five authors, the stream-approving body needs to consider if one or two editors should have primary responsibility for this document, with the other individuals listed in the Contributors or Acknowledgements section."   - "if there is a request ... needs to consider" doesn't seem to me like a strict "MUST be like that", indeed.

I'm also quite unhappy with the idea of moving to an editor - author mode. I think that for both the API and the implementation draft (I didn't really follow the arch. one), all authors have been making truly substantial contributions across the whole thing, more than just "send me a few paragraphs and we'll include them".

tfpauly:
@mwelzl I don't see how having an editor-author model reduces the perceived contribution of the authors. For documents like those in the QUIC WG, there are many people I know who are extremely active as authors, who I recognize are primary contributors, despite not being editors.

I raised this issue since its a point that Aaron brought up, and I agree that it gets fairly hard to read a document when the first page is half taken up with lines of names.

mwelzl:
@tfpauly - well I disagree about the first page being "half taken up" with this, and about the document getting hard to read; it also seems that the rule that Aaron was concerned about is not strict, plus it seems that the large majority of us is in favor of keeping it as-is.

gorryfair:
I still don't think this is the right thing to do.

philsbln:
@tfpauly For the academic contributors, author vs. being in the acknowledgment section often makes a big difference for internal and project evaluation – moving it (from an administrative perspective) from a project we get credit for to a fun/hobby project.

csperkins:
Agree with @gorryfair, @mwelzl, and @philsbln: moving authors to a contributors section is problematic and suggests a lesser contribution

[...]

PR #185, opened 2018-06-01
Move authors out of list to Contributors at end of document

tfpauly:
Satisfies #184

### DIFF
--- a/draft-ietf-taps-arch.md
+++ b/draft-ietf-taps-arch.md
@@ -32,48 +32,6 @@ author:
     street: Gloriastrasse 35
     city: 8092 Zurich
     country: Switzerland
-  -
-    ins: A. Brunstrom
-    name: Anna Brunstrom
-    org: Karlstad University
-    street: Universitetsgatan 2
-    city: 651 88 Karlstad
-    country: Sweden
-    email: anna.brunstrom@kau.se
-  -
-    ins: G. Fairhurst
-    name: Godred Fairhurst
-    org: University of Aberdeen
-    street: Department of Engineering
-    street: Fraser Noble Building
-    city: Aberdeen, AB24 3UE
-    country: Scotland
-    email: gorry@erg.abdn.ac.uk
-    uri: http://www.erg.abdn.ac.uk/
-  -
-    ins: C. Perkins
-    name: Colin Perkins
-    org: University of Glasgow
-    street: School of Computing Science
-    city: Glasgow  G12 8QQ
-    country: United Kingdom
-    email: csp@csperkins.org
-  -
-    ins: P. Tiesel
-    name: Philipp S. Tiesel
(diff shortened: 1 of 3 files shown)


draft-ietf-taps-arch.md
```diff
@@ -431,3 +389,7 @@ This work has been supported by the UK Engineering and Physical Sciences
 Research Council under grant EP/R04144X/1.
 
 Thanks to Stuart Cheshire, Josh Graessley, David Schinazi, and Eric Kinnear for their implementation and design efforts, including Happy Eyeballs, that heavily influenced this work.
+
+# Contributors
+
+The authors of this document include Tommy Pauly, Brian Trammell, Anna Brunstrom, Godred Fairhurst, Colin Perkins, Philipp S. Tiesel, and Chris Wood.
```
britram: I'd do this a bit differently (applies to all three docs)

- I'd place contributors above acknowledgments
- "include" sounds like there are other authors, which is not what we're saying.
- I haven't included editors in contributors' lists in the past "in addition to the editors, this document was written by:"
- I'd add back affiliations and (with author agreement, gdpr etc) email addresses
- and stylistically, I'd do this in list form.

Where specific sections have specific authors, this should also be noted (see 8095), but this is not the case here...

gorryfair: Just to be sure ...

Are you suggesting keeping the current list of authors in the boilerplate; diving text between contributors and acknowledgment sections. - That I would agree with.

If you were to be suggesting pruning the authors list - I would not want that for the document s were I am currenty listed as an author: specifically I *would* wish to be in the RFC-Ed last call for these documents.

mwelzl: See my answer to this in #184: I agree with Gorry here. Specifically, as long as we're not told that we MUST do such a thing, I think we should try to keep the author list as-is.

britram:
+1 @mwelzl I would also be okay with not doing this, and fighting the RPC about it. 

(One of the reasons the limit is being advertised more forcefully is that there have been instances where authorship doesn't necessarily reflect work done, in some cases because people are financially rewarded for getting their name on an RFC or even an Internet-Draft, and you have people who really have no clue what is in the document on the critical path in auth48. This is not the case here.)

(Doubleparenthetical: someone should have a talk with the RSE about how to handle actual large collaborations such as this one where not everyone needs to be on the RPC last call, but where people actually should show up in the metadata that gets pulled into various academically-important indices.)

mwelzl:
I hope we all agree that this can be closed; it's probably better addressed with an email to the chairs.

tfpauly:
Sure, I don't particularly care—I don't see why it's a huge problem, considering that the QUIC documents, etc, are doing this style, but if you guys want to fight the man, that's fine too =)

Issue #190, opened 2018-06-01
Connection properties in "Setting and Querying of Connection Properties" are not well structured and underspecified


philsbln:
In Section 9 – "Setting and Querying of Connection Properties" – the properties of an established connection and path are heavily underspecified.

Should be fixed along with #153.

theri:
I'm working on this issue within our pull request for #153, using the terminology we as simplified there.

For Section 9, I'm proposing the following:
* An application can query Connection Properties on a connection.
* Connection Properties include the status of the connection and whether it is possible to send and to receive data on it. (One of them might not be possible, e.g., in a unidirectional connection.)
* Additionally, Connection Properties include the Connection's Transport Properties. 
* Transport Properties include Selection Properties, so the application can know which Protocol Selection Properties had been specified, and which ones are actually supported by the protocol that has been chosen.
  * For Establishing connections (i.e., we are not done racing yet, the final protocols have not been selected yet), the application gets the Selection Properties that had been specified on the Preconnection earlier.
  * For Established, Closing, or Closed connections (i.e., the protocols have been selected), the application learns whether the Selection Properties are actually provided by the selected protocols. For example, if the application specified earlier, "Control Checksum Coverage: Prefer", now it will get "Control checksum coverage: True", indicating that on this connection it can indeed control the checksum coverage.
* Transport Properties also include Protocol Properties (Generic and Specific). These Protocol Properties can still be modified by the application. Basically, Protocol Properties are the only thing that the application can still set, everything else it can only query at this point.
* Finally, the application can query properties of the path. Not sure how well we have to specify them, probably this part will be implementation-specific just like, e.g., the Interface Instance or Type.

Does everyone agree, or is there anything to discuss here?
For me this is not a big change, it's just putting what was already there in terms of #153. The only thing I'm wondering if whether the Protocol Selection Properties of the finally chosen protocols are still "Selection Properties", as the selection is already done. That's why they were called "Transport Features" before, but now I want less terms and they're almost the same thing...

gorryfair:
OK - so I commented on these things before - but in these things I think I agree. I'd personally prefer to keep the features as attributes implemented in the transport (etc) and the properties as the things requested. Indeed this begs the question about what to call the resulting transport when you selected properties and now get a transport. I think still calling them (for consistency) properties seems really good to me. :-).

PR #191, opened 2018-06-01
support for unidirectional connections

### LINKED ISSUE #170: Do we need support for unidirectional connections
As QUIC has unidirectional streams, we might want to be able to accommodate them in our transport API. Do we want to handle them fundamentally different from a half-closed connection?
This should match the decisions made in #53.

philsbln:
This closed #170 

As we have no methods on a connection object that query properties, can_send/can_receive also became connection properties.

As with all the other connection properties, these are indeed a little underspecified. I want to keep it that way till we fix it for all connection properties with #153 and #190

### DIFF
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -459,6 +459,23 @@ useful to indicate its reliability requirements on a per-Message basis.
 This property applies to Connections and Connection Groups. This is not a
 strict requirement.  The default is to not have this option.
 
+### Direction of communication
+
+Type: Enum
+
+This property specifies whether an application wants to use the connection for sending and/or receiving data.  Possible values are:
+
+Bidirectional (default):
+: The connection must support sending and receiving data
+
+unidirectional send:
+: The connection must support sending data. 
+
+unidirectional receive:
+: The connection must support receiving data
+
+In case a unidirectional connection is requested, but the system should fall back to bidirectional transport if unidirectional connections are not  supported by the transport protocol.
+
 ### Use 0-RTT session establishment with an idempotent Message {#prop-0rtt}
 
 Type: Preference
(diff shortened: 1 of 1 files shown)


draft-ietf-taps-interface.md
```diff
@@ -459,6 +459,23 @@ useful to indicate its reliability requirements on a per-Message basis.
...
+: The connection must support sending and receiving data
+
+unidirectional send:
+: The connection must support sending data. 
+
+unidirectional receive:
+: The connection must support receiving data
+
+In case a unidirectional connection is requested, but the system should fall back to bidirectional transport if unidirectional connections are not  supported by the transport protocol.
+
```
britram: Though it probably is one for reasons of implementation expediency, I’m not sure this belongs as a 

Post had these as separate connection calls, which seems more natural to me, and maps to multicast more nicely: receive-only has a lot in common with listen, for instance.

philsbln:
Aehm… that comes surprising – @britram why didn't you bring this up in #170 ?
Do you think this pull-request is the right place to discuss this issue?

I would prefer we merge it now and make a new issue for the question whether the uni-directional connection should be more reflected in the connection. I am not particularly found on keeping it a selection property but would like to have the concept of unidirectional connections rather sooner than later.

britram:
>Aehm… that comes surprising – @britram why didn't you bring this up in #170 ? Do you think this pull-request is the right place to discuss this issue?

Happy to move the discussion back over to the issue if you like. Will continue the thread here for now though since you've got points to address below.

> I would prefer we merge it now and make a new issue for the question whether the uni-directional connection should be more reflected in the connection. I am not particularly found on keeping it a selection property but would like to have the concept of unidirectional connections rather sooner than later.

fine, but keep #170 open then, because adding the selection property is possibly-necessary but definitely-not-sufficient to fully address unidirectional connection semantics.

> Our consensus in #170 was that unidirectional connection should use the same abstraction for QUIC unidirectional streams, multicast sources/sinks, and half-closed connections – despite they are constructed in completely different ways.

good, I agree with this.

> The Selection property is just for the QUIC style ones…

then I'm confused why this PR purports to close #170, since it only addresses part of the issue...?

philsbln:
@britram agreed.

> then I'm confused why this PR purports to close #170, since it only addresses part of the issue...?

The question (and discussion) #170 was about support for including some support for unidirectional connections (with focus on QUIC and half-closed), but omitted details on multicast, as this has not laned yet. So indeed, this pull-requests does everything requested by #170, but does not fully capture the topic. 

I will move the open issues (Multicast support, is the Selection Property enough, find some place for more explanation of Uni/Bidi Streams) to #170 once this PR is merged.

Issue #172, opened 2018-04-18
Add architectural section on protocol equivalence as it pertains to racing


chris-wood:
And add a subsection specifically for security protocols. Determining what security protocols (and configurations) are equivalent is nontrivial and incredibly important if we are to race, e.g., TLS and gQUIC.

gorryfair:
I think the racing/gathering section needs a tiny little explanation text and this seems reasonable to add - if you add to this section, can you also make sure it mentions something of the terms "policy" and "cached state" as a part of this update?

chris-wood:
@gorryfair yes, will do!

chris-wood:
Addressed by #199.

PR #194, opened 2018-06-06
Prescribe policies for gathering multiple security protocols (or actually, not)

chris-wood:
Addresses #172.

We are limited to only recommending conditions for which multiple security protocols should be offered up. We cannot and should not attempt to define security protocol equivalence. Equivalent algorithms, for example, is not a sufficient condition for equivalence. I give an example of this for TLS 1.2 and 1.3. I think it might behoove us to consider a cross-area document that attempts to define security protocol equivalence to help guide implementations. It's definitely out of scope for these three documents.

### DIFF
--- a/draft-ietf-taps-arch.md
+++ b/draft-ietf-taps-arch.md
@@ -329,7 +329,18 @@ The Transport System Implementation Concepts define the set of objects used inte
 
 * Path Selection: Path Selection represents the act of choosing one or more paths that are available to use based on the Path Selection Properties provided by the application, and a Transport Services system's policies and heuristics.
 
-* Protocol Selection: Protocol Selection represents the act of choosing one or more sets of protocol options that are available to use based on the Protocol Properties provided by the application, and a Transport Services system's policies and heuristics.
+* Protocol Selection: Protocol Selection represents the act of choosing one or more sets of protocol options that are available to use based on the Protocol Properties provided by the application, and a Transport Services system's policies and heuristics. See {{security-equivalence}} for considerations regarding security Protocol Selection.
+
+#### Security Protocol Gathering and Equivalence {#security-equivalence}
+
+In some cases, Protocol Properties or a Transport Services system's policies and 
+heuristics may yield multiple candidate security protocols. For example, both
+DTLS and TLS may be viable if an application does not express options that
+are prohibited by either UDP or TCP. In such cases, only one security protocol
+SHOULD be used as the outcome of Protocol Selection as deemed appropriate by 
+System Policy. Multiple protocols MAY be used if they are equivalent in terms 
+of configuration and operation. See {{I-D.ietf-taps-transport-security}} for a 
+discussion of criteria that MUST be met for configuration and operation equivalence. 
 
 ### Racing {#racing}
 
--- a/draft-ietf-taps-arch.md
+++ b/draft-ietf-taps-arch.md
@@ -333,14 +333,17 @@ The Transport System Implementation Concepts define the set of objects used inte
 
 #### Security Protocol Gathering and Equivalence {#security-equivalence}
 
-In some cases, Protocol Properties or a Transport Services system's policies and 
-heuristics may yield multiple candidate security protocols. For example, both
-DTLS and TLS may be viable if an application does not express options that
+In some cases, a Transport Services system's policies and heuristics, or individual 
+Protocol Properties, heuristics may yield multiple candidate security protocols. For example, 
+both DTLS and TLS may be viable if an application does not express options that
 are prohibited by either UDP or TCP. In such cases, only one security protocol
 SHOULD be used as the outcome of Protocol Selection as deemed appropriate by 
 System Policy. Multiple protocols MAY be used if they are equivalent in terms 
-of configuration and operation. See {{I-D.ietf-taps-transport-security}} for a 
-discussion of criteria that MUST be met for configuration and operation equivalence. 
+of configuration, operation, and properties. Note that different versions of the same 
+protocol are not necessarily equivalent, even if the underlying cryptographic algorithms 
+are functionally the same. Handshake and wire format differences, among others, prohibit
+equivalence. As an example, TLS 1.2 and TLS 1.3 are NOT equivalent protocols, even if
+configured using the same essential cryptographic algorithms. 
 
 ### Racing {#racing}
 


draft-ietf-taps-arch.md
```diff
@@ -329,7 +329,21 @@ The Transport System Implementation Concepts define the set of objects used inte
...
+* Protocol Selection: Protocol Selection represents the act of choosing one or more sets of protocol options that are available to use based on the Protocol Properties provided by the application, and a Transport Services system's policies and heuristics. See {{security-equivalence}} for considerations regarding security Protocol Selection.
+
+#### Security Protocol Gathering and Equivalence {#security-equivalence}
+
+In some cases, a Transport Services system's policies and heuristics, or individual 
+Protocol Properties, heuristics may yield multiple candidate security protocols. For example, 
+both DTLS and TLS may be viable if an application does not express options that
+are prohibited by either UDP or TCP. In such cases, only one security protocol
+SHOULD be used as the outcome of Protocol Selection as deemed appropriate by 
+System Policy. Multiple protocols MAY be used if they are equivalent in terms
```
philsbln: I don't understand this should – does that mean we don't want racing between e.g. DTLS and QUIC because they use different security protocols? It is clear that stacking arbitrarily security protocols is a bad idea but this sentence does tell me no racing between security protocols with the next one realativising this.

britram: so there's apparent confusion here between "select for attempted racing" and "use in the configured stack". The latter is clearly nonsensical. The former is also potentially problematic, and I think that's what this SHOULD is trying to avoid...

draft-ietf-taps-arch.md
```diff
@@ -329,7 +329,21 @@ The Transport System Implementation Concepts define the set of objects used inte
...
+Protocol Properties, heuristics may yield multiple candidate security protocols. For example, 
+both DTLS and TLS may be viable if an application does not express options that
+are prohibited by either UDP or TCP. In such cases, only one security protocol
+SHOULD be used as the outcome of Protocol Selection as deemed appropriate by 
+System Policy. Multiple protocols MAY be used if they are equivalent in terms 
+of configuration, operation, and properties. Note that different versions of the same 
+protocol are not necessarily equivalent, even if the underlying cryptographic algorithms 
+are functionally the same. Handshake and wire format differences, among others, prohibit
+equivalence. As an example, TLS 1.2 and TLS 1.3 are NOT equivalent protocols, even if
+configured using the same essential cryptographic algorithms.
```
britram: I think we can probably be simpler about this for now: application MUST specify crypto protocols by name and preference. I don't think we have enough understanding yet to do crypto stack equivalence the same way we can do transport stack equivalence.

draft-ietf-taps-arch.md
```diff
@@ -329,7 +329,21 @@ The Transport System Implementation Concepts define the set of objects used inte
...
 
-* Protocol Selection: Protocol Selection represents the act of choosing one or more sets of protocol options that are available to use based on the Protocol Properties provided by the application, and a Transport Services system's policies and heuristics.
+* Protocol Selection: Protocol Selection represents the act of choosing one or more sets of protocol options that are available to use based on the Protocol Properties provided by the application, and a Transport Services system's policies and heuristics. See {{security-equivalence}} for considerations regarding security Protocol Selection.
+
+#### Security Protocol Gathering and Equivalence {#security-equivalence}
+
+In some cases, a Transport Services system's policies and heuristics, or individual 
+Protocol Properties, heuristics may yield multiple candidate security protocols. For example, 
+both DTLS and TLS may be viable if an application does not express options that
+are prohibited by either UDP or TCP. In such cases, only one security protocol
```
tfpauly: What do you mean by "only one security protocol SHOULD be used"? Also, we aren't using capital SHOULDs and MUSTs right now in the document.

draft-ietf-taps-arch.md
chris-wood: Yes, it's pointed at the former. Basically, do NOT race DTLS and QUIC.

draft-ietf-taps-arch.md
chris-wood: Agreed. That's what I'm trying to spell out here. Do not race security protocols since we don't know what's equal and what's not.

draft-ietf-taps-arch.md
chris-wood: Don't try to race DTLS and QUIC!

chris-wood:
@tfpauly I am opposed to adding any text that tries to articulate what security protocol equivalence is, and try to more specifically suggest circumstances in which it is safe to race competing security protocols. Brian's suggestion regarding specifying protocols by name is probably the safest and best way to go.

draft-ietf-taps-arch.md
tfpauly: Okay, that's not very clear

draft-ietf-taps-arch.md
csperkins: Do we need this much detail? This seems to be saying that once the set of candidates has been identified, those that don't provide appropriate security properties must be discarded before attempting to establish a connection. The implementation draft will need to provide guidance on what that means, but the architecture can be less specific.

philsbln:
I am not sure whether we want to call it "protocol equivalence" or just spin it as "find protocol candidates that satisfy requirements" – To be raced, protocols only need to fit (be equivalent with respect to the application's requirements).

I  have done this categorization for most transport aspects and plan to add this to draft-tiesel-communitgrany, but this gets a little lengthy to make it a section here. Defining this for security protocols would be possible, but probably hard to get consensus on.

chris-wood:
@philsbln defining that equivalence is outside the scope of these documents and TAPS as a whole, IMO.

philsbln:
@chris-wood I am not advocating for solving the security protocol equivalence problem here.

Still, I guess it would be good to have the TAPS not preventing an implementation that has a definition of security protocol equivalence to make use of it.

britram:
i would go so far as to say that, at this point in the development of security protocols, any implementation of a system purporting to provide security protocol equivalence is almost certainly wrong. the set of  ways in which these things differ from each other is much more subtle than transport protocols in ways that may appear to work without actually guaranteeing given properties. here, safety dictates that we do need to be able to reach into the stack here for now.

chris-wood:
+1 to Brian's comment! 

@all, Tommy is going to write a more general section on protocol equivalence that focuses primarily on the transport bits. This proposed text will be reworked and pulled in there. When done, I'll close this PR.

tfpauly:
@chris-wood Take a look at #199, and add what you'd like there; then we can close this PR once that's ready to go?

chris-wood:
Good to go! Closing this beast.

Issue #157, opened 2018-03-22
API needs a way to cancel Preconnection.Listen()


JonathanLennox:
As far as I can tell there's no way for an application to indicate that it wants to stop listening for an incoming connection.

As with my #156, this will need a way to know that it's complete, in case there's a race with Preconnection -> ConnectionReceived<> on another thread.

britram:
fixed by #195

Issue #155, opened 2018-03-22
Need relative ordering of API events


JonathanLennox:
The API document needs to specify the relative timing and ordering (i.e., for appropriate thread models, the threading properties) of events.

E.g., I think it should be guaranteed that a connection's Received<> event isn't called until its initiating event (Connected<>, ConnectionReceived<>, or RendezvousDone<>) has finished processing; and only one Received<> event is invoked at a time for a given connection, with the events being invoked with the appropriate logical order is required by the connection semantics.

Similarly, only one of a connection's Sent<> events should be invoked at a time.  However, I don't think there needs to be any restriction against Sent<> and Received<> being invoked simultaneously on the same connection.

No other events should be received for a connection after Closed<> or ConnectionError<>.

For different connections, It should be allowed for their events to be invoked simultaneously, even if the connections are entangled with each other.

JonathanLennox:
One particular consideration that came up in the May 16th call is that this last event (presumably Closed<> or ConnectionError<> or Aborted<> or whatever) needs to be enqueued to be executed after previous callbacks complete, rather than being executed synchronously on the stack of another callback.

britram:
fixed by #195

PR #195, opened 2018-06-11
API events

britram:
Suggestion to address #155 and #157 in the API doc.

### DIFF
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -1455,20 +1455,24 @@ Connection -> SoftError<>
 
 As this interface is designed to be independent of concurrency model, the
 details of how exactly actions are handled, and on which threads/callbacks
-events are dispatched, is an implementation detail. However, the interface
+events are dispatched, are implementation dependent. However, the interface
 does provide the following guarantees about the ordering of operations:
 
 - Received<> will never occur on a Connection before a Ready<> event on that
   Connection, or a ConnectionReceived<> or RendezvousDone<> containing that
   Connection.
 
-- No events or errors will occur on a Connection after a Closed<> event, an
+- No events will occur on a Connection after a Closed<> event, an
   InitiateError<> or ConnectionError<> on that connection. To ensure this
-  ordering, neither Closed<> nor ConnectionError<> will occur on a Connection
-  while other events on the Connection are still outstanding.
+  ordering, Closed<> will not occur on a Connection while other events on the
+  Connection are still locally outstanding (i.e., known to the interface and
+  waiting to be dealt with by the application). ConnectionError<> may occur
+  after Closed<>, but the interface must gracefully handle the application
+  ignoring these errors.
 
 - Sent<> events will occur on a Connection in the order in which the Messages
-  were sent.
+  were sent (i.e., delivered to the kernel or to the network interface,
+  depending on implementation).
 
 # IANA Considerations
 
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -1451,6 +1451,25 @@ stack supports it, there is no guarantee that a soft error will be signaled.
 Connection -> SoftError<>
 ~~~
 
+# Ordering of Operations and Events
+
+As this interface is designed to be independent of concurrency model, the
+details of how exactly actions are handled, and on which threads/callbacks
+events are dispatched, is an implementation detail. However, the interface
+does provide the following guarantees about the ordering of operations:
+
+- Received<> will never occur on a Connection before a Ready<> event on that
+  Connection, or a ConnectionReceived<> or RendezvousDone<> containing that
+  Connection.
+
+- No events or errors will occur on a Connection after a Closed<> event, an
+  InitiateError<> or ConnectionError<> on that connection. To ensure this
+  ordering, neither Closed<> nor ConnectionError<> will occur on a Connection
+  while other events on the Connection are still outstanding.
+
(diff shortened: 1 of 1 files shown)


draft-ietf-taps-interface.md
```diff
@@ -1436,6 +1451,25 @@ stack supports it, there is no guarantee that a soft error will be signaled.
 Connection -> SoftError<>
 ~~~
 
+# Ordering of Operations and Events
+
+As this interface is designed to be independent of concurrency model, the
+details of how exactly actions are handled, and on which threads/callbacks
+events are dispatched, is an implementation detail. However, the interface
```
mwelzl: "The details ... is a detail" isn't so great

draft-ietf-taps-interface.md
```diff
@@ -1436,6 +1451,25 @@ stack supports it, there is no guarantee that a soft error will be signaled.
...
+does provide the following guarantees about the ordering of operations:
+
+- Received<> will never occur on a Connection before a Ready<> event on that
+  Connection, or a ConnectionReceived<> or RendezvousDone<> containing that
+  Connection.
+
+- No events or errors will occur on a Connection after a Closed<> event, an
+  InitiateError<> or ConnectionError<> on that connection. To ensure this
+  ordering, neither Closed<> nor ConnectionError<> will occur on a Connection
+  while other events on the Connection are still outstanding.
```
mwelzl: How can you guarantee that?  An abort can happen at any time, and that would give you a ConnectionError. Maybe I don't understand what "outstanding" means... or maybe it needs to be defined?

draft-ietf-taps-interface.md
```diff
@@ -801,6 +801,15 @@ The Listen() Action consumes the Preconnection. Once Listen() has been
...
 subsequent establishment call may be made on the Preconnection.
 
+Listening continues until the global context shuts down, or until the Stop
+action is performed on the same Preconnection:
+
+~~~
+Preconnection.Stop()
+~~~
+
+After Stop() is called, the preconnection can be disposed of.
```
mwelzl: What's the point of this - how is it different from Abort?

draft-ietf-taps-interface.md
britram: This really depends on the concurrency model of the implementation platform. If the platform does callbacks, it ensures this by enqueueing dispatch of close/connection error after the dispatch of all other events is completed. If the platform does CSP dispatch, this naturally happens because a channel creates a queue. If the platform dispatches every event to a new thread, then this is hard, but it's kind of the platform's fault then...

draft-ietf-taps-interface.md
britram: Stop() is ordely. Abort() is more abrupt (at least, that should be the intent...)

mwelzl: Ah, sorry, missed this, it's for cleaning up a Preconnection. That makes sense to me. But semantically this seems quite similar to abort to me - a way to say "I don't want to use this anymore, from now on" - so why not give it the same name?  (though I don't care much about the name)

draft-ietf-taps-interface.md
mwelzl: Okay, you mean "outstanding" as in "enqueued locally". Then this should be phrased as "locally outstanding" or something like that, I think.  To be clear, I have no problem at all with the first sentence - it's the second one that makes me think of the protocol machinery when I read "outstanding" - as in outstanding packets, incoming messages, causing events...

csperkins: Does this mean `Closed<>` can't fire until you're sure there can be no errors from the peer (e.g., after the TIME_WAIT timeout for TCP)?

draft-ietf-taps-interface.md
philsbln: Is there a reason not calling it "Preconnection.Abort" ?

mwelzl: Exactly what I meant - sorry for not saying it clearer

draft-ietf-taps-interface.md
tfpauly: I agree with the approach of always having some final event (whether this is Closed or something else like Invalidated is another point).

If the connection has been dropped by the application, then I think the assumption is that if there are any errors that occurred later, the application would have no way of receiving them anyhow, so they must be silently discarded?

draft-ietf-taps-interface.md
```diff
@@ -1288,16 +1303,16 @@ Connection properties include:
 
 * The status of the Connection, which can be one of the following:
   Establishing, Established, Closing, or Closed.
```
tfpauly: See my comments on the `Closing` issue—I do think that this is of no benefit to the application, and doesn't provide a meaningful signal in the state machine. We can certainly prohibit send and receive after close, but that doesn't necessarily align with a closing state when things are asynchronous.

draft-ietf-taps-interface.md
britram: IMO `Abort()` should be used for exceptional or abnormal termination, and this is used for a normal termination of listening.

draft-ietf-taps-interface.md
britram: +1 to "locally outstanding", will fix. Will have to think about `Closed<>` versus `ConnectionError<>`.

draft-ietf-taps-interface.md
britram: ISTM there's probably another PR for the closing state issue once discussion completed; I won't address this here.

PR #197, opened 2018-06-19
Abstract security parameter configuration, and unify the interface with the security survey draft

chris-wood:


draft-ietf-taps-interface.md
```diff
@@ -661,43 +669,47 @@ Security configuration parameters and sample usage follow:
...
 ciphersuites, supported groups, and signature algorithms.
 
 ~~~
-SecurityParameters.AddSupportedGroup(22)    // secp256k1
-SecurityParameters.AddCiphersuite(0xCCA9)   // TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305_SHA256
-SecurityParameters.AddSignatureAlgorithm(7) // ed25519
+SecurityParameters.AddSupportedGroup(secp256k1)
+SecurityParameters.AddCiphersuite(TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305_SHA256)
+SecurityParameters.AddSignatureAlgorithm(ed25519)
 ~~~
```
mwelzl: Do these things need references?  E.g. "TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305_SHA256" ?

britram: eh... in examples, it's probably okay

chris-wood: I agree with Brian here, especially since these are meant to be abstract constants that each implementation could, on its on, redefine using some other constant or enum.

mwelzl: ok by me then. just wanted to check

PR #198, opened 2018-06-19
Split Message receive Metadata into a separate section

chris-wood:
And add an early data metadata value.

### DIFF
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -1054,7 +1054,7 @@ Idempotent is a boolean property. If true, the application-layer entity in the
 Message is safe to send to the remote endpoint more than once for a single
 Send Action. It is used to mark data safe for certain 0-RTT establishment
 techniques, where retransmission of the 0-RTT data may cause the remote
-
+application to receive the Message multiple times.
 
 ### Final {#send-final}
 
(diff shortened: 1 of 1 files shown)


draft-ietf-taps-interface.md
```diff
@@ -1054,7 +1054,7 @@ Idempotent is a boolean property. If true, the application-layer entity in the
 Message is safe to send to the remote endpoint more than once for a single
 Send Action. It is used to mark data safe for certain 0-RTT establishment
 techniques, where retransmission of the 0-RTT data may cause the remote
-application to receive the Message multiple times.
+
```
csperkins: Is this change intended?

draft-ietf-taps-interface.md
```diff
@@ -1224,6 +1218,32 @@ received that reception has failed. Such conditions that irrevocably lead the
...
+data streams. This is useful if applications need to treat early data separately, 
+e.g., if early data has different security properties than data sent after 
+connection establishment. In the case of TLS 1.3, client early data can be replayed 
+maliciously (see {{!I-D.ietf-tls-tls13}}). Thus, receivers may wish to perform additional 
+checks for early data to ensure it has not been replayed or restrict responses until 
+the handshake is finished. If TLS 1.3 is available and the recipient Message was 
+sent as part of early data, the corresponding metadata carries a flag indicating 
+as such. If early data is enabled, applications should check this metadata field 
+for Messages received during connection establishment and respond accordingly. 
+
```
csperkins: Better - although I wonder if it should be "Idempotent Early Data"?

draft-ietf-taps-interface.md
britram: +1, this looks like an editor regression

draft-ietf-taps-interface.md
britram: also +1 to "idempotent early data" (the former is implied by Early Data in a TLS context, but not necessarily in a transport context)

draft-ietf-taps-interface.md
chris-wood: Yeah, uh, or fat finger. Oops.

draft-ietf-taps-interface.md
chris-wood: That implies that the metadata signals whether or not the data is idempotent, which is something only applications can determine. I added some text clarifying this.

Issue #148, opened 2018-03-21
API: Add text for sending partial messages


tfpauly:
Send needs to take:
- Data 
- Which message (message object contains message properties)
- Flags: end-of-message, end-of-connection-send

csperkins:
Is data separate to the message?

britram:
@tfpauly: would you like to take a swing at this one?

britram:
Timing out. ietf-01 it is.

mwelzl:
What is the expected receive behavior - can an application also receive partial messages?

I'm asking because SCTP has this ability, and then needed to include means to inform the receiver that message delivery has been aborted (when not transmitting reliably and a message fragment is missing)

=> this is a path of complexity that I think we shouldn't take, just for the sake of allowing large messages to be transferred as such. The alternative is to force them to be small...

mwelzl:
I thought so (about receiving). So how do you handle a missing partial message? Do you assume that the app can always be told?  I’m trying to remember, but there was a problem with this...

mwelzl:
Another problem with this: receiver-side HOL blocking between streams. One stream gets a very long message, all the others get blocked. If you allow very large messages, SCTP needs to use NDATA to deal with this (which internally divides the messages). I think that's absolutely fair to expect, as this is really the right way to handle this. How does QUIC handle very large messages? Does it divide them, or will it get HOL blocking delay?

If all protocols that we'll use divide large messages into smaller chunks for prioritization and avoiding HOL delay between streams, than maybe this is a non-problem. Else - I'm just saying, a simple way around all of this is to put a small upper bound on the size of a message that we support. Or we document the risk...

tfpauly:
@mwelzl I'm not sure what the concern is for a missing partial message. When you receive, you get two pieces of information: the newly received data, and the associated Message object. If you could only receive part of the message (it's not complete), you will be able to check that this isn't the entire Message, and you'll be able to call receive on the connection again to get more data, which may be for the same Message. If there's an error before being able to receive the entire message, you'll get an error in the receive call letting you know that message was incomplete.

You don't need HOL blocking if each receive callback tells you which message the data is associated with.

mwelzl:
First, about HOL blocking:

I'll conclude: this is okay; it assumes that the transport below does the right thing, but I think that makes sense to assume.

[...]

PR #200, opened 2018-06-24
Rework partial sends and receives

tfpauly:
Satisfies #148 

- Support partial sends more naturally by splitting out into Message, messageData, and endOfMessage flag.
- Do the same for Receive
- Receive should also take minIncompleteLength to specify how partial receives are treated (or not, by default)

### DIFF
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -945,29 +945,46 @@ capacity that it sees fit.
 # Sending Data {#sending}
 
 Once a Connection has been established, it can be used for sending data. Data
-is sent by passing a Message Object and additional parameters
-{{send-params}} to the Send Action on an established Connection:
+is always sent in association with a Message Object and additional parameters
+({{send-params}}) using the Send Action on an established Connection:
 
 ~~~
-Connection.Send(Message, sendParameters)
+Connection.Send(Message, messageData, endOfMessage)
 ~~~
 
-The type of the Message to be passed is dependent on the implementation, and
-on the constraints on the Protocol Stacks implied by the Connection's
-transport parameters. It may itself contain an array of octets to be
-transmitted in the transport protocol payload, or be transformable to an array
-of octets by a sender-side framer (see {{send-framing}}).
+The Message parameter defines a single instance of a message and its parameters.
+The interpretation of a Message to be sent is dependent on the implementation, and 
+on the constraints on the Protocol Stacks implied by the Connection’s transport parameters.
+For example, a Message may be a single datagram for UDP Connections; or an HTTP
+Request for HTTP Connections. Various Send Parameters can be configured before using 
+a Message to send, which influence the priority and treatment of the Message's data upon 
+sending. See {{send-params}} for a description of each parameter.
+
+~~~
+Message := NewMessage()
+~~~
+
+The messageData parameter contains any octects to be sent for this message.
+
+~~~
+messageData := "hello".octets()
+~~~
+
+The endOfMessage parameter is a boolean that indicates whether or not this
+Send Action completes the Message, or if more messageData is expected.
+Applications using protocols with small Message or datagram lengths will generally
+always mark their Sends as complete. Applications that need to stream data,
+sometimes due to the data being too large to hold in memory at any one time, 
+will perform multiple Send Actions on the same Message with different portions of 
+messageData. All messageData sent with the same Message object will
+be treated as an in-order series, until the endOfMessage is marked. Once the end
+is marked, the Message object may be re-used as a new Message that has the
+same Send Parameters.
 
 Some transport protocols can deliver arbitrarily sized Messages, but other
 protocols constrain the maximum Message size. Applications can query the
-protocol property Maximum Message Size on Send to determine the maximum size.
-
-There may also be system and Protocol Stack dependent limits on the size of
-a Message which can be transmitted atomically. For that reason, the Message
-object passed to the Send action may also be a partial Message, either
-representing the whole data object and information about the range of bytes
-to send from it, or an object referring back to the larger whole Message.
-The details of partial Message sending are implementation-dependent.
+protocol property Maximum Message Size on Send to determine the maximum size
+allowed for a single Message.
 
 If Send is called on a Connection which has not yet been established, an
 Initiate Action will be implicitly performed simultaneously with the Send.
@@ -1015,33 +1032,24 @@ set of send parameters not consistent with the Connection's transport
 parameters. The SendError contains an implementation-specific reference to the
 Message to which it applies.
 
-## Send Parameters {#send-params}
+## Message Send Parameters {#send-params}
 
-The Send Action takes per-Message send parameters which control how the
-contents will be sent down to the underlying Protocol Stack and transmitted.
+Each Message can be marked with Send Parameters to control how Data
+associated with the Message will be sent down to the underlying Protocol Stack 
+and transmitted. Note that these properties are per-Message, not per-Send.
+All data portions associated with a single Message share properties. For example,
+it would not make sense to have the beginning of a Message expire, but allow
+the end of a Message to still be sent.
 
-If Send Parameters should be overridden for a specific Message, an
-empty sent parameter Object can be acquired and all desired Send Parameters
-can be added to that Object. A sendParameters Object can be reused for
-sending multiple contents with the same properties.
+Parameters may be added to a Message object before the Message is used
+for sendings.
 
 ~~~
-SendParameters := NewSendParameters()
-SendParameters.Add(parameter, value)
+Message := NewMessage()
+Message.Add(parameter, value)
 ~~~
 
-The Send Parameters share a single namespace with the Transport Parameters (see
-{{transport-params}}). This allows the specification of Protocol Properties that can
-be overridden on a per-Message basis.
-
-Send Parameters may be inconsistent with the properties of the Protocol Stacks
-underlying the Connection on which a given Message is sent. For example,
-a Connection must provide reliability to allow setting an infinitie value for the
-lifetime property of a Message. Sending a Message with Send Properties
-inconsistent with the
-Transport Preferences on the Connection yields an error.
-
-The following send parameters are supported:
+The following Message Send Parameters are supported:
 
 ### Lifetime {#send-lifetime}
 
@@ -1064,7 +1072,6 @@ wire for Protocol Stacks supporting prioritization.
 
 This encoding of the priority has a convenient property that the priority
 increases as both Niceness and Lifetime decrease.
-
 
 ### Ordered {#send-ordered}
 
@@ -1146,8 +1153,8 @@ in the batch is enqueued.
 
 ~~~
 Connection.Batch(
-    Connection.Send(Message, sendParameters)
-    Connection.Send(Message, sendParameters)
+    Connection.Send(Message, messageData, endOfMessage)
+    Connection.Send(Message, messageData, endOfMessage)
 )
 ~~~
 
@@ -1163,7 +1170,7 @@ it is bound to the Preconnection during the pre-establishment phase:
 ~~~
 Preconnection.FrameWith(Framer)
 
-OctetArray := Framer.Frame(Message)
+OctetArray := Framer.Frame(Message, messageData, endOfMessage)
 ~~~
 
 Sender-side framing is a convenience feature of the interface, for parity with
@@ -1171,10 +1178,11 @@ receiver-side framing (see {{receive-framing}}).
 
 # Receiving Data {#receiving}
 
-Once a Connection is established, Messages may be received on it. The application can indicate that it is ready to receive Messages by calling Receive() on the Connection.
+Once a Connection is established, Messages may be received on it. The application can
+indicate that it is ready to receive Messages by calling Receive() on the Connection.
 
 ~~~
-Connection.Receive(ReceiveHandler, maxLength)
+Connection.Receive(ReceiveHandler, minIncompleteLength, maxLength)
 ~~~
 
 Receive takes a ReceiveHandler, which can handle the Received Event and the
@@ -1185,30 +1193,41 @@ number of Messages with a single call. This allows an application to provide
 backpressure to the transport stack when it is temporarily not ready to
 receive messages.
 
+Receive takes an optional minIncompleteLength argument. This value indicates
+the smallest partial Message data size that should be delivered. By default, this
+value is infinite, which means that only complete, atomic Messages will be delivered.
+If this value is set to some smaller value, the associated ReceiveHandler will
+be triggered only when at least that many bytes are available, the Message is
+complete, or the system needs to free up memory.
+
 Receive also takes an optional maxLength argument, the maximum size (in bytes
 of data) Message the application is currently prepared to receive. The default
 value for maxLength is infinite. If an incoming Message is larger than the
 minimum of this size and the maximum Message size on receive for
 the Connection's Protocol Stack, it will be received as a partial Message.
-Note that maxLength does not guarantee that the application will receive that
-many bytes if they are available; the interface may return partial Messages
-smaller than maxLength according to implementation constraints.
+Partial Messages are indicated by marking the endOfMessage boolean value
+as false in the ReceivedHandler. Note that maxLength does not guarantee that 
+the application will receive that many bytes if they are available; the interface may
+return partial Messages smaller than maxLength according to implementation constraints.
 
 ~~~
-Connection -> Received<Message>
+Connection -> Received<Message, messageData, endOfMessage>
 ~~~
 
 As with sending, the type of the Message to be passed is dependent on the
 implementation, and on the constraints on the Protocol Stacks implied by the
 Connection's transport parameters. 
 
-The Message Object must provide some method to retrieve an octet array
-containing application data, corresponding to a single message within the
-underlying Protocol Stack's framing.  See {{receive-framing}} for handling
-framing in situations where the Protocol Stack provides octet-stream transport
-only.
+The Message Object identifies the logical message received by the transport
+protocol, and provides methods to access metadata about the received data.
+The octets of data associated with this message are delivered as messageData.
+Multiple invocations of the ReceivedHandler may devlier data for the same
+Message until the endOfMessage flag is delivered. See {{receive-framing}} for 
+handling Message framing in situations where the Protocol Stack provides 
+octet-stream transport only.
 
-The Message Object passed to Received is complete and atomic, unless one of the following
+If the minIncompleteLength was set to be infinite, the Message passed to Received
+will complete and atomic (that is, endOfMessage will be set), unless one of the following
 conditions holds:
 
 * the underlying Protocol Stack supports message boundary preservation, and
@@ -1220,17 +1239,9 @@ conditions holds:
 * the underlying Protocol Stack does not support message boundary
   preservation, and no deframer was supplied by the application
 
-The Message Object passed to Received will indicate one of the following:
-
-1. this is a complete message;
-2. this is a partial message containing a section of a message with a known message boundary (made partial for local buffering reasons, either by the underlying Protocol Stack or the deframer). In this case, the Message Object passed to Received may contain the byte offset of the data in the partial Message within the full Message, an indication whether this is the
-last (highest-offset) partial Message in the full Message, and an optional
-reference to the full Message it belongs to; or
-3. this is a partial message containing data with no definite message boundary, i.e. the only known message boundary is given by termination of the Connection
-
 Note that in the absence of message boundary preservation and without
-deframing, the entire Connection is represented as one large message of
-indeterminate length.
+deframing, all bytes received on the Connection are represented as one 
+large message of indeterminate length.
 
 ~~~
 Connection -> ReceiveError<>
@@ -1310,7 +1321,7 @@ layer, it is bound to the Preconnection during the pre-establishment phase:
 ~~~
 Preconnection.DeframeWith(Deframer)
 
-Message := Deframer.Deframe(OctetStream, ...)
+{Message, messageData} := Deframer.Deframe(OctetStream, ...)
 ~~~
 
 # Setting and Querying of Connection Properties {#introspection}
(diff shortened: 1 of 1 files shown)


draft-ietf-taps-interface.md
```diff
@@ -1015,33 +1032,24 @@ set of send parameters not consistent with the Connection's transport
...
+All data portions associated with a single Message share properties. For example,
+it would not make sense to have the beginning of a Message expire, but allow
+the end of a Message to still be sent.
 
-If Send Parameters should be overridden for a specific Message, an
-empty sent parameter Object can be acquired and all desired Send Parameters
-can be added to that Object. A sendParameters Object can be reused for
-sending multiple contents with the same properties.
+Parameters may be added to a Message object before the Message is used
+for sendings.
```
chris-wood: Maybe comment that message parameters cannot change after Send is invoked?

draft-ietf-taps-interface.md
```diff
@@ -1185,30 +1193,41 @@ number of Messages with a single call. This allows an application to provide
 backpressure to the transport stack when it is temporarily not ready to
 receive messages.
 
+Receive takes an optional minIncompleteLength argument. This value indicates
```
chris-wood: If it's optional, it should be the last parameter, or we should specify that a value of 0 is the default when one should use it.

draft-ietf-taps-interface.md
```diff
@@ -1185,30 +1193,41 @@ number of Messages with a single call. This allows an application to provide
...
+the smallest partial Message data size that should be delivered. By default, this
+value is infinite, which means that only complete, atomic Messages will be delivered.
+If this value is set to some smaller value, the associated ReceiveHandler will
+be triggered only when at least that many bytes are available, the Message is
+complete, or the system needs to free up memory.
```
chris-wood: Since the last case is exceptional, perhaps we should mention that applications must check the received length before use, as it may be less than minIncompleteLength?

draft-ietf-taps-interface.md
```diff
@@ -1185,30 +1193,41 @@ number of Messages with a single call. This allows an application to provide
...
-The Message Object must provide some method to retrieve an octet array
-containing application data, corresponding to a single message within the
-underlying Protocol Stack's framing.  See {{receive-framing}} for handling
-framing in situations where the Protocol Stack provides octet-stream transport
-only.
+The Message Object identifies the logical message received by the transport
+protocol, and provides methods to access metadata about the received data.
+The octets of data associated with this message are delivered as messageData.
+Multiple invocations of the ReceivedHandler may devlier data for the same
+Message until the endOfMessage flag is delivered. See {{receive-framing}} for
```
chris-wood: Partial message data is sent in-order. Received data may be received out-of-order, right? Do we need to clarify that?

theri:
Looks like a good approach.
I have a general question about setting Message Properties:
This Pull Request gets rid of an explicit MessageProperties object and instead sets any Message Properties directly on the Message. For (Pre-)Connections, we do have an explicit TransportProperties object, which has the advantage that it can be reused across multiple Connections without being tied to any specific Connection. I think that the same would be useful for Messages, e.g., across multiple threads. 
What is the advantage of setting properties directly on a Message, versus having an explicit MessageProperties object, which is then assigned to the Message?

draft-ietf-taps-interface.md
```diff
@@ -1185,30 +1193,41 @@ number of Messages with a single call. This allows an application to provide
...
-many bytes if they are available; the interface may return partial Messages
-smaller than maxLength according to implementation constraints.
+Partial Messages are indicated by marking the endOfMessage boolean value
+as false in the ReceivedHandler. Note that maxLength does not guarantee that 
+the application will receive that many bytes if they are available; the interface may
+return partial Messages smaller than maxLength according to implementation constraints.
 
 ~~~
-Connection -> Received<Message>
+Connection -> Received<Message, messageData, endOfMessage>
```
britram: as with send, having a thing called a Message as well as messageData also on receive is... confusing. Renaming this to `MessageContext` would fix this at the expense of verbosity and slight obtuseness...

draft-ietf-taps-interface.md
theri: I'm wondering how this works with, e.g., "Final" messages. If Message Properties cannot be changed after Send, the application cannot reuse a Message object with an additional "Final" parameter added, but needs to create a new Message object just for this purpose, adding all properties of the previous messages to it, and "Final". Is this intended?

draft-ietf-taps-interface.md
```diff
@@ -1015,33 +1032,24 @@ set of send parameters not consistent with the Connection's transport
...
 
-## Send Parameters {#send-params}
+## Message Send Parameters {#send-params}
 
-The Send Action takes per-Message send parameters which control how the
-contents will be sent down to the underlying Protocol Stack and transmitted.
+Each Message can be marked with Send Parameters to control how Data
+associated with the Message will be sent down to the underlying Protocol Stack 
+and transmitted. Note that these properties are per-Message, not per-Send.
+All data portions associated with a single Message share properties. For example,
```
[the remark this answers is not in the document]
mwelzl: Data blocks?

draft-ietf-taps-interface.md
mwelzl: This is THE problem I have with this whole thing - it's also the question that I raised at the end of #200 and don't see addressed. Partial messages may arrive out of order, or they may never arrive at all, if we map them on an unreliable transport. At least, applications must have a number to identify that they're now getting message fragment 3 out of 7, something like that.

draft-ietf-taps-interface.md
```diff
@@ -945,29 +945,46 @@ capacity that it sees fit.
...
 
 Once a Connection has been established, it can be used for sending data. Data
-is sent by passing a Message Object and additional parameters
-{{send-params}} to the Send Action on an established Connection:
+is always sent in association with a Message Object and additional parameters
+({{send-params}}) using the Send Action on an established Connection:
 
 ~~~
-Connection.Send(Message, sendParameters)
+Connection.Send(Message, messageData, endOfMessage)
```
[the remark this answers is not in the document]
mwelzl: I find the split between Message and messageData awkward. Also, parameters are logically separate from the message itself, so it makes sense to write this as we had it - but "endOfMessage" is just one out of many message parameters, so why have this as a "special" parameter in the send call?

draft-ietf-taps-interface.md
mwelzl: And what about partial reliability?  E.g., fragments 1 and 2 out of a total of 3 arrive in time, but for the third, the time expires and the transport gives up. Shouldn't it then tell the receiving application that it will never get the third fragment, and should throw away the first two?

mwelzl:
@csperkins - your idea sounds interesting, but from your description I don't understand why this reduces complexity in the API.

I think the complexity is related to ordering. If we assume that, on the receiver side, message fragments are only delivered in order, things become easier. Then we may only need an extra signal to tell the receiver that remaining fragments will never arrive.

philsbln:
I like the general idea and I agree having an explicit message object really is a way out, but I also think the API in this PR exposes too many details and makes the "common" case unnecessarily complicated.

I like @csperkins Idea to move that partial send complexity into the message object.

draft-ietf-taps-interface.md
```diff
@@ -1185,30 +1193,41 @@ number of Messages with a single call. This allows an application to provide
...
+protocol, and provides methods to access metadata about the received data.
+The octets of data associated with this message are delivered as messageData.
+Multiple invocations of the ReceivedHandler may devlier data for the same
+Message until the endOfMessage flag is delivered. See {{receive-framing}} for 
+handling Message framing in situations where the Protocol Stack provides 
+octet-stream transport only.
 
-The Message Object passed to Received is complete and atomic, unless one of the following
+If the minIncompleteLength was set to be infinite, the Message passed to Received
+will complete and atomic (that is, endOfMessage will be set), unless one of the following
```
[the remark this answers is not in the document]
tfpauly: i.e. literally means "that is", so I think this is about the same?

draft-ietf-taps-interface.md
tfpauly: If you need to do a write close with Final, you should mark that on your Message in the first send. If you need to send a partial Message that isn't necessarily the Final one, then don't mark it Final. You can later send a complete message that is final with no data.

draft-ietf-taps-interface.md
tfpauly: My inclination here is to only receive chunks in order and not deliver gaps. Does that work for you?

britram:
on second thought, let me +1 @csperkins on "move partial send complexity into the message object":

- to send a partial message, get a message and keep writing bytes to it.

- to receive a partial message, get a message that indicates that the octets read from it are not the whole message.

draft-ietf-taps-interface.md
```diff
@@ -1015,33 +1032,24 @@ set of send parameters not consistent with the Connection's transport
...
-empty sent parameter Object can be acquired and all desired Send Parameters
-can be added to that Object. A sendParameters Object can be reused for
-sending multiple contents with the same properties.
+Parameters may be added to a Message object before the Message is used
+for sendings.
 
 ~~~
-SendParameters := NewSendParameters()
-SendParameters.Add(parameter, value)
+Message := NewMessage()
```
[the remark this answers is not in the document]
tfpauly: The data needs to be specific to the send action, not the Message, or else the send action becomes somewhat meaningless. You could call the Message object just MessageContext or MessageParameters or MessageMetadata, which would be a bit more accurate.

draft-ietf-taps-interface.md
tfpauly: I like MessageContext

draft-ietf-taps-interface.md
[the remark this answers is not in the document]
tfpauly: Fixed

draft-ietf-taps-interface.md
```diff
@@ -945,29 +945,46 @@ capacity that it sees fit.
 # Sending Data {#sending}
 
 Once a Connection has been established, it can be used for sending data. Data
-is sent by passing a Message Object and additional parameters
-{{send-params}} to the Send Action on an established Connection:
+is always sent in association with a Message Object and additional parameters
```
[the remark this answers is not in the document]
tfpauly: Fixed

draft-ietf-taps-interface.md
```diff
@@ -945,29 +945,46 @@ capacity that it sees fit.
...
+For example, a Message may be a single datagram for UDP Connections; or an HTTP
+Request for HTTP Connections. Various Send Parameters can be configured before using 
+a Message to send, which influence the priority and treatment of the Message's data upon 
+sending. See {{send-params}} for a description of each parameter.
+
+~~~
+Message := NewMessage()
+~~~
+
+The messageData parameter contains any octects to be sent for this message.
```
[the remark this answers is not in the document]
tfpauly: Fixed

draft-ietf-taps-interface.md
tfpauly: @chris-wood Added text

draft-ietf-taps-interface.md
tfpauly: Added text

tfpauly:
@csperkins Right, having Message objects be the active objects for sending and receiving is the other option. That's the one that we had implemented back when we were working on Minion, years ago. It certainly can work, but adds a ton of complexity to the application workflow, and reduces the ability to have meaningful back pressure and flow control.

I think that to a large degree, this PR is just writing out the underlying way that sending and receiving in the current draft works. Here's the long form of sending that works for partial and complete:

Send(messageContext, messageData, endOfMessage)

The default for endOfMessage should be true, therefore most applications will just see:

Send(messageContext, messageData)

And the the default for messageContext should be the default parameters for a new message, so most
applications (using UDP, for example) will just see:

Send(messageData)

The same is true for receive. We can give the variants of Send to be partial or not different names, but they boil down to convenience/default sets of the same thing.

The key points here are these:
1. The properties of a message need to be the same across multiple partial sends if there are multiple partial sends. If the application is streaming a long message, then the properties of that message apply to all of the data.
2. The completion handlers for Send actions that provide notions of back pressure and errors *must* be per data chunk sent, not per message. If I am streaming a long message, then I will likely need to handle flow control, and the completion handler for send can't just be on the whole message, it needs to be on the partial chunks.

If you move to sending data on a new Message object rather than the Connection, you would need to have back pressure on those sends too. It becomes very complicated for the application:

Connection.Send(Message) -> MessageSentHandler[May have error but can't be used for flow control]
Message.Send(bytes) -> DataSentHandler[Used for per-message flow control, which now becomes difficult to coordinate with other messages if you have a stream with HOL blocking]

Going down this road means that partial sends become front-and-center for applications dealing with the API, even when they don't want it, and adds an extra layer of indirection just to get basic bytes on the wire for a transport.

The approach in this PR ends up being a way to allow expressiveness for partial sends when they are needed, but allows applications to ignore it most of the time with defaults and write the simplest code.

tfpauly:
Right, having the default support makes things much easier. For languages that support defaults (Swift, for us), you can do the scheme like I had above. For languages like C, you can write out the common sets of calls, which all can be implemented using the most complex:

`connection_partial_send(data_t data, message_context_t message, bool end_of_message)`
`connection_send(data_t data, message_context_t message)`
`connection_default_message_send(data_t data)`

draft-ietf-taps-interface.md
mwelzl: @tfpauly I think that's a good direction, as it makes things much easier. I think we should then still add a way for the transport system to signal to the application: "throw away your chunks, the rest is never going to arrive" - e.g. to support partial reliability.

But I have a bigger concern, then: what about UDP? What if I send a large message, the chosen transport is UDP, and the message is split it into 3 packets - first, this is perhaps not a great idea; second, the receiver won't be able to know about the order, so can't even do what you suggest here. How are you going to handle this case?

abrunstrom:
+1 on introducing defaults. The API is beginning to get quite complex. Maybe we should explicitly split up the send section in "Basic Send Operations" and "Advanced Send Functionality" or something like that. Splitting up other relevant sections in a similar way.

tfpauly:
@abrunstrom Sure, I can work tomorrow on splitting out the definition of the simple send operations, versus the more full story for partial sends. We can mention as an implementation detail that everything in the simpler approach for atomic messages can be expressed in terms of defaults in the variants that support partial sends, and leave the specific symbol bindings as a per-language choice.

draft-ietf-taps-interface.md
tfpauly: So, in our implementation, we allow the protocol to require that data for a single message fits within a limit (for UDP, whatever it can fit in a datagram). Splitting up a message over multiple UDP datagrams without any other framing layer on top doesn't make any sense, since there isn't any way for the other side to understand the message. So, in our implementation, we throw an error if you try to do this.

Protocols on top of UDP, like IKE, may support fragmenting large messages across datagrams, but then the message that you send is first interpreted by a protocol in the stack that supports unbounded message sizes.

I also do agree that there should be a signal that the chunks won't ever arrive—our implementation passes an optional error along with the ReceiveHandler to indicate this.

mwelzl: Ah ok!  Great, I think that solves it. I'll conclude by requesting the following text:
1) in the API draft, some kind of a signal that the chuncks won't error arrive (as we agreed)
2) in the API draft, for the receiver side, a hint that message chunks will only arrive in order
3) in the implementation draft, a hint that message fragmentation of course requires support by the underlying protocol and thus the size of any message, even with fragmentation, will have to be limited to the message size that the underlying protocol supports. For example, messages can not be spread across multiple UDP packets because the receiver would not understand that the message parts belong together.

I can take care of 2 and 3 (but I'm fine with you doing it too if you want!  Just offering to help instead of always only criticizing!) .... but for 1), I think it would be better for you to do it since your implementation already does it.  Ok?

draft-ietf-taps-interface.md
chris-wood: Yep -- I made the comment before getting to the next paragraph. Maybe add: "Both minIncompleteLength and maxLength are optional parameters." at the beginning of this paragraph, and then proceed to describe them both?

chris-wood:
+1 to @abrunstrom's recommendation.

britram:
addressing @abrunstrom's comment here would make this ready to land IMO. I'd like to get this and #201 in soon so we can submit -01 in time, and keep working on this in person in Montreal.

tfpauly:
Okay, just updated the Sending side to break out basic sends, sending with context, and partial sends. I'll do a bit more work on receiving too today, but take a look at the sending side and see if that addresses the comments there.

abrunstrom:
Thanks @tfpauly for the restructuring, I like the new structure! I find it much more clear and readable.

One question on the Events in the relation to the partial sends. If I understand it correctly Sent Events relate to specific send calls whereas Expired Events relate to an entire Message. Not sure about SendError Events? Seems there it may depend on what the error is. Not sure if this is something that needs to be clarified further.

tfpauly:
@abrunstrom I personally think it's simplest if the events correspond to Send events, so if you did a 2 partial sends for a single message that expired, you should get two Expired events if neither could be sent, etc.

abrunstrom:
Makes sense, thanks.

tfpauly:
I've added the rest of my reworked text, to bring Receives into the same style as Send (with the basic approach up front, and partial phrased as an extension later on). Thanks for all of the good feedback, I think this text is certainly clearer!

@abrunstrom I added some text about how Send events are delivered when there are multiple Sends per Message, please check.

@mwelzl I added text to explain that receiving is only in-order contiguous chunks, and clarified that you can't span a message over multiple datagrams directly, etc.

abrunstrom:
Thanks @tfpauly , the added text looks good to me!

draft-ietf-taps-interface.md
```diff
@@ -1163,53 +1212,94 @@ it is bound to the Preconnection during the pre-establishment phase:
...
+The application can set a minIncompleteLength value to indicates the smallest partial 
+Message data size in bytes that should be delivered in response to this Receive. By default, 
+this value is infinite, which means that only complete Messages should be delivered. 
+If this value is set to some smaller value, the associated recieve event will be triggered 
+only when at least that many bytes are available, or the Message is complete with fewer
+bytes, or the system needs to free up memory. Applications should always
+check the length of the data delivered to the receive event and not assume
+it will be as long as minIncompleteLength in the case of shorter complete Messages
+or memory issues.
+
```
abrunstrom: I understand that the applications should derive the length of the delivered data from the messageData. Perhaps this should be made more explicit? Before I got to the ReceivedPartial Event when reading, where the messageContext was reused for multiple events, I was thinking that the length of the data delivered may be provided to the application as part of the messageContext meta data.

tfpauly: Good point, I'll add an explicit comment about this.

draft-ietf-taps-interface.md
```diff
@@ -944,38 +944,55 @@ capacity that it sees fit.
...
-The details of partial Message sending are implementation-dependent.
+protocol property Maximum Message Size on Send to determine the maximum size
+allowed for a single Message. If a Message is too large to fit in the Maximum Message
+Size for the Connection, the Send will fail with a SendError event ({{send-error}}). For
+example, it is invalid to send a Message over a UDP connection that is larger than
+the available datagram sending size.
 
 If Send is called on a Connection which has not yet been established, an
 Initiate Action will be implicitly performed simultaneously with the Send.
 Used together with the Idempotent property (see {{send-idempotent}}), this can
```
mwelzl: A nit: I suggest to remove the "Used" at the beginning of this sentence

draft-ietf-taps-interface.md
```diff
@@ -1163,53 +1213,97 @@ it is bound to the Preconnection during the pre-establishment phase:
...
-convenience functions to indicate readiness to receive a larger but finite
-number of Messages with a single call. This allows an application to provide
-backpressure to the transport stack when it is temporarily not ready to
-receive messages.
+By default, Receive will try to deliver complete Messages in a single event ({{receive-complete}}).
+
+The application can set a minIncompleteLength value to indicates the smallest partial 
+Message data size in bytes that should be delivered in response to this Receive. By default, 
+this value is infinite, which means that only complete Messages should be delivered. 
+If this value is set to some smaller value, the associated recieve event will be triggered
```
mwelzl: typo "recieve"

csperkins:
@tfpauly wrote:
>The key points here are these:
>
>The properties of a message need to be the same across multiple partial sends if there are multiple partial sends. If the application is streaming a long message, then the properties of that message apply to all of the data.
>The completion handlers for Send actions that provide notions of back pressure and errors must be per data chunk sent, not per message. If I am streaming a long message, then I will likely need to handle flow control, and the completion handler for send can't just be on the whole message, it needs to be on the partial chunks.

Agree with these.

>If you move to sending data on a new Message object rather than the Connection, 

I wasn't arguing to move the sending data onto the Message object. In my approach, the Connection would still handle sending the data, but rather than have the application provide chunks of data, I was suggesting to have the connection pull data from the message object as needed.

>you would need to have back pressure on those sends too. It becomes very complicated for the application:

Don't we need that back pressure anyway? If one sends a message object from which data is pulled as needed, then I agree the connection would need to generate `PartiallySent` events before the final `Sent` event. But, we don't avoid that need by splitting the data up at the application level and making partial send requests: with a transport like TCP the connection can always block, and will need to generate a `PartiallySent` event, no matter whether the application has split the data into chunks before calling send or not.

It's perhaps easier for the application to change what it sends, or cancel the send, if it send chunk-by-chunk – that I agree – but otherwise I don't see a complexity saving.

draft-ietf-taps-interface.md
tfpauly: Fixed! Thanks for catching.

draft-ietf-taps-interface.md
tfpauly: Agreed, removed.

tfpauly:
@csperkins Regarding backpressure, you're right that we do need backpressure on all chunks that are sent. However, by allowing partial sends to be expressed in the native send action of the Connection, they automatically get the same backpressure that complete sends get. Each Sent event is one-to-one with a Send action, whether partial or complete.

tfpauly:
Okay, I think based on the reviews we're in a good state to land this, and I do want to unblock merging for #201. As such, I'm going to go ahead and merge.

PR #201, opened 2018-06-25
Restructure Transport Properties – closes #153 and #190

### LINKED ISSUE #153: Restructure Transport Parameters Section

philsbln:
This PR should close Issue #153 and #190.

Change summary:
 - Unify terminology: all (Selection/Protocol/Message)-(Preferences/Properties/Parameters) become Properties
   - Adjusted Section 5.2 
   - Adjusted Section 7.1
   - Adjusted Section 9 
 - Add proper Property Types
   - Describe how Preferences from Preconnections are reflected on Connections (selection results)
 - Text on Property Classification
   - Added "Control Properties" to catch transport system state like "Final" or notifications
   - Added description of Intents (to classify "Capacity Profile")
 - Move all Properties to a single section (Section 12)
   - Unify Property Formatting
   - Merge some message/connection properties
   - Fix property types

There are some ToDos left in the text – mostly already existing inconsistencies that have become obvious due to the new structure/formatting.

### DIFF
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -369,16 +369,17 @@ TransportProperties := NewTransportProperties()
 
 The Individual properties are then added to the TransportProperties Object.
 While all Properties use the `add` call, Transport Properties of Preference
-Type can use special calls for the levels defined in {{transport-props-preference}}.
+Type can use special calls to add a Property with a specific preference level,
+see also {{transport-props-preference}}.
 
 ~~~
 TransportProperties.Add(property, value)
 
-TransportProperties.Require(preference)
-TransportProperties.Prefer(preference)
-TransportProperties.Ignore(preference)
-TransportProperties.Avoid(preference)
-TransportProperties.Prohibit(preference)
+TransportProperties.Require(property)
+TransportProperties.Prefer(property)
+TransportProperties.Ignore(property)
+TransportProperties.Avoid(property)
+TransportProperties.Prohibit(property)
 ~~~
 
 For an existing Connection, the Transport Properties can be queried any time
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -91,6 +91,7 @@ normative:
 
 informative:
   I-D.ietf-taps-transport-security:
+  RFC7556:
 
 --- abstract
 
@@ -228,8 +229,9 @@ programming interface to the Transport Services Architecture defined in
 two Objects, Preconnections and Connections. A Preconnection represents a set of
 parameters and constraints on the selection and configuration of paths and
 protocols to establish a Connection with a remote endpoint. A Connection
-represents a transport Protocol Stack on which data can be sent to and received
-from a remote endpoint. Connections can be created from Preconnections in three
+represents a transport Protocol Stack on which data can be sent to and/or received
+from a remote endpoint (i.e., depending on the kind of transport, connections can be
+bi-directional or unidirectional). Connections can be created from Preconnections in three
 ways: by initiating the Preconnection (i.e., actively opening, as in a client),
 through listening on the Preconnection (i.e., passively opening, as in a
 server), or rendezvousing on the Preconnection (i.e. peer to peer establishment).
@@ -327,7 +329,8 @@ The API does not need the application to resolve names, and premature name
 resolution can damage performance by limiting the scope for alternate path
 discovery during Connection establishment.
 The Resolve() method is, however, provided to resolve a Local Endpoint or a
-Remote Endpoint in cases where this is required, for example with some NAT
+Remote Endpoint in cases where this is required, for example with some Network
+Address Translator (NAT)
 traversal protocols (see {{rendezvous}}).
 
 ## Specifying Transport Parameters {#transport-params}
@@ -459,6 +462,23 @@ useful to indicate its reliability requirements on a per-Message basis.
 This property applies to Connections and Connection Groups. This is not a
 strict requirement.  The default is to not have this option.
 
+### Direction of communication
+
+Type: Enum
+
+This property specifies whether an application wants to use the connection for sending and/or receiving data.  Possible values are:
+
+Bidirectional (default):
+: The connection must support sending and receiving data
+
+unidirectional send:
+: The connection must support sending data. 
+
+unidirectional receive:
+: The connection must support receiving data
+
+In case a unidirectional connection is requested, but the system should fall back to bidirectional transport if unidirectional connections are not  supported by the transport protocol.
+
 ### Use 0-RTT session establishment with an idempotent Message {#prop-0rtt}
 
 Type: Preference
@@ -521,18 +541,60 @@ strict requirement, as it signifies a reduction in reliability. The
 default is full checksum coverage without being able to change it, and
 requiring a checksum when receiving.
 
-### Interface Type {#prop-intf-type}
+### Interface Instance or Type {#prop-interface}
 
 Type: Tuple (Enumeration, Preference)
 
-This property specifies which kind of access network interface,
-e.g., WiFi, Ethernet, or LTE, to prefer over others for this Connection, in
-case they are available. In general, Interface Types should be used only with
-the `Prefer` and `Prohibit` preference level. Specifically, using the
-`Require` preference level for Interface Type may limit path selection in a
-way that is detrimental to connectivity. The default is to use the default
-interface configured in the system policy.
-The valid values for the access network interface kinds are implementation specific.
+This property allows the application to select which specific network interfaces
+or categories of interfaces it wants to `Require`, `Prohibit`, `Prefer`, or `Avoid`.
+
+If a system supports discovery of specific interface identifiers, such as `en0` or `eth0`
+on Unix-style systems, an implemention should allow using these identifiers to
+define path preferences. Note that marking a specific interface as `Required`
+strictly limits path selection to a single interface, and leads to less flexible and
+resilient connection establishment.
+
+The set of valid interface types is implementation- and system-specific. For example,
+on a mobile device, there may be `Wi-Fi` and `Cellular` interface types available;
+whereas on a desktop computer, there may be `Wi-Fi` and `Wired Ethernet`
+interface types available. Implementations should provide all types that are supported
+on some system to all systems, in order to allow applications to write generic code.
+For example, if a single implementation is used on both mobile devices and desktop
+devices, it should define the `Cellular` interface type for both systems, since an
+application may want to always `Prohibit Cellular`. Note that marking a specific 
+interface type as `Required` limits path selection to a small set of interfaces, and leads
+to less flexible and resilient connection establishment.
+
+The set of interface types is expected to change over time as new access technologies
+become available.
+
+Interface types should not be treated as a proxy for properties of interfaces such as
+metered or unmetered network access. If an application needs to prohibit metered
+interfaces, this should be specified via Provisioning Domain attributes {{prop-pvd}}
+or another specific property.
+
+### Provisioning Domain Instance or Type {#prop-pvd}
+
+Type: Tuple (Enumeration, Preference)
+
+Similar to interface instances and types {{prop-interface}}, this property allows
+the application to control path selection by selecting which specific Provisioning Domains
+or categories of Provisioning Domains it wants to  `Require`, `Prohibit`, `Prefer`, or `Avoid`.
+Provisioning Domains define consistent sets of network properties that may be more
+specific than network interfaces {{RFC7556}}.
+
+The indentification of a specific Provisioning Domain (PvD) is defined to be implementation-
+and system-specific, since there is not a portable standard format for a PvD identitfier.
+For example, this identifier may be a string name or an integer. As with 
+requiring specific interfaces, requiring a specific PvD strictly limits path selection.
+
+Categories or types of PvDs are also defined to be implementation- and system-specific.
+These may be useful to identify a service that is provided by a PvD. For example, if an application
+wants to use a PvD that provides a Voice-Over-IP service on a Cellular network, it can use
+the relevant PvD type to require some PvD that provides this service, without needing to
+look up a particular instance. While this does restrict path selection, it is more broad than
+requiring specific PvD instances or interface instances, and should be preferred over those
+options.
 
 ### Capacity Profile {#prop-cap-profile}
 
@@ -563,13 +625,25 @@ The following values are valid for Capacity Profile:
   expense of bandwidth efficiency. This implies that the Connection may fail
   if the desired rate cannot be maintained across the Path. A transport
   may interpret this capacity profile as preferring a circuit breaker
-  {{?RFC8084}} to a rate adaptive congestion controller.
+  {{?RFC8084}} to a rate-adaptive congestion controller.
 
   Scavenger/Bulk:
   : The application is not interactive. It expects to send/receive a large
-  amount of data, without any urgency. This can be used to select protocol
+  amount of data, without any urgency. This can, for example, be used to select protocol
   stacks with scavenger transmission control, to signal a preference for
-  less-than-best-effort treatment, and so on.
+  less-than-best-effort treatment, or to assign the traffic to a lower-effort service.
+
+### Congestion control {#prop-cc}
+
+Type: Preference
+
+This property specifies whether the application would like the Connection to be
+congestion controlled or not. Note that if a Connection is not congestion controlled,
+an application using such a Connection should itself perform congestion control in
+accordance with {{?RFC2914}}. Also note that reliability is usually combined with
+congestion control in protocol implementations, rendering "reliable but not congestion
+controlled" a request that is unlikely to succeed.
+
 
 ## Specifying Security Parameters and Callbacks {#security-parameters}
 
@@ -788,9 +862,11 @@ for listening, when the Local Endpoint or Remote Endpoint cannot be resolved,
 when no transport-layer connection can be established to the Remote Endpoint,
 or when the application is prohibited from rendezvous by policy.
 
-When using some NAT traversal protocols, e.g., ICE {{?RFC5245}}, it is
+When using some NAT traversal protocols, e.g., Interactive Connectivity
+Establishment (ICE) {{?RFC5245}}, it is
 expected that the Local Endpoint will be configured with some method of
-discovering NAT bindings, e.g., a STUN server. In this case, the
+discovering NAT bindings, e.g., a Session Traversal Utilities for NAT (STUN) server.
+In this case, the
 Local Endpoint may resolve to a mixture of local and server reflexive
 addresses. The Resolve() method on the Preconnection can be used to
 discover these bindings:
@@ -838,9 +914,9 @@ lower Niceness values. An ideal transport system implementation would assign
 the Connection the capacity share (M-N) x C / M, where N is the Connection's
 Niceness value, M is the maximum Niceness value used by all Connections in the
 group and C is the total available capacity. However, the niceness setting is
-purely advisory, and no guarantees are given about capacity allocation and
-each implementation is free to implement exact capacity allocation as it sees
-fit.
+purely advisory, and no guarantees are given about the way capacity is shared.
+Each implementation is free to implement a way it shares
+capacity that it sees fit.
 
 # Sending Data {#sending}
 
@@ -936,8 +1012,9 @@ be overridden on a per-Message basis.
 
 Send Parameters may be inconsistent with the properties of the Protocol Stacks
 underlying the Connection on which a given Message is sent. For example,
-infinite Lifetime is not possible on a Message over a Connection not providing
-reliability. Sending a Message with Send Properties inconsistent with the
+a Connection must provide reliability to allow setting an infinitie value for the
+lifetime property of a Message. Sending a Message with Send Properties
+inconsistent with the
 Transport Preferences on the Connection yields an error.
 
 The following send parameters are supported:
@@ -952,17 +1029,18 @@ implementation-specific.
 
 ### Niceness {#send-niceness}
 
-Niceness represents an unbounded hierarchy of priorities of Messages, relative
+Niceness is a numeric (non-negative) value that represents an
+unbounded hierarchy of priorities of Messages, relative
 to other Messages sent over the same Connection and/or Connection Group (see
-{{groups}}). It is most naturally represented as a non-negative integer.
+{{groups}}).
 A Message with Niceness 0 will yield to a Message with Niceness 1, which will
 yield to a Message with Niceness 2, and so on. Niceness may be used as a
 sender-side scheduling construct only, or be used to specify priorities on the
 wire for Protocol Stacks supporting prioritization.
 
-Note that this inversion of normal schemes for expressing priority has a
-convenient property: priority increases as both Niceness and Lifetime
-decrease.
+This encoding of the priority has a convenient property that the priority
+increases as both Niceness and Lifetime decrease.
+
 
 ### Ordered {#send-ordered}
 
@@ -1017,7 +1095,7 @@ The following values are valid for Transmission Profile:
 
   Low Latency:
   : Response time (latency) should be optimized at
-  the expense of bandwidth efficiency and delay variation when sending this
+  the expense of efficiently using the available capacity when sending this
   message. This can be used by the system to disable the coalescing of
   multiple small Messages into larger packets (Nagle's algorithm); to prefer
   immediate acknowledgment from the peer endpoint when supported by the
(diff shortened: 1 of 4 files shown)


britram:
this is a monster PR and will take a while even to review. As it stands, I don't anticipate it landing before Montreal. Alternately, we could just land it as is in order to stop blocking, and file a flurry of issues on anything we find wrong with it (I have a few already, in a pending review...)

britram:
(Other reviewers: I suggest just looking at the document in its entirety with this PR here: https://taps-api.github.io/drafts/properties/draft-ietf-taps-interface.html -- I'm finding it much easier to read it that way than trying to navigate the giant diff on the MD)

philsbln:
Sorry for the giant PR @britram – we already tried to split it up in more or less useful commits, but tracking text moves is always tedious. 

I would be in favour of landing it and ripping the Properties apart afterwards. No matter what the consensus of the WG will finally be, in this form it should be much easier to discuss the problems and inconsistencies regarding properties than in the previous structure.

I also see the categories and data types rather as something to start discussing than something I will fight for in every detail.

britram:
> we already tried to split it up in more or less useful commits, but tracking text moves is always tedious.

no problem, I don't think there was any other way to do this; just trying to figure out how to make best use of our time on this in Montreal.

britram:
(I will say that seeing the properties expressed like this was much more useful to me than the Venn diagram in understanding the arrangement of things, and halfway through my review I think it needs a bit of tweaking, on which more later.)

philsbln:
Thank you for your high-level feedback @abrunstrom. 

 - I just did another pass through the document to fix left-over parameters, so I hope your first issue is addressed now. Security Parameters still are Parameters to not divert from TLS terminology. 
 - To your second issue, I agree that a little more cross-references and explanation int he might be needed for better readability, but in my opinion, moving the vast amount of properties to Section 12 really improved readability of most parts of the document and makes the general idea of the API clearer.
 - As many Protocol Properties apply to connections as well as messages, we find the term "Message Properties" misleading for classification of properties. The new terminology is really orthogonal:
   Everything is a Transport Property
     - Transport properties can be applied to different objects
         - Preconnections (Connection Properties) 
         - Connections (also Connection Properties) 
         - Messages (Message Properties) 
      - Transport properties can be classified by what and how they Effect things
         - Selection Properties (effect the selection process)
         - Protocol Properties (effect protocols selected) 
         - Control Properties  (effect the transport system / protocol state machine)
         - Intents (may effect anything after being interoreted by the transport system)

draft-ietf-taps-interface.md
```diff
@@ -1502,6 +1097,732 @@ does provide the following guarantees about the ordering of operations:
...
+  {{?RFC8084}} to a rate-adaptive congestion controller.
+
+  Scavenger/Bulk:
+  : The application is not interactive. It expects to send/receive a large
+  amount of data, without any urgency. This can, for example, be used to select protocol
+  stacks with scavenger transmission control, to signal a preference for
+  less-than-best-effort treatment, or to assign the traffic to a lower-effort service.
+
+
+### Congestion control {#prop-cc}
```
britram: This seems potentially dangerous.

draft-ietf-taps-interface.md
```diff
@@ -1502,6 +1096,732 @@ does provide the following guarantees about the ordering of operations:
...
 
+
+# Transport Properties {#transport-props}
+
+Transport Properties allow an application to control most aspects of the
+transport system.
+
+Properties are structured in two ways: 
+
+ - By how they influence the transport system, which leads to a classification
```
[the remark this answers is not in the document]
mwelzl: I very much agree with this. @abrunstrom also suggested to remove "intent" but I agree that even "protocol" vs. "control" doesn't seem a very clear line. Can we just have these two types - "selection properties" and, e.g., "configuration properties", to start with, and think about dividing the latter further only if it makes sense?

mwelzl:
Hi,

Thank you very much for this PR! I have now also given this a close read; my first reaction is that this structural re-working was much needed and does indeed clean things up nicely. While I understand the concern of @abrunstrom about things being harder to read with all these forward references, I also agree with @philsbln 's answer that this move was worth it and made it all clearer.

I don't understand this concern of @britram:

> The arrangement of properties -- most importantly, that Message properties are simply protocol properties -- is IMO probably incorrect.

... as I can't see this PR stating that Message properties are protocol properties.

Generally I'd also be in favor of landing this and then fine-tuning it, either way I think it's a better basis for the document structure, better get this in earlier than later.

draft-ietf-taps-interface.md
mwelzl: What?

britram:
reading @mwelzl's comment, IMO we should converge very soon on "land this for -01 and fix it later" versus "hold for post-Montreal". 

My 2c: I think the structural changes to the document are correct. I'm less sure of the organization of properties (sharing @abrunstrom's and @mwelzl's concern) but these are details. I have a few picky questions I would like to see addressed before merging though (in the review). IMO we can land, submit -01, discuss in the WG meeting, and fix for a -02 after Montreal.

On this:

> ... as I can't see this PR stating that Message properties are protocol properties.

I'm possibly misreading the classification, and possibly not expressing myself very well. I'll try to draw a picture to say what I mean soon.

britram:
on making a fast decision: @tfpauly please come have a look and chime in on "land and fix" versus "hold for fix"

draft-ietf-taps-interface.md
```diff
@@ -1502,6 +1097,732 @@ does provide the following guarantees about the ordering of operations:
...
+supply a Message to the transport protocol before Connection
+establishment, which will then be reliably transferred to the other side
+before or during Connection establishment, potentially multiple times.
+See also {{send-idempotent}}. The default is to not have this option.
+
+
+### Idempotent {#send-idempotent}
+
+Classification: 
+: Protocol Property (Generic)
```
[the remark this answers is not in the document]
philsbln: Ack - Fixed

draft-ietf-taps-interface.md
```diff
@@ -1502,6 +1097,732 @@ does provide the following guarantees about the ordering of operations:
...
+: Preconnection, Connection (read only)
+
+This property specifies whether an application would like to
+supply a Message to the transport protocol before Connection
+establishment, which will then be reliably transferred to the other side
+before or during Connection establishment, potentially multiple times.
+See also {{send-idempotent}}. The default is to not have this option.
+
+
+### Idempotent {#send-idempotent}
```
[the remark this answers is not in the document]
philsbln: It could be (as Final) – exactly that is what makes it a "Control Propery"

draft-ietf-taps-interface.md
```diff
@@ -361,296 +349,63 @@ transport protocol selection is necessarily tied to path selection. This may
...
 ~~~
-TransportParameters.Add(parameter, value)
+TransportProperties.Add(parameter, value)
 
-TransportParameters.Require(preference)
-TransportParameters.Prefer(preference)
-TransportParameters.Ignore(preference)
-TransportParameters.Avoid(preference)
-TransportParameters.Prohibit(preference)
+TransportProperties.Require(preference)
```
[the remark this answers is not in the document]
theri: Basically, TransportProperties.Require(property) is syntactic sugar for TransportProperties.add(property, value=Require). 
I'm not sure if this syntactic sugar should be in there, but I just pushed a commit to hopefully make it clearer that of course the property is the argument to the call, and the call just sets a specific preference level for the property.

abrunstrom:
I think landing this now is fine to allow other parts to progress after restructuring. I think we will find out if we in the end can push all information about the Transport Properties to one section, or if some parts need to be introduced earlier, but I agree it is good to have it all there for the discussion. 
If we land it for Montreal I would like us to simplify the structure of the properties before Montreal though. As also pointed out by @mwelzl  it is better to start simple and then add additional classes if really needed. So I am very much in favor of having just two classes to start with: Selection and one more (Protocol or Configuration).

draft-ietf-taps-interface.md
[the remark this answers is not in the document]
abrunstrom: Why is this a control property? Then most message properties should be control properties?

draft-ietf-taps-interface.md
```diff
@@ -1502,6 +1096,732 @@ does provide the following guarantees about the ordering of operations:
...
+
+The application can expect the transport system to take appropriate actions
+involving protocol selection, path selection and, setting of protocol flags.
+For example, if an application sets the "Capacity Profile" to "bulk" on a 
+Preconnection, this will likely influence path selection, DSCP flags in
+the IP header as well as niceness for multi-streaming connections.
+When using Intents, the application must not expect consistent behavior across 
+different environments, implementations or versions of the same implementation.
+
+## Mandatory Transport Properties
```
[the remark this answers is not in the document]
philsbln: We basically took the old order from the Connection Properties and merged everything else in a way that related stuff is directly adjacent. Afterward, we moved some things up that may need special attention, e.g., final.

The resulting order is, in our opinion, somewhat useful for discussion but far from final.

draft-ietf-taps-interface.md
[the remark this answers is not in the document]
philsbln: This is just syntactic sugar now - got just clarified in the text

draft-ietf-taps-interface.md
```diff
@@ -1502,6 +1096,732 @@ does provide the following guarantees about the ordering of operations:
...
+The representation is implementation dependent.
+
+### Integer {#transport-props-int}
+
+Integer types are used to represent integer numbers. 
+The representation is implementation dependent.
+
+### Preference {#transport-props-preference}
+
+The Preference type is used in most Selection properties on a Preconnection object to constrain Path Selection and Protocol Selection.
```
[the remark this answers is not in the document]
philsbln: see above

draft-ietf-taps-interface.md
```diff
@@ -1502,6 +1096,732 @@ does provide the following guarantees about the ordering of operations:
   were sent (i.e., delivered to the kernel or to the network interface,
   depending on implementation).
 
+
+# Transport Properties {#transport-props}
+
+Transport Properties allow an application to control most aspects of the
```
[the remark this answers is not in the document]
philsbln: ack; addressed

draft-ietf-taps-interface.md
```diff
@@ -1502,6 +1096,732 @@ does provide the following guarantees about the ordering of operations:
...
+Preconnection, this will likely influence path selection, DSCP flags in
+the IP header as well as niceness for multi-streaming connections.
+When using Intents, the application must not expect consistent behavior across 
+different environments, implementations or versions of the same implementation.
+
+## Mandatory Transport Properties
+
+The following properties are mandatory to implement in a transport system:
+  
+### Final {#send-final}
```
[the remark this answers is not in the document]
philsbln: True for most control properties that apply to messages – putting it here is a trade-off between cluttering the `Send()` call and "hiding" it in the properties.

draft-ietf-taps-interface.md
```diff
@@ -1502,6 +1096,732 @@ does provide the following guarantees about the ordering of operations:
...
+Type: 
+: Boolean
+
+Applicability: 
+: Message 
+
+This property specifies that a message should be sent in such a way
+that the transport protocol ensures all data is received on the other side
+without corruption. Reliable Data Transfer can only be used on Messages if the
+transport protocol supports partial reliability (see {{prop-partially-reliable}}).
```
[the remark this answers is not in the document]
philsbln: clarified

draft-ietf-taps-interface.md
```diff
@@ -1502,6 +1096,732 @@ does provide the following guarantees about the ordering of operations:
...
+Classification: 
+: Protocol Property (Generic)
+
+Type: 
+: Boolean
+
+Applicability: 
+: Message 
+
+This property specifies that a Message should be delivered to the other side
```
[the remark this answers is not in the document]
philsbln: Clarified

draft-ietf-taps-interface.md
```diff
@@ -1502,6 +1096,732 @@ does provide the following guarantees about the ordering of operations:
...
+after the previous Message which was passed to the same Connection via the Send
+Action.
+
+
+### Direction of communication
+
```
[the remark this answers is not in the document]
philsbln: Agreed - but let's do this discussion for all properties in Montreal

philsbln:
I would prefer to discuss the classification of Transport Properties in Montreal – the work has been done in one way and classes can be easily merged if we agree that we need fewer classes, but have to be split up again if we are able to convince you that we need more.

draft-ietf-taps-interface.md
philsbln: This has been in the draft for since trammell-taps-iterface-00 - nothing we added

abrunstrom:
The reason I think it is better to simplify first is that it seems all that have read it so far have had difficulties in understanding the separation between the classes. As it is not a difficult fix, I think it is better to simplify things before pushing it to the wg as other readers will likely have similar difficulties. We should also try to align the terminology in the other drafts which will be simpler with less constructs.

We also have several outstanding pull requests that need to be merged. Having seen the new structure for the Send section in Tommy's branch I think the send parameters should probably stay in that section. That would remove the message properties.

tfpauly:
@britram I still need to do my thorough pass on the PR, but my thoughts regarding schedule:

- I'm in favor of landing this as part of -01
- I disagree with some of the changes made to the sending of parameters section. I'd like to land #200 first, and then see how it merges into this PR overall. Many of the properties around sending have to do with scheduling a particular message, which I don't necessarily believe belong in the generic list of properties, or at least need a separate section.

mwelzl:
1) simplification vs. keeping the classification as-is: I'd also like to insist on simplifying because a) I don't see it as hugely complicated to re-insert the properties into more categories again if we must, and b) at least I, and it seems some more people, are going to disagree with the classification anyway. So what's the point of first keeping it, then having us re-iterate that we don't like it, and then having to change it later?   Starting simple is always good.

2) regarding PR #200, I read it carefully once more now, and I also agree that some message properties don't need to be pulled out because they strictly only appear there. I see only 2 1/2 exceptions: the first two are "capacity profile" and "niceness", for which it may be better to keep it as in this PR and simply refer forward to it in the sending section  (or vice versa!! Maybe even better?). The 1/2 is the "corruption protection length", which naturally fits together with "control checksum coverage on sending or receiving" (*) but is split apart in #200. I don't think this is a big issue though, and maybe better to keep them separate.

(*) Side note, I dislike the naming conflict between these two; wouldn't "checksum coverage" be clearer than "corruption protection length" ?

mwelzl:
I've just been discussing with @philsbln for a while - I think a good way forward is to first land #200, then land this but without deleting stuff in the sending section. This will give us duplicate text - and then we start to look at these duplicates one by one. @philsbln says that they found several inconsistencies, but on the other hand, it seems that @tfpauly has improved some text in #200. Sure we could just compare the branches but I think we should keep things from diverging more at this point, and first pull these two PR's in one by one.

(I'm only suggesting but hoping for @britram to actually pull the trigger because I'm afraid to mess something up - I've never manually merged a PR...)

theri:
Thanks for all the comments so far. 
Regarding classification, I agree with making it simpler for now, and then discuss in Montreal. This would mean going back to Selection Properties, Protocol Properties, and Message Properties for now, which is also in line with the other drafts. 
I will move the classification for Control Properties and Intents to the Appendix for now, so we can still see these definitions, modify them, and move them back up if we decide we want them.
After we decide on the classification we want, we can also discuss about in which category the individual Properties belong.
About #200 I'd be okay with landing it and then adjusting this PR, but I'm still not entirely sure about getting rid of the explicit MessageProperties...

philsbln:
@tfpauly I agree to land PR #200 ASAP and re-merge your changes to this one.

I think we will need all "Control Properties" at least listed in Section 12. This makes keeping related properties much easier. I agree to move the ones important to control flow, e.g., final, back to Section 7 and only list/cross-reference them in Section 12. Will do so after merging PR #200.

theri:
Looking at the document, it seems like it will be easier to discuss the classification when having it all in one place, so I did not move them to the Appendix, but instead labeled them as "TODO: Discuss".

philsbln:
@mwelzl and @abrunstrom – I totally agree that we need to simplify things.

@theri and I do not agree on the current classification --- but we agree that we a need for discussion about how to structure Properties.

I see this classification rather as a basis for discussion in Montreal than as a final thingy I want to push. Therefore, I would like to have the proposed classification in -01 point will add the rationale behind it to this PR. After discussion for Montreal, I am pretty sure that we will have a totally different classification (different from -00 and this proposal for -01).

theri:
Now that #200 landed, I merged the changes, changing all "parameters" to "properties" in line with this PR. 
After this initial merge, we now duplicate the following Properties in Section 7.3 and Section 12.3:

- Lifetime is exactly the same, just adds Classification, Type, and Applicability in 12.3
- Niceness, in Section 12.3, can not only apply on a Message but also on a Connection relative to other Connections (in the same group).
- Ordered has been slightly reworded in Section 12.3 in an attempt to make it more clear
- Idempotent is basically similar, just the start of the sentence has been slightly modified
- Final has been slightly reworded in Section 12.3 in an attempt to make it more clear
- Corruption protection length is exactly the same, just adds Classification, Type, and Applicability
- Transmission Profile from 7.3 basically became per-Message Capacity Profile in 12.3, and has more levels in 12.3

Now we could move all Properties to 12.3 while forward-referencing them in 7.3, merging the ones that are the same, and splitting Niceness and Capacity Profile into per-Connection and per-Message again.
Alternatively, we could keep the duplicates for now and discuss further after landing this, but I would prefer to have the Properties all in one place.

gorryfair:
I like this shorter list of 2 categories - selection looks like it is special, and then the stack needs configured.

Gorry

theri:
I marked the Properties in 7.3 as "TODO: De-duplicate". 
With this, @philsbln and I think that this PR is ready to land. We are planning to prepare slides for Montréal summarizing the current status and the discussion points.

draft-ietf-taps-interface.md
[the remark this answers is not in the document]
tfpauly: Right, the previous title of being a message “parameter” makes more sense for items like Final. Not everything is a property... one option as long as you’re adding these todos is to change the section back to parameters, which justifies its own section, and mention that the duplication that should be resolved is between some of the message sending parameters which are property-like, and the list of properties below.

draft-ietf-taps-interface.md
```diff
@@ -1594,114 +1309,887 @@ does provide the following guarantees about the ordering of operations:
...
+   | Prohibit   | Select only protocols/paths not providing the property, fail otherwise |
+
+
+When used on a Connection, this type becomes a (read-only) Boolean representing whether the selected transport supports the requested feature.
+
+
+## Transport Property Classification {#transport-props-classes}
+
+Note:
+: This section is subject to WG discussion on ITEF-102.
```
tfpauly: Typo: ITEF->IETF

draft-ietf-taps-interface.md
```diff
@@ -1594,114 +1309,887 @@ does provide the following guarantees about the ordering of operations:
...
+
+Transport Properties – not matter whether they apply to connections, preconnections, or messages – differ in the way they effect the transport system and protocols exposed through the transport system.
```
tfpauly: Grammar: remove "not matter"

tfpauly: Grammar: effect->affect

draft-ietf-taps-interface.md
```diff
@@ -1594,114 +1309,887 @@ does provide the following guarantees about the ordering of operations:
...
+This classification emphasizes two aspects of what and how properties
+effect the transport system in order to correctly set their expectation
```
tfpauly: effect->affect

draft-ietf-taps-interface.md
```diff
@@ -1594,114 +1309,887 @@ does provide the following guarantees about the ordering of operations:
...
+When used on a Connection, this type becomes a (read-only) Boolean representing whether the selected transport supports the requested feature.
+
+
+## Transport Property Classification {#transport-props-classes}
+
+Note:
+: This section is subject to WG discussion on ITEF-102.
+
+Transport Properties – not matter whether they apply to connections, preconnections, or messages – differ in the way they effect the transport system and protocols exposed through the transport system.
+This classification emphasizes two aspects of what and how properties
```
tfpauly: Remove "what and"

draft-ietf-taps-interface.md
```diff
@@ -1594,114 +1309,887 @@ does provide the following guarantees about the ordering of operations:
...
+effect the transport system in order to correctly set their expectation
+towards these properties:
```
tfpauly: In general, this sentence is quite confusing. Perhaps:

"Properties are classified based on how they take effect in the system based on application preferences. There are two factors that determine the classification of a property:"

draft-ietf-taps-interface.md
```diff
@@ -1594,114 +1309,887 @@ does provide the following guarantees about the ordering of operations:
...
+
+ - Whether properties effect protocols exposed through the transport system (Protocol Properties) or the transport system itself (Control Properties)
```
tfpauly: effect -> affect

draft-ietf-taps-interface.md
```diff
@@ -1594,114 +1309,887 @@ does provide the following guarantees about the ordering of operations:
...
+
+Note that many protocol properties have a corresponding selection property
+which asks for a protocol providing a specific transport feature that is
+controlled by the protocol property.
+
+### Control Properties {#control-props}
+
+\[TODO: Discuss]
+
+Control properties control the local transport system behavior or request
```
tfpauly: "Control properties control" is a bit redundant

draft-ietf-taps-interface.md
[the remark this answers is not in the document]
tfpauly: I'd leave this out, and just let it be a Message Context Parameter

draft-ietf-taps-interface.md
```diff
@@ -1032,15 +790,15 @@ Connection -> SendError<msgRef>
...
 an attempt to send a Message which is too large for the system and
 Protocol Stack to handle, some failure of the underlying Protocol Stack, or a
-set of send parameters not consistent with the Connection's transport
-parameters. The SendError contains an implementation-specific reference to the
+set of Message Properties not consistent with the Connection's transport
+properties. The SendError contains an implementation-specific reference to the
 Message to which it applies.
 
-## Message Context Parameters {#send-params}
+## Message Context Properties {#message-props}
```
tfpauly: I'd prefer to leave this category as Parameters, since they're not all "properties"

draft-ietf-taps-interface.md
theri: Fixed.

draft-ietf-taps-interface.md
theri: Fixed.

draft-ietf-taps-interface.md
theri: Fixed.

draft-ietf-taps-interface.md
theri: Fixed.

draft-ietf-taps-interface.md
theri: I reworded the second sentence, trying to convey the meaning that there may be multiple aspects in which properties differ, and here we emphasize two of these aspects, so applications know what to expect.

draft-ietf-taps-interface.md
theri: Fixed.

draft-ietf-taps-interface.md
theri: True. Changed it to "manage", not perfect but hopefully still works for what is meant here.

draft-ietf-taps-interface.md
theri: Renamed this part back to what it was in #200, and left "Final" up as a Message Context Parameter.
I think the distinction between "parameters" and "properties" goes along the same lines as the distinction between "Control Properties" and "Protocol Properties", and I'm looking forward to the discussion in Montreal.

philsbln:
In my opinion, this PR should now address all changes requested or clearly mark them as discussion points. Slides for the Properties classification discussions in Montreal are ready.

Can we land it?

tfpauly:
@theri @philsbln Thanks for the changes! I’m fine with landing this now, and look forward to the discussion it will create. Having this concrete set of items is a great framing to hold the conversation around.

PR #214, opened 2018-09-03
InitiateAndSend() / RendezvousAndSend()

britram:
This is my attempt to address #112 with something that's (IMO) less clunky in the common case than #124. 

I'm not a big fan of this way of doing things either -- specifically, I think the semantics of RendezvousAndSend(), while consistent, are probably wrong.

### DIFF
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -666,9 +666,6 @@ the Preconnection will listen for incoming Connections. This list can be
 passed to a peer via a signalling protocol, such as SIP {{?RFC3261}} or WebRTC
 {{?RFC7478}}, to configure the remote.
 
-See also {{rendezvous-and-send}} to combine rendezvous and transmission of the
-first message in a single action.
-
 ## Connection Groups {#groups}
 
 Groups of Connections can be created using the Clone Action:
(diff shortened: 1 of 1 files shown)


draft-ietf-taps-interface.md
```diff
@@ -989,6 +990,75 @@ Connection.Batch(
...
+
+~~~
+Connection := Preconnection.InitiateAndSend(messageData)
+~~~
+
+With a message context (as in {{message-props}}):
+
+~~~
+Connection := Preconnection.InitiateAndSend(messageData, messageContext)
+~~~
```
csperkins: I like the general approach of `InitiateAndSend`. It keeps the types right, and avoids issues with trying to call `Send()` on `Preconnection` objects and with enqueued data.

Maybe suggest changing the name to `InitiateAndSendIdempotent()` though, to be explicit? There are security issues with idempotent data, so we need to (try to) make sure people don't use without thinking through the consequences.

draft-ietf-taps-interface.md
```diff
@@ -989,6 +990,75 @@ Connection.Batch(
...
+
+The Events that may be sent after InitiateAndSend() are equivalent to those
+that would be sent by an invocation of Initate() followed immediately by an
+invocation of Send(), with the caveat that a send failure that occurs because
+the Connection could not be established will not result in a
+SendError separate from the InitiateError signaling the failure of Connection
+establishment.
+
+## Send on Rendezvous: RendezvousAndSend {#rendezvous-and-send}
+
```
csperkins: I guess we need to think more about the rendezvous model. 

If the goal is to support traditional TCP simultaneous open, then 0-RTT idempotent data could work. For something like an ICE-driven rendezvous, with NAT traversal, I wonder if it still makes sense? STUN doesn't support delivering data in the binding discovery messages, so it's no longer 0-RTT. And, if some future protocol did support data in binding discovery messages, there's no guarantee that all those messages hit the same server, so the semantics get weird. 

I wonder if we should not try to support idempotent, 0-RTT, rendezvous for the first version?

draft-ietf-taps-interface.md
abrunstrom: I  agree, being explicit here would be important.

draft-ietf-taps-interface.md
abrunstrom: Sounds reasonable to leave it out for now.

draft-ietf-taps-interface.md
britram: +1 to change to `InitiateAndSendIdempotent()`, since it forces that message property.

draft-ietf-taps-interface.md
britram: Yeah, I tried this out primarily to see what would happen if `RendezvousAndSend()` were kept consistent with `InitiateAndSend()` and you get this weird buffered-banner behavior that (1) probably doesn't do anything anyone wants and (2) also probably doesn't work anyway.

Will cut it for now, and we can leave rendezvous acceleration as an area for future research.

draft-ietf-taps-interface.md
britram: <bikeshed>How about `InitiateWithIdempotentSend()`?</bikeshed>

csperkins: wfm

mirjak:
After some discussion with Brian agree that the only solution is to provide some initial date to the preconnection when calling initiate. However, why do we need to have two different function. If we only have initiate and provide idempotent data as an optional input parameter that should be fine, right?

britram:
@mirjak: your question is now #224

PR #215, opened 2018-09-03
Small changes to "receive"

mwelzl:
1 tiny fix suggested to make the text a little clearer (because the default "complete message" delivery may need a de-framer, or the whole Connection may function as a message), and pointed at a possible error when delivering partial data (it's already written elsewhere, but the suggestion makes a sentence more correct, I believe).


tfpauly:
Looks good, thanks!

PR #216, opened 2018-09-03
Set/query conn. properties (sec 9) structure change

mwelzl:
Rather than saying "this can only be queried" everywhere, I split this into setting and querying now, and I also moved SoftErrors in there. It's not precisely "querying" for a Property, but it's the closest match - better than "Connection Termination", I think (this is where SoftErrors were before).

### DIFF
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -1204,15 +1204,15 @@ Preconnection.DeframeWith(Deframer)
 
 # Setting and Querying Connection Properties {#introspection}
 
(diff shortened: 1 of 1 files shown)


abrunstrom:
Maintenance sounds a bit strange to me, "Managing Connections: Setting Properties and Obtaining Information" ?

mwelzl:
I like it (and I agree about "maintenance" being slightly strange)

theri:
Managing sounds good to me too.

mwelzl:
Ack, done

PR #217, opened 2018-09-03
Transport Property clean-up after the re-org. De-duplication etc.

mwelzl:
I gave the document a very careful back-to-back read. My take-away was that removing the Message Context Parameters from where they are, and just pointing to a later section, would make for a bad reading experience. Basically, it would feel like an empty shell - at some point, the reader wants to see something tangible, and so, reading all these Message properties where they appear is satisfying, in a way. The document would get quite tough to digest if it would just provide the "framework" for sending, receiving etc., and we have to wait for the "meat" until the end, where it is then offered in a very organized fashion.

Hence, I decided to leave all the message properties described where they are (as Message Context Parameters). That was my approach to de-duplication in this PR.

Side comment if someone's surprised by the removal of A.1.9 "Immediate": that's because this is already fully covered by the Transmission Profile in Section 7.3.7.

### DIFF
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -861,9 +861,7 @@ implementation-specific.
 
 This property is a numeric (non-negative) value that represents an unbounded
 hierarchy of priorities. It can specify the priority of a Message, relative to
-other Messages sent over the same Connection or Connection Group (see
-{{groups}}), or the priority of a Connection, relative to other Connections in
-the same Connection Group.
+other Messages sent over the same Connection.
 
 A Message with Niceness 0 will yield to a Message with Niceness 1, which will
 yield to a Message with Niceness 2, and so on. Niceness may be used as a
@@ -1889,8 +1887,11 @@ Type:
 Applicability:
 : Connection, Message
 
-See {{msg-niceness}}.
-As an exception from the per-Connection behavior described in {{groups}}, even when set on a Connection, this property is not entangled when Connections are cloned.
+This Property is a non-negative integer representing the relative inverse
+priority of this Connection relative to other Connections in the same
+Connection Group. It has no effect on Connections not part of a Connection
+Group. As noted in {{groups}}, this property is not entangled when Connections
+are cloned.
 
 
 ### Timeout for aborting Connection {#timeout}
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -849,11 +849,7 @@ inconsistent with the Selection Properties of the Connection yields an error.
 
 The following Message Context Parameters are supported:
 
-\[TODO: De-Duplicate with Properties in {{transport-props}}, find consensus on which Section to put them]
-
-### Lifetime
-
-\[TODO: De-Duplicate with {{send-lifetime}}]
+### Lifetime {#msg-lifetime}
 
 Lifetime specifies how long a particular Message can wait to be sent to the
 remote endpoint before it is irrelevant and no longer needs to be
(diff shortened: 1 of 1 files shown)


draft-ietf-taps-interface.md
```diff
@@ -849,26 +849,22 @@ inconsistent with the Selection Properties of the Connection yields an error.
...
 
-### Niceness
+### Niceness {#msg-niceness}
 
-\[TODO: De-Duplicate with {{send-niceness}}]
+This property is a numeric (non-negative) value that represents an unbounded
+hierarchy of priorities. It can specify the priority of a Message, relative to
+other Messages sent over the same Connection or Connection Group (see
+{{groups}}), or the priority of a Connection, relative to other Connections in
+the same Connection Group.
```
theri: Previously, we had merged per-Message and per-Connection Niceness, but then come to realize that we rather want them separate.

I think this sentence is an artifact of this. 
The latter part refers to per-Connection Niceness (within a Connection Group), as opposed to per-Message Niceness (within the same Connection). I think here we only want per-Message Niceness, so the per-Connection part should be taken out.

draft-ietf-taps-interface.md
```diff
@@ -1920,20 +1889,7 @@ Type:
...
-
-A Message with Niceness 0 will yield to a Message with Niceness 1, which will
-yield to a Message with Niceness 2, and so on. Niceness may be used as a
-sender-side scheduling construct only, or be used to specify priorities on the
-wire for Protocol Stacks supporting prioritization.
-
-This encoding of the priority has a convenient property that the priority
-increases as both Niceness and Lifetime decrease.
-
+See {{msg-niceness}}.
```
theri: I think here we still want the per-Connection Niceness as it was before it was merged with the per-Message Niceness.

I dug the old text out of the repo, and it said:
"This Property is a non-negative integer representing the relative inverse
priority of this Connection relative to other Connections in the same
Connection Group. It has no effect on Connections not part of a Connection
Group. As noted in {{groups}}, this property is not entangled when Connections
are cloned."

draft-ietf-taps-interface.md
mwelzl: ACK!  I'm changing it only to "over the same Connection", dropping the group because this can make it confusing: what if I give a message a lower Niceness than all others in the group, yet the Connection the Message is on has a higher Niceness than the others? Then what does this mean?  All of this gets clear if the context of per-Message Niceness is only a single Connection. I think this is what you probably identified and I didn't think of so far - great catch!

draft-ietf-taps-interface.md
mwelzl: ACK - and thanks for digging, this was very useful!

draft-ietf-taps-interface.md
```diff
@@ -1889,8 +1887,11 @@ Type:
 Applicability:
 : Connection, Message
```
theri: Very minor thing, but here we want to drop "Message" as well.

mwelzl:
Fixed - thanks!

philsbln:
Sorry for being late with the review, but I dislike some details of this PR as is only one part of what we agreed on in Montreal an adds inconsistencies:

Inconsistencies:
 - The list-style labeling of all properties with their data-type has removed for those properties that have been moved up to section 7.3 and replaced by prose.

Missing Steps
 - Those Transport Properties which we decided to call "Connection Properties" have not been renamed.
 - Those Transport Properties which we decided to call "Selection Properties" have not been renamed.
 - Moving the other Transport Properties up

After the complete thing, Section 12 should contain the following things:
 - Overview what Properties are and when they can be specified
 - Definition of types
 - List/Table of all properties with references to the earlier chapters where there are defined

I will start preparing PRs fixing these…

PR #223, opened 2018-09-06
This fixes inconcistencies from PR #217 - Message Properties move to Section 7.3

philsbln:
- de-duplicates some properties in 7.3. and 12
- moves up some message properties missing in 7.3.
- make type definition consistent across all properties

This ignores Classification and Applicability labels which are subject
to removal in another PR moving Selection and Connections properties up.
Section 12.3 will most likely be replaces with a table then.

### DIFF
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -687,7 +687,7 @@ the Connection Group, and so on. Connections in a Connection Group share all
 Protocol Properties that are not applicable to a Message.
 
 Changing one of these Protocol Properties on one Connection in the group changes it for all others. Per-Message Protocol Properties, however, are not entangled.
-For example, changing "Timeout for aborting Connection" (see {{timeout}}) on one Connection in a group will automatically change this Protocol Property for all Connections in the group in the same way. However, changing "Lifetime" (see {{send-lifetime}}) of a Message will only affect a single Message on a single Connection, entangled or not.
+For example, changing "Timeout for aborting Connection" (see {{timeout}}) on one Connection in a group will automatically change this Protocol Property for all Connections in the group in the same way. However, changing "Lifetime" (see {{msg-lifetime}}) of a Message will only affect a single Message on a single Connection, entangled or not.
 
 If the underlying protocol supports multi-streaming, it is natural to use this functionality to implement Clone. In that case, entangled Connections are multiplexed together, giving them similar treatment not only inside endpoints but also across
 the end-to-end Internet path.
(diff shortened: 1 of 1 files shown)


draft-ietf-taps-interface.md
```diff
@@ -1925,6 +1916,12 @@ Group. As noted in {{groups}}, this property is not entangled when Connections
 are cloned.
 
 
+### Niceness (Message)
+
+Integer Message Property - see {{msg-niceness}}.
+This property overrides the connection niceness - see {{conn-niceness}} - on a per-message basis.
```
theri: Mhh, I don't think this is true.
I think per-Connection Niceness and per-Message Niceness are different Properties and refer to different things (the priority of Connections within a Connection group, and the priority of Messages within a Connection), so they cannot override each other.
I'd be in favor of removing this sentence.

philsbln: Sorry, forgot about that - hope the new sentence captures the discussion.
Just removing it does will bring us to new discussions about this sooner or later.

philsbln:
@theri I want to convert 12.3 to a table once all properties have been moved up.

I guess it is useful to include types in the table and put relations as footnotes into the table.

draft-ietf-taps-interface.md
theri: Good point. I think like this it's fine.

gorryfair:
I think I understand "connection niceness", I do know how that would be used.
As I look through things today, I'm less sure about how expressing message niceness is actually needed, and if so - what sort of machinery you'd need to properly handle a message niceness property. My own mental model is that a sender that cares about "niceness" will keep its output queue low, and that it can itself maintain some notions about which threads/messages/whatever are most important to send next. What does the stack need to do in addition?
- Maybe I am really missing something here (please tell me).

mwelzl:
I think per-message niceness is an artefact of not being sure how this will map onto multi-streaming. E.g., if this is NOT implemented as "one (TAPS) Connection = one stream of a transport", then maybe multiple streams could be used underneath the API, and per-message niceness could be used to assign a message to a stream based on a Niceness value. Then, however, per-connection Niceness becomes irrelevant.

My conclusion is that we should keep it in there for now, but mark it with a "DISCUSS" so we remember to decide for *either* per-message or per-connection Niceness depending on how we intend to implement multi-streaming.

(my personal preference is to have per-connection Niceness only, and hence map "one (TAPS) connection is one stream")

britram:
let's land this now, though, and have the niceness discussion in another issue / at the interim...

mwelzl:
I see I'm late for the party... but just for the record, now that I managed to give it a close read, I also agree with this  :)

PR #225, opened 2018-09-06
Some clarifications in property text

mwelzl:
I didn't like something about reliability (let's not throw an error when we can't be UNreliable, only when we can't be reliable). I updated the text on ordering a bit, to say that the Property is ignored for the first Message on a Connection, and rephrased it slightly to talk about the receiver-side transport system and the receiving application - before, the text might have talked about a sender-side behavior.  I added a TODO DISCUSS comment about per-message Niceness.

### DIFF
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -882,7 +882,7 @@ being interpreted as a higher priority anywhere.]
 Type:
 : Boolean
 
-If true, it specifies that the receiver-side transport system should only deliver the Message to the receiving application after the previous ordered Message which was passed to the same Connection via the Send
+If true, it specifies that the receiver-side transport protocol stack only deliver the Message to the receiving application after the previous ordered Message which was passed to the same Connection via the Send
 Action, when such a Message exists. If false, the Message may be delivered to the receiving application out of order.
 This property is used for protocols that support preservation of data ordering,
 see {{prop-ordering}}, but allow out-of-order delivery for certain messages.
@@ -1547,7 +1547,7 @@ Type:
 Applicability:
 : Preconnection, Connection (read only)
 
-This property specifies whether the application wishes to use a transport
+This property specifies whether the application needs to use a transport
 protocol that ensures that all data is received on the other side without
 corruption. This also entails being notified when a Connection is closed or
 aborted. The default is to enable Reliable Data Transfer.
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -882,8 +882,8 @@ being interpreted as a higher priority anywhere.]
 Type:
 : Boolean
 
-If true, it specifies that a Message should be delivered to the other side after the previous Message which was passed to the same Connection via the Send
-Action. If false, the Message may be delivered out of order.
+If true, it specifies that the receiver-side transport system should only deliver the Message to the receiving application after the previous Message which was passed to the same Connection via the Send
+Action. If false, the Message may be delivered to the receiving application out of order.
 This property is used for protocols that support preservation of data ordering,
 see {{prop-ordering}}, but allow out-of-order delivery for certain messages.
 
@@ -937,9 +937,8 @@ Type:
 This property specifies that a message should be sent in such a way
 that the transport protocol ensures all data is received on the other side
 without corruption. Changing the ´Reliable Data Transfer´ property on Messages
-is only possible if the transport protocol supports
-partial reliability (see {{prop-partially-reliable}}).
-Therefore, for protocols that always transfer data reliably, this property is always true and for protocols that always transfer data unreliably, this property is always false. Changing it may generate an error.
+is only possible if the transport protocol supports reliability.
+When this is not the case, changing it will generate an error.
 
 ### Transmission Profile {#send-profile}
 
@@ -1917,11 +1916,13 @@ are cloned.
 
 
 ### Niceness (Message)
+\[TODO: Discuss: should we remove this? Whether we need this or the other depends
+on how we want to implement multi-streaming. We don't need both, so we should make a decision.]
 
 Integer Message Property - see {{msg-niceness}}.
 
 Note that this property is not a per-message override of the connection Niceness - see {{conn-niceness}}.
-Both Niceness properties may interact, but can be used indepentendly and be realized by different mechanisms.
+Both Niceness properties may interact, but can be used indepently and be realized by different mechanisms.
 
 
 ### Timeout for aborting Connection {#timeout}
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -886,6 +886,7 @@ If true, it specifies that the receiver-side transport system should only delive
 Action. If false, the Message may be delivered to the receiving application out of order.
 This property is used for protocols that support preservation of data ordering,
 see {{prop-ordering}}, but allow out-of-order delivery for certain messages.
+It is ignored on the first Message that is sent on a Connection.
 
 
 ### Idempotent {#msg-idempotent}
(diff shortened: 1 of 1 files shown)


draft-ietf-taps-interface.md
```diff
@@ -1917,11 +1917,13 @@ are cloned.
...
 
 ### Niceness (Message)
+\[TODO: Discuss: should we remove this? Whether we need this or the other depends
+on how we want to implement multi-streaming. We don't need both, so we should make a decision.]
 
 Integer Message Property - see {{msg-niceness}}.
 
 Note that this property is not a per-message override of the connection Niceness - see {{conn-niceness}}.
-Both Niceness properties may interact, but can be used indepentendly and be realized by different mechanisms.
+Both Niceness properties may interact, but can be used indepently and be realized by different mechanisms.
```
abrunstrom: Spelling is still incorrect :-)

draft-ietf-taps-interface.md
```diff
@@ -882,10 +882,11 @@ being interpreted as a higher priority anywhere.]
 Type:
 : Boolean
 
-If true, it specifies that a Message should be delivered to the other side after the previous Message which was passed to the same Connection via the Send
-Action. If false, the Message may be delivered out of order.
+If true, it specifies that the receiver-side transport system should only deliver the Message to the receiving application after the previous Message which was passed to the same Connection via the Send
+Action. If false, the Message may be delivered to the receiving application out of order.
 This property is used for protocols that support preservation of data ordering,
 see {{prop-ordering}}, but allow out-of-order delivery for certain messages.
+It is ignored on the first Message that is sent on a Connection.
```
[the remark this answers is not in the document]
mwelzl: I don't understand why it's not correct but it doesn't matter - I prefer your "when such a message exists" wording. However, why say "after the previous _ordered_ Message" ? If the previous one (msg 1) doesn't have the ordered flag set, yet the next one (msg 2) does, shouldn't that mean that msg 2 should be delivered after msg 1?

draft-ietf-taps-interface.md
```diff
@@ -937,9 +938,8 @@ Type:
 This property specifies that a message should be sent in such a way
 that the transport protocol ensures all data is received on the other side
 without corruption. Changing the ´Reliable Data Transfer´ property on Messages
-is only possible if the transport protocol supports
-partial reliability (see {{prop-partially-reliable}}).
-Therefore, for protocols that always transfer data reliably, this property is always true and for protocols that always transfer data unreliably, this property is always false. Changing it may generate an error.
+is only possible if the transport protocol supports reliability.
```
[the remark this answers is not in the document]
mwelzl: Agreed! will fix

draft-ietf-taps-interface.md
mwelzl: Sigh  :)  thanks, will fix!

draft-ietf-taps-interface.md
```diff
@@ -937,9 +937,8 @@ Type:
 This property specifies that a message should be sent in such a way
 that the transport protocol ensures all data is received on the other side
 without corruption. Changing the ´Reliable Data Transfer´ property on Messages
-is only possible if the transport protocol supports
-partial reliability (see {{prop-partially-reliable}}).
-Therefore, for protocols that always transfer data reliably, this property is always true and for protocols that always transfer data unreliably, this property is always false. Changing it may generate an error.
+is only possible if the Connection supports reliability.
+When this is not the case, changing it will generate an error.
```
theri: Agree that if the Connection does not support reliability and the application requests reliable, we must throw an error so that sending fails, but we don't need to fail the other way round. Maybe a warning makes sense, but we don't need to standardize that.
However, I think this property makes most sense with partially reliable transport protocols, and I'd like to keep a reference to {{prop-partially-reliable}} here. Maybe something like "if the Connection supports reliability, see {{prop-reliable}}, or partial reliability, see {{prop-partially-reliable}}"?

draft-ietf-taps-interface.md
```diff
@@ -882,8 +882,8 @@ being interpreted as a higher priority anywhere.]
 Type:
 : Boolean
 
-If true, it specifies that a Message should be delivered to the other side after the previous Message which was passed to the same Connection via the Send
-Action. If false, the Message may be delivered out of order.
+If true, it specifies that the receiver-side transport system should only deliver the Message to the receiving application after the previous ordered Message which was passed to the same Connection via the Send
```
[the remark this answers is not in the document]
philsbln: From my perspective, this is a minor knit. Perhaps try the following instead:

> If true, it specifies that the receiver-side transport system is expected to only deliver the Message to
> the receiving application after the previous ordered Message which was passed to the same Connection.
> The actual signaling of this is protocol specific.

draft-ietf-taps-interface.md
philsbln: For me, this is clear.

draft-ietf-taps-interface.md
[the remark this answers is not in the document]
gorryfair: What you say is true, but then TAPS uses transports that also have receiver sematics: 

True - seems like receiver TCP semantics? (so if there is a TCP receiver at the remote, TAPS can use that). SCTP can also support this.

False - seems like UDP. If the stack supports this, the SCTP can alternatively do this.

... My choice would be I'd prefer to standardise "does not need to be ordered", if I had a choice of semantics.

draft-ietf-taps-interface.md
gorryfair: ... My choice would be I'd prefer to standardise "needs reliable", if I had a choice of semantics.
Relaxing "needs reliable" can be done on a per-message basis for SCTP stacks that allow this.

draft-ietf-taps-interface.md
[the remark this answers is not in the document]
mwelzl: Indeed, the sender side alone can never guarantee ordering - at best, it can know that there is a receiver that takes care of it, or (e.g. in case of SCTP) signal the wish to the receiver. I updated the text to talk about the receiver behavior to avoid indicating sender-side logic that shouldn't be there. I thought the receiver-oriented text is clearer.

draft-ietf-taps-interface.md
mwelzl: I see -- I misunderstood the intention behind "Reliable Data Transfer (Connection)". I suggest to include the reference to partially reliability as @theri suggests, but also update the text after "Reliable Data Transfer (Connection)" from: "This property specifies whether the application wishes to use a transport protocol that ensures that all data is received on the other side without corruption."  to "This property specifies whether the application requires that all data it sends on the Connection will be received on the other side without corruption."

The main point here is the use of "requires" as opposed to "wishes".

theri: Well, "Reliable Data Transfer (Connection)" has the data type "Preference" anyway, so an application can Require, Prefer, Avoid, or Prohibit it. Out of these, I guess "Require" makes the most sense, and I think it can replace "wishes", which sounds like "Prefers".
However, we have lots and lots of formulations like this in the other properties as well, so maybe we want to agree on a more general non-awkward way to phrase this (in another PR?).

draft-ietf-taps-interface.md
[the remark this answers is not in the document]
theri: I agree with what you're saying, it's clearer.
My only comment (perhaps a nit) is to change "receiver-side transport system" to "receiver-side transport protocol stack", so we do not imply that there is a transport system (which, I think, we define as a TAPS implementation) running there, it could also just be a vanilla stack.

draft-ietf-taps-interface.md
mwelzl: Ah. Understood.  But let me just call it "needs" for now.

draft-ietf-taps-interface.md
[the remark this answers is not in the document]
mwelzl: I agree, that's better

PR #227, opened 2018-09-12
Moved Connection Properties up to #introspection

philsbln:
This is the second step towards applying our new Properties
Classification. It moves up the Connection Properties and annotates most
places with stuff to do.

### DIFF
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -1256,8 +1256,11 @@ Preconnection.DeframeWith(Deframer)
 
 # Managing Connections: Setting Properties and Obtaining Information {#introspection}
 
-The application can set per-connection Protocol Properties (see {{protocol-props}}).
-Certain Procotol Properties may be read-only, on a protocol- and property-specific basis.
+
+## Connection Properties
+
+The application can set per-connection Properties.
+Certain Connection Properties may be read-only, on a protocol- and property-specific basis.
 ~~~
 Connection.SetProperty(property, value)
 ~~~
@@ -1267,8 +1270,11 @@ At any point, the application can query Connection Properties.
 ConnectionProperties := Connection.GetProperties()
 ~~~
 
+
 Depending on the status of the connection, the queried Connection
 Properties will include different information:
+
+\[TODO: turn this list into actual properties or move up into the explaining text]
 
 * The status of the connection, which can be one of the following:
   Establishing, Established, Closing, or Closed.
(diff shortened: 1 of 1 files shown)


draft-ietf-taps-interface.md
```diff
@@ -1306,6 +1312,126 @@ Properties will include different information:
...
+
+### Notification of excessive retransmissions {#conn-retrans-notify}
+
+Type:
+: Boolean
+
+This property specifies whether an application considers it useful to be informed in case sent data was retransmitted more often than a certain threshold.
+When set to true, the effect is twofold:
+The application may receive events in case excessive retransmissions.
+In addition, the transport system considers this as a preference to use transports stacks that can provide this notification. This is not a strict requirement.
```
theri: This basically says that this property is two things at the same time:
1. A Connection Property saying that the application wants these notifications in case the transport stack can provide them,
2. A Selection Property to "Prefer" transport stacks that can provide these notifications.

I'm wondering if this property really should be both in one.
I think this text got written before we had an actual concept of Preferences in Selection Properties. Maybe we want to make this an actual Selection Property instead, with data type "Preference"? And then once the Connection is instantiated, it becomes a Connection property like all other Selection Properties, providing a Boolean of whether the feature is there or not?
I'd also be okay with it staying like this for this PR though.

draft-ietf-taps-interface.md
```diff
@@ -1306,6 +1312,126 @@ Properties will include different information:
...
+: Boolean
+
+This property specifies whether an application considers it useful
+to be informed when an ICMP error message arrives that does not force
+termination of a connection.
+When set to true, received ICMP errors will be available as SoftErrors.
+Note that even if a protocol supporting this property is selected, not all
+ICMP errors will necessarily be delivered, so applications cannot rely
+on receiving them.
+Setting this option also implies a preference to prefer transports stacks that can provide this notification.
```
theri: Same question as for "Notification of excessive retransmissions" above.

draft-ietf-taps-interface.md
```diff
@@ -1378,19 +1504,19 @@ does provide the following guarantees about the ordering of operations:
 
 # Transport Properties {#transport-props}
 
-Having discussed Transport Properties in {{connection-props}} and listed a
-number of Message Properties ("Ordered", "Idempotent" etc.) in {{message-props}},
+Having discussed Selection Properties in {{connection-props}}, Connection Properties {{connection-props}} and,  Message Properties ({{message-props}}),
```
theri: Two nits here: 
1. The first link should point to {{selection-props}}.
2. The comma should be before "and", not after.

draft-ietf-taps-interface.md
```diff
@@ -1378,19 +1504,19 @@ does provide the following guarantees about the ordering of operations:
...
 we now provide a complete overview of a transport system's defined Transport Properties.
 
-Properties are structured in two ways:
-
- - By how they influence the transport system, which leads to a classification
-   into "Selection Properties", "Protocol Properties", "Control Properties" and "Intents".
- - By the object they can be applied to: Preconnections, see {{connection-props}},
-   Connections, see {{introspection}}, and Messages, see {{message-props}}.
+Transport Properties are structured by the context they can apply on:
```
theri: Why change "the object they apply on" to "context they apply on"? I think "object" was clearer.

draft-ietf-taps-interface.md
```diff
@@ -1446,15 +1572,9 @@ When used on a Connection, this type becomes a (read-only) Boolean representing
...
 
-Transport Properties – whether they apply to connections, preconnections, or messages – differ in the way they affect the transport system and protocols exposed through the transport system.
-The classification proposed below emphasizes two aspects of how properties
-affect the transport system, so applications know what to expect:
-
- - Whether properties affect protocols exposed through the transport system (Protocol Properties) or the transport system itself (Control Properties)
-
- - Whether properties have a clearly defined behavior that is likely to be
-   invariant across implementations and environments (Protocol Properties and Control Properties) or whether the properties are interpreted by the transport system to provide a best effort service that matches the application needs as closely as possible (Intents).
+Transport Properties are classified by the the objects that are
```
theri: This sentence is incomplete.

draft-ietf-taps-interface.md
```diff
@@ -1661,64 +1781,19 @@ within a Connection Group to be provided by streams of a single underlying
...
-
-Type:
-: Integer
-
-Applicability:
-: Preconnection, Connection
-
-This property specifies after how many retransmissions to inform the
-application about "Excessive Retransmissions".
+Interger Connection Property - see {{conn-excss-retransmit}}
```
theri: Nit: s/Interger/Integer/

draft-ietf-taps-interface.md
philsbln: Thanks for this findings - fixed

PR #229, opened 2018-09-20
fix #219, Bindings is now an array

### LINKED ISSUE #219: WTF is a PreconnectionBinding?
section 6.3 (Rendezvous) has the following line:

PreconnectionBinding := Preconnection.Resolve()

If i read the text correctly, I'd propose to change this to 

Preconnection<> := Preconnection.Resolve()

britram:


draft-ietf-taps-interface.md
```diff
@@ -653,23 +660,24 @@ when no transport-layer connection can be established to the Remote Endpoint,
...
+Traversal Utilities for NAT (STUN) server. In this case, the Local Endpoint
+may resolve to a mixture of local and server reflexive addresses. The
+Resolve() action on the Preconnection can be used to discover these bindings:
 
 ~~~
-PreconnectionBindings := Preconnection.Resolve()
+[]Preconnection := Preconnection.Resolve()
 ~~~
 
 The Resolve() call returns a list of Preconnection Objects, that represent the
```
mwelzl: Just a nit: above, it's said that an action creates an array of objects, but here it's called a list. For consistency, it would be better to call it either a list or an array everywhere, I guess.

PR #232, opened 2018-10-07
segmentation and fragmentation prohibited: fix #226

### LINKED ISSUE #226: "Fragmentation Prohibited": message/connection property?
Discussion on #203 identified that we probably want a Message (and/or Connection) property that forces DF on for IPv4 (and, I guess, ensures that datagram sends on IPv6 below MTU will fail, by disabling sender fragmentation, as well).

britram:
Add segmentation and fragmentation prohibited message properties

### DIFF
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -990,7 +990,8 @@ Type:
 
 This property specifies that a message should be sent without network-layer
 fragmentation, either at the sending endpoint or along the path. When used with
-connections running over IP version 4, the Don't Fragment bit will be set. 
+transports supporting this functionality and running over IP version 4, the Don't
+Fragment bit will be set.
 
 ## Partial Sends {#send-partial}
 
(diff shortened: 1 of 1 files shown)


draft-ietf-taps-interface.md
```diff
@@ -974,6 +974,23 @@ The following values are valid for Transmission Profile:
...
+current estimate of its maximum transmission segment size will result in a `SendError`.
+
+### Fragmentation Prohibited
+
+Type:
+: Boolean
+
+This property specifies that a message should be sent without network-layer
+fragmentation, either at the sending endpoint or along the path. When used with
+connections running over IP version 4, the Don't Fragment bit will be set.
```
mwelzl: Suggest: "When used with transports supporting this functionality, running over IP version 4, the Don't Fragment bit will be set."

britram: ack fill wix.

britram:
all right let's do the other one for now.

Issue #142, opened 2018-03-11
Clean up Capacity Profile


gorryfair:
TEXT:
      application's expectation of the dominating traffic pattern for...

- I am calling this out in case I think different to others... and then I want to discuss, but it could also be that this was an oversight on which we alreday agree... either way there's a lot of comment here on just a few words.

- I really do not like this. To me this is expressing QoS properties, and I think this is the wrong way to do this. I’d be OK if the application instructs the network to use a particular treatment. When an endpoint requests a treatment, the network can either treat it in a default way, or give it the treatment it requests. If the application sends data incompatible with the treatment, then the performance observed may suffer - but that is exactly correct because that is how the application asked its data to be treated. I'd for example prefer:
     application’s expectation of the treatment the network should provide for...

I am really against the application trying to label it’s data as being of some type (which I think this part currently does). To me, a QoS framework on the other hand, sets certain parameters in the network and then “labels” the flows in some way to use them. That’s a different approach, and one that demands more support from the network.
- Similarly, I did not like:
     The application is not interactive.  
- I’d prefer:
     The application does not wish traffic to be treated as interactive.  
  ... or something alike.

philsbln:
The application labeling what they are going to do and that traffic pattern they expect is one of the key ideas of the original Socket Intents work. It allows optimizing locally for the expected pattern in a way that is impossible without. 
 - This kind of information is intended to be used for path, endpoint, and protocol selection logic.
 - This information may also be used to automatically derive defaults for protocol parameters.
 - This information is _not_ intended to be used for polarization or for QoS labeling.

I guess we must include this as a clear warning to make sure it is not mixed up with QoS.

For your example: "The application is not interactive." is easy for a developer.
What it really states: "The programmer opening this socket suggests that latency and bandwidth variations will most likely have no observable user impact" – therefore the OS might, when combined with other hints like Resilience and Traffic Category, establish the connection over an unstable, but high bandwidth path.

britram:
One concrete proposal (around the table in London):

Replace "Capacity Profile" and "Instantaneous Capacity Profile" with a three-state variable for traffic classification which can be used for DSCP marking and local AQM:

1. Default
2. Scavenger
3. Low-latency

mwelzl:
[...]

PR #233, opened 2018-10-07
clean up capacity profile per #142

britram:
This is a (work in progress) attempt to clean up Capacity Profile as a selection property along the lines of the state of the discussion in #142.

@gorryfair: you had a suggestion to add "High Throughput Data" as a separate profile mapping to AF2n, but I don't know anything else about what this would actually imply for transport selection; can you help me out here?

### DIFF
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -1922,49 +1922,50 @@ values are valid for the Capacity Profile:
   Default: : The application makes no representation about its expected capacity
   profile. No special optimizations of the tradeoff between delay, delay
   variation, and bandwidth efficiency should be made when selecting and
-  configuring transport protocol stacks. When the underlying transport protocol
-  stack(s) support per-connection DSCP signaling, transmitted traffic SHOULD
-  carry the DSCP Default Forwarding {{?RFC2474}} PHB; when the Connection is
-  multiplexed, the guidelines in section 6 of {{?RFC7657}} apply.
+  configuring transport protocol stacks. Transport system implementations that
+  map the requested capacity profile onto per-connection DSCP signaling without
+  multiplexing SHOULD assign the DSCP Default Forwarding {{?RFC2474}} PHB; when
+  the Connection is multiplexed, the guidelines in section 6 of {{?RFC7657}}
+  apply.
 
   Scavenger/Bulk: : The application is not interactive. It expects to send
   and/or receive data without any urgency. This can, for example, be used to
   select protocol stacks with scavenger transmission control and/or to assign
-  the traffic to a lower-effort service. When the underlying transport protocol
-  stack(s) support per-connection DSCP signaling, transmitted traffic SHOULD
-  carry the DSCP Less than Best Effort {{?LE-PHB=I-D.ietf-tsvwg-le-phb}} PHB;
-  when the Connection is multiplexed, the guidelines in section 6 of
-  {{?RFC7657}} apply.
+  the traffic to a lower-effort service. Transport system implementations that
+  map the requested capacity profile onto per-connection DSCP signaling without
+  multiplexing SHOULD assign the DSCP Less than Best Effort
+  {{?LE-PHB=I-D.ietf-tsvwg-le-phb}} PHB; when the Connection is multiplexed, the
+  guidelines in section 6 of {{?RFC7657}} apply.
 
   Low Latency/Interactive: : The application is interactive, and prefers loss to
   latency. Response time should be optimized at the expense of bandwidth
   efficiency and delay variation when sending on this connection. This can be
   used by the system to disable the coalescing of multiple small Messages into
   larger packets (Nagle's algorithm); to prefer immediate acknowledgment from
-  the peer endpoint when supported by the underlying transport; and so on. When
-  the underlying transport protocol stack(s) support per-connection DSCP
-  signaling, transmitted traffic SHOULD carry the DSCP Expedited Forwarding
-  {{?RFC3246}} PHB; when the Connection is multiplexed, the guidelines in
-  section 6 of {{?RFC7657}} apply.
+  the peer endpoint when supported by the underlying transport; and so on.
+  Transport system implementations that map the requested capacity profile onto
+  per-connection DSCP signaling without multiplexing SHOULD assign the DSCP
+  Expedited Forwarding {{?RFC3246}} PHB; when the Connection is multiplexed, the
+  guidelines in section 6 of {{?RFC7657}} apply.
 
   Low Latency/Non-Interactive: : The application prefers loss to latency but is
   not interactive. Response time should be optimized at the expense of bandwidth
-  efficiency and delay variation when sending on this connection. When the
-  underlying transport protocol stack(s) support per-connection DSCP signaling,
-  transmitted traffic SHOULD carry a DSCP Assured Forwarding
-  (AF21,AF22,AF23,AF24) {{?RFC2597}} PHB; when the Connection is multiplexed,
-  the guidelines in section 6 of {{?RFC7657}} apply.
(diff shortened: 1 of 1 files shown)


draft-ietf-taps-interface.md
```diff
@@ -1904,47 +1904,70 @@ options.
...
 : Enumeration
 
 Applicability:
 : Preconnection, Connection, Message
 
-This property specifies the application's expectation of the dominating traffic pattern for this Connection.
-This implies that the transport system should optimize for the capacity profile specified. This can influence path and protocol selection.
-The following values are valid for the Capacity Profile:
+This property specifies the traffic interaction pattern expected by the
+application, the desired network treatment for traffic sent by the application,
```
mwelzl: I think Gorry took issue with "pattern expected by the application" in #142 (i.e., I don't think this re-wording fixes his problem with the old version). Specifically, he wrote that he'd prefer: "application’s expectation of the treatment the network should provide for..."

draft-ietf-taps-interface.md
```diff
@@ -1904,47 +1904,70 @@ options.
...
+  : The application makes no representation about its expected capacity
+  profile. No special optimizations of the tradeoff between delay, delay
+  variation, and bandwidth efficiency should be made when selecting and
+  configuring transport protocol stacks. When the underlying transport protocol
+  stack(s) support per-connection DSCP signaling, transmitted traffic SHOULD carry
+  the DSCP Best Effort PHB. \[EDITOR'S NOTE: cite]
+
+  Scavenger/Bulk: 
+  : The application is not interactive. It expects to send
+  and/or receive a large amount of data, without any urgency. This can, for
```
mwelzl: This was also in the old text and I only notice it now: why do we say "a large amount of data"?  It could also be tiny?  That seems quite unnecessary here.

draft-ietf-taps-interface.md
```diff
@@ -1904,47 +1904,70 @@ options.
...
+  Connection may fail if the desired rate cannot be maintained across the Path.
+  A transport may interpret this capacity profile as preferring a circuit
+  breaker {{?RFC8084}} to a rate-adaptive congestion controller. When the
+  underlying transport protocol stack(s) support per-connection DSCP signaling,
+  transmitted traffic SHOULD carry a DSCP Assured Forwarding
+  (AF31,AF32,AF33,AF34) PHB.
+
+  High Throughput Data: : \[EDITOR'S NOTE: Gorry wants this, but bht has no idea
+  how to implement it beyond DSCP AF2n. Cut, or specity how this is really
+  different from one of the other profiles.]
```
mwelzl: What a strange beast. Oh well. It exists and there's probably a point. As for implementation, see:  https://trac.tools.ietf.org/html/rfc4594#section-4.8   AF1n, it seems, not AF2n.

draft-ietf-taps-interface.md
britram: ah, yep, will fix.

draft-ietf-taps-interface.md
britram: agreed, will fix.

draft-ietf-taps-interface.md
britram: will change to AF1n, then wait for @gorryfair to tell me what it means.

britram:
> Transport system implementations that map the requested capacity profile onto per-connection DSCP signaling without multiplexing SHOULD assign .... to the transmitted traffic. For multiplexed connections, the guidelines in RFC 75657, Section 6 apply.

+1 to this

mwelzl:
Thanks for the commit that addresses my comments!

But this goes only half-way along my proposal, and I suspect that wasn't by intention.

My proposed text was:
***
Transport system implementations that map the requested capacity profile onto per-connection DSCP signaling without multiplexing SHOULD assign .... to the transmitted traffic. For multiplexed connections, the guidelines in RFC 75657, Section 6 apply.
***


You incorporated the following for all DSCP bits:
"For multiplexed connections, the guidelines in RFC 75657, Section 6 apply."
... but the text still says "When the underlying transport protocol stack(s) support per-connection DSCP signaling, transmitted traffic SHOULD ...", which I find too restrictive  (not suitable if I don't want to use the DSCP at all and instead implement it all only with a sender-side scheduler, for example).
Hence, please also use the first half of my proposed text, or something like it.

draft-ietf-taps-interface.md
abrunstrom: Actually, I find Scavenger/Bulk very confusing. Bulk is a large amount of data, and could still be urgent. Scavenger could be tiny or large and is not urgent. So if this is what we are talking about the name should be only Scavenger?

draft-ietf-taps-interface.md
```diff
@@ -1904,47 +1904,72 @@ options.
...
+  Connection may fail if the desired rate cannot be maintained across the Path.
+  A transport may interpret this capacity profile as preferring a circuit
+  breaker {{?RFC8084}} to a rate-adaptive congestion controller. Transport
+  system implementations that map the requested capacity profile onto
+  per-connection DSCP signaling without multiplexing SHOULD assign a DSCP
+  Assured Forwarding (AF31,AF32,AF33,AF34) {{?RFC2597}} PHB; when the Connection
+  is multiplexed, the guidelines in section 6 of {{?RFC7657}} apply.
+
+  High Throughput Data: : \[EDITOR'S NOTE: Gorry wants this, but bht has no idea
+  how to implement it beyond DSCP AF1n {{?RFC2597}}. Cut, or specity how this is
```
tfpauly: Sp: specity => specify

tfpauly: Yeah, not really sure what to do with "High Throughput". Perhaps we can just specify that this enum list of profiles is allowed to grow in the future when we have better ideas?

mwelzl: I'm concerned with us simply shrugging off stuff that has been previously defined in the IETF; the "High Throughput" service class is described in RFC 4594, Section 4.8. While I personally also find the service as described there a little awkward, I find it hard to believe that RFC 4594 was simply published with no prior discussion. If you want something newer, it's also discussed in RFC 8325. By mapping it to AF1n as described in RFC 4594, it seems simple enough to implement.

So, I'd rather keep it with a reference to this RFC and this simple mapping. But I'm not going to fight for it!

mwelzl:
@theri - "Please give me the maximum available bandwidth, possibly at the cost of latency" is not how this is defined. See RFC 4594, section 4.8: "The High-Throughput Data service class is RECOMMENDED for elastic applications that require timely packet forwarding of variable rate traffic sources and, more specifically, is configured to provide good throughput for TCP longer-lived flows."

I don't interpret "require timely packet forwarding" as "possibly at the cost of latency".

I'm also in favor of keeping it - because it has a defined meaning and implementation method as described in at least one RFC, and is hence perhaps even already used by some ISPs to treat traffic in a certain way.

theri:
@mwelzl Yes, but for a high-throughput flow, "timely packet forwarding" will more likely be achieved on a high-bandwidth path with higher latency than on a low-latency path with lower bandwidth, that's what I mean. In RFC 4593, Figure 2, it says that High-Throughput data has a "Medium-High tolerance to Delay", I think that means the same - for this traffic, you don't optimize for low latency on the path, but for high bandwidth.

draft-ietf-taps-interface.md
britram: If we want to keep it, we need text as to why an application would want this.

britram:
just committed text for High Throughput, making it clear that the *only* thing the transport system can do with this is stick a DSCP mark on it. Is this ready to go?

britram:
@tfpauly is this good to merge?

gorryfair:
Seemed you reached the correct conclusion scavenger =/= bulk.

Issue #208, opened 2018-07-19
Re-Structure Transport Properties


philsbln:
After getting a better understanding, we need to push the properties in a next-step structure.

We actually agreed on three kinds of Properties:
 - Selection Properties (stuff that effects the pre-connection)
 - Connection Properties (stuff that effects the connection)
 - Message Properties (stuff that effects the individual message)

Each of them can be set on the objects above in the list and are used as defaults then.

Each property will be marked as specific (clear protocol specific meaning) or generic (needs implementation and protocol specific meaning, but otherwise consistent behavior).
More fancy Intents are deferred to another layer above the TAPS system.

Selection Properties requesting per-Message behavior are just cross-references.

britram:
see the resolution to #218; I would also suggest moving protocol selection properties up to a section closer to the pre-establishment phase section of the document. I didn't see a response from @philsbln today so I didn't do this, not wanting to collide with any work you may have in progress. I'll probably get cycles to do this again on 15.10. though if you like; I would like to get these out for -03.

philsbln:
Fine with me – please retain data types when moving and leave overview in Section 12 analogous to the other properties.

As soon as I get spare cycles again, I will turn Section 12 in a proposal for a registry

britram:
Taking the token here. Note that Section 12 will largely be replaced a list/table showing relationships among properties and containing references back into the document where each property (split by class) will be defined. "Classification" will then disappear in the text, being implicit in the document structure.

britram:
will keep this open until there's consensus on the restructure; removing assignments and milestones to reflect that this is the state we'll have for BKK though.

mwelzl:
I think we had consensus before and in BKK, so I'm closing this.

PR #238, opened 2018-10-16
First pass at property reorganization (see #208)

britram:
This is my suggestion for cleaning up properties. It makes the following changes:

1. Section 12 Transport Properties goes away. Its introduction to what transport properties are moves into a new section 4.1 (since Properties are key to how the API works, they need to come first). Selection Properties move into Section 5.2. Some of the explanation of each class has moved to the section of the document where the class is defined (5.2 Selection Properties, 7.3 Message Context Properties, 9.1 Connection Properties).

2. Capacity Profile is a Connection Property, not a Selection Property, so it moves to Connection Properties. 

3. Directionality moves back into the appendix, because we're not done discussing it yet.

Things that (might) still need to happen, not part of this PR:

1. Reconcile Transmission Profile and Capacity Profile

2. Figure out the best way to cross reference Selection Properties with Connection and Message Context Properties they enable. Maybe this happens in a new Section 12 (as the registry work @philsbln wants to do / an 8095-style table), and/or with direct references back to Selection Properties in each Connection and Message Context Property.

3. Rename Message Context Properties

### DIFF
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -249,21 +249,61 @@ deframing callback which finds message boundaries in a stream. Messages are
 received asynchronously through a callback registered by the application.
 Errors and other notifications also happen asynchronously on the Connection.
 
-In the following sections, we describe the details of application interaction
-with Objects through Actions and Events in each phase of a Connection, following
-the phases described in {{I-D.ietf-taps-arch}}.
+{{pre-establishment}}, {{establishment}}, {{sending}}, {{receiving}}, and
+{{termination}} describe the details of application interaction with Objects
+through Actions and Events in each phase of a Connection, following the phases
+described in {{I-D.ietf-taps-arch}}.
 
-# Pre-Establishment Phase
+## Transport Properties
+
+Each application using the Transport Services Interface declares its preferences
+for how the transport service should operate using properties at each stage of
+the lifetime of a connection. During pre-establishment, Selection Properties
+{{selection-props}} are used to specify which paths and protocol stacks can be
+used and are preferred by the application, and Connection Properties
+{{connection-props}} can be used to fine-tune the eventually established
+connection. These Connection Properties can also be used to monitor and
+fine-tune established connections. The behavior of the selected protocol
+stack(s) when sending Messages is controlled by Message Context Properties
+{{message-props}}.
+
+Collectively, Selection, Connection, and Message Context Properties can be
+referred to as Transport Properties. All Transport Properties, regardless of the
+phase in which they are used, are organized within a single namespace. This
+enables setting them as defaults in earlier stages and querying them in later
+stages:
+ - Connections Properties can be set on Preconnections
+ - Message Properties can be set on Preconnections and Connections
+ - The effect of Selection Properties can be queried on Connections and Messages.
+
+Transport Properties can have one of a set of data types:
+
+- Boolean: can take the values "true" and "false"; representation is
+  implementation-dependent.
+- Integer: can take positive or negative numeric values; range and
+  representation is implementation-dependent.
+- Enumeration: can take one value of a finite set of values, dependent on the
+  property itself. The representation is implementation dependent; however,
+  implementations MUST provide a method for the application to determine the
+  entire set of possible values for each property.
+- Preference: can take one of five values (Prohibit, Avoid, Ignore, Prefer,
+  Require) for the level of preference of a given property during protocol
+  selection; see {{selection-props}}.
+
+
+
+# Pre-Establishment Phase {#pre-establishment}
 
 The pre-establishment phase allows applications to specify properties for
 the Connections they are about to make, or to query the API about potential
 connections they could make.
 
 A Preconnection Object represents a potential Connection. It has state that
-describes properties of a Connection that might exist in the future.  This
-state comprises Local Endpoint and Remote Endpoint Objects that denote the
-endpoints of the potential Connection (see {{endpointspec}}), the transport
-properties (see {{transport-props}}), and the security parameters (see
+describes properties of a Connection that might exist in the future.  This state
+comprises Local Endpoint and Remote Endpoint Objects that denote the endpoints
+of the potential Connection (see {{endpointspec}}), the Selection Properties
+(see {{selection-props}}), any preconfigured Connection Properties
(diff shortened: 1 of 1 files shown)


britram:
i'm also inclined to land this and fix the problems with it later. will wait for one more green check before doing so. 

(For those trying to review this, the important bits to look at are the new [section 4.1](https://taps-api.github.io/drafts/property-structure-208-218/draft-ietf-taps-interface.html#rfc.section.4.1) and [section 5.2](https://taps-api.github.io/drafts/property-structure-208-218/draft-ietf-taps-interface.html#rfc.section.5.2).)

draft-ietf-taps-interface.md
```diff
@@ -344,46 +384,53 @@ The Resolve() action on Preconnection can be used by the application to force
...
 Internally, the transport system will first exclude all protocols and paths
 that match a Prohibit, then exclude all protocols and paths that do not match
 a Require, then sort candidates according to Preferred properties, and then
-use Avoided properties as a tiebreaker. In case of conflicts between Protocol
-and Path Selection Properties, Path Selection Properties take precedence. For
-example, if an application indicates a preference for a specific path, but
-also a preference for a protocol not available on this path, the transport
-system will try the path first, so the Protocol Selection Property might not
-have an effect.
+use Avoided properties as a tiebreaker.
```
theri: Can this conflict between Protocol and Path Selection really just be removed? We don't distinguish between these anymore in the API, but we added this because of the observation that path and protocol selection may be at odds and this might be a relevant part of the API contract, see #149. 
I'd suggest to add this again.

draft-ietf-taps-interface.md
```diff
@@ -400,16 +447,145 @@ by using the following call on the Connection Object:
 TransportProperties := Connection.GetTransportProperties()
 ~~~
 
-{{transport-props}} provides a list of Transport Properties.
+A Connection gets its Transport Properties either by being explicitly configured
+via a Preconnection, by configuration after establishment, or by inheriting them
+from an antecedent via cloning; see {{groups}} for more.
+
+{{connection-props}} provides a list of Connection Properties, while Selection
+Properties are listed in the subsections below. . Note that many properties are
```
theri: nit: Redundant ". "

draft-ietf-taps-interface.md
```diff
@@ -344,46 +384,53 @@ The Resolve() action on Preconnection can be used by the application to force
...
-endpoints that support only specific network addresses or transports,
-transport protocol selection is necessarily tied to path selection. This may
-involve choosing between multiple local interfaces that are connected to
-different access networks.
+Properties for selecting protocol stacks and paths, as well as Connection
+Properties for configuration of the detailed operation of the selected Protocol
+Stacks.
+
+The protocol(s) and path(s) selected as candidates during establishment are
+determined and configufed using these properties. Since there could be paths
```
theri: Nit: Fix "configufed"

draft-ietf-taps-interface.md
```diff
@@ -1273,13 +1441,24 @@ Preconnection.DeframeWith(Deframer)
 {messageData} := Deframer.Deframe(OctetStream, ...)
 ~~~
 
-# Managing Connections: Setting Properties and Obtaining Information {#introspection}
+# Configuring and Querying Connection Properties {#introspection}
```
theri: Wait, this got renamed again? 
We changed it because this section also talks about Errors, which are not properties.
Maybe the title isn't ideal, but I suggest we keep the title starting with "Managing Connections".

draft-ietf-taps-interface.md
britram:
```suggestion
Properties are listed in the subsections below. Note that many properties are
```

draft-ietf-taps-interface.md
britram:
```suggestion
# Managing Connections {#introspection}
```

draft-ietf-taps-interface.md
britram:
```suggestion
determined and configured using these properties. Since there could be paths
```

draft-ietf-taps-interface.md
britram: Okay, will rephrase to capture the fact that we don't have two kinds of selection properties.

draft-ietf-taps-interface.md
britram: Properties and errors are discussed in the subheads, so maybe just Managing Connections?

draft-ietf-taps-interface.md
```diff
@@ -249,21 +249,61 @@ deframing callback which finds message boundaries in a stream. Messages are
...
+fine-tune established connections. The behavior of the selected protocol
+stack(s) when sending Messages is controlled by Message Context Properties
+{{message-props}}.
+
+Collectively, Selection, Connection, and Message Context Properties can be
+referred to as Transport Properties. All Transport Properties, regardless of the
+phase in which they are used, are organized within a single namespace. This
+enables setting them as defaults in earlier stages and querying them in later
+stages:
+ - Connections Properties can be set on Preconnections
```
tfpauly: Connections Properties, or Connection Properties?

draft-ietf-taps-interface.md
```diff
@@ -249,21 +249,61 @@ deframing callback which finds message boundaries in a stream. Messages are
...
+ - Message Properties can be set on Preconnections and Connections
+ - The effect of Selection Properties can be queried on Connections and Messages.
```
tfpauly: Nit: let's be consistent about periods at the end of bullet lists

philsbln:
Whooop… that was fast. 

I like the overall result, except deleting the overview from Section 12.
Re-introducing it for the registry proposal will be another 2-3h of unnecessary work.

britram:
IMO the useful subset of the overview content (for this document) lives in 4.1 now, and information on each class to the respective section for the class. The reorganization into Selection, Connection, and Message properties makes a lot of the boilerplate in each entry unnecessary. What, in your opinion, is still missing here?

ISTM the next step toward the registry is assigning names-in-code for each of the properties (for which I'll file a separate issue to be done post-BKK); after that's done, summary information for each property should be collected into the IANA Considerations section, with normative references to the subsections within the document for each registry entry (assuming, of course, we come to consensus that a registry is the right way to go).

PR #239, opened 2018-10-16
fix #237

### LINKED ISSUE #237: Another does of NiTs for Architecture
From /Introudction/
Please define/API/ in  /Many APIs to perform/.
---
First Para mentions TCP, but not UDP - some people could draw conclusions (very silly perhaps) that this is about TCP. It would be super nice to include an e.g. with Datagrams mentioning QUIC and UDP. ...

---
From /Design principles/
Curent text:
/and it can communicate with different Remote Endpoints to optimize performance./
My new proposed text:
/and it can communicate with different Remote Endpoints to optimize performance, robustness to failure, or some other metric/.

---
From /Access to Specialized Features/
This says:
/Since applications will often need to control fine-grained details of transport protocols to optimize their behavior and ensure compatibility with remote peers, a Transport Services system also needs to allow more specialized protocol features to be used./
I quibble. I don’t believe this is /often/, and anyway I do not think this is the most prevalent case, just and important usage, so I would suggest we write something:
/There are applications that will need to control fine-grained details of transport protocols to optimize their behavior and ensure compatibility with remote peers,. A Transport Services system will therefore also needsto allow more specialized protocol features to be used./
---
From /Access to Specialized Features/
This says:
/A specialized feature may be required by an application only when using a specific protocol, /
I suggest:
/A specialized feature could be required by an application only when it uses a specific transport protocol, /
and change:
/For example, if an application is using UDP, it may require/
to:
/For example, if an application is using UDP, it could require/
---
Please remove /In order to preserve flexibility and compatibility .../
Suggest:
/To preserve flexibility and compatibility .../
---
Please change:
/Mappings of these API features in the Implementation document, on the other hand, must explain the ramifications of each feature on existing protocols. It is expected that the Implementation document will be updated and supplemented as new protocols and protocol features are developed./
to this:
/The mappings of these API features is explained in the TAPS Implementation [REF], which also explain the implications of the feature provided by existing protocols. It is expected that this documentation will be updated and supplemented as new protocols and protocol features are developed./
---
Please change:
/require any changes on remote hosts./
suggest:
/require any changes to a remote host./
---
Please change:
/Transport protocols such as SCTP provide/
add commas:
/Transport protocols, such as SCTP, provide/
---

britram:


draft-ietf-taps-arch.md
```diff
@@ -215,9 +204,9 @@ Any Transport Services API must allow access to the distilled minimal set of fea
 
 ## Access to Specialized Features
 
-Since applications will often need to control fine-grained details of transport protocols to optimize their behavior and ensure compatibility with remote peers, a Transport Services system also needs to allow more specialized protocol features to be used. The interface for these specialized options should be exposed differently from the common options to ensure flexibility.
+There are applications that will need to control fine-grained details of transport protocols to optimize their behavior and ensure compatibility with remote peers,. A Transport Services system will therefore also needsto allow more specialized protocol features to be used. The interface for these specialized options should be exposed differently from the common options to ensure flexibility.
```
mwelzl: you copy+pasted a typo  :) the "s" in  "needsto"

PR #240, opened 2018-10-16
A minset appendix

mwelzl:
The point of this is to draw a line to minset, to make sure we're not missing something. Names of properties in the current document may be wrong, and I can fix them after the current round of updates by @britram - but this is more about what's missing than what's there.

I'm happy to report that very, very little is missing!


mwelzl:
JFYI, updates after your approval were only me playing around with the list format to make it look a little nicer. no content changes.

Issue #45, opened 2018-02-14
Do we need to make state storage explicit in the architecture and API?


britram:
Post Sockets had an Association concept, which contained all the information about a pair of endpoints, including security parameters, cached security state, cached path state information, and so on. ISTM that we implicitly decided that this was implementation-specific in the Cambridge meeting (though it lived on in the architecture as Association Cache in §4.2).

It turns out that the Connectable proposed by @csperkins in #42 shares enough of its properties with the non-cacheable bits of an Association that we might want to make this an explicit object in the API again. 

Discuss.

abrunstrom:
I think we should leave this out for v1, I think it is a whole other discussion. On the taps-level there is nothing on the other end so I find it confusing to talk about a pair of endpoints. I am not sure where the term Association Cache came from, but can we call it just Cache or at least something other than Association Cache? I find the term Association very confusing both because it seem to imply a taps counterpart on the other side and because connection and association are very similar to me (just depends on if one use TCP or SCTP :) ). The cache can contain many things, some of them related to a remote endpoint some of them not.

britram:
ok, punting to post-London

tfpauly:
Grabbing this one, as I've been working with the association concept a bunch recently, so I'd like to take a stab at editing the text.

tfpauly:
@britram I'm thinking we should punt this out to -02, that OK with you?

britram:
@tfpauly Yep. basically at this point I'm thinking that -01 is going to be #200 and #201 and we're done.

tfpauly:
@britram I'll try to get something out today; if I don't, then we can skip.

britram:
see discussion on #244 -- I think we (y'all ;) ) should spend some time on this in Bangkok, then take another pass on writing up the consensus that falls out of that.

tfpauly:
Call the connection group a connection context instead?

tfpauly:
Create explicit context to hand to preconnect

tfpauly:
Not passing any context puts you in the implicit context

britram:
On review, this should have been closed with #294 -- closing.

PR #244, opened 2018-10-19
Add CacheContext to architecture and API

tfpauly:
Address #45 by adding explicit caching context for defining state sharing boundaries. Doesn't expose knobs for accessing the cached state externally, but we can discuss and add that later if needed.

### DIFF
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -328,14 +328,15 @@ describes properties of a Connection that might exist in the future.  This state
 comprises Local Endpoint and Remote Endpoint Objects that denote the endpoints
 of the potential Connection (see {{endpointspec}}), the Selection Properties
 (see {{selection-props}}), any preconfigured Connection Properties
-({{connection-props}}), and the security parameters (see
-{{security-parameters}}):
+(see {{connection-props}}), the security parameters (see
+{{security-parameters}}), and a cache context (see {{cache-context}}):
 
 ~~~
    Preconnection := NewPreconnection(LocalEndpoint,
                                      RemoteEndpoint,
                                      TransportProperties,
-                                     SecurityParams)
+                                     SecurityParams,
+                                     CacheContext)
 ~~~
 
 The Local Endpoint MUST be specified if the Preconnection is used to Listen()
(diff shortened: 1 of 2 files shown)


draft-ietf-taps-interface.md
```diff
@@ -705,6 +706,65 @@ ChallengeCallback := NewCallback({
...
+
+Some examples of cached state include:
+
+- DNS query answers may be cached for use in subsequent Connections
+to the same hostname.
+
+- TCP round-trip-time (RTT) metrics may be cached to tune delays for
+protocol stack racing.
+
+- TLS state may be cached for use in session resumption.
```
chris-wood: Maybe drop a reference to RFC 5077?

draft-ietf-taps-interface.md
```diff
@@ -705,6 +706,65 @@ ChallengeCallback := NewCallback({
...
+
+Within a single application, there may need to be several distinct and separate
```
chris-wood: s/to be/to maintain?

draft-ietf-taps-interface.md
```diff
@@ -705,6 +706,65 @@ ChallengeCallback := NewCallback({
...
+caches to ensure that data and state is not shared between certain Connections.
+Each Preconnection is therefore associated with a CacheContext, which defines
+a distinct set of cached information. All Connections and Preconnections that are
```
chris-wood: Possibly add here, or in the implementation draft: Moreover, certain cache state may not be appropriate for re-use for all applications. For example, to improve privacy and limit client tracking, TLS 1.3 {{!RFC8446}} connection state should not be re-used absent early data to write, as this introduces an unnecessary tracking vector without saving any round trips. (It may however cost more CPU.)

draft-ietf-taps-interface.md
```diff
@@ -705,6 +706,65 @@ ChallengeCallback := NewCallback({
...
+associated with a single CacheContext are allowed to share state.
```
chris-wood: I'd piggyback this with a blurb similar to the above, wherein we only want use that shared state if necessary.

draft-ietf-taps-interface.md
```diff
@@ -705,6 +706,65 @@ ChallengeCallback := NewCallback({
...
+~~~
+CacheContext.SetDefaultEntryLifetime(SECONDS_PER_DAY)
+CacheContext.SetTotalCapacity(MAX_CACHE_ELEMENTS)
+~~~
+
+A custom CacheContext may also be configured to disable all caching
+for privacy-sensitive connections that should not preserve state.
+
+~~~
+CacheContext.SetDisableCaching(true)
```
chris-wood: I'm not sure if it's cleaner to give Preconnections a caching context that disables caching, or simply not give them a caching context.

draft-ietf-taps-interface.md
```diff
@@ -328,14 +328,15 @@ describes properties of a Connection that might exist in the future.  This state
...
+(see {{connection-props}}), the security parameters (see
+{{security-parameters}}), and a cache context (see {{cache-context}}):
 
 ~~~
    Preconnection := NewPreconnection(LocalEndpoint,
                                      RemoteEndpoint,
                                      TransportProperties,
-                                     SecurityParams)
+                                     SecurityParams,
+                                     CacheContext)
```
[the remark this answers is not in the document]
mwelzl: After discussion with Tommy in Bangkok: CacheContext should be an optional parameter.

draft-ietf-taps-interface.md
theri: In Bangkok we said that the default is that all sharing is allowed: 
If the application does not provide any explicit caching context, all Connections are associated with the same implicit caching context, so they can share state.

If we do it like this, let's make it very explicit in the text:
I suggest to move the sentence "By default, a Preconnection is associated with an implicit CacheContext." up here, clarify that it's always the same implicit CacheContext, and say that this means that by default, all Connections within the same application context are allowed to share state. 
After this, the warning about sharing TLS 1.3 connection state fits quite nicely imo.

Alternatively, we could set the default that all state can be shared _except_ TLS 1.3 connection state ...?

draft-ietf-taps-interface.md
philsbln: I tend to have a global, singleton ``DoNotCacheContext`` as a shortcut for that.

tfpauly:
Will create new PR

britram:
many thanks, please try again. :)

Per discussion at the interim meeting, we need to have a slot in the architecture that allows applications control over which connections/preconnections might share cached state, and which connections/preconnections must not share state. 

Proposal is for @tfpauly to write a new PR for -arch only, which we'll discuss on the list

PR #250, opened 2018-11-14
Add 2119 language

tfpauly:
This PR contains the architecture revisions Gorry and I went through at IETF 103. The main change is adding official 2119 language and cleaning up all other uses of should/may/must.

This does *not* contain the necessary rework of the framing section, which will come in a future PR.


draft-ietf-taps-arch.md
```diff
@@ -186,55 +188,63 @@ The Transport Services architecture is based on the survey of Services Provided
...
 
-A specialized feature could be required by an application only when using a specific protocol, and not when using others. For example, if an application is using UDP, it could require control over the checksum or fragmentation behavior for UDP; if it used a protocol to frame its data over a byte stream like TCP, it would not need these options. In such cases, the API should expose the features in such a way that they take effect when a particular protocol is selected, but do not imply that only that protocol could be used if there are equivalent options.
+A specialized feature could be required by an application only when using a specific protocol, and not when using others. For example, if an application is using UDP, it could require control over the checksum or fragmentation behavior for UDP; if it used a protocol to frame its data over a byte stream like TCP, it would not need these options. In such cases, the API ought to expose the features in such a way that they take effect when a particular protocol is selected, but do not imply that only that protocol could be used if there are equivalent options.
```
mwelzl: "if there are equivalent options" sounds a bit more limiting / stricter than I think it should. Just as a suggestion, why not say it like this: "... that only that protocol could be used. For example, allowing an application to specify constrained checksum usage will not make communication fail when using a protocol such as TCP, which always uses a checksum covering the entire payload."

draft-ietf-taps-arch.md
```diff
@@ -325,90 +335,90 @@ The diagram below provides a high-level view of the actions taken during the lif
...
 
 ### Event Handling {#events}
 
-This list of events that can be delivered to an application is not exhaustive, but gives the top-level categories of events. The API may expand this list.
+This list of events that can be delivered to an application is not exhaustive, but gives the top-level categories of events. The API MAY expand this list.
 
 * Connection Ready: Signals to an application that a given Connection is ready to send and/or receive Messages. If the Connection relies on handshakes to establish state between peers, then it is assumed that these steps have been taken.
 
-* Connection Finished: Signals to an application that a given Connection is no longer usable for sending or receiving Messages. This should deliver an error to the application that describes the nature of the termination.
+* Connection Finished: Signals to an application that a given Connection is no longer usable for sending or receiving Messages. The event SHOULD deliver an error to the application that describes the nature of the termination.
```
mwelzl: Why an "error"? If it can also be due to a graceful shutdown, then "notification" may be a better word.

draft-ietf-taps-arch.md
```diff
@@ -325,90 +335,90 @@ The diagram below provides a high-level view of the actions taken during the lif
...
+* Close: The action an application takes on a Connection to indicate that it no longer intends to send data, is no longer willing to receive data, and that the protocol SHOULD signal this state to the remote system if the transport protocol allows this.
 
-* Abort: The action the application may take on a Connection to indicate a Close, but with the additional indication that the transport system should not attempt to deliver any outstanding data.
+* Abort: The action the application takes on a Connection to indicate a Close and also indicate that the transport system SHOULD NOT attempt to deliver any outstanding data.
 
 ## Transport System Implementation Concepts
 
-The Transport System Implementation Concepts define the set of objects used internally to a system or library to provide the functionality required to provide a transport service across a network, as required by the abstract interface.
+The Transport System Implementation Concepts define the set of objects used internally to a system or library to provide the functionality needed to provide a transport service across a network, as required by the abstract interface.
```
mwelzl: double "provide" here... maybe replace the second with "implement" ?

draft-ietf-taps-arch.md
```diff
@@ -325,90 +335,90 @@ The diagram below provides a high-level view of the actions taken during the lif
...
 
 * Connection Group: A set of Connections that share properties. For multiplexing transport protocols, the Connection Group defines the set of Connections that can be multiplexed together.
 
-* Path: Represents an available set of properties that a Local Endpoint may use to send or receive packets with a Remote Endpoint.
+* Path: Represents an available set of properties that a local system can use to send or receive packets with a remote system.
```
mwelzl: I find "send or receive .... with" a weird communication. Maybe replace with "to / from" ?

draft-ietf-taps-arch.md
```diff
@@ -325,90 +335,90 @@ The diagram below provides a high-level view of the actions taken during the lif
...
+* Path Selection Properties: The Path Selection Properties consist of the options that an application can set to influence the selection of paths between the local and remote systems. These options can take  the form of requirements, prohibitions, or preferences. Examples of options that influence path selection include the interface type (such as a Wi-Fi Ethernet connection, or a Cellular LTE connection), characteristics of the path that are locally known like Maximum Transmission Unit (MTU) or discovered like Path MTU (PMTU), or predicted based on cached information like expected throughput or latency.
```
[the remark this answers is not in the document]
gorryfair: ... adding /the/ sounds fine to me.

draft-ietf-taps-arch.md
gorryfair: The approach is OK, is this better:
"For example, if an application were to specify constrained checksum usage, communication would not fail when a protocol such as TCP is selected, which always uses a checksum covering the entire payload.""

draft-ietf-taps-arch.md
gorryfair: Do you mean "send packets to or receive packets from"?

mwelzl: I meant to replace it with the term "to / from", but I can see that this doesn't read well. I suggest: "... to send packets to a Remote Endpoint or receive packets from it." in the first case (s/Remote Endpoint/remote system in the latter).

gorryfair: Woks for me:-)

draft-ietf-taps-arch.md
tfpauly: Swapping the first "provide" for "implement", actually.

draft-ietf-taps-arch.md
tfpauly: Great, fixed!

draft-ietf-taps-arch.md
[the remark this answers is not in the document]
tfpauly: Added `the`

draft-ietf-taps-arch.md
tfpauly: Made this "reason or error"

draft-ietf-taps-arch.md
tfpauly: Changed to "communicate with"

PR #252, opened 2018-11-30
Additions to the abstract and intro text

mirjak:


tfpauly:
Thanks for these fixes! I did a bit of reworking of the abstract, but it keeps the main part of the changes you made.

mirjak:
Regarding the sentence you removed, that was an adaption of an existing sentence, which I believe aimed to mention the words implementation and API to implicitly refere to the other docs (which was probably too subtle for anybody to notice anyway). However, the other change I made to that sentence was to mention one of the benefits / deployment incentives for taps (dynamic protocol and path selection). I wasn't sure about the wording here, but I think it would be good to (re-)add a sentence to the abstract saying why taps is so great and that everybody should deploy it.

PR #253, opened 2018-11-30
small wording edit on event driven API

mirjak:
I guess here is not nessecarily a notification required in sychronuous communication


tfpauly:
Merged in master to avoid conflict, and tweaked working to remove "some times", which felt a bit too colloquial in tone.

PR #259, opened 2018-11-30
Change "Basic Objects" to "Connection(s)" in diagram

mirjak:
After reading the whole doc, this seems more logical for me.

### DIFF
--- a/draft-ietf-taps-arch.md
+++ b/draft-ietf-taps-arch.md
@@ -249,7 +249,7 @@ The following diagram summarizes the top-level concepts in the architecture and 
     |        establishment  |   termination  |
     |                |      |       |        |
     |             +--v------v-------v+       |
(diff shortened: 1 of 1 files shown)


tfpauly:
The reason this had been marked as "Basic Objects" is to match with the section below on Basic Objects, which includes Preconnection, Connection, and Listener: https://taps-api.github.io/drafts/draft-ietf-taps-arch.html#rfc.section.4.1.1

However, the arrows into the diagram don't quite line up—you're right that these arrows describe the connection parts, and that pre-establishment would be going into Preconnection...

mirjak:
I understood where "Basic Objects" came from, but a) as you said that's not what's shown the diagram, and b) I find it strange to have this abstract term in the diagram. I think it'd be okay to only show connection though because it's a central concept, or alternatively you can add another box for pre-connection.

PR #264, opened 2018-11-30
Re-base 2119 language diff

tfpauly:
Re-based #250 to avoid build issue/conflict


tfpauly:
Didn't actually need this

PR #265, opened 2018-12-03
IETF 103 discussion results. Closes #37

### LINKED ISSUE #37: Make some choices about §5.2.1 Transport Selection Parameters
Do all of the transport selection parameters in §5.2.1 need to be listed in the proposed standard interface? Discuss.

mwelzl:
This does the following things with the properties from the appendix, based on https://datatracker.ietf.org/meeting/103/materials/minutes-103-taps-00.html

- direction of communication: moved to the main text
- "Suggest a timeout to the Remote Endpoint": moved to the main text but explicitly described as a TCP-specific property; also changed it to contain all the parameters that UTO really has
- traffic category: removed. Mirja said "Let's see if there's anything in the implementation document that actually needs this information" but I figured that chances are low, and if this really happens we can still re-introduce this later.
- size to be sent or received: removed
- duration: removed
- send or receive bit-rate: changed this to specify an upper and/or lower bound
- cost preferences: kept in the appendix, but inserted mirja's comment "Not convinced cost preferences are on the right level."  (@mirjak: what did you mean with "on the right level"?)

### DIFF
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -616,7 +616,52 @@ particular instance. While this does restrict path selection, it is broader than
 requiring specific PvD instances or interface instances, and should be preferred
 over these options.
 
+### Direction of communication
 
+This property specifies whether an application wants to use the connection for sending and/or receiving data.  Possible values are:
+
+Bidirectional (default):
+: The connection must support sending and receiving data
+
+Unidirectional send:
+: The connection must support sending data
+
+Unidirectional receive:
+: The connection must support receiving data
+
+In case a unidirectional connection is requested, but unidirectional connections are not supported by the transport protocol,
+the system should fall back to bidirectional transport.
+
+
+### Bounds on Send or Receive Rate
+
+This property specifies an upper-bound rate that a transfer is not expected to
+exceed (even if flow- and congestion control allow higher rates), and/or a
+lower-bound rate below which the application does not deem
+a data transfer useful. It is given in bits per second.
+
+
+### TCP-specific option: User Timeout
+
+This property specifies, for the case TCP becomes the chosen transport protocol:
+
+User Timeout:
+: the time after which the connection will be aborted if data could not be delivered
+
+Advertised User Timeout:
+: a time value to be advertised via the User Timeout Option (UTO) for the TCP at the remote endpoint
+to adapt its own user User Timeout value accordingly
+
+User Timeout Enabled:
+: a boolean (default false) to control whether the UTO option is enabled for a
+connection. This applies to both sending and receiving.
+
+Changeable:
+: a boolean (default true) which controls whether the User Timeout may be changed
+based on a UTO option received from the remote peer. This boolean becomes false when
+the 'User Timeout' is used.
+
+All of the above parameters are optional.
 
 
 ## Specifying Security Parameters and Callbacks {#security-parameters}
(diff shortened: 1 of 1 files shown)


mirjak:
Cost doesn't seem to mean too much for the transport layer. I guess it could e.g. make more sense to specified which link technology you prefer (e.g. wifi over cellular) (which would be on a less abstract level). However, still not sure what the right level is...

draft-ietf-taps-interface.md
```diff
@@ -616,7 +616,52 @@ particular instance. While this does restrict path selection, it is broader than
...
+User Timeout Enabled:
+: a boolean (default false) to control whether the UTO option is enabled for a
+connection. This applies to both sending and receiving.
+
+Changeable:
+: a boolean (default true) which controls whether the User Timeout may be changed
+based on a UTO option received from the remote peer. This boolean becomes false when
+the 'User Timeout' is used.
+
+All of the above parameters are optional.
```
theri: Same question as above: Should this be in Section 9.1? That's where we have all the other timeouts etc.

mirjak:
Thinking about cost prefernce a bit further and with respect to @theri's comment that this could part of an layer on top of taps, I would prefer to not include it expect we have some concrete implementation example guidance in the implementation draft.

draft-ietf-taps-interface.md
```diff
@@ -616,7 +616,52 @@ particular instance. While this does restrict path selection, it is broader than
...
+In case a unidirectional connection is requested, but unidirectional connections are not supported by the transport protocol,
+the system should fall back to bidirectional transport.
+
 
+### Bounds on Send or Receive Rate
+
+This property specifies an upper-bound rate that a transfer is not expected to
+exceed (even if flow- and congestion control allow higher rates), and/or a
+lower-bound rate below which the application does not deem
+a data transfer useful. It is given in bits per second.
```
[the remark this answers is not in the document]
mwelzl: I agree! will do

draft-ietf-taps-interface.md
mwelzl: I agree! will do

mwelzl:
@mirjak that all sounds very reasonable to me, thanks for clarifying!  So it seems fit where the cost preferences are now in this PR - I left them in the appendix for now, with a mark saying that we'll see if there will suitable text in the implementation draft.

draft-ietf-taps-interface.md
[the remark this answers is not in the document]
mwelzl: Regarding optional implementation, I think everything in the API draft is mandatory to offer. What's optional is to do something useful with the information, and it's also optional for the application to use it. But I think all APIs should offer all the primitives - e.g. to make applications more portable or be able to retro-fit functionality below them.

draft-ietf-taps-interface.md
```diff
@@ -1740,6 +1755,36 @@ per-Message basis using the Transmission Profile Message Property; see
 {{send-profile}}.
 
 
+### Bounds on Send or Receive Rate
+
+This property specifies an upper-bound rate that a transfer is not expected to
+exceed (even if flow- and congestion control allow higher rates), and/or a
```
tfpauly: Having `flow-` and `congestion control` seems unbalanced. Either:
- `flow- and congestion-control`
- `flow control and congestion control`

PR #267, opened 2018-12-06
Last missing bits from minset. Closes #109. Closes #246.

### LINKED ISSUE #109: API 5.2 Connection timeout
This API says:
```
* Suggest a timeout to the Remote Endpoint:
  This boolean property specifies whether an application considers it
  useful to propose a timeout until the connection is assumed to be lost.
  This property applies to Connections and Connection Groups. This is not a
  strict requirement. The default is to have this option.
```
Need to clarify if this is a timeout for connection establishment (i.e., how long should the endpoint wait before declaring that `Initiate()` failed), or it it's a timeout for data transfer after a connection has been established?

### LINKED ISSUE #246: address TODO's from minset appendix
these are a few minor changes to apply, and properties to add (e.g. "disable MPTCP")

mwelzl:
Notable changes:
- added control over using multipath (which was a late addition to minset)
- added a timeout for conn. establishment, separate from the other conn. abort timeout (see #109)
- renamed InitiateWithIdempotentSend into InitiateWithSend, and adjusted the text to recommend declaring idempotence. This addresses the minset function that goes back to TCP's send-with-SYN.

### DIFF
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -633,6 +633,14 @@ In case a unidirectional connection is requested, but unidirectional connections
 the system should fall back to bidirectional transport.
 
 
+### Timeout for aborting Connection Establishment {#conn-establish-timeout}
+
+Type:
+: Integer
+
+This property specifies how long to wait before aborting a Connection during establishment.
+It is given in seconds.
+
 
 ## Specifying Security Parameters and Callbacks {#security-parameters}
 
@@ -1245,36 +1253,32 @@ Connection.Batch(
 )
 ~~~
 
-## Send on Active Open: InitiateWithIdempotentSend {#initiate-and-send}
+## Send on Active Open: InitiateWithSend {#initiate-and-send}
 
 For application-layer protocols where the Connection initiator also sends the
-first message, the InitiateWithIdempotentSend() action combines Connection initiation with
-a first Message sent, provided that message is idempotent.
+first message, the InitiateWithSend() action combines Connection initiation with
+a first Message sent.
 
 Without a message context (as in {{send-basic}}):
 
 ~~~
-Connection := Preconnection.InitiateWithIdempotentSend(messageData)
+Connection := Preconnection.InitiateWithSend(messageData)
 ~~~
 
 With a message context (as in {{message-props}}):
 
 ~~~
-Connection := Preconnection.InitiateWithIdempotentSend(messageData, messageContext)
+Connection := Preconnection.InitiateWithSend(messageData, messageContext)
 ~~~
 
-The message passed to InitiateWithIdempotentSend() is, as suggested by the
-name, considered to be idempotent (see {{msg-idempotent}}) regardless of
-declared message properties or defaults. If protocol stacks supporting 0-RTT
-establishment with idempotent data are available on the Preconnection, then
-0-RTT establishment may be used with the given message when establishing
-candidate connections. For a non-idemponent initial message, or when the
-selected stack(s) do not support 0-RTT establishment,
-InitiateWithIdempotentSend is identical to Initiate() followed by Send().
+Whenever possible, a messageContext should be provided to declare the message passed to InitiateWithSend
+as idempotent. This allows the transport system to make use of 0-RTT establishment in case this is supported
+by the available protocol stacks. When the selected stack(s) do not support transmitting data upon connection
+establishment, InitiateWithSend is identical to Initiate() followed by Send().
 
-Neither partial sends nor send batching are supported by InitiateWithIdempotentSend().
+Neither partial sends nor send batching are supported by InitiateWithSend().
 
-The Events that may be sent after InitiateWithIdempotentSend() are equivalent to those
+The Events that may be sent after InitiateWithSend() are equivalent to those
 that would be sent by an invocation of Initate() followed immediately by an
 invocation of Send(), with the caveat that a send failure that occurs because
 the Connection could not be established will not result in a
@@ -1633,9 +1637,8 @@ are cloned.
 Type:
 : Integer
 
-This property specifies how long to wait before aborting a Connection during
-establishment, or before deciding that a Connection has failed after
-establishment. It is given in seconds.
+This property specifies how long to wait before deciding that a Connection has
+failed after establishment. It is given in seconds.
 
 ### Connection group transmission scheduler {#conn-scheduler}
 
@@ -1763,6 +1766,20 @@ lower-bound rate below which the application does not deem
 a data transfer useful. It is given in bits per second.
 
 
+### Parallel Use of Multiple Paths
+
+Type:
+: Boolean
+
+This property specifies whether an application considers it useful to
+transfer data across multiple paths between the same end hosts. Generally,
+in most cases, this will improve performance (e.g., achieve greater throughput).
+One possible side-effect is increased jitter, which may be problematic for
+delay-sensitive applications.
+
+The recommended default is to have this option.
+
+
 ### TCP-specific Property: User Timeout
 
 This property specifies, for the case TCP becomes the chosen transport protocol:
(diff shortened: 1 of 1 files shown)


draft-ietf-taps-interface.md
```diff
@@ -1763,6 +1766,20 @@ lower-bound rate below which the application does not deem
 a data transfer useful. It is given in bits per second.
 
 
+### Parallel Use of Multiple Paths
+
+Type:
+: Boolean
```
philsbln: Why Boolean? This is a selection property and should be a Preference as most others

draft-ietf-taps-interface.md
```diff
@@ -633,6 +633,14 @@ In case a unidirectional connection is requested, but unidirectional connections
 the system should fall back to bidirectional transport.
 
 
+### Timeout for aborting Connection Establishment {#conn-establish-timeout}
+
+Type:
+: Integer
+
+This property specifies how long to wait before aborting a Connection during establishment.
+It is given in seconds.
```
philsbln: Really seconds here? I can imagine time-critical applications that want to use timeouts on the range of 200-500ms

mwelzl: Good one!  We specify it in seconds, yet don't allow floats  :-D    For now, I'll just remove the statement about seconds and the Integer type here, and from "Timeout for aborting Connection", which has the same mistake.

draft-ietf-taps-interface.md
mwelzl: Why limit this to a selection property? FWIW, an application could turn this on or off later. Not in minset, but in the real MPTCP API:  https://datatracker.ietf.org/meeting/96/materials/slides-96-mptcp-4

tfpauly: There can be different levels of "aggressiveness" in using multiple paths; anything actually using MPTCP or QUIC (migration or multipath) will likely need more than just a bool to set this meaningfully.

mwelzl: TL;DR: can't we just keep the boolean for now, to have minset addressed, and discuss the rest later?  Because that may not be an easy one.

The longer version:
Ohhhhh.... I feel like I opened a gate that I shouldn't have opened, by pointing at the beyond-the-standards API slides. I mean, the thing that I'm guessing you have in mind (duplicating data across multiple paths to minimize latency vs. not doing that) really needs full control of subflows, which this API offers, but the (proposed) standard doesn't. Hence, we ended up with only the "disable" boolean in RFC 8303, and then minset - the absolute minimum that an application really needs, I guess.

As a way to keep the API simple, rather than adding more "aggression levels" or something like that, wouldn't it be better to let the application be in control, by using multiple TAPS connections as these subflows? We could maybe let it express a wish to couple the congestion control of connections if possible (because I'm guessing that, for the data duplication use case, that's the only thing that MPTCP buys you here? And... is that even a win?).

philsbln: @tfpauly I tend to rather have a "Do I want Multipath" preference in the first place (like the one @mwelzl proposed) and have the aggressiveness either controlled by an additional enum or by an Intents layer on top

abrunstrom:
If you make it a boolean it cannot be used as a selection property, so I do not think that is a good choice. Or how should it be interpreted for the selection?

mwelzl:
> If you make it a boolean it cannot be used as a selection property, so I do not think that is a good choice. Or how should it be interpreted for the selection?

@abrunstrom what do you mean, that it should be a (binary - yes/no) Preference instead?  I thought this would be more limiting but I'm fine with this as well.

draft-ietf-taps-interface.md
```diff
@@ -1245,36 +1250,32 @@ Connection.Batch(
...
-declared message properties or defaults. If protocol stacks supporting 0-RTT
-establishment with idempotent data are available on the Preconnection, then
-0-RTT establishment may be used with the given message when establishing
-candidate connections. For a non-idemponent initial message, or when the
-selected stack(s) do not support 0-RTT establishment,
-InitiateWithIdempotentSend is identical to Initiate() followed by Send().
+Whenever possible, a messageContext should be provided to declare the message passed to InitiateWithSend
+as idempotent. This allows the transport system to make use of 0-RTT establishment in case this is supported
+by the available protocol stacks. When the selected stack(s) do not support transmitting data upon connection
+establishment, InitiateWithSend is identical to Initiate() followed by Send().
```
theri: Perhaps just a question to understand this:
What if the InitiateWithSend() message is not marked as idempotent, but the protocol stack supports 0-RTT?
In this case, no 0-RTT happens, right?

draft-ietf-taps-interface.md
theri: FWIW I agree that there should be different levels, like "no multipath", "only as backup path", and "balance load" or something. But I'm fine with keeping this a Boolean to start with and discuss later.

draft-ietf-taps-interface.md
mwelzl: Exactly

draft-ietf-taps-interface.md
philsbln: @theri really boolean (as in the original version) or preference (which would match the other selection properties)?

abrunstrom:
@mwelzl I meant that it should be a Preference as pointed out in the original comment by @philsbln. I do not understand what a boolean means during selection (is it a requirement or a preference?) I also do not understand what it means in the middle of a connection. You mean you should try and use a single or several subflows for scheduling data, assuming you already have a multipath connection? I guess this is possible, but then the description needs to be more clear that this is what it means.
In the appendix about minset your have "Disable MPTCP: "Parallel Use of Multiple Paths" Property." I assume this refers to the Protocol Selection phase? It does not make sense to disable or enable MPTCP in the middle of a connection I think.
 
I guess part of how to specify the property depends on how to interpret this part from the intro to the Managing Connections section:
"The application can set and query Connection Properties on a per-Connection basis. Connection Properties that are not read-only can be set during pre-establishment (see {{selection-props}}), as well as on connections directly using the SetProperty action"
I interpreted this as properties set during pre-establishment are selection properties, but maybe this is not the intention as they may need to be expressed differently? If they are not intended as selection properties, we are missing a selection property for multipath I guess.

mwelzl:
@abrunstrom 

> @mwelzl I meant that it should be a Preference as pointed out in the original comment by @philsbln. 

Ok.


> I do not understand what a boolean means during selection (is it a requirement or a preference?) I also do not understand what it means in the middle of a connection. You mean you should try and use a single or several subflows for scheduling data, assuming you already have a multipath connection? I guess this is possible, but then the description needs to be more clear that this is what it means.

That's what I meant, yes - but I agree that's not covered by the description, it really is a selection preference only as it's written now.


> In the appendix about minset your have "Disable MPTCP: "Parallel Use of Multiple Paths" Property." I assume this refers to the Protocol Selection phase? It does not make sense to disable or enable MPTCP in the middle of a connection I think.

OK  (though you CAN do it with the "real" API, by closing subflows, but this goes beyond the spec'd protocol, and I'm also not sure it's a good idea, especially when limited to only enabling / disabling it altogether).


> I guess part of how to specify the property depends on how to interpret this part from the intro to the Managing Connections section:
> "The application can set and query Connection Properties on a per-Connection basis. Connection Properties that are not read-only can be set during pre-establishment (see {{selection-props}}), as well as on connections directly using the SetProperty action"
> I interpreted this as properties set during pre-establishment are selection properties, but maybe this is not the intention as they may need to be expressed differently? If they are not intended as selection properties, we are missing a selection property for multipath I guess.

I would have understood this as in your first interpretation: "properties set during pre-establishment are selection properties".

abrunstrom:
@mwelzl 
> OK (though you CAN do it with the "real" API, by closing subflows, but this goes beyond the spec'd protocol, and I'm also not sure it's a good idea, especially when limited to only enabling / disabling it altogether).
Closing a subflow is not the same as disabling MPTCP to me. You will still be using MPTCP, only with a single subflow. And you can certainly not open a subflow unless MPTCP is already enabled.

But seems we now have a common understanding.

Issue #256, opened 2018-11-30
Reconcile Property groupings between Arch and API


mirjak:
Reviewing the arch doc right now which talkes about path and protocol selection properties, however, the API doc now has transport properties, connection properties, and message properties... I think we need to realign the terms here....

mwelzl:
I admit to being slightly lost now with the naming and re-organization process... so maybe this is a left-over thing that should no longer exist... but: the API draft says "Collectively, Selection, Connection, and Message Properties can be referred to as Transport Properties."

Unless I'm mistaken, this is what's needed:

- API draft: it seems strange how the list of selection properties (from 5.2.1 onward) appears out of nowhere. There should be a new heading, 5.3, just before what is now 5.2.1, called "Selection properties"

- Arch draft: with some exceptions that should perhaps be removed, the API draft doesn't really distinguish between "path" and "protocol" selection properties. Can we remove this distinction from the arch draft? Because of the availability of a protocol with certain abilities, a transport system might also make a decision about a certain path, and a path selection property might lead to a decision for a path that rules out a certain protocol... so this distinction seems to me to be a bit artificial anyway.

britram:
closed by 5b173e2

PR #272, opened 2018-12-20
Re-align Properties in Architecture - closes Issue #256

philsbln:
- fixes Properties Terminology and aligns them to the API 
- adds Properties as "Basic Object"
- updates Figure 2

### DIFF
--- a/draft-ietf-taps-arch.md
+++ b/draft-ietf-taps-arch.md
@@ -292,7 +292,7 @@ Fundamentally, a Transport Services API needs to provide basic objects ({{object
 
 Beyond the basic objects, there are several high-level groups of actions that any Transport Services API implementing this specification MUST provide:
 
-* Pre-Establishment ({{preestablishment}}) encompasses the properties that an application can pass to describe its intent, requirements, prohibitions, and preferences for its networking operations. For any system that provides generic Transport Services, these properties SHOULD be defined to apply to multiple transport protocols. Some properties are specific to a single transport protocol, and these are be exposed as non-generic interfaces (see Specific Protocol Properties {{preestablishment}}). Properties can have a large impact on the rest of the aspects of the interface: they modify how establishment occurs, they influence the expectations around data transfer, and they determine the set of events that will be supported.
+* Pre-Establishment ({{preestablishment}}) encompasses the properties that an application can pass to describe its intent, requirements, prohibitions, and preferences for its networking operations. For any system that provides generic Transport Services, these Selection Properties SHOULD be defined to apply to multiple transport protocols. Selection Properties can have a large impact on the rest of the aspects of the interface: they modify how establishment occurs, they influence the expectations around data transfer, and they determine the set of events that will be supported.
 
 * Establishment ({{establishment}}) focuses on the actions that an application takes on the basic objects to prepare for data transfer.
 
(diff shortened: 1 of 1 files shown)


draft-ietf-taps-arch.md
```diff
@@ -336,12 +331,19 @@ The diagram below provides a high-level view of the actions taken during the lif
...
 
 * Connection: A Connection object represents an active transport protocol instance that can send and/or receive Messages between local and remote systems. It holds state pertaining to the underlying transport protocol instance and any ongoing data transfer. This represents, for example, an active connection in a connection-oriented protocol such as TCP, or a fully-specified 5-tuple for a connectionless protocol such as UDP.
 
 * Listener: A Listener object accepts incoming transport protocol connections from remote systems  and generates corresponding Connection objects. It is created from a Preconnection object that specifies the type of incoming connections it will accept.
 
+* Transport Properties: Transport Properties allow the application to configure the Transport System and express their requirements, prohibitions, and preferences. There are three kinds of Transport Properties: 
+  Selection Properties are used influence connection establishment.
```
britram: does this render correctly as a sublist when compiled to txt/html?

draft-ietf-taps-arch.md
```diff
@@ -311,22 +311,17 @@ The diagram below provides a high-level view of the actions taken during the lif
...
- |   +----------+          :      |                        :
- +-->| Listener |-----------------+                        :
+ |                         :       ^  ^  ^  |    Finished  :
+ +-- Local Endpoint        :       |  |  |  |              :
+ |                         :       |  |  |  +---------+    :
+ +-- Remote Endpoint       :  +----+  |  +-----+      |    :
+ |                         :  |       |Send()  |      |    :
+ +-- Selection Properties  :  |  +---------+   |      v    :
+ +-- Connection Properties ---+  | Message |   |  Message  :
+ +-- Message Properties -------->| to send |   |  Received :
```
britram: why are Message Properties bound to Preconnections? The availability of some Message Properties may depend on Selection Properties, and one can set default Message Properties, but this rendering makes it look like setting defualt Message Properties on a Preconnection is non-optional...

draft-ietf-taps-arch.md
```diff
@@ -336,12 +331,19 @@ The diagram below provides a high-level view of the actions taken during the lif
...
 
 * Connection: A Connection object represents an active transport protocol instance that can send and/or receive Messages between local and remote systems. It holds state pertaining to the underlying transport protocol instance and any ongoing data transfer. This represents, for example, an active connection in a connection-oriented protocol such as TCP, or a fully-specified 5-tuple for a connectionless protocol such as UDP.
 
 * Listener: A Listener object accepts incoming transport protocol connections from remote systems  and generates corresponding Connection objects. It is created from a Preconnection object that specifies the type of incoming connections it will accept.
 
+* Transport Properties: Transport Properties allow the application to configure the Transport System and express their requirements, prohibitions, and preferences. There are three kinds of Transport Properties:
```
britram: I'm (still) not really convinced of the value of a terminological container for all properties -- does this grouping exist because there really is something useful to be said about all properties together, or as an artifact of how we edited the API doc?

draft-ietf-taps-arch.md
philsbln: At the moment, it is just a wall of text – but if turned into a sublist, it renders fine. Will update it to become a sublist for the sake of readability.

draft-ietf-taps-arch.md
philsbln: We do not distinguish between optional and mandatory in the drawing at all, e.g., we have listen and imitate paths in there.

The only problem I see is that we might have to clarify in API that "default message properties specified on a pre-connection that do not apply do not cause errors – if you need the property, set the corresponding selection property to require "

draft-ietf-taps-arch.md
britram: ok, will drop this objection.

draft-ietf-taps-arch.md
britram: +1, it makes sense to note this at the level of the archietcture; can you add this sentence when doing the sublist change above, then I'll merge...

philsbln:
@britram's comments should be addressed now.

Thanks for the reviews

Issue #230, opened 2018-09-21
Split "Notification of excessive retransmission" in a Connection and a Selection Property


philsbln:
As discussed in #227

PR #274, opened 2019-01-04
Updating implementation

mwelzl:
Several small updates, mostly to clean things up a bit, and align property names and such.


draft-ietf-taps-impl.md
```diff
@@ -327,31 +327,29 @@ Branching for derived endpoints is the final step, and may have multiple layers
 
 Implementations should sort the branches of the tree of connection options in order of their preference rank. 
 Leaf nodes on branches with higher rankings represent connection attempts that will be raced first.
-Implementations should order the branches to reflect the preferences expressed by the application for its new connection, including Protocol and Path Selection Properties, which are specified in {{I-D.ietf-taps-interface}}. 
+Implementations should order the branches to reflect the preferences expressed by the application for its new connection, including Selection and Connection Properties, which are specified in {{I-D.ietf-taps-interface}}. 
 
 In addition to the properties provided by the application, an implementation may include additional criteria such as cached performance estimates, see {{performance-caches}}, or system policy, see {{role-of-system-policy}}, in the ranking.
-Two examples of how the Protocol and Path Selection Properties may be used to sort branches are provided below:
+Two examples of how the Selection and Connection Properties may be used to sort branches are provided below:
```
abrunstrom: Connection properties should not be used for this.

draft-ietf-taps-impl.md
```diff
@@ -327,31 +327,29 @@ Branching for derived endpoints is the final step, and may have multiple layers
...
 
-* Interface Type:
+* Selection Property "Interface Instance or Type":
 If the application specifies an interface type to be preferred or avoided, implementations should rank paths accordingly.
```
abrunstrom: Selection Property should not be needed here.

draft-ietf-taps-impl.md
```diff
@@ -327,31 +327,29 @@ Branching for derived endpoints is the final step, and may have multiple layers
...
 If the application specifies an interface type to be required or prohibited, we expect an implementation to not include the non-conforming paths into the three.
 
-* Capacity Profile:
+* Connection Property "Capacity Profile":
 An implementation may use the Capacity Profile to prefer paths optimized for the application's expected traffic pattern according to cached performance estimates, see {{performance-caches}}:
```
abrunstrom: I think there is a matching Selection Property missing in the API description, if it should be used for selection.

draft-ietf-taps-impl.md
```diff
@@ -327,31 +327,29 @@ Branching for derived endpoints is the final step, and may have multiple layers
...
-   * Scavenger/Bulk:
-     Prefer paths with the highest expected available bandwidth, based on observed maximum throughput
-
-\[Note: See {{branch-sorting-non-consensus}} for additional examples related to Properties under discussion.]
 
 Implementations should process properties in the following order: Prohibit, Require, Prefer, Avoid.
-If Protocol or Path Selection Properties contain any prohibited properties, the implementation should first purge branches containing nodes with these properties. For required properties, it should only keep branches that satisfy these requirements. Finally, it should order branches according to preferred properties, and finally use avoided properties as a tiebreaker.
+If Selection or Connection Properties contain any prohibited properties, the implementation should first purge branches containing nodes with these properties. For required properties, it should only keep branches that satisfy these requirements. Finally, it should order branches according to preferred properties, and finally use avoided properties as a tiebreaker.
```
abrunstrom: Again only selection properties.

draft-ietf-taps-impl.md
```diff
@@ -327,31 +327,29 @@ Branching for derived endpoints is the final step, and may have multiple layers
...
+For Require and Avoid, path selection takes precedence over protocol selection.
+For example, if the application has indicated both a preference for WiFi over LTE and for a feature only available in SCTP, branches will be first sorted accord to path selection, with WiFi at the top. Then, branches with SCTP will be sorted to the top within their subtree according to the properties influencing protocol selection. However, if the implementation has cached the information that SCTP is not available on the path over WiFi, there is no SCTP node in the WiFi subtree. Here, the path over WiFi will be tried first, and, if connection establishment succeeds, TCP will be used. So the Selection Property of preferring WiFi takes precedence over the Property that led to a preference for SCTP.
```
abrunstrom: This example should actually be moved to the previous paragraph where the  branchingorder is discussed or deleted.

draft-ietf-taps-impl.md
```diff
@@ -444,9 +448,9 @@ However, if a peer is not reachable over the network using the unconnected proto
...
 
-If the Path Selection Properties do not require a single network interface or path, but allow the use of multiple paths, the Listener object should register for incoming traffic on all of the network interfaces or paths that conform to the Path Selection Properties. The set of available paths can change over time, so the implementation should monitor network path changes and register and de-register the Listener across all usable paths. When using multiple paths, the Listener is generally expected to use the same port for listening on each.
+If the Selection or Connection Properties do not require a single network interface or path, but allow the use of multiple paths, the Listener object should register for incoming traffic on all of the network interfaces or paths that conform to the Properties. The set of available paths can change over time, so the implementation should monitor network path changes and register and de-register the Listener across all usable paths. When using multiple paths, the Listener is generally expected to use the same port for listening on each.
```
abrunstrom: Selection Properties only.

draft-ietf-taps-impl.md
```diff
@@ -444,9 +448,9 @@ However, if a peer is not reachable over the network using the unconnected proto
...
 
-If the Protocol Selection Properties allow multiple protocols to be used for listening, and the implementation supports it, the Listener object should register across the eligble protocols for each path. This means that inbound Connections delivered by the implementation may have heterogeneous protocol stacks. 
+If the Selection or Connection Properties allow multiple protocols to be used for listening, and the implementation supports it, the Listener object should register across the eligble protocols for each path. This means that inbound Connections delivered by the implementation may have heterogeneous protocol stacks.
```
abrunstrom: Same here.

draft-ietf-taps-impl.md
```diff
@@ -522,23 +527,31 @@ It is also possible that protocol stacks within a particular leaf node use 0-RTT
...
+If the Protocol Stack includes a transport protocol that supports multipath connectivity, an update to the available paths should inform the Protocol Instance of the new set of paths that are permissible based on the Selection or Connection Properties passed by the application. A multipath protocol can establish new subflows over new paths, and should tear down subflows over paths that are no longer available. If the Protocol Stack includes a transport protocol that does not support multipath, but support migrating between paths, the update to available paths can be used as the trigger to migrating the connection. For protocols that do not support multipath or migration, the Protocol Instances may be informed of the path change, but should not be forcibly disconnected if the previously used path becomes unavailable. An exception to this case is if the System Policy changes to prohibit traffic from the Connection based on its properties, in which case the Protocol Stack should be disconnected.
```
abrunstrom: Again only Selection Properties.

mwelzl:
@abrunstrom - I think you're wrong: my understanding (and also the way the text is written - otherwise we'd have to change it in the API draft) was that Connection Properties can (and should) also be set on Preconnections, to give as much input as possible for Connection establishment. In the API draft, the Selection Properties in Section 5.2 can *only* be used on Preconnections: you can't ask for congestion control after establishment. You can't say that you'd rather have multistreaming support after establishment. But that doesn't exclude Connection properties from being usable earlier.

Required minimum checksum coverage could influence choosing UDP-Lite vs. UDP. Bounds on Send or Receive Rate could influence the interface choice. The capacity profile could influence various things (and no, there isn't a matching Selection Property now - there is a per-Message override).

So... I think that either you're wrong, or I'm missing something and the API draft heavily lags behind a decision that was made. I hope not!

abrunstrom:
@mwelzl The API drafts says: 

> Each application using the Transport Services Interface declares its preferences for how the transport service should operate using properties at each stage of the lifetime of a connection. During pre-establishment, Selection Properties (see {{selection-props}}) are used to specify which paths and protocol stacks can be used and are preferred by the application, and Connection Properties (see {{connection-props}}) can be used to fine-tune the eventually established connection. These Connection Properties can also be used later, to monitor and fine-tune established connections -- but configuring Connection Properties on Preconnections is preferred because they can then can influence decisions made during establishment.

I interpret this as only the Selection Properties influence selection and the Connection properties are used to further configure the connections according to user preferences. Otherwise what is the difference between them?

mwelzl:
If you interpret it as you do, I think the text isn't clear enough (indeed it isn't super clear - in particular, "eventually established" is misleading IMO). Though, note that it says that Connection properties can *also* be used later.

Regarding the difference between them: Selection Properties *only* make sense for Preconnections.

draft-ietf-taps-impl.md
abrunstrom: Sorry, my comment was messed up. I meant to say that the example should move to the previous section (not paragraph), where the branching order is defined. It is not connected to Require and Avoid so that sentence can be removed.

mwelzl:
@abrunstrom the other PR means nothing about what I said here: it only tried to address issue #230 which is about a wording inconsistency in a specific property.

You say that Connection properties can also be used during establishment - if you agree with allowing this, then why not allow to configure them as early as possible and allow them to influence selection? What's bad about having more information available that can guide the selection process?

draft-ietf-taps-impl.md
mwelzl: understood, done

abrunstrom:
@mwelzl the problem is that the Connection properties are not expressed in a way suitable for selection. If they should also be used for Selection they would have to be reworked to be of type Preference. Otherwise I do not know what they mean for selection. What does it mean if I pick say a particular connection group transmission scheduler? Is it required or preferred? Most connection properties also do not look very useful to me for selection so in case there are some we need I think it would be simpler to just define a matching selection property.

mwelzl:
@abrunstrom we're only discussing whether it's allowed to use them as additional input or not. I agree, some (like the scheduler choice) don't seem useful, but some do... generally we may not have good guidance on how they should be used by an implementation, but giving the system freedom to use as many inputs as possible is probably a good thing.

Regarding the type, require/prefer//ignore/avoid/prohibit obviously works fine for a "wishlist" of boolean character: having something, or not having it. The selection properties are generally like that - except "Interface Instance or Type" and "Provisioning Domain Instance or Type" which are, well... wishlist-collections.

Consider the capacity profile. Surely configuring this early can be useful input - e.g. if someone picks low latency / interactive, you may even want to put this traffic on a specific network interface!  But if you'd want to make a copy of the capacity profile as a "proper" selection property, how would you do it?  Apply require/prefer/ignore/avoid/prohibit for each of its values? Then I could say "prefer" low latency/non-interactive but "avoid" low latency/non-interactive, which would surely be odd. Apply the qualifier to the whole thing? Then it's about, e.g. "avoiding" to use a capacity profile at all. That's odd as well.

I think it's in the nature of the current selection properties that they merely say "I want to be able to do this, and this is how important it is to me", which is different from configuring something. That's also why, as written, they simply wouldn't work for later connection configuration.

Connection properties really are about configuring something, but I still say that being able to use them as early input is good. I agree that the implementation isn't super clear as they don't come with a require/prefer/... etc. qualifier, but then we just offer some freedom to implementers.

draft-ietf-taps-impl.md
mwelzl: I just made a commit that updates all other occurrences of "Selection and Connection", but I left this one standing on purpose: I think, because Connection Properties CAN be used to guide the decisions later, it is useful to have them in the example here as a way to remind the reader of this possibility.

draft-ietf-taps-impl.md
```diff
@@ -150,9 +152,9 @@ It is important to fail as early as possible in such cases in order to avoid all
 
 ## Role of system policy
 
-The properties specified during pre-establishment has a close connection to system policy. The implementation is responsible for combining and reconciling several different sources of preferences when establishing Connections. These include, but are not limited to:
+The properties specified during pre-establishment have a close connection to system policy. The implementation is responsible for combining and reconciling several different sources of preferences when establishing Connections. These include, but are not limited to:
 
-1. Application preferences, i.e., preferences specified during the pre-establishment such as Local Endpoint, Remote Endpoint, Path Selection Properties, and Protocol Selection Properties.
+1. Application preferences, i.e., preferences specified during the pre-establishment via Selection Properties.
```
philsbln: nit: Again mention other properties: "or derived from early specification of Connection or Message Properties"

draft-ietf-taps-impl.md
philsbln: I think this would duplicate this property a second time and leads to confusion. I really prefer using the early-provided connection property here.

draft-ietf-taps-impl.md
```diff
@@ -474,23 +480,24 @@ Protocols that provide the framing (such as length-value protocols, or protocols
...
 
-- Niceness: this represents the ability to de-prioritize a Message in favor of other Messages. This can be implemented by the system re-ordering Messages that have yet to be handed to the Protocol Stack, or by giving relative priority hints to protocols that support priorities per Message. For example, an implementation of HTTP/2 could choose to send Messages of different niceness on streams of different priority.
+- Niceness: this represents the ability to de-prioritize a Message in favor of other Messages. This can be implemented by the system re-ordering Messages that have yet to be handed to the Protocol Stack, or by giving relative priority hints to protocols that support priorities per Message.
```
philsbln: This has been renamed to "Priority" in the API (after writing this PR)

draft-ietf-taps-impl.md
```diff
@@ -807,29 +822,3 @@ Research Council under grant EP/R04144X/1.
...
-
-* Send / Receive Bitrate:
-If the application indicates an expected send or receive bitrate, an implementation may prefer a path that can likely provide the desired bandwidth, based on cached maximum throughput, see {{performance-caches}}. The application may know the Send or Receive Bitrate from metadata in adaptive HTTP streaming, such as MPEG-DASH.
-
-* Cost Preferences:
-If the application indicates a preference to avoid expensive paths, and some paths are associated with a monetary cost, an implementation should decrease the ranking of such paths. If the application indicates that it prohibits using expensive paths, paths that are associated with a cost should be purged from the decision tree.
```
philsbln: The Cost Preference is still in the API as experimental, so I would like to also keep it in the Implementation

draft-ietf-taps-impl.md
```diff
@@ -807,29 +822,3 @@ Research Council under grant EP/R04144X/1.
...
-
-* Size to be Sent or Received:
-An implementation may use the Size to be Sent or Received in combination with cached performance estimates, see {{performance-caches}}, e.g. the observed Round Trip Time and the observed maximum throughput, to compute an estimate of the completion time of a transfer over different available paths. It may then prefer the path with the shorter expected completion time. This property may be used instead of the Capacity profile, as the application does not always know whether its transfer will be latency-bound or bandwidth-bound, and thus may not be able to specify a Capacity Profile. However, the application may know the Size to be Sent or Received from metadata, e.g., in adaptive HTTP streaming such as MPEG-DASH, or in operating system upgrades. A related paper is currently under submission.
-
-* Send / Receive Bitrate:
-If the application indicates an expected send or receive bitrate, an implementation may prefer a path that can likely provide the desired bandwidth, based on cached maximum throughput, see {{performance-caches}}. The application may know the Send or Receive Bitrate from metadata in adaptive HTTP streaming, such as MPEG-DASH.
```
philsbln: This has been included in the API as "Bounds on Send or Receive Rate" and should stay

draft-ietf-taps-impl.md
abrunstrom: @philsbln The earlier text covers the fact that you may also use connection properties if available. Let us not put them back into the text everywhere again. And there is also a distinction as you must take selection properties into account, but the connection properties are additional input that you can use if you like.

draft-ietf-taps-impl.md
```diff
@@ -135,7 +135,9 @@ Listener objects are created with a Preconnection, at which point their configur
 
 # Implementing Pre-Establishment
 
-During pre-establishment the application specifies the Endpoints to be used for communication as well as its preferences regarding Protocol and Path Selection. The implementation stores these objects and properties as part of the Preconnection object for use during connection establishment. For Protocol and Path Selection Properties that are not provided by the application, the implementation must use the default values specified in the Transport Services API ({{I-D.ietf-taps-interface}}).
+During pre-establishment the application specifies the Endpoints to be used for communication as well as its preferences via Selection Properties and, if desired, also Connection Properties. Generally, Connection Properties should be configured as early as possible, as they may serve as input to decisions that are made by the implementation (the Capacity Profile may guide usage of a protocol offering scavenger-type congestion control, for example). In the remainder of this document, we only refer to Selection Properties because they are the more typical case.
```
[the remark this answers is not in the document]
abrunstrom: Perhaps we should add to the last sentence "we only refer to Selection Properties because they are the more typical case and have to be handled by all implementations" or similar.  

An implementation could ignore the Connection properties, but not the Selection properties (it then depend on their setting if they can be ignored in the second step or not, but they have to be considered and if set to require or prohibit also satisfied).

draft-ietf-taps-impl.md
```diff
@@ -135,7 +135,9 @@ Listener objects are created with a Preconnection, at which point their configur
...
+
+The implementation stores these objects and properties as part of the Preconnection object for use during connection establishment. For Selection Properties that are not provided by the application, the implementation must use the default values specified in the Transport Services API ({{I-D.ietf-taps-interface}}).
```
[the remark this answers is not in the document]
abrunstrom: @philsbln This may depend on the property. I do not think that it can be changed to a general "should". You can for instance not change from reliable to unreliable, even if your platform is constrained. Properties that can break the application if changed must follow the default if not provided by the application. I think a must in combination with good defaults should work. For properties that are set to prefer or avoid the implementation can always choose to ignore them in favor of other constraints.

draft-ietf-taps-impl.md
```diff
@@ -379,7 +379,13 @@ Each approach is appropriate in different use-cases and branch types. However, t
 
 The timing algorithms for racing should remain independent across branches of the tree. Any timers or racing logic is isolated to a given parent node, and is not ordered precisely with regards to other children of other nodes.
 
-### Delayed Racing
+
+### Immediate
+
+\[MICHAEL: This seems like an easy thing to do, and its description may be obvious, but I don't think we can skip it -- it just looks odd to first introduce the three variants and then only describe two of them.]
```
tfpauly: The reason for leaving out immediate is that we recommend not using it, `implementations should not use immediate racing`.

draft-ietf-taps-impl.md
tfpauly: Good point, let's update that.

draft-ietf-taps-impl.md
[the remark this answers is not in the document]
mwelzl: Done with the wording you propose

draft-ietf-taps-impl.md
mwelzl: I don't see the problem but removed "Selection Property" here and "Connection Property" in the next item, that should be a safe way out either way.

draft-ietf-taps-impl.md
mwelzl: @philsbln this isn't really a discussion for this PR (which is just trying to reflect the API in the implementation draft).
Anyway, to answer your concern: the capacity profile now exists in the interface draft as a Connection Property and as a per-Message override; the latter contains a statement about being inconsistent, and being fixed in the future.

Let's not get hung up on this now in this PR and fix the capacity profile later.

draft-ietf-taps-impl.md
mwelzl: Understood. I removed it.

draft-ietf-taps-impl.md
mwelzl: I agree to include the bounds but that is different text.

draft-ietf-taps-impl.md
mwelzl: Done, sorry

tfpauly:
Fixed up branch building and resolved merge conflicts. I'll merge this in later today before the posting deadline.

mwelzl:
Cool thanks!  Sorry if I messed something up

PR #275, opened 2019-01-08
Split Notification of Excessive Retransmission, and make the preference to receive Soft Errors a Selection Property

### LINKED ISSUE #230: Split "Notification of excessive retransmission" in a Connection and a Selection Property

mwelzl:
This closes #230 and addresses a comment about soft errors by @theri in #227:

- "Notification of Excessive Retransmission" is made a Selection Property
- There's now an "Excessive Retransmission" Event (just like the one for Soft Errors)
- The preference to receive Soft Errors is also made a Selection Property

### DIFF
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -589,6 +589,7 @@ such as metered or unmetered network access. If an application needs to prohibit
 metered interfaces, this should be specified via Provisioning Domain attributes
 (see {{prop-pvd}}) or another specific property.
 
+
 ### Provisioning Domain Instance or Type {#prop-pvd}
 
 Similar to interface instances and types (see {{prop-interface}}), this property
@@ -618,6 +619,7 @@ particular instance. While this does restrict path selection, it is broader than
 requiring specific PvD instances or interface instances, and should be preferred
 over these options.
 
+
 ### Parallel Use of Multiple Paths
 
 This property specifies whether an application considers it useful to
@@ -646,7 +648,30 @@ In case a unidirectional connection is requested, but unidirectional connections
 the system should fall back to bidirectional transport.
 
 
-### Timeout for aborting Connection Establishment {#conn-establish-timeout}
+### Notification of excessive retransmissions {#prop-establish-retrans-notify}
+
+Type:
+: Boolean
+
+This property specifies whether an application considers it useful to be
+informed in case sent data was retransmitted more often than a certain
+threshold. The recommended default is to have this option.
+
+
+### Notification of ICMP soft error message arrival {#prop-soft-error}
+
+Type:
+: Boolean
+
+This property specifies whether an application considers it useful to be
+informed when an ICMP error message arrives that does not force termination of a
+connection. When set to true, received ICMP errors will be available as
+SoftErrors. Note that even if a protocol supporting this property is selected,
+not all ICMP errors will necessarily be delivered, so applications cannot rely
+on receiving them. The recommended default is not to have this option.
+
+
+### Timeout for aborting Connection Establishment {#prop-establish-timeout}
 
 This property specifies how long to wait before aborting a Connection during establishment.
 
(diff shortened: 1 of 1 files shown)


draft-ietf-taps-interface.md
```diff
@@ -646,7 +648,30 @@ In case a unidirectional connection is requested, but unidirectional connections
 the system should fall back to bidirectional transport.
 
 
-### Timeout for aborting Connection Establishment {#conn-establish-timeout}
+### Notification of excessive retransmissions {#prop-establish-retrans-notify}
+
+Type:
```
britram: This is a selection property, so Type is implied (preference)

draft-ietf-taps-interface.md
```diff
@@ -646,7 +648,30 @@ In case a unidirectional connection is requested, but unidirectional connections
...
+: Boolean
+
+This property specifies whether an application considers it useful to be
+informed in case sent data was retransmitted more often than a certain
+threshold. The recommended default is to have this option.
+
+
+### Notification of ICMP soft error message arrival {#prop-soft-error}
+
+Type:
```
britram: same as above

mwelzl:
Oops sorry, will fix

mwelzl:
@theri - the latter was exactly my thinking. if someone doesn’t want these notifications they can just ignore them.

abrunstrom:
I agree with Gorry that you can probably remove them. If anything I would see it as a Connection properties, not as Selection properties. How likely is it that you will have two transport stacks to select from where one supports soft errors the other not and this is what you will want to use to make a decision? Does not seem like a strong use case.

tfpauly:
To Gorry's point, how widely do we expect these to be supported for various protocols? And is the meaning of excessive retransmits consistent across protocols that do retransmission (TCP, SCTP, QUIC)?

draft-ietf-taps-interface.md
```diff
@@ -646,7 +648,24 @@ In case a unidirectional connection is requested, but unidirectional connections
...
+
+
+### Notification of ICMP soft error message arrival {#prop-soft-error}
+
+This property specifies whether an application considers it useful to be
+informed when an ICMP error message arrives that does not force termination of a
+connection. When set to true, received ICMP errors will be available as
+SoftErrors. Note that even if a protocol supporting this property is selected,
+not all ICMP errors will necessarily be delivered, so applications cannot rely
+on receiving them. The recommended default is not to have this option.
```
philsbln: As ICMP is unreliable anyway, the remark is somewhat redundant.

gorryfair: There could be a perception issue here, because we see this differently.

* First, transports ignore soft errors (as specified by the IETF), primarily because after years of experience this was found to be the correct thing to do. Sure this extra information *could* help in some cases, it can also be very unhelpful in many cases (one obvious case is where there is some parallelism and packets are forwarded across more than one set of nodes - e.g. after a failover,  ... when the path changes, this is made worse when there are delayed messages due to intermittent forwarding ). I'd suggest the end to argument identifies these messages are for diagnostics, not to influence the forwarding/transport.

(2) I think you suggest the logic does not exist to forward this presently through the API. That is as I expected, and adding this as a requirement to TAPS is adding another requirement to the stack developer, TCP implementor, etc - all of which seems like a hurdle to deploying TAPS rather than a selling point. Many UDP API's don't even signal hard errors.

Based on this, I argue soft errors from the network (ICMP) must not be passed to an application. Soft errors from the transport (such as retry failures, timeouts) could be passed if there is a need.

Gorry

mwelzl: @gorry - I'm not sure you answered in the right place - I think you're answering my comment, not Phil's?  Or maybe it is Phil's, because of your point (2), where I don't understand why you say "I think you suggest the logic does not exist to forward this presently through the API. "

Either way: your argument of not passing ICMP messages to the application at all makes sense to me, yet it's a divergence from minset and as such not offering everything that TCP and UDP offer. In a way it's quite a serious divergence, as we're stepping away from RFC 1122 here.

An option I see is to take this discussion to the mailing list, and if there's agreement to do this, do it and capture the reasoning in one of our documents.

gorryfair: P.S. I just re-read RFC 5461, and I like the phrase:

 "When the connection timer expires and an ICMP soft error message has
   been received before the timeout, TCP can use this information to
   provide the user with a more specific error message (see [Stevens],
   pp. 317-319)."

P.P.S It wouldn't be a deployment hurdle, if it were optional to implement;-).

mwelzl: I like that phrase too but I'm not sure this changes anything about the API... it's a TCP-internal thing. I like the idea of going optional !

philsbln: I agree that _requiring_ an TAPS implementation to forward soft errors is a deployment hurdle, but I also think it is a useful feature. 
What I would like to see is _requiring the support of the property_ but making the functionality optional (so the property can be a NO-OP). I am fine with adding a warning against relying on ICMP generating soft errors.

mwelzl: @philsbln @gorryfair I wanted to update the text to address your comment now, but then I saw that the text about "Soft Errors" already says:

"This will only happen if the underlying protocol stack supports
access to soft errors; however, even if the underlying stack supports it, there
is no guarantee that a soft error will be signaled."

The part after the semicolon is what you want, right?

So ... all is optional. Which text changes are you then asking for?

philsbln: @mwelzl I am fine with the text for soft errors. 

We should just add a general remark about optional / NO-OP selection properties in a separate PR.

PR #276, opened 2019-01-11
Define a IANA registry for Transport Property Names

philsbln:
This PR contains a rough skeleton for an IANA registry fro property names.

It is mainly a proposal and starting point and will be discussed on the Jan'19 Interim.


philsbln:
Superseded by PR #291

PR #279, opened 2019-01-23
Minor adjustment to the semantics of Reliable Data Transfer (Message)…

### LINKED ISSUE #273: 7.3.7. Reliable Data Transfer (Message) redundant?
Because we also have 7.3.1 Lifetime, with the statement "When a Message’s Lifetime is infinite, it must be transmitted reliably."

mwelzl:
… and Lifetime. Closes #273.

### DIFF
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -1113,9 +1113,12 @@ Type:
 
 Lifetime specifies how long a particular Message can wait to be sent to the
 remote endpoint before it is irrelevant and no longer needs to be
-(re-)transmitted. When a Message's Lifetime is infinite, it must be
-transmitted reliably. The type and units of Lifetime are
-implementation-specific.
+(re-)transmitted. This is a hint to the transport system -- it is not guaranteed
+that a Message will not be sent when its Lifetime has expired.
+Setting a Message's Lifetime to infinite indicates that the application does
+not wish to apply a time constraint on it, but it does not express a need for
+reliable delivery; reliability is adjustable per Message via the "Reliable Data Transfer (Message)"
+property (see {{#msg-reliable-message}}). The type and units of Lifetime are implementation-specific.
 
 ### Niceness {#msg-niceness}
 
@@ -1192,11 +1195,14 @@ guaranteed, any other requests are advisory.
 Type:
 : Boolean
 
-This property specifies that a message should be sent in such a way
+This property, if set to true, specifies that a message should be sent in such a way
 that the transport protocol ensures all data is received on the other side
 without corruption. Changing the ´Reliable Data Transfer´ property on Messages
 is only possible if the Connection supports reliability.
 When this is not the case, changing it will generate an error.
+Disabling this property indicates that the transport system may disable retransmissions
+or other reliability mechanisms for this particular Message, but such disabling is not guaranteed.
+
 
 ### Transmission Profile {#send-profile}
 
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -1115,10 +1115,11 @@ Lifetime specifies how long a particular Message can wait to be sent to the
 remote endpoint before it is irrelevant and no longer needs to be
 (re-)transmitted. This is a hint to the transport system -- it is not guaranteed
 that a Message will not be sent when its Lifetime has expired.
+
 Setting a Message's Lifetime to infinite indicates that the application does
 not wish to apply a time constraint on it, but it does not express a need for
 reliable delivery; reliability is adjustable per Message via the "Reliable Data Transfer (Message)"
-property (see {{#msg-reliable-message}}). The type and units of Lifetime are implementation-specific.
+property (see {{msg-reliable-message}}). The type and units of Lifetime are implementation-specific.
 
 ### Niceness {#msg-niceness}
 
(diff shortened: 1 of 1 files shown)


draft-ietf-taps-interface.md
```diff
@@ -1113,9 +1113,12 @@ Type:
...
 remote endpoint before it is irrelevant and no longer needs to be
-(re-)transmitted. When a Message's Lifetime is infinite, it must be
-transmitted reliably. The type and units of Lifetime are
-implementation-specific.
+(re-)transmitted. This is a hint to the transport system -- it is not guaranteed
+that a Message will not be sent when its Lifetime has expired.
+Setting a Message's Lifetime to infinite indicates that the application does
+not wish to apply a time constraint on it, but it does not express a need for
+reliable delivery; reliability is adjustable per Message via the "Reliable Data Transfer (Message)"
+property (see {{#msg-reliable-message}}). The type and units of Lifetime are implementation-specific.
```
britram: do this to unbreak the build:

```suggestion
property (see {{msg-reliable-message}}). The type and units of Lifetime are implementation-specific.
```

draft-ietf-taps-interface.md
```diff
@@ -1113,9 +1113,12 @@ Type:
 
 Lifetime specifies how long a particular Message can wait to be sent to the
 remote endpoint before it is irrelevant and no longer needs to be
-(re-)transmitted. When a Message's Lifetime is infinite, it must be
-transmitted reliably. The type and units of Lifetime are
-implementation-specific.
+(re-)transmitted. This is a hint to the transport system -- it is not guaranteed
+that a Message will not be sent when its Lifetime has expired.
+Setting a Message's Lifetime to infinite indicates that the application does
```
britram: new paragraph?

draft-ietf-taps-interface.md
```diff
@@ -1192,11 +1195,14 @@ guaranteed, any other requests are advisory.
 Type:
 : Boolean
 
-This property specifies that a message should be sent in such a way
+This property, if set to true, specifies that a message should be sent in such a way
```
britram: tiny style nit:

```suggestion
When true, this property specifies that a message should be sent in such a way
```

draft-ietf-taps-interface.md
mwelzl: Very sorry for this one!

draft-ietf-taps-interface.md
mwelzl: ok

draft-ietf-taps-interface.md
mwelzl: ack

mwelzl:
Thanks @britram, all done

draft-ietf-taps-interface.md
```diff
@@ -1113,9 +1113,13 @@ Type:
...
 Lifetime specifies how long a particular Message can wait to be sent to the
 remote endpoint before it is irrelevant and no longer needs to be
-(re-)transmitted. When a Message's Lifetime is infinite, it must be
-transmitted reliably. The type and units of Lifetime are
-implementation-specific.
+(re-)transmitted. This is a hint to the transport system -- it is not guaranteed
+that a Message will not be sent when its Lifetime has expired.
+
+Setting a Message's Lifetime to infinite indicates that the application does
+not wish to apply a time constraint on it, but it does not express a need for
```
abrunstrom: I find "a time constraint on it" confusing. Perhaps this should be "does not wish to apply a time constraint on the transmission of the message". An infinite lifetime does mean it should be transmitted at least once.

draft-ietf-taps-interface.md
```diff
@@ -1192,11 +1196,14 @@ guaranteed, any other requests are advisory.
 Type:
 : Boolean
 
-This property specifies that a message should be sent in such a way
+When true, this property specifies that a message should be sent in such a way
 that the transport protocol ensures all data is received on the other side
 without corruption. Changing the ´Reliable Data Transfer´ property on Messages
 is only possible if the Connection supports reliability.
```
abrunstrom: What does the 

> Connection supports reliability

 mean? The connection is using a protocol that support reliability or the application has asked for a reliable connection? If I ask for an unreliable connection and it happens to get implemented with TCP, can I ask for a reliable message?  I would say no, as this would fail most of the time. So maybe this should be "Connection is relable"?

draft-ietf-taps-interface.md
mwelzl: I'm sorry, I merged this PR too quickly. I agree with your comment and will address it with a direct commit to the master.

draft-ietf-taps-interface.md
mwelzl: I agree with the idea of what you say, but not with the wording "Connection is reliable" because that sounds contradictory - we'd have a sentence saying that changing reliable data transfer per message is only possible if the Connection is reliable.  (but if it's reliable, why can you then change it?)

I think you mean that this should only be possible for Connections that were established with the Selection Property "Reliable Data Transfer (Connection)" enabled. So I propose to phrase it as: "Changing the ´Reliable Data Transfer´ property on Messages
is only possible for Connections that were established with the Selection Property 'Reliable Data Transfer (Connection)' enabled."

Please ACK or NACK - then I'll also apply this as a direct commit to the master. Sorry again for merging too fast.

abrunstrom: @mwelzl -Yes that is what I  I meant, thanks.

mwelzl: @abrunstrom: great, thanks - I now committed both these changes to the master.

PR #280, opened 2019-01-29
Change Niceness to Priority, as agreed with @britram. Closes #228

### LINKED ISSUE #228: Rename either Conn. or Msg. Niceness
I suggest:
Connections have Priorities
Messages have Niceness

Also, for Message Niceness (if that's the name that stays), we need to clarify that this can, for example, influence how messages are ordered in the send buffer (as per Mirja's suggestion).

This issue is the result of the discussion on Niceness at the interim. See:
https://etherpad.tools.ietf.org/p/notes-ietf-taps-interim-2018-09-12

mwelzl:
Not many text fixes needed - quite a bit of text reflected the old choice of Priority anyway and wasn't correctly updated for Niceness   :-)

### DIFF
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -966,14 +966,14 @@ result in a CloneError:
 Connection -> CloneError<>
 ~~~
 
-The Protocol Property "Niceness" operates on entangled Connections as in {{msg-niceness}}:
+The Protocol Property "Priority" operates on entangled Connections as in {{msg-priority}}:
 when allocating available network
 capacity among Connections in a Connection Group, sends on Connections with
-higher Niceness values will be prioritized over sends on Connections with
-lower Niceness values. An ideal transport system implementation would assign
+higher Priority values will be prioritized over sends on Connections with
+lower Priority values. An ideal transport system implementation would assign
 each Connection the capacity share (M-N) x C / M, where N is the Connection's
-Niceness value, M is the maximum Niceness value used by all Connections in the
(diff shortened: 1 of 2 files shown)


draft-ietf-taps-interface.md
```diff
@@ -1121,22 +1121,22 @@ not wish to apply a time constraint on the transmission of the Message, but it d
...
+This property represents a hierarchy of priorities.
 It can specify the priority of a Message, relative to other Messages sent over the
 same Connection.
 
-A Message with Niceness 0 will yield to a Message with Niceness 1, which will
-yield to a Message with Niceness 2, and so on. Niceness may be used as a
+A Message with Priority 0 will yield to a Message with Priority 1, which will
+yield to a Message with Priority 2, and so on. Priorities may be used as a
 sender-side scheduling construct only, or be used to specify priorities on the
 wire for Protocol Stacks supporting prioritization.
```
britram: should we say something here about what the max prio is?

mwelzl: I think not - this should be application dependent. Giving a max of 10 or 100 or whatever they want. I only removed the statement to say that it's unbounded - with Niceness or Priority, this could be done, but is that better or worse than limiting it? I don't know. Let them decide.

philsbln: I see no reason to specify a concrete "max prio" – but perhaps we should define a read-only property to query it

britram: okay, we'll just leave this as undefined (and a nonportable value in the abstract API, which is also fine)

PR #281, opened 2019-01-29
Closes #207. Added "preservation of message boundaries" selection property; turned…

### LINKED ISSUE #207: There is no Selection Property for Preservation of Message Boundaries
Just realized that during Tommy's Presentation

mwelzl:
… establishment timeout into a parameter of init calls (it stood out strangely in the list of selection properties because it really isn't one); updated defaults for selection properties.

### DIFF
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -1644,7 +1644,7 @@ are cloned.
 
 This property specifies how long to wait before deciding that a Connection has
 failed when trying to reliably deliver data to the destination. Adjusting this Property
-is only take effect when the underlying stack supports reliability.
+will only take effect when the underlying stack supports reliability.
 
 
 ### Connection group transmission scheduler {#conn-scheduler}
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -503,21 +503,27 @@ implemented over UDP.
 This property specifies whether the application needs to use a transport
 protocol that ensures that all data is received on the other side without
 corruption. This also entails being notified when a Connection is closed or
-aborted. The recommended default is to enable Reliable Data Transfer.
+aborted. The recommended default is to Require Reliable Data Transfer.
+
+### Preservation of Message Boundaries {#prop-boundaries}
+
+This property specifies whether the application needs to use a transport
+protocol that preserves message boundaries. The recommended default
+is to Prefer Preservation of Message Boundaries.
 
 ### Configure per-Message reliability {#prop-partially-reliable}
 
 This property specifies whether an application considers it useful to indicate
 its reliability requirements on a per-Message basis. This property applies to
-Connections and Connection Groups. The recommended default is to not have this
+Connections and Connection Groups. The recommended default is to Prefer this
 option.
 
 ### Preservation of data ordering {#prop-ordering}
 
 This property specifies whether the application wishes to use a transport
 protocol that can ensure that data is received by the application on the other
-end in the same order as it was sent. The recommended default is to preserve
-data ordering.
+end in the same order as it was sent. The recommended default is to Require
+Preservation of data ordering.
 
 ### Use 0-RTT session establishment with an idempotent Message {#prop-0rtt}
 
@@ -525,22 +531,21 @@ This property specifies whether an application would like to supply a Message to
 the transport protocol before Connection establishment, which will then be
 reliably transferred to the other side before or during Connection
 establishment, potentially multiple times. See also {{msg-idempotent}}. The
-recommended default is to not have this option.
+recommended default is to Prefer this option.
 
 ### Multistream Connections in Group {#prop-multistream}
 
 This property specifies that the application would prefer multiple Connections
 within a Connection Group to be provided by streams of a single underlying
-transport connection where possible. The recommended default is to not have
+transport connection where possible. The recommended default is to Prefer have
 this option.
 
 ### Control checksum coverage on sending or receiving {#prop-checksum-control}
 
 This property specifies whether the application considers it useful to enable,
 disable, or configure a checksum when sending a Message, or configure whether to
-require a checksum or not when receiving.  The recommended default is full
-checksum coverage without the option to configure it, and requiring a checksum
-when receiving.
+require a checksum or not when receiving.  The recommended default is to Prefer
+this option for sending, and Require a checksum when receiving.
 
 ### Congestion control {#prop-cc}
 
@@ -550,7 +555,7 @@ controlled, an application using such a Connection should itself perform
 congestion control in accordance with {{?RFC2914}}. Also note that reliability
 is usually combined with congestion control in protocol implementations,
 rendering "reliable but not congestion controlled" a request that is unlikely to
-succeed. The recommended default is that the Connection is congestion
+succeed. The recommended default is to Require that the Connection is congestion
 controlled.
 
 
@@ -560,7 +565,7 @@ This property allows the application to select which specific network interfaces
 or categories of interfaces it wants to `Require`, `Prohibit`, `Prefer`, or
 `Avoid`.
 
-In contrast to other Selection Properties, this property is tuple of an
+In contrast to other Selection Properties, this property is a tuple of an
 (Enumerated) interface identifier and a preference, and can either be
 implemented directly as such, or for making one preference available for each
 interface and interface type available on the system.
@@ -599,7 +604,7 @@ Provisioning Domains or categories of Provisioning Domains it wants to
 consistent sets of network properties that may be more specific than network
 interfaces {{RFC7556}}.
 
-As with interface instances and types, this property is tuple of an (Enumerated)
+As with interface instances and types, this property is a tuple of an (Enumerated)
 PvD identifier and a preference, and can either be implemented directly as such,
 or for making one preference available for each interface and interface type
 available on the system.
@@ -626,16 +631,14 @@ This property specifies whether an application considers it useful to
 transfer data across multiple paths between the same end hosts. Generally,
 in most cases, this will improve performance (e.g., achieve greater throughput).
 One possible side-effect is increased jitter, which may be problematic for
-delay-sensitive applications.
-
-The recommended default is to have this option.
+delay-sensitive applications. The recommended default is to Prefer this option.
 
 
 ### Direction of communication
 
 This property specifies whether an application wants to use the connection for sending and/or receiving data.  Possible values are:
 
-Bidirectional (default):
+Bidirectional (default: Prefer):
 : The connection must support sending and receiving data
 
 Unidirectional send:
@@ -652,7 +655,7 @@ the system should fall back to bidirectional transport.
 
 This property specifies whether an application considers it useful to be
 informed in case sent data was retransmitted more often than a certain
-threshold. The recommended default is to have this option.
+threshold. The recommended default is to Prefer this option.
 
 
 ### Notification of ICMP soft error message arrival {#prop-soft-error}
@@ -662,13 +665,7 @@ informed when an ICMP error message arrives that does not force termination of a
 connection. When set to true, received ICMP errors will be available as
 SoftErrors. Note that even if a protocol supporting this property is selected,
 not all ICMP errors will necessarily be delivered, so applications cannot rely
-on receiving them. The recommended default is not to have this option.
-
-
-### Timeout for aborting Connection Establishment {#prop-establish-timeout}
-
-This property specifies how long to wait before aborting a Connection during establishment.
-
+on receiving them. The recommended default is to Avoid this option.
 
 
 ## Specifying Security Parameters and Callbacks {#security-parameters}
@@ -775,9 +772,10 @@ client-server interactions. Active open is supported by this interface through t
 Initiate Action:
 
 ~~~
-Connection := Preconnection.Initiate()
+Connection := Preconnection.Initiate(timeout)
 ~~~
 
+The timeout parameter specifies how long to wait before aborting Active open.
 Before calling Initiate, the caller must have populated a Preconnection
 Object with a Remote Endpoint specifier, optionally a Local Endpoint
 specifier (if not specified, the system will attempt to determine a
@@ -816,8 +814,8 @@ available Paths and/or Protocol Stacks meeting the constraints is empty) or
 reconciled with the local and/or remote endpoints; when the remote specifier
 cannot be resolved; or when no transport-layer connection can be established to
 the remote endpoint (e.g. because the remote endpoint is not accepting
-connections, or the application is prohibited from opening a Connection by the
-operating system).
+connections, the application is prohibited from opening a Connection by the
+operating system, or the establishment attempt has timed out for any other reason).
 
 See also {{initiate-and-send}} to combine Connection establishment
 and transmission of the first message in a single action.
@@ -1298,13 +1296,13 @@ a first Message sent.
 Without a message context (as in {{send-basic}}):
 
 ~~~
-Connection := Preconnection.InitiateWithSend(messageData)
+Connection := Preconnection.InitiateWithSend(messageData, timeout)
 ~~~
 
 With a message context (as in {{message-props}}):
 
 ~~~
-Connection := Preconnection.InitiateWithSend(messageData, messageContext)
+Connection := Preconnection.InitiateWithSend(messageData, messageContext, timeout)
 ~~~
 
 Whenever possible, a messageContext should be provided to declare the message passed to InitiateWithSend
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -291,7 +291,7 @@ Transport Properties can have one of a set of data types:
   property itself. The representation is implementation dependent; however,
   implementations MUST provide a method for the application to determine the
   entire set of possible values for each property.
-- Preference: can take one of five values (Prohibit, Avoid, Ignore, Prefer,
+- Preference: can take one of five values (Prohibit, Avoid, Neutral, Prefer,
   Require) for the level of preference of a given property during protocol
   selection; see {{selection-props}}.
 
@@ -434,7 +434,7 @@ preference levels:
    |------------|------------------------------------------------------------------------|
    | Require    | Select only protocols/paths providing the property, fail otherwise     |
    | Prefer     | Prefer protocols/paths providing the property, proceed otherwise       |
-   | Ignore     | Cancel any system default preference for this property                 |
+   | Neutral    | No preference                                                          |
    | Avoid      | Prefer protocols/paths not providing the property, proceed otherwise   |
    | Prohibit   | Select only protocols/paths not providing the property, fail otherwise |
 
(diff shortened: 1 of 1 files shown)


draft-ietf-taps-interface.md
```diff
@@ -503,44 +503,49 @@ implemented over UDP.
 This property specifies whether the application needs to use a transport
 protocol that ensures that all data is received on the other side without
 corruption. This also entails being notified when a Connection is closed or
-aborted. The recommended default is to enable Reliable Data Transfer.
+aborted. The recommended default is to Require Reliable Data Transfer.
```
philsbln: Good point - I take this as editorial change

draft-ietf-taps-interface.md
```diff
@@ -503,44 +503,49 @@ implemented over UDP.
...
+
+### Preservation of Message Boundaries {#prop-boundaries}
+
+This property specifies whether the application needs to use a transport
+protocol that preserves message boundaries. The recommended default
+is to Prefer Preservation of Message Boundaries.
```
philsbln: s/needs/needs or prefers/

draft-ietf-taps-interface.md
```diff
@@ -626,16 +631,14 @@ This property specifies whether an application considers it useful to
 transfer data across multiple paths between the same end hosts. Generally,
 in most cases, this will improve performance (e.g., achieve greater throughput).
 One possible side-effect is increased jitter, which may be problematic for
-delay-sensitive applications.
-
-The recommended default is to have this option.
+delay-sensitive applications. The recommended default is to Prefer this option.
 
 
 ### Direction of communication
```
philsbln: Despite this being a selection property, this is rather an Enum then a Preference. This was most probably lost on the pass that moved this property up in the document and removed the types.
As unidirectional applications can just ignore the other direction, using this as an Enum is safe.

draft-ietf-taps-interface.md
```diff
@@ -775,9 +772,10 @@ client-server interactions. Active open is supported by this interface through t
 Initiate Action:
 
 ~~~
-Connection := Preconnection.Initiate()
+Connection := Preconnection.Initiate(timeout)
```
philsbln: Why do we need a timeout as a mandatory parameter here? I am fine with adding an option one, as we have no other means for this yet, but mandatory?

theri: I'd say this is a Connection Property rather than a Selection Property, so it should go in Section 9.1 -- in fact it already is there: 9.1.4 is "Timeout for aborting Connection" (which is missing a Type, but we can easily add that).

So let's not add it as a mandatory parameter here, but have it as a Connection Property with some default.

draft-ietf-taps-interface.md
theri: I'm fine with this staying a simple selection property. Having more granular control would be nice, but I'm not sure if we need to standardize that.

@philsbln 
By Enum you mean something like "Use second path as fallback", "aggregate bandwidth", or "do not use multiple paths"? Or are there more possible values?
I'm not sure what "the other direction" would mean here?

draft-ietf-taps-interface.md
mwelzl: I didn't mean to make it mandatory, sorry - happy to make it optional. Having it here seems better: all the other properties express a wish (or requirement, or wish to avoid, ... etc.) about the Connection that we're about to get - but this one doesn't. It really stands out in the list, as being nothing but a detail of how init works - so I thought it seems much cleaner to have it here as a parameter.

draft-ietf-taps-interface.md
mwelzl: I'm lost - this is about "Direction of communication", right? But it already has multiple values (bidirectional vs. unidirectional)?

theri: I thought this was about "Parallel Use of Multiple Paths".

draft-ietf-taps-interface.md
theri: Wait, Wishes (or requirements etc) are Selection Properties, not Connection properties. 
Connection properties do not express a wish (or requirement etc), but they set a specific variable, such as a timeout, to a specific value, such as a number of seconds.

I see your point that it's not really the property of a Connection, but of a connection attempt or something like this.
But then, does the Connection property in 9.1.4 make sense? Or is this yet something else?

mwelzl: It's something else - but now I notice that the wording there ("before deciding that a Connection has failed after establishment.") is a bit misleading. This (9.1.4) is the timeout that decides to abort a connection when data could not be delivered too long. It can be changed at any time, whereas the init. timeout just decides how long to try establishing before giving up.

mwelzl: .... and now I noticed how my answer to you was confusing. Sorry - indeed, you talked about the Connection Property which is already there. This update (of making the timeout a (with the next commit, optional) parameter of init) is about removing the init timeout only, 5.2.14.

theri: Okay, makes sense :)
Then I'm fine with this timeout being an optional parameter to Initiate() (Maybe it's better to say explicitly that it's optional?), but I'd like it if 9.1.4 was made a bit more clear.

draft-ietf-taps-interface.md
```diff
@@ -503,44 +503,49 @@ implemented over UDP.
...
+This property specifies whether the application needs to use a transport
+protocol that preserves message boundaries. The recommended default
+is to Prefer Preservation of Message Boundaries.
 
 ### Configure per-Message reliability {#prop-partially-reliable}
 
 This property specifies whether an application considers it useful to indicate
 its reliability requirements on a per-Message basis. This property applies to
-Connections and Connection Groups. The recommended default is to not have this
+Connections and Connection Groups. The recommended default is to Prefer this
```
britram: Really? I mean I'm all for a set of defaults that drives SCTP (or future PR-QUIC) usage, but I don't think most applications care about this.

draft-ietf-taps-interface.md
```diff
@@ -503,44 +503,49 @@ implemented over UDP.
...
 establishment, potentially multiple times. See also {{msg-idempotent}}. The
-recommended default is to not have this option.
+recommended default is to Prefer this option.
 
 ### Multistream Connections in Group {#prop-multistream}
 
 This property specifies that the application would prefer multiple Connections
 within a Connection Group to be provided by streams of a single underlying
-transport connection where possible. The recommended default is to not have
+transport connection where possible. The recommended default is to Prefer have
```
britram: Same comment as above.

(It actually seems to me that smarter TAPS implementations could even extract the preferences/requirements of an application from static code analysis. An app that never clones connections doesn't need multistreaming, as an app that never sets deadlines/reliability doesn't need partial reliability)

draft-ietf-taps-interface.md
```diff
@@ -503,44 +503,49 @@ implemented over UDP.
...
 
 ### Control checksum coverage on sending or receiving {#prop-checksum-control}
 
 This property specifies whether the application considers it useful to enable,
 disable, or configure a checksum when sending a Message, or configure whether to
-require a checksum or not when receiving.  The recommended default is full
-checksum coverage without the option to configure it, and requiring a checksum
-when receiving.
+require a checksum or not when receiving.  The recommended default is to Prefer
+this option for sending, and Require a checksum when receiving.
```
britram: This language was IMO not correct in the initial version; the fix to change preference is also not right... I think this is two separate properties: require checksum on receive, control checksum coverage on send.

Prefer checksum coverage control on sending will tend to always select UDP-Lite when e.g. TAPS-enable Linux machines are talking to each other. Is this what we want?

draft-ietf-taps-interface.md
```diff
@@ -652,7 +655,7 @@ the system should fall back to bidirectional transport.
 
 This property specifies whether an application considers it useful to be
 informed in case sent data was retransmitted more often than a certain
-threshold. The recommended default is to have this option.
+threshold. The recommended default is to Prefer this option.
```
britram: not Require?

draft-ietf-taps-interface.md
britram: In general, I think we need to handle optional parameters better in the API doc. Will file an issue.

draft-ietf-taps-interface.md
mwelzl: Prefer only means that you'd rather get a protocol that can do this if possible. It's harmless to request. A part of "not caring" is perhaps also the fact that this is so often not available?

draft-ietf-taps-interface.md
mwelzl: The above (supporting per-message reliability) is a matter of "does the application care", as in "will it use this functionality"?  However this is a very different case!  This is about being allowed to group them without the application caring about it - I think we definitely want to do this, to benefit from multistreaming whenever possible. The benefits can be large even without app involvement, as both Connections get a single congestion control context.

For our API design, let's not presume that the underlying system can see into the future by analyzing the app code. Sequentially, this is before an application could do "clone".

draft-ietf-taps-interface.md
mwelzl: I agree that these should be two properties. Will fix.

About always selecting UDP-Lite: why not? The problem is reachability, and that's for the TAPS transport system to figure out. It's certainly correct to use UDP - not much bad will happen.

draft-ietf-taps-interface.md
mwelzl: No, because it's only available in TCP (I traced it back in the minset => RFC 8303 path). E.g., SCTP can't do that. So this is then up to the transport system to weigh against other preferences... I thought "Prefer" because it seems to make sense to have.

draft-ietf-taps-interface.md
mwelzl: So I think that @philsbn was talking about "Direction of communication" but got confused. Then @theri thought it's about "Parallel Use of Multiple Paths". So altogether we're all confused, I think. Here's a suggestion: let me continue with this as-is on this PR, and fix it afterwards.

draft-ietf-taps-interface.md
mwelzl: done

draft-ietf-taps-interface.md
mwelzl: done - I added "optional: " in front of the parameter. I suggest to accept this as-is for now and fix afterwards because it's a general style thing - if you don't like this specific style proposal, no problem to change it.

draft-ietf-taps-interface.md
philsbln: It will be difficult to do multi-streaming without explicit action by the application (calling clone).

The "automagic multi-streaming mode" works for my connection pool proposal (I should just write an PR for that), but not for regular connections. We tried searching for "matching connections" on ```socketconnect()```in our Socket Intents Prototype, but this was really tedious and error prone.
So requiring/preferring it really means "the application may want to call ```clone()```"

draft-ietf-taps-interface.md
philsbln: Sorry – I was talking about "Direction of communication" … made the comment above the section.

Direction of communication is not a preference – It has three distinct values whereby one of these values can be used as an obvious fallback.

draft-ietf-taps-interface.md
philsbln: It is not clear to me why an SCTP implementation should not be able to gather this kind of information, but I think Prefer is just right. Many stacks will not implement this feature and applications need to be careful in using this signal anyway.

draft-ietf-taps-interface.md
mwelzl: "Automagic" streaming mode is possible and has always been part of the model - see minset. Also, NEAT is able to do this. For details, see:
F. Weinrank and M. Tüxen. “Transparent Flow Mapping for NEAT“. In Workshop on Future of Internet Transport (FIT 2017), Stockholm, June 2017.
http://dl.ifip.org/db/conf/networking/networking2017/1570349380.pdf

draft-ietf-taps-interface.md
mwelzl: SCTP: it's simply not part of the spec. See RFC 8303.

draft-ietf-taps-interface.md
abrunstrom: I think this discussion around defaults, here and at other places in the document, indicates that we need to have a "not defined" option as a default for several of them. 
The indicated preferences should be used to make the selection of the transport. This should not be influenced by various parameters that are not relevant to the application. I would not like to get or not get a transport becuase of a default setting of a parameter that is not relevant for my application. On the other hand, if I prefer a protocol that provides for instance partial relaiability I would like that to have some impact on what is selected. So it should not be "harmless" to request something.

mwelzl: I totally agree!  Actually I think the issue is - I tried to make a choice between the defined preference levels, and couldn't find a reasonable "neutral" one: the middle one is defined as:

> Ignore  - Cancel any system default preference for this property

... but "Cancel any system default" hardly makes sense for a recommended default. I think we want the middle preference level to mean "don't care".  (which I'd prefer over "not defined" because, well, there must be a default value, so what does "not defined" mean?)

So I suggest:

> Neutral - No preference for or against this property

draft-ietf-taps-interface.md
philsbln: @mwelzl in case of NEAT, this is explicitly negotiated with the peer, which is a contradiction to the TAPS requirement to work with a non-TAPS peer.
If we accept that Transparent Flow Mapping is a protocol feature that might be present, I am unsure how does this will interact with an application that explicitly uses ```clone()``` to initiate another stream? 
 - Do we need a parameter to turn off Transparent Flow Mapping to allow manual flow mapping with clone?
 - What about protocols like HTTP/3 that have a fixed stream-id to protocol semantic mapping?

draft-ietf-taps-interface.md
britram: hm, i was confused here. question withdrawn.

draft-ietf-taps-interface.md
britram: We already have "Ignore", which could be slightly redefined to be used for this purpose.

draft-ietf-taps-interface.md
britram: yep, let's rename Ignore to Neutral.

draft-ietf-taps-interface.md
mwelzl: The negotiation happens in SCTP - NEAT itself doesn't know or care about the negotiation, it just locally tries to use SCTP with all necessary features. So saying this contradicts TAPS is like saying "using MPTCP contradicts TAPS because it requires negotiation"; it's just a local thing to try. But the idea is to definitely allow (even encourage!) people implementing TAPS systems to support this, as it can yield a lot of benefits (reduced FCTs, less overall delay from less congestion control interactions, ..) without the application having to worry about everything.

About your questions:
1) No, you can still do this. (Why do you think you couldn't?)

2) Does it, *still*? Sigh... well I guess it's just a matter of giving a stream number to a Preconnection. That should be a protocol-specific property, I guess. Moreover, the requirement level for "Multistream Connections in Groups" will have to be "Require" if you want that level of control.

draft-ietf-taps-interface.md
mwelzl: Got it - fixed it by changing the text back (just removing the "Prefer").

draft-ietf-taps-interface.md
mwelzl: Changed to @britram style optional in the last commit.

draft-ietf-taps-interface.md
mwelzl: Done (and applied to the (now two separate) checksum-related Selection properties.

draft-ietf-taps-interface.md
mwelzl: Done

draft-ietf-taps-interface.md
philsbln: HTTP/3 still does, but the mapping has become less ugly. I think we should not worry too much here as HTTP ugliness can be easily hid with connection pools and pseudo transports. An average application programmer has no change of implementing HTTP/1.1 + 2 + 3 including negotiation and fallback anyway.

draft-ietf-taps-interface.md
abrunstrom: Sorry, but I actually think Ignore is a better name. Perhaps better defined as  

> Ignore - Not used during protocol/path selection

> 

And we should use it :-)

mwelzl: Sigh  :)   but I agree, the problem that I had was the "cancel any system default preference" description, not the name. I'm fine with going back to Ignore. Will do in the next commit.

draft-ietf-taps-interface.md
```diff
@@ -1640,7 +1643,9 @@ are cloned.
 ### Timeout for aborting Connection {#conn-timeout}
 
 This property specifies how long to wait before deciding that a Connection has
-failed after establishment.
+failed when trying to reliably deliver data to the destination. Adjusting this Property
+is only take effect when the underlying stack supports reliability.
+
```
theri: Nit: s/is/will/

draft-ietf-taps-interface.md
```diff
@@ -434,7 +434,7 @@ preference levels:
    |------------|------------------------------------------------------------------------|
    | Require    | Select only protocols/paths providing the property, fail otherwise     |
    | Prefer     | Prefer protocols/paths providing the property, proceed otherwise       |
-   | Ignore     | Cancel any system default preference for this property                 |
+   | Neutral    | No preference                                                          |
```
tfpauly: I agree that `neutral` is better (and less overloaded) than `ignore`, but I'm not 100% sold on it.

I'm also wondering if it is useful to have an explicit way in the API to request to get the `default` system behavior for a property. I know that defaults are specified for individual properties below, but it may be useful for an application being ported between different implementation to say "just do what the system/implementation wants for this one", which is effectively neutral and passes the decision down to the lower layer.

If we replaced `neutral` with `default`, of course, it would be silly to say that the default disposition is `default`... but it's something to think about.

For various APIs we've exposed in the past, we've later had to add an explicit way to set the default value in this way.

mwelzl: But now we have two voices in favor of "Ignore" again - the latest agreement was that it should be "Ignore", but without the pesky text "Cancel any system default preference for this property".  Combining this idea with what you say here, I suggest to call this:

> Ignore   -   No preference; use system defaults for this property

(just the opposite of "Cancel system defaults", actually  :)  )

draft-ietf-taps-interface.md
mwelzl: Done, I changed it back to "Ignore" but kept the newer description.

draft-ietf-taps-interface.md
mwelzl: done

britram:
Looks good to me now. Fixed the merge conflict with #284.

abrunstrom:
With the updated definitions I think the defaults for "Notification of excessive retransmissions" and "Notification of ICMP soft error message arrival" should probably be Ignore?
If you are not interested in these properties they should not affect the selection.

mwelzl:
@abrunstrom I completely agree; done. @britram thanks a lot for fixing the conflict! I was worried about this :)  Now waiting for an ok from @philsbn and @abrunstrom before this lands.

PR #284, opened 2019-02-07
add optional parameter notation, fix #283 

### LINKED ISSUE #283: Common notation for optional parameters
Even though all the languages we want to support don't have convenient syntax for optional parameters to functions/methods, it would be nice to note that we intend for some parameters to not be mandatory on every call.

britram:


draft-ietf-taps-interface.md
```diff
@@ -189,6 +190,12 @@ immediately, according to the error handling idioms of the implementation
 platform, for errors which can be immediately detected, such as inconsistency
 in Transport Properties.
 
+Not all languages have convenient syntax for optional parameters. In the
+interests of ease of use for the most common case(s), it is recommended that
+implementations of this API in such languages use two or more concrete entry
+points per Action, and that the "most common" entry point not include
+any of the optional parameters.
```
philsbln: I am a little bit unsure whether this is the best resolution – in some languages, e.g., C, it would be much more common to use a constant for "don't care" – I would mention both as examples.

mwelzl: +1, I don't like this much for sounding too language-specific. It sure sounds like method overloading as in Java...   I think developers will have their own ways to handle these kinds of things, it's enough to tell them that a parameter is optional.

britram: okay, can cut this and let implementors figure it out.

Issue #247, opened 2018-11-08
API document should begin with a simple usage example


tfpauly:
Similar to the way we added simple diagrams to the architecture document to make the "story" of the document more compelling, we should begin the interface document with a simple workflow for an example connection. This way, we can show concretely up front how this is simple and useful, while being quite flexible and powerful. This can also inform how we then continue to explain depth to the details.

theri:
I'm going to prepare a PR this week.

philsbln:
closed by #293

PR #293, opened 2019-03-07
API usage examples

theri:
This PR provides usage examples of our abstract interface, to address #247.

The examples show how an application might use Listen(), Initiate(), and Rendezvous(), send a Message, and receive a Message.
There is some redundancy between these examples, and I introduces some subtle differences to make them plausible, e.g., the "server" provides its identity, and the "client" verifies the other endpoint's identity using a trust callback.

Overall the section became rather long. Should we instead just stick to one example, e.g., Initiate()? (Or Rendezvous() because it covers the most aspects? Or Initiate() but with all the aspects from Rendezvous()?)

Otherwise I'm particularly interested in feedback on the Rendezvous() example -- I think it's correct wrt the draft, but is there a "more plausible" port number, and should anything else be set there that is not in the Initiate() and Listen() examples?

### DIFF
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -312,9 +312,8 @@ Preconnection.Listen()
 
 Preconnection -> ConnectionReceived<Connection>
 
-// Only receive complete messages:
-// minIncompleteLength and maxLength are infinite by default
-Connection.Receive(minIncompleteLength, maxLength)
+// Only receive complete messages
+Connection.Receive()
 
 Connection -> Received(messageDataRequest, messageContext)
 
@@ -359,9 +358,8 @@ Connection -> Ready<>
 
 Connection.Send(messageDataRequest)
 
-// Only receive complete messages:
-// minIncompleteLength and maxLength are infinite by default
-Connection.Receive(minIncompleteLength, maxLength)
+// Only receive complete messages
+Connection.Receive()
 
 Connection -> Received(messageDataResponse, messageContext)
 
@@ -406,9 +404,8 @@ Preconnection -> RendezvousDone<Connection>
 
 Connection.Send(messageDataRequest)
 
-// Only receive complete messages:
-// minIncompleteLength and maxLength are infinite by default
-Connection.Receive(minIncompleteLength, maxLength)
+// Only receive complete messages
+Connection.Receive()
 
 Connection -> Received(messageDataResponse, messageContext)
 
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -261,14 +261,15 @@ described in {{I-D.ietf-taps-arch}}.
 ## Usage Examples
 
 The following usage examples illustrate how an application might use a
-Transport Services Interface to
+Transport Services Interface to:
 
-- Act as a server: Listen for incoming connections, receive a request, and send
-  a response, see {{server-example}}.
-- Act as a client: Connect to a remote endpoint, send a request, and receive a
-  response, see {{client-example}}.
-- Act as a peer: Connect to a remote endpoint using Rendezvous, send a Message,
-  and receive a Message, see {{peer-example}}.
+- Act as a server, by listening for incoming connections, receiving requests,
+  and sending responses, see {{server-example}}.
+- Act as a client, by connecting to a remote endpoint using Initiate, sending
+  requests, and receiving responses, see {{client-example}}.
+- Act as a peer, by connecting to a remote endpoint using Rendezvous while
+  simultaneously waiting for incoming Connections, sending Messages, and
+  receiving Messages, see {{peer-example}}.
 
 The examples in this section presume that a transport protocol is available
 between the endpoints which provides Reliable Data Transfer, Preservation of
@@ -290,13 +291,12 @@ using the Transport Services Interface, receive a request, and send a response.
 
 ~~~
 LocalSpecifier := NewLocalEndpoint()
-LocalSpecifier.WithInterface("en0")
+LocalSpecifier.WithInterface("any")
 LocalSpecifier.WithService("https")
 
 TransportProperties := NewTransportProperties()
-TransportProperties.Require(reliability)
-TransportProperties.Require(preserve-order)
 TransportProperties.Require(preserve-msg-boundaries)
+// Reliable Data Transfer and Preserve Order are Required by default
 
 SecurityParameters := NewSecurityParameters()
 SecurityParameters.AddIdentity(identity)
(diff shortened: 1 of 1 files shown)


draft-ietf-taps-interface.md
```diff
@@ -320,6 +483,7 @@ following recommendations:
   because transport protocols implementing a given Property are not available on
   the platform.
 
+
```
tfpauly: Random extra line? =)

draft-ietf-taps-interface.md
```diff
@@ -258,6 +258,169 @@ Errors and other notifications also happen asynchronously on the Connection.
 through Actions and Events in each phase of a Connection, following the phases
 described in {{I-D.ietf-taps-arch}}.
 
+## Usage Examples
+
+The following usage examples illustrate how an application might use a
+Transport Services Interface to
```
tfpauly: Editorial nit, feels like this should be structured like:

```
how an application might use a Transport Services Interface to:

- Act as a server, by listening for incoming connections, receiving requests, and sending responses. See {{server-example}}.
```
etc. The main point being we should have a `:` on the dangling sentence, which is filled out by the bullets.

draft-ietf-taps-interface.md
```diff
@@ -258,6 +258,169 @@ Errors and other notifications also happen asynchronously on the Connection.
...
+
+~~~
+LocalSpecifier := NewLocalEndpoint()
+LocalSpecifier.WithInterface("en0")
+LocalSpecifier.WithService("https")
+
+TransportProperties := NewTransportProperties()
+TransportProperties.Require(reliability)
+TransportProperties.Require(preserve-order)
+TransportProperties.Require(preserve-msg-boundaries)
```
tfpauly: I'd actually argue that we should not show setting these values. If we do believe that these are the defaults:

```
The recommended default is to Require Reliable Data Transfer.
```

Then why would the application need to set these explicitly? That just adds to boiler-plate code bulk.

draft-ietf-taps-interface.md
```diff
@@ -258,6 +258,169 @@ Errors and other notifications also happen asynchronously on the Connection.
...
+
+
+### Server Example
+
+This is an example of how an application might listen for incoming Connections
+using the Transport Services Interface, receive a request, and send a response.
+
+~~~
+LocalSpecifier := NewLocalEndpoint()
+LocalSpecifier.WithInterface("en0")
```
tfpauly: I don't like ever recommending using an interface by name; people often make incorrect assumptions about which interface is named which.

draft-ietf-taps-interface.md
```diff
@@ -258,6 +258,169 @@ Errors and other notifications also happen asynchronously on the Connection.
...
+                                  TransportProperties,
+                                  SecurityParameters)
+
+Preconnection.Listen()
+
+Preconnection -> ConnectionReceived<Connection>
+
+// Only receive complete messages
+minIncompleteLength = Infinite
+maxLength = Infinite
```
tfpauly: Again, aren't these the defaults?

draft-ietf-taps-interface.md
theri: Oops, yea, that happened when I moved this part around on my branch...

draft-ietf-taps-interface.md
theri: Agree, will change.

draft-ietf-taps-interface.md
theri: True for reliability and preserve order -- I wanted to show them explicitly, but I can also do that in a comment, and that's probably better.
For preserve-msg-boundaries, the default is Prefer, so I do think I need to set it here, otherwise I think the example will be somewhat ambiguous.

draft-ietf-taps-interface.md
theri: Agreed, hardcoding these in an application might be a bad idea.
I took this example from the Transport Properties Section, so if we decide it's bad, maybe we should change it there as well.

What about: LocalSpecifier.WithInterface("any")? Or is there a better way to specify an interface to listen on than with interface names?

draft-ietf-taps-interface.md
theri: Yes, they are.
I wanted to show them explicitly, so readers are not confused when they read the parameters of the Receive call. But again, I can do that in a comment and that's probably better.

theri:
Thanks for the review, I addressed the comments.
Maybe the default for Preserve Message Boundaries should be Require as well, if we receive only complete Messages by default...

draft-ietf-taps-interface.md
tfpauly: Hm, okay. I think keeping one example of a changed property is good, so this one is fine. May want to change it in the future, but we can see.

draft-ietf-taps-interface.md
```diff
@@ -258,6 +258,163 @@ Errors and other notifications also happen asynchronously on the Connection.
...
+                                  TransportProperties,
+                                  SecurityParameters)
+
+Preconnection.Listen()
+
+Preconnection -> ConnectionReceived<Connection>
+
+// Only receive complete messages:
+// minIncompleteLength and maxLength are infinite by default
+Connection.Receive(minIncompleteLength, maxLength)
```
tfpauly: Would the defaults not make this collapse down to Connection.Receive()?

tfpauly:
I'm good with these changes. Regarding how we show the pseudo-code, I'm not totally sure about the notation (how rigorous are we on that?), so someone else should validate it makes sense.

draft-ietf-taps-interface.md
theri: You're right, I overlooked the "?".

draft-ietf-taps-interface.md
britram: I _think_ we're leaning toward underscore-, as opposed to dash-separation for property names? in any case I don't think this has been decided yet so can leave it for now...

draft-ietf-taps-interface.md
```diff
@@ -258,6 +258,160 @@ Errors and other notifications also happen asynchronously on the Connection.
...
+
+// Specifying a remote endpoint is optional when using Listen()
+Preconnection := NewPreconnection(LocalSpecifier,
+                                  None,
+                                  TransportProperties,
+                                  SecurityParameters)
+
+Preconnection.Listen()
+
+Preconnection -> ConnectionReceived<Connection>
```
britram: this means that "the following lines will happen after this event triggers"? 

We should probably add a paragraph or two to the notation section for notations we want to use in examples. As above, see no reason to block merging on this though.,

PR #294, opened 2019-03-08
Caching Boundaries using Connection Groups

tfpauly:
Alternate proposal to address #45 (old PR was "Add CacheContext to architecture and API", #244)

- Define explicit Connection Groups as a way to define caching boundaries
- Update TLS 1.3 references to RFC 8446

### DIFF
--- a/draft-ietf-taps-arch.md
+++ b/draft-ietf-taps-arch.md
@@ -447,15 +447,17 @@ If two different Protocol Stacks can be safely swapped, or raced in parallel (se
 
 ### Separating Connection Groups {#groups}
 
-By default, all stored properties of the Implementation are shared within a process, such as cached protocol state, cached path state, and heuristics. This provides efficiency and convenience for the application, since the Transport System Implementation will be able to automatically optimize behavior.
+By default, all stored properties of the Implementation are shared within a process, such as cached protocol state, cached path state, and heuristics. This provides efficiency and convenience for the application, since the Transport System Implementation can automatically optimize behavior.
 
 There are several reasons, however, that an application might want to isolate some Connections within a single process. These reasons include:
 
-- Privacy concerns about re-using cached protocol state that can lead to linkability. Sensitive state may include TLS session resumption state {{RFC8446}} and HTTP cookies {{RFC6265}}.
+- Privacy concerns about re-using cached protocol state that can lead to linkability. Sensitive state may include TLS session state {{RFC8446}} and HTTP cookies {{RFC6265}}.
 - Privacy concerns about allowing Connections to multiplex together, which can tell a Remote Endpoint that all of the Connections are coming from the same application (for example, when Connections are multiplexed HTTP/2 or QUIC streams).
 - Performance concerns about Connections introducing head-of-line blocking due to multiplexing or needing to share state on a single thread.
 
-The Transport Services API SHOULD allow applications to explicitly define Connection Groups to force separation of Cached State and Protocol Stacks. The interface to specify these Groups can optionally expose fine-grained tuning for which properties and cached state is allowed to be shared with other Connections. For example, an application might want to allow sharing TCP Fast Open cookies across groups, but not TLS resumption state. 
+The Transport Services API SHOULD allow applications to explicitly define Connection Groups to force separation of Cached State and Protocol Stacks. For example, a web browser application might use separate Connection Groups for different tabs in the browser to decrease linkability.
+
+The interface to specify these Groups can optionally expose fine-grained tuning for which properties and cached state is allowed to be shared with other Connections. For example, an application might want to allow sharing TCP Fast Open cookies across groups, but not TLS session state.
 
 # IANA Considerations
 
--- a/draft-ietf-taps-arch.md
+++ b/draft-ietf-taps-arch.md
@@ -403,7 +403,7 @@ This list of events that can be delivered to an application is not exhaustive, b
 
 The Transport System Implementation Concepts define the set of objects used internally to a system or library to implement the functionality needed to provide a transport service across a network, as required by the abstract interface.
 
-* Connection Group: A set of Connections that share properties. For multiplexing transport protocols, the Connection Group defines the set of Connections that can be multiplexed together. Groups can be defined implicitly by the Implementation, based on which properties are compatible. However, applications SHOULD be able to explicitly define Connection Groups, as discussed in {{groups}}.
+* Connection Group: A set of Connections that share properties and caches. For multiplexing transport protocols, only Connections within the same Connection Group are allowed be multiplexed together. Applications can use their explicitly defined Connection Groups to control caching boundaries, as discussed in {{groups}}.
 
 * Path: Represents an available set of properties that a local system can use to communicate with a remote system, such as routes, addresses, and physical and virtual network interfaces.
 
@@ -455,9 +455,9 @@ There are several reasons, however, that an application might want to isolate so
 - Privacy concerns about allowing Connections to multiplex together, which can tell a Remote Endpoint that all of the Connections are coming from the same application (for example, when Connections are multiplexed HTTP/2 or QUIC streams).
 - Performance concerns about Connections introducing head-of-line blocking due to multiplexing or needing to share state on a single thread.
 
-The Transport Services API SHOULD allow applications to explicitly define Connection Groups to force separation of Cached State and Protocol Stacks. For example, a web browser application might use separate Connection Groups for different tabs in the browser to decrease linkability.
+The Transport Services API SHOULD allow applications to explicitly define Connection Groups that force separation of Cached State and Protocol Stacks. For example, a web browser application might use Connection Groups with separate caches for different tabs in the browser to decrease linkability.
 
-The interface to specify these Groups can optionally expose fine-grained tuning for which properties and cached state is allowed to be shared with other Connections. For example, an application might want to allow sharing TCP Fast Open cookies across groups, but not TLS session state.
+The interface to specify these Groups MAY expose fine-grained tuning for which properties and cached state is allowed to be shared with other Connections. For example, an application might want to allow sharing TCP Fast Open cookies across groups, but not TLS session state.
 
 # IANA Considerations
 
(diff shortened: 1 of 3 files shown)


draft-ietf-taps-arch.md
```diff
@@ -442,7 +443,19 @@ If two different Protocol Stacks can be safely swapped, or raced in parallel (se
...
+
+### Separating Connection Groups {#groups}
+
+By default, all stored properties of the Implementation are shared within a process, such as cached protocol state, cached path state, and heuristics. This provides efficiency and convenience for the application, since the Transport System Implementation will be able to automatically optimize behavior.
```
chris-wood: nit: s/will be able to/can

draft-ietf-taps-arch.md
```diff
@@ -442,7 +443,19 @@ If two different Protocol Stacks can be safely swapped, or raced in parallel (se
...
+
+There are several reasons, however, that an application might want to isolate some Connections within a single process. These reasons include:
+
+- Privacy concerns about re-using cached protocol state that can lead to linkability. Sensitive state may include TLS session resumption state {{RFC8446}} and HTTP cookies {{RFC6265}}.
```
chris-wood: I'd drop "resumption" here.

draft-ietf-taps-arch.md
```diff
@@ -442,7 +443,19 @@ If two different Protocol Stacks can be safely swapped, or raced in parallel (se
...
+- Privacy concerns about allowing Connections to multiplex together, which can tell a Remote Endpoint that all of the Connections are coming from the same application (for example, when Connections are multiplexed HTTP/2 or QUIC streams).
```
chris-wood: This is a bit of a double edged sword since multiplexing hides information from the on-path observer. I might drop this one, since coalescing to the same origin is (IMO) fine.

draft-ietf-taps-arch.md
```diff
@@ -442,7 +443,19 @@ If two different Protocol Stacks can be safely swapped, or raced in parallel (se
...
+- Performance concerns about Connections introducing head-of-line blocking due to multiplexing or needing to share state on a single thread.
+
+The Transport Services API SHOULD allow applications to explicitly define Connection Groups to force separation of Cached State and Protocol Stacks. The interface to specify these Groups can optionally expose fine-grained tuning for which properties and cached state is allowed to be shared with other Connections. For example, an application might want to allow sharing TCP Fast Open cookies across groups, but not TLS resumption state.
```
chris-wood: Maybe add an example which says that connections for a single browser tab would be grouped?

philsbln:
This is a light-weight approach, but seems a little heavy-handed.

I thought connection groups were a tool to model multi-streaming. With this PR, they become a fuzzy thing between cache context and multi-streaming abstraction. This may be the better choice, but ist definitely an Architecture decision. 

I liked the cache original context more.

tfpauly:
The result of the interim discussion was that we didn't like a separate cache context object as much.

Based on the discussion, we need some abstraction to say "these connections don't share state with the rest of the connections in my process, and can share between one another". The Group is a convenient way to do this. Note that you can just have one connection in its own group even.

In some ways, it solves a lot of problems—it isn't sensical to say that connections in a Group can not share cache state. It I multiplex, I'm by default sharing TLS state. So if we have a separate cache context, we'll need some prohibition that you can't have multiple connections in a group have different cache contexts. Using the group as the way to define the boundary solves this.

philsbln:
I agree that it solves a lot of Problems and is an easy-to-use solution. The more I think about it, the more I like it. If we agree that we want to go down this road, definitely a +1 from my side modulo  @chris-wood's comments.

Still, if we land this one, we need to add some means of manually managing connection groups to the API in the next round.

draft-ietf-taps-arch.md
britram:
```suggestion
By default, all stored properties of the Implementation are shared within a process, such as cached protocol state, cached path state, and heuristics. This provides efficiency and convenience for the application, since the Transport System Implementation can automatically optimize behavior.
```

...implementing @chris-wood's suggestion as a Suggestion

draft-ietf-taps-arch.md
britram: does the transport system implementation know enough when autocoalescing to ensure that it's the same origin, though? Thinking about cdn/cloud load balancers here...

tfpauly: Yeah, I'd like to leave this. It's given as a concern an application may have; and when one is being paranoid (like for a private browsing case) it's nice to have a guarantee in usage like this.

draft-ietf-taps-arch.md
tfpauly: Added example!

draft-ietf-taps-arch.md
tfpauly: Dropped

draft-ietf-taps-arch.md
tfpauly: Fixed

tfpauly:
@chris-wood please re-review!

draft-ietf-taps-arch.md
```diff
@@ -402,7 +403,7 @@ This list of events that can be delivered to an application is not exhaustive, b
 
 The Transport System Implementation Concepts define the set of objects used internally to a system or library to implement the functionality needed to provide a transport service across a network, as required by the abstract interface.
 
-* Connection Group: A set of Connections that share properties. For multiplexing transport protocols, the Connection Group defines the set of Connections that can be multiplexed together.
+* Connection Group: A set of Connections that share properties. For multiplexing transport protocols, the Connection Group defines the set of Connections that can be multiplexed together. Groups can be defined implicitly by the Implementation, based on which properties are compatible. However, applications SHOULD be able to explicitly define Connection Groups, as discussed in {{groups}}.
```
abrunstrom: Defining the Connection Groups implicitly is a bit different from what we have in the API draft at the moment where I would read it as Connection Groups are formed explicitly by calling Clone, So the API text should probbaly be aligned a bit if we change it here. 

Related to this, how do I know as an application what connections I affect when changing a protocol property if connections are grouped implicitly? We have in the API draft: 

> Connections in a Connection Group share all Protocol Properties that are not applicable to a Message. Changing one of these Protocol Properties on one Connection in the group changes it for all others. For example, changing "Timeout for aborting Connection" (see {{conn-timeout}}) on one Connection in a group will automatically change this Protocol Property for all Connections in the group in the same way. 

I think we would need some comment about the effect of the implicit grouping in the text above. With implicit grouping changing a protocol property may affect multiple connectione even if they were not explicitly grouped with Clone. This may not be a problem, but the semantics should be clear.

mwelzl: I also wanted such implicit grouping some time ago, but @abrunstrom convinced me that this doesn't add value (and indeed creates problems like she describes here). That's because "clone" only expresses a wish for Connections to be grouped in some form, but there is no promise that much is really happening under the hood. I.e., after calling "clone", an application doesn't really know if its Connections WILL be multiplexed, and if priorities between them WILL play out correctly. However, there's a consistent behavior now regarding the things that ARE promised - e.g. the Connection Properties that she describes here: if one is changed, it affects all the Connections within a Group.

As a conclusion, rather than changing the text she quotes, I think this implicit grouping should be removed. The default behavior for an application should just be to use Clone whenever possible ... whenever there's no good reason NOT to use it.

draft-ietf-taps-arch.md
```diff
@@ -442,7 +443,21 @@ If two different Protocol Stacks can be safely swapped, or raced in parallel (se
...
+
+There are several reasons, however, that an application might want to isolate some Connections within a single process. These reasons include:
+
+- Privacy concerns about re-using cached protocol state that can lead to linkability. Sensitive state may include TLS session state {{RFC8446}} and HTTP cookies {{RFC6265}}.
+- Privacy concerns about allowing Connections to multiplex together, which can tell a Remote Endpoint that all of the Connections are coming from the same application (for example, when Connections are multiplexed HTTP/2 or QUIC streams).
+- Performance concerns about Connections introducing head-of-line blocking due to multiplexing or needing to share state on a single thread.
+
+The Transport Services API SHOULD allow applications to explicitly define Connection Groups to force separation of Cached State and Protocol Stacks. For example, a web browser application might use separate Connection Groups for different tabs in the browser to decrease linkability.
```
mwelzl: This would automatically be solved if you agree with my comment regarding clone above: if this is never implicitly happening but always only a result of clone, you don't even need this SHOULD here.

draft-ietf-taps-arch.md
```diff
@@ -442,7 +443,21 @@ If two different Protocol Stacks can be safely swapped, or raced in parallel (se
...
+
+The interface to specify these Groups can optionally expose fine-grained tuning for which properties and cached state is allowed to be shared with other Connections. For example, an application might want to allow sharing TCP Fast Open cookies across groups, but not TLS session state.
```
mwelzl: I suppose anyone can optionally add whatever bells and whistles they want to their transport system... else, I wonder, is this important?  If it is, we could e.g. add a statement to the grouping text of the interface draft to say that, for every Connection Property, a "sharing allowed" boolean could also be offered by the API... or something like that.

draft-ietf-taps-arch.md
tfpauly: Clarified what the SHOULD is here. The SHOULD is now that the interface should allow apps to mark their Groups as having an isolated cache. So, it says you need to be able to specify a knob to have an isolated group.

draft-ietf-taps-arch.md
tfpauly: Changed the language slightly to just say the interface MAY expose fine-grained tuning. I want to keep this in here to mitigate concerns that readers of the document may have about the architecture lumping layers together.

tfpauly:
@mwelzl I updated the text to avoid the implication about cloning. I didn't intend to change the behavior for Clone() being required to get multiplexing. I was more thinking that the use of the cache could be managed implicitly, while still requiring calling Clone() to actually multiplex. But, for now, I removed the text about implicit groups.

Please check that this resolves your concerns!

draft-ietf-taps-arch.md
chris-wood: 👍

draft-ietf-taps-arch.md
chris-wood: :+1

tfpauly:
Thanks for the reviews and input!

Issue #266, opened 2018-12-05
Do we want to add Connection Pools to the Architecture / API


philsbln:
During the discussion of possible QUIC mappings for TAPS in Bangkok, we have seen that there might be two ways to map QUIC Streams / Connections to TAPS Messages / Connections / Connection groups – Mapping QUIC Streams to TAPS Message is more convenient for HTTP style applications, mapping QUIC streams to TAPS Connections looks more natural. 

One way to solve this tussle is to introduce a new kind of object – a Connection Pool:
- An application can send a message using a connection pool and get events for a reply without touching a connection object
- A connection pool can represent things like a multi-streaming QUIC connection or a bunch of TCP connections between equivalent endpoints.
- To make this logic happen, a Connection Pool needs to be stuffed with framers to make this work
- A Send on a connection pool may take Selection Properties to aid selection between multiple connections within a pool.

Do we want to add Connection Pools to the Architecture / API ?

csperkins:
Mapping a TAPS message to a QUIC stream seems odd to me. QUIC/HTTP streams can carry multiple requests, surely? I'd like to understand the mapping much better before introducing connection pools.

philsbln:
The QUIC/HTTP mapping opens a new bi-directional stream for each request.
In reply, it can send the HTTP reply and multiple push-promises related to the request.
The pushed objects then go on a separate uni-directional server-initiated stream each.

philsbln:
I view framers more as some kind of input filters that can do anything from looking at message boundaries to implementing protocol layers – but I guess this does not matter for this case.

In both cases above, the framer parses HTTP/3 – the only question is wether on a HTTP/3 frame or an HTTP/3 Request level. The latter has the advantage that Requests are neatly aligned with QUIC streams (except for server push that needs additional application logic) and allow to turn HTTP/3 Requests/Responses into TAPS messages.

csperkins:
It seems we perhaps need to define the framer more clearly

mwelzl:
I'm afraid that the framer is turning into a magic box that can potentially do whatever we currently can't agree on. Like @csperkins, I didn't expect a framer to do connection handling. However, it seems we do need a certain amount of functionality beyond "just framing", like changing them in-flight, as per #249.

[...]

PR #295, opened 2019-03-09
Add Connection Pool Object

philsbln:
This PR adds a first proposal for Connection Pools to TAPS as discussed in Issue #266.

**This is an alternative to PR #298.**

Pooled Connections are only useful for request/response protocols like HTTP and serve the following purposes:
 - Provide a clean abstraction for HTTP/3 that is backward compatible to HTTP/0.9
 - Remove the motivation to map QUIC streams to Messages (by removing the drawback of complication HTTP/3 mapping when they are mapped to TAPS connections)
 - Allow Per-Request Path- and Endpoint Selection

See [my post on the mailing list](https://mailarchive.ietf.org/arch/msg/taps/u4imn1IKFBeSxwJRJK_HNwEoT98) for a more detailed description of this topic.

### DIFF
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -947,6 +947,61 @@ the resolution process. This list can be passed to a peer via a signalling
 protocol, such as SIP {{?RFC3261}} or WebRTC {{?RFC7478}}, to configure the
 remote.
 
+## Connection Pools {#pools}
+
+When using Request/Response style protocols, such as HTTP, applications
+can let the Transport System manage their connections by using Connection Pools.
+In this case, the Transport System can automatically distribute requests across multiple connections towards equivalent endpoints or be multiplexed on multi-streaming connections.
+A connection pool is created by the ConnectionPool Action:
+
+~~~
+ConnectionPool := Preconnection.ConnectionPool()
+~~~
+
+The caller should specify all properties necessary for candidate selection that are shared by all requests send or received using this connection pool.
+By specifying selection properties on a send call later, the application can influence which Connection within the Connection Pool should be used. 
+
+The Connection Pool can be used for sending and receiving requests and responses like a connection, see {{sending}} and {{receiving}}.
+
+~~~
+ConnectionPool.Stop()
+~~~
+
+After Stop() is called, the Connection Pool can be disposed of, all listening will be stopped, and all backing connections will be terminated.
+
+### Initiating Pools
+
+When using a connection pool to send requests, the caller must have populated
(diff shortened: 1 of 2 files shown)


draft-ietf-taps-interface.md
```diff
@@ -997,17 +1052,22 @@ and takes optional per-Message properties (see {{send-basic}}). All Send actions
 are asynchronous, and deliver events (see {{send-events}}). Sending partial
 Messages for streaming large data is also supported (see {{send-partial}}).
 
-Messages are sent on a Connection using the Send action:
+Messages are sent on a Connection or a Connection Pool using the Send action:
 
 ~~~
-Connection.Send(messageData, messageContext?, endOfMessage?)
+msgRef := Connection.Send(messageData, messageContext?, endOfMessage?)
+msgRef := ConnectionPool.Send(reqRef?, messageData, messageContext?, endOfMessage?)
```
britram: Exposing the distinction between a Connection and a ConnectionPool in this way seems to add complexity to the API for no benefit. Why does the application need to know a connection pool might be in use? Isn't one of the points of TAPS to allow multiplexing and pooling when possible but to provide nominal functionality when not?

The request reference is indeed useful here (if we accept that TAPS should be useful to build interfaces atop request/response semantics, which I do think is the case) to tell the TAPS system that one message should use the same low-level transport (Transport Stack Instance in Post terminology, IIRC) as another, but ISTM this could be added to every Received event and Send call as an optional parameter.

tfpauly:
Thanks for writing this up!

I agree very much with @britram that the notion of a separate object for a ConnectionPool does not seem to add direct value, but makes the interface more complex.

One of the key notions of the architecture is that the "Connection" is not necessarily analogous to a TCP connection, and I think that really makes the Connection abstraction able to do the same thing you're trying to add with a "Pool". For example, in HTTP/2, a Connection will be just one stream, that's multiplexed. And since each stream is effectively a single HTTP request/response, that means that we're really just treating a Connection as an HTTP request/response pair. So, over HTTP/1.1, or the like, a Connection for a request/response pair ends up "pooling" onto existing TLS/TCP connections. This can and should be automatic, perhaps with the ability to define boundaries of when not to automatically group.

philsbln:
I just updated the Connection Pool objects to be more Connection like and added the ```reqRef``` as parameter to the normal send and receive. So a connection pool really works like a connection and only differs in creation by calling ```RequestorPool()``` instead of ```Initiate()``` and ```ResponderPool()``` instead of ```Listen()```.

philsbln:
Closing this based on the result of the IETF-104 discussion

PR #320, opened 2019-04-29
Clarify that a TAPS connection has no 1:1 relation to an underlying transport connection

philsbln:
This updates the definition of a TAPS connection to allow a TAPS implementation to do connection pooling and multi-streaming. It removes the limitation from a TAPS connection to relate to exactly one underlying transport connection and hints that underlying transport connections my be created automatically.

This is one result of the discussion around PR #298.


draft-ietf-taps-arch.md
```diff
@@ -182,6 +182,7 @@ Providing a message-based abstraction provides many benefits, such as:
 * the ability to associate deadlines with messages, for applications that care about timing;
 *  the ability to provide control of reliability, choosing which messages to retransmit in the event of packet loss, and how best to make use of the data that arrived;
 * the ability to manage dependencies between messages, when the transport system could decide to not deliver a message, either following packet loss or because it has missed a deadline. In particular, this can avoid (re-)sending data that relies on a previous transmission that was never received.
+* the ability to automatically assign messages and connections to underlaying transport connections to exploit multi-streaming and pooled connections.
```
tfpauly: I don't love using the word `exploit` here; it's a bit of a negative connotation. How about using `utilize`?

draft-ietf-taps-arch.md
```diff
@@ -338,7 +339,9 @@ The diagram below provides a high-level view of the actions taken during the lif
...
   * Transport Properties: Transport Properties can be specified as part of a Preconnection to allow the application to configure the Transport System and express their requirements, prohibitions, and preferences. There are three kinds of Transport Properties: Selection Properties ({{preestablishment}}), Connection Properties ({{preestablishment}}), and Message Properties ({{datatransfer}}). Message Properties can also be specified during data transfer to affect specific Messages.
 
-* Connection: A Connection object represents an active transport protocol instance that can send and/or receive Messages between local and remote systems. It holds state pertaining to the underlying transport protocol instance and any ongoing data transfer. This represents, for example, an active connection in a connection-oriented protocol such as TCP, or a fully-specified 5-tuple for a connectionless protocol such as UDP.
+* Connection: A Connection object represents one or more active transport protocol instance that can send and/or receive Messages between local and remote systems. It holds state pertaining to the underlying transport protocol instance(s) and any ongoing data transfer(s).
```
tfpauly: If you switch to `one or more`, you also need to make it `transport protocol instances` in the first sentence. I'd also just remove the `(s)` and use `s` instead.

draft-ietf-taps-arch.md
```diff
@@ -338,7 +339,9 @@ The diagram below provides a high-level view of the actions taken during the lif
...
+  This represents, for example, an active connection in a connection-oriented protocol such as TCP, or a fully-specified 5-tuple for a connectionless protocol such as UDP.
+  However, a connection can also represent a pool of transport protocol instance, for example, a bunch of TCP and QUIC connections to equivalent endpoints provided by a CDN, or a stream of a multi-streaming transport protocol instance.
```
tfpauly: Let's make this all one paragraph again, and I'd drop the `However,`. Just say `A connection can also...`

Rather than `a bunch of`, please use `multiple` or `a set of`.

I'd also drop the reference to a CDN.

draft-ietf-taps-interface.md
```diff
@@ -1048,7 +1048,10 @@ candidate transport-layer connections to be created to the specified remote
 endpoint. The caller may immediately begin sending Messages on the Connection
 (see {{sending}}) after calling Initate(); note that any idempotent data sent
 while the Connection is being established may be sent multiple times or on
-multiple candidates.
+multiple candidates. 
+Also note that transport-layer connections that match the original
+Selection Properties may automatically be created at any later time,
```
tfpauly: This is a little too open/scary— don't say that they may automatically be created at any time, that sounds like there's no way to opt out, etc. Do we need a comment here in the API? This section isn't about transport layer stuff, just about initiate.

gorryfair: I don't think this is a wise place to dig: saying transport connections can be opened later is not something I would like to see in TAPS (Maybe I share the view this COULD happen, but I'd really like to avoid saying anyting in this topic at all in the API.

mwelzl: +1

philsbln: I would like to give implementations this kind of freedom, e.g., to automatically re-initiate a  transport connection after the server closed an old, idle one. But maybe this hint fits better into implementation.

abrunstrom: Yes, I think implementations can have the freedom to do many things without the need to put it in the base spec.

draft-ietf-taps-arch.md
philsbln: addressed

draft-ietf-taps-arch.md
philsbln: ack; fixed

draft-ietf-taps-arch.md
philsbln: yep, sounds better. applied

philsbln:
All comments should be addressed – ready for merge

tfpauly:
Thanks for the changes! LGTM.

gorryfair:
That seems OK, can we fix: / It can also represent a pool of transport protocol instance/ to
be /instances/ or whatever the plural is of a transport protocol instance?

Gorry

PR #321, opened 2019-04-29
Merge Message References and Message Contexts and enable their use for Requests/Response protocols

### LINKED ISSUE #310: msgref is never specified
In the interface draft, msgref is part of 4 callbacks (RendezvousError, Sent, Expired and SendError), however it is never mentioned what msgref is. As the Send action does not have any return value it is also unclear how the application is actually supposed to know what msgref references to.

philsbln:
This merges Message References with the Message Context (closes #310) and adds a reply call to the Message Context to enable the transport system to assign a response to the same underlying transport the request was received on.

This is another result of the discussion around PR #298 and implements most o the API part for #266


draft-ietf-taps-interface.md
```diff
@@ -1555,11 +1573,11 @@ messageContext.add(parameter, value)
 
 messageData := "hel".octets()
 endOfMessage := false
-Connection.Send(messageData, messageContext, endOfMessage)
+Connection.Send(messageData, messageContext, reqRef?, endOfMessage)
```
tfpauly: The messageContext parameter is exactly what this proposed message reference does. How about just adding more text to better explain the message context?

mwelzl:
The technical idea behind this PR is fine IMO; though I agree with @tfpauly that the messageContext parameter could be used instead of adding an extra parameter.

draft-ietf-taps-interface.md
philsbln: I agree that the functionality of linking requests to responses may be a better done using the message context than using this parameter. Doing so would require to answer a few additional questions (see below).

philsbln:
I agree with @tfpauly that linking replies with requests should be better done using the message context. However, this requires answering some design questions:
 1. Should one be able to pass the message context from a message we received to the send of the reply? This would be really elegant, but may limit implementations.
 2. Should the ```msgref``` returned by send really be an opaque identifier (as specified in this request) or should it me a message context object?

If we answer both with yes, I am happy to write more text in message contexts, otherwise, I am happy to move setting the request-reference to the message context.

philsbln:
I updated the PR to address @tfpauly's comments.  
- merged the functionality of msgRef (which was not described anyway) into the Message Context.
- merged send and receive Message Context.
- realised the reply functionality using the Message Context.
- clarified what the the Message Context is / does

Sorry for the force pushes, I accidentally fixed too many whitespace errors.

draft-ietf-taps-interface.md
```diff
@@ -1278,6 +1281,19 @@ Size for the Connection, the Send will fail with a SendError event ({{send-error
 example, it is invalid to send a Message over a UDP connection that is larger than
 the available datagram sending size.
 
+## Sending Replies {#send-replies}
+
+When a message is sent in responses to a message received, the application
```
mwelzl: tiny fix: s/responses/response

draft-ietf-taps-interface.md
```diff
@@ -1278,6 +1281,19 @@ Size for the Connection, the Send will fail with a SendError event ({{send-error
...
+may use the Message Context of the received Message to construct a Message Context for the reply.
+
+~~~
+replyMessgeContext := reqestMessageContext.reply()
+~~~
+
+By using the ```replyMessgeContext```, the transport system is informed that
```
mwelzl: just a nit: here, and in the "code" line above, s/replyMessgeContext/replyMessageContext

draft-ietf-taps-interface.md
```diff
@@ -1334,16 +1350,21 @@ set of Message Properties not consistent with the Connection's transport
...
-end of a Message to still be sent.
+Connection. Therefore a message context containing these properties can be passed to the Send Action. For other uses of the message context, see {{msg-ctx}}.
+
+Note that message properties are per-Message, not per-Send if partial Messages
+are sent ({{send-partial}}). All data blocks associated with a single Message
+share properties specified in the Message Contexts. For example, it would not
+make sense to have the beginning of a Message expire, but allow the end of a
+Message to still be sent.
+
+A MessageContext object contains  and metadata for  Messages to be sent or received.
```
mwelzl: nit: remove "and" after "contains"

draft-ietf-taps-interface.md
```diff
@@ -1681,28 +1703,33 @@ when it is temporarily not ready to receive messages.
...
+objects, the received bytes as messageData, and the metadata and properties of
+the received Message as messageContext.
+
+The messageData object provides access to the bytes that were received for this
+Message, along with the length of the byte array.
 
-The messageData object provides access to the bytes that were received for this Message,
-along with the length of the byte array.
+The messageContext is provided to retrieve metadata about the message, see
+{{msg-ctx}}, and to refer to the message later on, e.g., to enable to map
```
mwelzl: nit: I think this would read better as "to enable mapping" (to reduce the number of "to" occurrences in the sentence). But anyway, I'd rather not repeat the lower-level details here and instead say: "The messageContext is provided to retrieve metadata about the message, see {{msg-ctx}}, and to refer to the message later on, e.g., to map responses to prior requests."   That's just a suggestion, it's ok for me either way.

draft-ietf-taps-interface.md
```diff
@@ -1681,28 +1703,33 @@ when it is temporarily not ready to receive messages.
...
-See {{receive-framing}} for handling Message framing in situations where the Protocol
-Stack provides octet-stream transport only.
+See {{receive-framing}} for handling Message framing in situations where the
+Protocol Stack provides octet-stream transport only.
 
 ### ReceivedPartial {#receive-partial}
 
 ~~~
-Connection -> ReceivedPartial<messageData, messageContext, endOfMessage>
+Connection -> ReceivedPartial<messageData, messageContext, messageContext, endOfMessage>
```
mwelzl: huh? two messageContexts, with the same name even?

draft-ietf-taps-interface.md
```diff
@@ -1681,28 +1703,33 @@ when it is temporarily not ready to receive messages.
...
 ~~~
 
 If a complete Message cannot be delivered in one event, one part of the Message
 may be delivered with a ReceivedPartial event. In order to continue to receive more
 of the same Message, the application must invoke Receive again.
 
 Multiple invocations of ReceivedPartial deliver data for the same Message by
-passing the same MessageContext, until the endOfMessage flag is delivered or a
+passing the same MessageContext and messageContext, until the endOfMessage flag is delivered or a
```
mwelzl: why two? and why is one written with a capital M?

draft-ietf-taps-interface.md
```diff
@@ -1740,9 +1767,29 @@ The ReceiveError event passes an optional associated MessageContext. This may
...
 
-## Message Receive Context {#receive-context}
+## Message Contexts {#msg-ctx}
+
+A MessageContext object contains properties, see {{message-props}}, and other
+metadata for Messages. Therefore, a MessageContext object can be passed to the
+Send Action and is retuned by each Send action as well as Send and Receive
+related events.
+
+The application can set, see {{message-props}} and query Message Properties using the Message Context.
```
mwelzl: I think this is grammatically challenged. PERHAPS it's correct if you add a comma after {{message-props}}, dunno; I'm quite sure it's correct if you use parantheses instead, i.e.: "The application can set (see {{message-props}}) and query..."

draft-ietf-taps-interface.md
```diff
@@ -1740,9 +1767,29 @@ The ReceiveError event passes an optional associated MessageContext. This may
...
+~~~
+
+The application can also query information about the local and remote endpoint.
+
+~~~
+RemoteEndpoint := MessageContext.GetRemoteEndpoint()
+LocalEndpoint := MessageContext.GetLocalEndpoint()
+~~~
+
+Message Contexts can also be used to send messages that are flagged as reply to other messages, see {{send-replies}} for details.
```
mwelzl: I think (not 100% sure) that there must be an "a" before "reply" here. Anyway, if you add the "a", I'm quite sure it's correct.

draft-ietf-taps-interface.md
philsbln: rewriting artefact - fixed

draft-ietf-taps-interface.md
philsbln: rewriting artefact - fixed

draft-ietf-taps-interface.md
philsbln: Rewrote the paragraph to make it more readable

philsbln:
I did a merge and update pass to address @mwelzl's comments and also added a way to get the original request's Message Context from the reply's MessageContext

draft-ietf-taps-interface.md
```diff
@@ -1257,11 +1257,14 @@ Messages for streaming large data is also supported (see {{send-partial}}).
 Messages are sent on a Connection using the Send action:
 
 ~~~
-Connection.Send(messageData, messageContext?, endOfMessage?)
+messageContext := Connection.Send(messageData, messageContext?, endOfMessage?)
```
tfpauly: What is this messageContext returned by Send()? If you pass in a messageContext, is it the same one?

draft-ietf-taps-interface.md
```diff
@@ -1290,6 +1293,19 @@ Size for the Connection, the Send will fail with a SendError event ({{send-error
 example, it is invalid to send a Message over a UDP connection that is larger than
 the available datagram sending size.
 
+## Sending Replies {#send-replies}
+
+When a message is sent in response to a message received, the application
+may use the Message Context of the received Message to construct a Message Context for the reply.
+
+~~~
+replyMessgeContext := replyMessageContext.reply()
```
tfpauly: Typo: `replyMessgeContext` -> `replyMessageContext`

tfpauly: Also, it feels odd to have replyMessageContext come from the reply of itself. Should these be two separate variables?

draft-ietf-taps-interface.md
philsbln: Yes – you only need it if you don't provide one. 
This is necessary to get rid of the msgRef

draft-ietf-taps-interface.md
philsbln: Sorry - It seems it was way too hot (40°C+) in my office… 
```replyMessageContext := requestMessageContext.reply()```
fixed

**Comment by philsbln** [comment]
> It would be good have some cleanup on the usage of contexts, and if we want to change the functions like send to return a context, explain in more detail what the returned value means.

Agreed, but this is not a problem introduced by the PR, but that existed before for the ```msgRef``` and the MessageContext object.
Making the returned value a MessageContext object and clarifying that the MessageContexts used in ```send``` and ```recv``` are the same kind of object improves this in my opinion, but may in fact not be sufficient.
The question is whether we need to fix this in this PR or (if there are no other things blocking this) should merge the PR and do the cleanup in a separate PR.

**Comment by mwelzl** [comment]
@tfpauly, @britram - should we press the Merge button on this one?

File: draft-ietf-taps-interface.md (hunk 10 above)
Comment (tfpauly) [reply draft-ietf-taps-interface.md]: This is to allow replying, correct? I'd say that if the application needed to get a reply, it should use an explicit messageContext instead.

My concern with returning a value that most often is not needed is that you then are often ignoring a return value, which is a warning often.

File: draft-ietf-taps-interface.md [hunk 12]
```diff
@@ -1811,6 +1832,37 @@ that the other endpoint is done sending on a connection.
...
+~~~
+
+Message Contexts can also be used to send messages that are flagged as a reply to other messages, see {{send-replies}} for details.
+If the message received was send by the remote endpoint as a reply to an earlier message and the transports provides this information, the MessageContext of the original request can be accessed using the Message Context of the reply:
+
+~~~
+RequestMessageContext := MessageContext.GetOriginalRequest()
+~~~
+
+
```
tfpauly: Looks like there's some extra whitespace here

draft-ietf-taps-interface.md
```diff
@@ -1811,6 +1832,37 @@ that the other endpoint is done sending on a connection.
...
+PropertyValue := MessageContext.get(scope?, property)
+~~~
+
+To get or set Message Properties, the optional scope parameter is left empty, for framing meta-data, the framer is passed.
+
+The application can also query information about the local and remote endpoint.
+
+~~~
+RemoteEndpoint := MessageContext.GetRemoteEndpoint()
+LocalEndpoint := MessageContext.GetLocalEndpoint()
```
tfpauly: Are these only queryable on received messages?

tfpauly:
I'd like to hear the response for returning a message from Send... I'm OK to merge, but would want an issue to revisit that point.

draft-ietf-taps-interface.md
philsbln: I'd say they should be queryable on any MessageContext, but for sent messages, this is currently only possible after send. I'll add a note for this.

draft-ietf-taps-interface.md
philsbln: Okay … I see your point.

This design was not made for replying, but to allow to match errors to sends without creating an explicit message context first. I have no strong opinion what is worse - having a return value that is often ignored or needing to create a message context for matching error events to sends - so let's make this an issue and decide in Montreal whether to drop the return value.

For replying, one would use the ```MessageContext``` returned by the ```recv``` event for this. Also for matching replies, using  I would suspect rather using an explicit message context than saving the implicitly generated one.

draft-ietf-taps-interface.md
philsbln: fixed

PR #323, opened 2019-05-06
Add details to the Framer Interface

tfpauly:
This PR is based on the presentation given at IETF 104 and the received feedback regarding Framers. This enhances and clarifies the API text around how framers are set up and used.

### DIFF
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -1806,7 +1806,7 @@ into individual transport datagrams.
 
 ## Defining Message Framers
 
-Applications can define a Message Framer by creating a object to represent a unique
+Applications can define a Message Framer by creating an object to represent a unique
 definition of a framing protocol. Event handlers for inbound and outbound data
 are registered on this object.
 
@@ -1860,6 +1860,7 @@ the FramerInstance. This call can be delayed in case the protocol needs to do so
 simple handshake exchange before handling application Messages.
 
 ~~~
+// instance is a FramerInstance reference
 instance.ready()
 ~~~
 
@@ -1868,6 +1869,15 @@ to fail and provide an error.
 
 ~~~
 instance.failed(Error)
+~~~
+
+Before a Message Framer marks itself as ready, it can also dynamically add a protocol or framer
+above it in the stack. This allows protocols like STARTTLS, that need to add TLS conditionally,
+to modify the Protocol Stack based on a handshake result.
+
+~~~
+otherFramer := NewMessageFramer()
+instance.prependProtocol(otherFramer)
 ~~~
 
 ## Sender-side Message Framing {#send-framing}
@@ -1890,7 +1900,7 @@ instance.send(Data)
 
 To provide an example, a simple protocol that adds a length as a header would receive
 the `NewOutputMessage` event, create a data representation of the length of the Message
-data, and then send a data that is the concatenation of the length header and the original
+data, and then send a block of data that is the concatenation of the length header and the original
 Message data.
 
 ## Receiver-side Message Framing {#receive-framing}
@@ -1902,7 +1912,7 @@ is first notified whenever new data is available to parse.
 FramerInstance -> HandleReceivedData<>
 ~~~
 
-Upon receiving this even, a Message Framer can inspect the inbound data. The
+Upon receiving this event, a Message Framer can inspect the inbound data. The
 data is parsed from a particular cursor representing the unprocessed data. The
 Framer requests a specific amount of data it needs to have available in order to parse.
 If the data is not available, the parse fails.
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -78,10 +78,19 @@ author:
     ins: C. Wood
     name: Chris Wood
     org: Apple Inc.
-    street: 1 Infinite Loop
+    street: One Apple Park Way
     city: Cupertino, California 95014
     country: United States of America
     email: cawood@apple.com
+  -
+    ins: T. Pauly
+    name: Tommy Pauly
+    role: editor
+    org: Apple Inc.
+    street: One Apple Park Way
+    city: Cupertino, California 95014
+    country: United States of America
+    email: tpauly@apple.com
 
 normative:
   I-D.ietf-tsvwg-sctp-ndata:
@@ -249,7 +258,7 @@ peer to peer establishment).
 Once a Connection is established, data can be sent on it in the form of
 Messages. The interface supports the preservation of message boundaries both
 via explicit Protocol Stack support, and via application support through a
-deframing callback which finds message boundaries in a stream. Messages are
+Message Framer definition which finds message boundaries in a stream. Messages are
 received asynchronously through a callback registered by the application.
 Errors and other notifications also happen asynchronously on the Connection.
 
@@ -280,8 +289,8 @@ If none of the available transport protocols provides Preservation of Message
 Boundaries, but there is a transport protocol which provides a reliable ordered
 octet stream, an application may receive this octet stream as partial
 Messages and transform it into application-layer Messages.  Alternatively,
-an application may provide a Deframer, which is a function that transforms an
-octet stream into a sequence of Messages, see {{receive-framing}}.
+an application may provide a Message Framer, which can transform an
+octet stream into a sequence of Messages ({{receive-framing}}).
 
 
 ### Server Example
@@ -540,8 +549,8 @@ incoming Connections.
 The Local Endpoint and the Remote Endpoint MUST both be specified if a
 peer-to-peer Rendezvous is to occur based on the Preconnection.
 
-Framers (see {{send-framing}}) and deframers (see {{receive-framing}}), if
-necessary, should be bound to the Preconnection during pre-establishment.
+Message Framers (see {{framing}}), if required, should be added to the 
+Preconnection during pre-establishment.
 
 ## Specifying Endpoints {#endpointspec}
 
@@ -1606,24 +1615,6 @@ the Connection could not be established will not result in a
 SendError separate from the InitiateError signaling the failure of Connection
 establishment.
 
-## Sender-side Framing {#send-framing}
-
-Sender-side framing allows a caller to provide the interface with a function
-that takes a Message of an appropriate application-layer type and returns an
-array of octets, the on-the-wire representation of the Message to be handed down
-to the Protocol Stack. It consists of a Framer Object with a single Action,
-Frame. Since the Framer depends on the protocol used at the application layer,
-it is bound to the Preconnection during the pre-establishment phase:
-
-~~~
-Preconnection.FrameWith(Framer)
-
-OctetArray := Framer.Frame(messageData)
-~~~
-
-Sender-side framing is a convenience feature of the interface, for parity with
-receiver-side framing (see {{receive-framing}}).
-
 # Receiving Data {#receiving}
 
 Once a Connection is established, it can be used for receiving data. As with
@@ -1650,7 +1641,8 @@ By default, Receive will try to deliver complete Messages in a single event ({{r
 
 The application can set a minIncompleteLength value to indicates the smallest partial
 Message data size in bytes that should be delivered in response to this Receive. By default,
-this value is infinite, which means that only complete Messages should be delivered (see {{receive-partial}} and {{receive-framing}} for more information on how this is accomplished).
+this value is infinite, which means that only complete Messages should be delivered (see {{receive-partial}} 
+and {{receive-framing}} for more information on how this is accomplished).
 If this value is set to some smaller value, the associated receive event will be triggered
 only when at least that many bytes are available, or the Message is complete with fewer
 bytes, or the system needs to free up memory. Applications should always
@@ -1689,7 +1681,7 @@ The messageData object provides access to the bytes that were received for this 
 along with the length of the byte array.
 
 See {{receive-framing}} for handling Message framing in situations where the Protocol
-Stack provides octet-stream transport only.
+Stack only provides an octet-stream transport.
 
 ### ReceivedPartial {#receive-partial}
 
@@ -1715,14 +1707,14 @@ delivered if one of the following conditions is true:
   the size of the Message is larger than the buffers available for a single
   message;
 * the underlying Protocol Stack does not support message boundary
-  preservation, and the deframer (see {{receive-framing}}) cannot determine
+  preservation, and the Message Framer (see {{receive-framing}}) cannot determine
   the end of the message using the buffer space it has available; or
 * the underlying Protocol Stack does not support message boundary
-  preservation, and no deframer was supplied by the application
+  preservation, and no Message Framer was supplied by the application
 
 Note that in the absence of message boundary preservation or
-deframing, all bytes received on the Connection will be represented as one
-large message of indeterminate length.
+a Message Framer, all bytes received on the Connection will be represented as one
+large Message of indeterminate length.
 
 ### ReceiveError
 
@@ -1731,7 +1723,7 @@ Connection -> ReceiveError<messageContext>
 ~~~
 
 A ReceiveError occurs when data is received by the underlying Protocol Stack
-that cannot be fully retrieved or deframed, or when some other indication is
+that cannot be fully retrieved or parsed, or when some other indication is
 received that reception has failed. Such conditions that irrevocably lead to
 the termination of the Connection are signaled using ConnectionError instead
 (see {{termination}}).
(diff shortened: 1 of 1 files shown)


draft-ietf-taps-interface.md
```diff
@@ -1780,38 +1772,170 @@ that the other endpoint is done sending on a connection.
 
 Any calls to Receive once the Final Message has been delivered will result in errors.
 
-## Receiver-side De-framing over Stream Protocols {#receive-framing}
+# Message Framers {#framing}
+
+Message Framers are implementations of simple protocols that can convert
```
mwelzl: A protocol usually includes more than just adding and parsing headers: it also defines a sequence of communication (and that may come with timeouts and such). I think framers can be simpler than that - if a framer just adds and removes a header, I think it's wrong to call it a protocol. You talk about simple application protocols as framers by example later in the text, and I think that's okay - I just don't like the statement here that says that they "are implementations of simple protocols". Also, there's no need for them to be simple... I also wouldn't want to claim that. Can we just rephrase this to something like "Message Framers are converters that translate between application Message data and raw transport protocol data."  ?

draft-ietf-taps-interface.md
```diff
@@ -1780,38 +1772,170 @@ that the other endpoint is done sending on a connection.
...
+Note that while Message Framers add the most value when placed above
+a protocol that otherwise does not preserve message boundaries, they can
+also be used with datagram- or message-based protocols. In these cases,
+they add an additional transformation to further encode or encapsulate,
+and can potentially support packing multiple application-layer Messages
+into individual transport datagrams.
+
+## Defining Message Framers
+
+Applications can define a Message Framer by creating a object to represent a unique
```
mwelzl: First, a nit: this should be "an object".

Second, more importantly: this is now written for object-oriented languages. I think there's no good way to write text about framers without deciding either to talk about objects or to talk about callbacks... so my proposal would be to preface this sentence with a statement like the following:

We describe Message Framers assuming an object-oriented programming language. In the case of programming languages that are not object-oriented, a Message Framer can be thought of as a number of functions and static data that they operate on. Adding a Framer to a MessageContext, as below, would then mean to register a callback to a function of the framer.

draft-ietf-taps-interface.md
```diff
@@ -1780,38 +1772,170 @@ that the other endpoint is done sending on a connection.
...
+protocol. Implementations SHOULD ensure that there is a way to pass the original data
+through without copying to improve performance.
+
+~~~
+instance.send(Data)
+~~~
+
+To provide an example, a simple protocol that adds a length as a header would receive
+the `NewOutputMessage` event, create a data representation of the length of the Message
+data, and then send a data that is the concatenation of the length header and the original
```
mwelzl: "send a data" sounds broken. "send a block of data" or "send data" ?

draft-ietf-taps-interface.md
```diff
@@ -1780,38 +1772,170 @@ that the other endpoint is done sending on a connection.
...
+## Receiver-side Message Framing {#receive-framing}
+
+In order to parse an received flow of data into Messages, the Message Framer
+is first notified whenever new data is available to parse.
+
+~~~
+FramerInstance -> HandleReceivedData<>
+~~~
+
+Upon receiving this even, a Message Framer can inspect the inbound data. The
```
mwelzl: s/even/event

draft-ietf-taps-interface.md
```diff
@@ -1780,38 +1772,170 @@ that the other endpoint is done sending on a connection.
...
+FramerInstance -> Stop()
 ~~~
 
+When Message Framer gets a start message, it sets up its state and then indicates
+to the connection that it is ready to handle sending Messages by calling `ready` on
+the FramerInstance. This call can be delayed in case the protocol needs to do some
+simple handshake exchange before handling application Messages.
+
+~~~
+instance.ready()
```
mwelzl: Here, and below, I wonder about this use of "instance". Above you talk about a a FramerInstance and call Start and Stop on it. Why has this FramerInstance only become "instance" all of a sudden?

draft-ietf-taps-interface.md
philsbln: @mwelzl: This is true for most of our API description and not specific to framers.
I would not add this kind of hit here, but maybe at a more general place.

draft-ietf-taps-interface.md
```diff
@@ -1780,38 +1772,170 @@ that the other endpoint is done sending on a connection.
...
+~~~
+
+When sending Messages, applications can add specific Message
+values to a MessageContext ({{message-props}}) that is intended for a Framer.
+This can be used, for example, to set the type of a Message for a TLV protocol.
+The namespace of values is custom for each unique Message Framer.
+
+~~~
+messageContext := NewMessageContext()
+messageContext.add(framer, key, value)
```
philsbln: This syntax overlaps with the syntax for message properties – while this is very elegant, it might add a lot of confusion for languages without polymorphism.

draft-ietf-taps-interface.md
```diff
@@ -1780,38 +1772,170 @@ that the other endpoint is done sending on a connection.
...
-Preconnection.DeframeWith(Deframer)
+framer := NewMessageFramer()
+framer.setEventHandlers()
+~~~
+
+## Adding Message Framers to Connections
 
-{messageData} := Deframer.Deframe(OctetStream)
+The Message Framer object can then be added to one or more Preconnections
+to run on top of transport protocols. Multiple Framers may be added.
```
philsbln: How do we do the ordering of framers here?
How can I slide in framers later, e.g., for STARTTLS?

draft-ietf-taps-interface.md
```diff
@@ -1780,38 +1772,170 @@ that the other endpoint is done sending on a connection.
...
+Framer to handle a new Message being sent by the application.
+
+~~~
+FramerInstance -> NewSentMessage<messageData, messageContext, endOfMessage>
+~~~
+
+Upon receiving this event, a Message Framer implementation is responsible for
+performing any necessary transformations and sending the resulting data to the next
+protocol. Implementations SHOULD ensure that there is a way to pass the original data
+through without copying to improve performance.
```
philsbln: While this is a valid comment, is this really something that belongs in the API?

draft-ietf-taps-interface.md
```diff
@@ -1780,38 +1772,170 @@ that the other endpoint is done sending on a connection.
...
+data, and then send a data that is the concatenation of the length header and the original
+Message data.
+
+## Receiver-side Message Framing {#receive-framing}
+
+In order to parse an received flow of data into Messages, the Message Framer
+is first notified whenever new data is available to parse.
+
+~~~
+FramerInstance -> HandleReceivedData<>
```
philsbln: This example broken:
- FramerInstance is the receiver of the event, not the sender
- We have no other events with "handle" in the name
- Arguments (Data + Context are missing)

draft-ietf-taps-interface.md
```diff
@@ -1780,38 +1772,170 @@ that the other endpoint is done sending on a connection.
...
+~~~
+
+Upon receiving this even, a Message Framer can inspect the inbound data. The
+data is parsed from a particular cursor representing the unprocessed data. The
+Framer requests a specific amount of data it needs to have available in order to parse.
+If the data is not available, the parse fails.
+
+~~~
+instance.parse(minIncompleteLength, maxLength) -> (Data?, endOfMessage?)
```
philsbln: why is this a call on the framer, and not on the connection?

draft-ietf-taps-interface.md
tfpauly: Yeah, that's a typo. Should be `send a block of data`.

draft-ietf-taps-interface.md
tfpauly: Yup, thanks.

draft-ietf-taps-interface.md
tfpauly: Not sure on our pseudocode parlance, but this was representing the call that an app would make on a local variable that holds their reference to a FramerInstance. Having `FramerInstance.ready()` looks like a "class method" rather than an "instance method".

draft-ietf-taps-interface.md
tfpauly: They are one of the simplest forms of protocol, I'd argue. They certainly can have flags, etc, in their headers that do some communication. For example, WebSocket is a protocol and certainly is something you'd implement as a framer.

draft-ietf-taps-interface.md
tfpauly: Thanks for the typo.

Mainly, I'm using object since we are using object in the rest of the interface description. Even for a non-object oriented language, you can still refer to an allocated structure with defined functions to use as an "object". A Connection is an object, etc.

tfpauly:
@mwelzl I agree that we should have an example somewhere, but I also thought this should be a different PR.

draft-ietf-taps-interface.md
tfpauly: No, FramerInstance is not the receiver of the event, it *is* the sender. It's the handle to the representation of the framer within the Connection that you interact with. It's your state. It says: your framer now has data available.

The data for new inbound data should not be provided in this event—the implementation of the framer needs to parse the data out. The data may not be in a form that is efficiently consumable as a single data object (it's an entire stream).

draft-ietf-taps-interface.md
tfpauly: This is based on the discussion at IETF 104 that indicated we wanted to ensure that the API allows for 0-copy. That is an API semantic decision.

draft-ietf-taps-interface.md
tfpauly: Yes, you should be able to dynamically add. We'd need new text for that.

draft-ietf-taps-interface.md
tfpauly: Sure, in another language, you could have it be `addFramerMessageValue` or something.

draft-ietf-taps-interface.md
tfpauly: Added text for this.

draft-ietf-taps-interface.md
```diff
@@ -1780,37 +1772,179 @@ that the other endpoint is done sending on a connection.
...
+Message Framers define an event that gets called on an instance of a
+Framer to handle a new Message being sent by the application.
+
+~~~
+FramerInstance -> NewSentMessage<messageData, messageContext, endOfMessage>
+~~~
+
+Upon receiving this event, a Message Framer implementation is responsible for
+performing any necessary transformations and sending the resulting data to the next
+protocol. Implementations SHOULD ensure that there is a way to pass the original data
```
MaxF12: What happens if the final protocol has been reached? How does the message that went through all layers of framers get returned to the connection so it can be send out?

draft-ietf-taps-interface.md
```diff
@@ -1780,37 +1772,179 @@ that the other endpoint is done sending on a connection.
...
+through without copying to improve performance.
+
+~~~
+instance.send(Data)
```
MaxF12: What exactly is this call supposed to do? Does it actually send out the message? If it does, why should the framer and not the connection be responsible for sending now?

draft-ietf-taps-interface.md
```diff
@@ -1780,37 +1772,179 @@ that the other endpoint is done sending on a connection.
...
+
+To deliver a Message to the application, the Framer can either directly
+send data that it has created, or deliver a range of data directly from the underlying
+transport and simulatenously advance the receive cursor.
+
+~~~
+instance.advanceReceiveCursor(length)
+instance.deliverAndAdvanceReceiveCursor(messageContext, length, endOfMessage)
+instance.deliver(messageContext, data, endOfMessage)
+~~~
```
MaxF12: This is another point where I am unsure on how the framer actually interfaces with the connection. Who calls instance.advanceReceiveCursor?

draft-ietf-taps-interface.md
```diff
@@ -1780,37 +1772,179 @@ that the other endpoint is done sending on a connection.
...
+## Message Framer Lifetime
+
+When a Connection establishment attempt begins, an event is delivered to
+notify the framer that a new Connection is being created and has a new
+instance of a Message Framer. Similarly, a stop event is delivered when a
+Connection is being torn down.
+
+~~~
+FramerInstance -> Start()
+FramerInstance -> Stop()
```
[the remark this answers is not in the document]
tfpauly: FramerInstance is not created by the application. It is created by the connection, as the handle to the instance of the framer. The MessageFramer object is the definition created by the application.

draft-ietf-taps-interface.md
```diff
@@ -1780,37 +1772,179 @@ that the other endpoint is done sending on a connection.
...
-layer, it is bound to the Preconnection during the pre-establishment phase:
+## Defining Message Framers
 
+Applications can define a Message Framer by creating an object to represent a unique
+definition of a framing protocol. Event handlers for inbound and outbound data
+are registered on this object.
+
+~~~
+framer := NewMessageFramer()
+framer.setEventHandlers()
```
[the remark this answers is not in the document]
tfpauly: You need something to add to the connection. The MessageFramer is the object that the application creates to do this.

draft-ietf-taps-interface.md
```diff
@@ -1780,37 +1772,179 @@ that the other endpoint is done sending on a connection.
...
+instance of a Message Framer. Similarly, a stop event is delivered when a
+Connection is being torn down.
+
+~~~
+FramerInstance -> Start()
+FramerInstance -> Stop()
+~~~
+
+When Message Framer gets a start message, it sets up its state and then indicates
+to the connection that it is ready to handle sending Messages by calling `ready` on
```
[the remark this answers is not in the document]
tfpauly: This is the event that the application's implementation of the framer sends to influence the main ready callbacks on the connection itself. It causes the event on the connection.

draft-ietf-taps-interface.md
tfpauly: The framers are part of the connection protocol stack.. they go down to the transport next.

draft-ietf-taps-interface.md
tfpauly: It sends data down to the next protocol down in the stack. The application sends original data on the connection, it gets sent to the first framer, which sends it to the next, which eventually sends it to the transport, which sends it out the device.

draft-ietf-taps-interface.md
tfpauly: The application's implementation of the framer.

draft-ietf-taps-interface.md
philsbln: Okay - I got it. Also makes sense not to specify who receives this Event, as it could be another Framer or the connection object to forward it to the application.

Can we try to make this event match the Received/ReceivedPartial event somehow? 
This would look much more consistent and allows to directly send it from the Framer.

draft-ietf-taps-interface.md
philsbln: As with the ```HandleReceivedData<>```, making it analogous to ``` Received/ReceivedPartial``` would cause less confusion here

draft-ietf-taps-interface.md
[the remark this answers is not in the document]
philsbln: Why this? This deprives the application of the ability to pass arguments to the Framer's constructor.

draft-ietf-taps-interface.md
philsbln: Firing a Send-Event to the next instance would look much more consistent here…

draft-ietf-taps-interface.md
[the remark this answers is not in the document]
MaxF12: Oh I finally I think I understand where my confusion comes from. I assumed that the application was free to create their own framer class that had to expose certain functions for interaction with the connection, so I was looking for that interface. Sorry, my bad.

The place where the application now implements the behavior of framer is in the eventHandlers it sets, is that correct?

draft-ietf-taps-interface.md
```diff
@@ -78,10 +78,19 @@ author:
...
     org: Apple Inc.
-    street: 1 Infinite Loop
+    street: One Apple Park Way
     city: Cupertino, California 95014
     country: United States of America
     email: cawood@apple.com
+  -
+    ins: T. Pauly
+    name: Tommy Pauly
+    role: editor
```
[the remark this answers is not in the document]
tfpauly: Whoops, that was meant to just be author

draft-ietf-taps-interface.md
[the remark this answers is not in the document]
MaxF12: Thank you Brian, thats exactly what I couldn't quite put into words. 

The previous iteration of framers made it very simple for applications that had no interest in implementing them to grab a framer implementation someone else provided and use it. All the calls and callbacks between the implementation and the TAPS system stayed the same, independent of whether or not a framer was used. I think its important to keep the part of the interface application developers use simple while adding complexity in the part someone that implements a framer would use, because at least to me these are two different groups of people.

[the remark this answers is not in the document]
tfpauly: Agreed. The code here is just how you implement a custom framer. The calls to actually add it to a connection (the common action) is below.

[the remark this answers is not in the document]
MaxF12: Of course, my bad. I totally missed that part, that makes things a lot clearer and I now understand the advantage of doing it this way around, thanks.

On a different note, why is the framer exposing a setEventHandlers call while neither the preconnection nor the connection expose something similar even though they issue events as well?

draft-ietf-taps-interface.md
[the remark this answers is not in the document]
tfpauly: Fixed

draft-ietf-taps-interface.md
```diff
@@ -249,7 +258,7 @@ peer to peer establishment).
 Once a Connection is established, data can be sent on it in the form of
 Messages. The interface supports the preservation of message boundaries both
 via explicit Protocol Stack support, and via application support through a
-deframing callback which finds message boundaries in a stream. Messages are
+Message Framer definition which finds message boundaries in a stream. Messages are
```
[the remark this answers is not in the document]
tfpauly: Fixed

tfpauly:
Not done, but did some cleanup to remove the Instance object, etc. Now there's just a MessageFramer that interacts with the Connection, which simplifies things a bunch.

I'd like to add some diagrams, but also a code example of a simple length-value header framer.

I was looking at our "sample" code for a client and server, and it's not clear to me in our syntax how we show what the code does upon receiving an event. Let's say, I want to show that my message framer is called to handle a new sent message. Where do I write that "block" of code? @britram @mwelzl ?

mwelzl:
Earlier when we discussed an example, we agreed that we need it but that we could leave it out of this PR. Another answer: I think the receiving code is also changed by PR #332. ISTM that landing this one as well as #332 first, and then writing this example code would be a good sequence.

draft-ietf-taps-interface.md
```diff
@@ -1784,37 +1775,183 @@ that the other endpoint is done sending on a connection.
...
+
+Upon receiving this event, a Message Framer can inspect the inbound data. The
+data is parsed from a particular cursor representing the unprocessed data. The
+Framer requests a specific amount of data it needs to have available in order to parse.
+If the data is not available, the parse fails.
+
+~~~
+MessageFramer.Parse(Connection, MinimumIncompleteLength, MaximumLength) -> (Data?, IsEndOfMessage?)
+~~~
+
```
philsbln: This implies a Framer would always operate on a byte stream / partial message.
Should we add an Interface to de-frame data that is already a Message?

tfpauly: Hm, so it should be able to handle messages fine, since this is handling partial messages. I think the most common case of framer is over a bye stream, so I think handling that by default is important. But if you have a high `MinimumIncompleteLength`, any shorter Message will just be delivered along with `IsEndOfMessage`.

I suppose the thing to add here is the Message object itself.

draft-ietf-taps-interface.md
```diff
@@ -1784,37 +1775,183 @@ that the other endpoint is done sending on a connection.
...
+## Adding Message Framers to Connections
+
+The Message Framer object can then be added to one or more Preconnections
+to run on top of transport protocols. Multiple Framers may be added. If multiple
+Framers are added, the last one added runs first when framing outbound messages,
+and last when parsing inbound data.
+
 ~~~
+Preconnection.AddFramer(framer)
+~~~
```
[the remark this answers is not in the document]
tfpauly: Added!

draft-ietf-taps-interface.md
```diff
@@ -1784,37 +1775,183 @@ that the other endpoint is done sending on a connection.
...
+notify the Framer that a new Connection is being created.
+Similarly, a stop event is delivered when a Connection is being torn down.
+The Framer can use the Connection object to look up specific properties
+of the connection or the network being used that may influence how
+to frame Messages.
+
+~~~
+MessageFramer -> Start(Connection)
+MessageFramer -> Stop(Connection)
+~~~
```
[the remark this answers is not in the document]
tfpauly: Clarified event delivery

draft-ietf-taps-interface.md
```diff
@@ -1784,37 +1775,183 @@ that the other endpoint is done sending on a connection.
...
+
+Before a Message Framer marks itself as ready, it can also dynamically add a protocol or framer
+above it in the stack. This allows protocols like STARTTLS, that need to add TLS conditionally,
+to modify the Protocol Stack based on a handshake result.
+
+~~~
+otherFramer := NewMessageFramer()
+MessageFramer.PrependFramer(Connection, otherFramer)
+~~~
```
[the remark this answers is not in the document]
tfpauly: I think, since handshakes build from the bottom up, and you don't want outstanding data in the stream below you, it's only safe to prepend.

draft-ietf-taps-interface.md
```diff
@@ -1784,37 +1775,183 @@ that the other endpoint is done sending on a connection.
...
+
+## Sender-side Message Framing {#send-framing}
+
+Message Framers deliver an event whenever a Connection sends a new Message.
```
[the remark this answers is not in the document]
tfpauly: Clarified event delivery

draft-ietf-taps-interface.md
tfpauly: Added MessageContext here.

tfpauly:
@philsbln thanks for the comments! I added more clarification that there's a Message Framer object that delivers events to the framer implementation code, and made it more clear that this "framer implementation" code is something that the application or some other library does, and it's what receives the events to actual do the framing and run its code.

draft-ietf-taps-interface.md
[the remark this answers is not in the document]
philsbln: Right. So if a would do some kind of multiplexing within my framer, i can not easily reuse other framers that further process my messages. Anyway, this would require to dispatch stuff to another framer, not appending it.

tfpauly:
@mwelzl @britram can you take another look at the state of the PR? I'd like to see this merged and then iterate more if we need, to make sure that other work (like the message context PR) make sense with this and don't get out of sync.

mwelzl:
So sorry for leaving this dangling!  I looked at this PR recently but was too clumsy to notice that my requests have long been addressed  :(    now, I'd be fine with merging this, but again I'm afraid of my clumsiness because this talks of conflicts...   my suggestion: give @britram a day or two too, and then just go ahead and merge this in yourself. Either way, I think it's easy to agree that we should merge this now, and if there would be small fixes needed they can still be done later.

tfpauly:
Okay, based on @mwelzl's request, merging in now. Please file new issues for further changes!

PR #326, opened 2019-05-15
Note that Idempotent is needed for UDP like protocols

### LINKED ISSUE #302: UDP idempotence
This came up in a dinner among some of us yesterday: all UDP messages should be understood as being idempotent, because UDP doesn't protect the receiver against getting duplicates. This is different from unreliable transmission with, e.g., SCTP (or, probably, a future such extension of QUIC?), where you can be sure you're not getting a duplicate.

philsbln:
closes #302

### DIFF
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -1422,7 +1422,6 @@ Action, when such a Message exists. If false, the Message may be delivered to th
 This property is used for protocols that support preservation of data ordering,
 see {{prop-ordering}}, but allow out-of-order delivery for certain messages.
 
-
 ### Idempotent {#msg-idempotent}
 
 Name:
@@ -1437,9 +1436,11 @@ certain 0-RTT establishment techniques, where retransmission of the 0-RTT data
 may cause the remote application to receive the Message multiple times.
 
 Note that for protocols that do not protect against duplicated messages,
-e.g., UDP, all messages MUST be marked as Idempotent. 
-In order to enable protocol selection to choose such a protocol, 
-Idempotent MUST be added to the TransportProperties passed to the Preconnection.
+e.g., UDP, all messages MUST be marked as Idempotent.
+In order to enable protocol selection to choose such a protocol,
+Idempotent MUST be added to the TransportProperties passed to the
+Preconnection. If such a protocol was chosen, disabling Idempotent on
+individual messages MUST result in a SendError.
 
 ### Final {#msg-final}
 
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -644,10 +644,8 @@ indicates a preference for a specific path by specifying an interface, but also 
 preference for a protocol not available on this path, the transport system will
 try the path first, ignoring the preference.
 
-Both Selection and Connection Properties can be added to a Preconnection to
-configure the selection process, and to further configure the eventually
-selected protocol stack(s). They are collected into a TransportProperties object
-to be passed into a Preconnection object:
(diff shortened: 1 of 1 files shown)


draft-ietf-taps-interface.md
```diff
@@ -644,10 +644,8 @@ indicates a preference for a specific path by specifying an interface, but also
...
 
-Both Selection and Connection Properties can be added to a Preconnection to
-configure the selection process, and to further configure the eventually
-selected protocol stack(s). They are collected into a TransportProperties object
-to be passed into a Preconnection object:
+Selection, and Connection Properties, as well as defaults for Message Properties, can be added to a Preconnection to configure the selection process, and to further configure the eventually selected protocol stack(s).
+They are collected into a TransportProperties object to be passed into a Preconnection object:
 
 ~~~
 TransportProperties := NewTransportProperties()
```
mwelzl: I would add an extra sentence here to say, e.g.: "For protocols that do not protect against duplicated messages, e.g., UDP, a default Message Property of Idempotent=enabled must be assigned."

draft-ietf-taps-interface.md
```diff
@@ -1438,6 +1436,11 @@ more than once for a single Send Action. It is used to mark data safe for
 certain 0-RTT establishment techniques, where retransmission of the 0-RTT data
 may cause the remote application to receive the Message multiple times.
 
+Note that for protocols that do not protect against duplicated messages,
+e.g., UDP, all messages MUST be marked as Idempotent. 
+In order to enable protocol selection to choose such a protocol, 
+Idempotent MUST be added to the TransportProperties passed to the Preconnection.
```
mwelzl: This is good, but I think it's important to add something like this: "If Idempotent has been chosen as a default for the Preconnection, it may not be possible to enable it per Message later."

I mean, once UDP has been chosen, disabling Idempotent can't work, right?  So if the transport system still offers it, this produces a wrong expectation for an application. We should think if there are other Message Properties that have the same kind of behavior. I guess so - what if reliability = off by default? We MAY be able to enable it later (e.g. if SCTP with partial reliability was chosen, or even TCP, simply sending everything reliable), but it can't be promised (e.g. if UDP was chosen).

draft-ietf-taps-interface.md
philsbln: I thought about it, but decided against putting it there. This is a general section, and does not describe special cases. We may add it as an example here, but still looks like an odd place for this.

draft-ietf-taps-interface.md
philsbln: If you disable idempotent later, this may or may not work (as most message properties) I will add a hint here.

draft-ietf-taps-interface.md
mwelzl: That works for me

abrunstrom: I agree with @philsbln that it does not fit to add such details here.

MaxF12: Should you be able to change the default message properties after connection establishment? Right now you would not be able to as the preconnection gets consumed.

gorryfair: I'm not sure this is necessary to specify. Clearly a connection can change the per-message properties at any time - but I suggest that changing the default established in the preconnection adds more  text which isn't strictly necessary. Omitting this does not prevent an API providing it as an additional feature, but then it would then need to work out what to do when the connection is cloned and preconnection properties are replicated. I'd rather not delve there.

philsbln: @MaxF12 This is possible by changing the properties on the Connection object. Doing it on the Preconnection would be harmful, as we want to make the preconnection reusable. If you think this needs discussion, please open a new Issue for that.

Issue #324, opened 2019-05-15
Add explicit protocol selection.


philsbln:
This issue was discussed in the May 2019 interim.

Some Applications may really need a specific protocol and can not rely on the transport system  to choose an appropriate one.

gorryfair:
Agreed. Although that does not mean we have to or even should standardise a way to express this.

philsbln:
I guess we should standardise a way to express this, to avoid ossifying the selection properties on the assumption "combination x1 x2 x3 is guaranteed to give you YTP".
Nevertheless, we should discourage its use by having profiles as a more convenient alternative.

GrumpyOldTroll:
I agree with @philsbln; if we do not provide a standard way to express this, the workarounds people will end up using when they need this will be much worse for taps's adoption and goals than giving them a way to express an explicit need for a specific protocol.

Even though I also see that providing this will sometimes result in apps using the same old protocols instead of making use of new and better ones, we can't get rid of the need for picking specific protocols in some cases, and we can't serve the need by ignoring it.

IMO the solution is to accept that developers will make the wrong choice on this sometimes, and to make it so that there's a good 1-line fix when they realize it.  Making it harder to express won't prevent the mistakes, it'll just make the mistakes harder to unwind.

tfpauly:
I agree that apps need to be able to restrict this. That has been part of the architecture from the start, although the API doc doesn't specify how to do it.

Based on the architecture, my suggestion is:
- Specify a set of allowed protocol stacks if you want to restrict; if you leave it empty, we choose for you
- The protocol stacks are composed of per-protocol options, to express "if you use TCP, use these TCP flags"

philsbln:
Please take a look at PR #327 wether it addresses the first of @tfpauly's points.

For the second one, I think the current design with protocol specific properties is sufficient, as protocol specific properties on a preconnection are ignored for other protocols.

britram:
Suggestion from the authors' meeting in Montréal: leave this underspecified in the standard, and state that implementations SHOULD provide the application a way to override transport selection and instantiate a specific stack, but make this functionality explicitly non-portable.

britram:
(also explain why you may want to force a stack per @gorryfair)

britram:
Michael: figure out where to put this. Note that selection properties are also used for listening, and that (especially servers) might want to provide explicit protocol stack selection, but that methods for explicit selection are out of scope for the abstract interface. michael suggests this belongs up top, near/in intro.

PR #327, opened 2019-05-17
Adding a way of forcing the use of specific protocols

philsbln:
This fixes issue #324.

The solution involves specifying protocol sets on the local endpoint.
Note that the text is dependent the introduction of Transport Property Profiles.


draft-ietf-taps-interface.md
```diff
@@ -1011,6 +1011,30 @@ ChallengeCallback := NewCallback({
...
+~~~
+
+Multiple sets can be specified as alternatives:
+
+~~~
+LocalSpecifier := NewLocalEndpoint()
+LocalSpecifier.WithProtocols("tcp", "ipv4")
+LocalSpecifier.WithProtocols("tcp", "ipv6")
+~~~
+
```
MaxF12: Why is this done on the local endpoint object rather than on the transport properties object?

philsbln: The reason is that really forcing a protocol end-to-end is hard and requires crypto. Therefore, we only force the protocol at the local endpoint. On the path, there may be proxies or protocol converters, so that the Remote Endpoint may use another protocol. Adding it on the Local Endpoint object makes this explicit.

GrumpyOldTroll:
I like this extension, and thanks for opening the issue.

As written, it leaves me guessing about at least one assumption, which I think would be:

If the constrained protocol choice conflicts with the default TransportProperties, the TransportProperties _also_ must be set to something compatible, or you'd get an error when calling connect or listen, correct?

 In particular, I think this would always fail with error:

~~~
   Preconnection := NewPreconnection(
           NewLocalEndpoint.WithProtocols("udp"),
           NewRemoteEndpoint.WithHostname("example.com").WithPort(9000),
           NewTransportProperties(),
           NewSecurityParameters())
~~~

Fixing it would require explicitly suppressing the several default-require transport properties that conflict with udp in order to have a chance at not erroring:

~~~
   Preconnection := NewPreconnection(
           NewLocalEndpoint.WithProtocols("udp"),
           NewRemoteEndpoint.WithHostname("example.com").WithPort(9000),
           NewTransportProperties()
                .Ignore(reliability)
                .Ignore(preserve-order)
                .Ignore(congestion-control),
           NewSecurityParameters())
~~~

Is that expected?  Does it need to be made explicit?

And was there an intent for this to be made simpler with protocol-based profiles for TransportProperties?

philsbln:
I suspect with the addition of profiles, the solution should rather look like:

~~~
   Preconnection := NewPreconnection(
           NewLocalEndpoint.WithProtocols("udp"),
           NewRemoteEndpoint.WithHostname("example.com").WithPort(9000),
           NewTransportProperties(unreliable-datagram)
           NewSecurityParameters())
~~~

with "unreliable datagram" implying

~~~
           NewTransportProperties()
                .avoid(reliability)
                .avoid(preserve-order)
                .avoid(congestion-control)
                .add(idempotent, true)
           }
~~~

philsbln:
See PR #328 for a concrete proposal for profiles

draft-ietf-taps-interface.md
abrunstrom: I also find using the endpoint object confusing and agree with Max that it should be a transport property. It is used as a property for the selection and racing so we should keep all this together. (I am not convinced by the argument for why we should use the endpoint object. The path could violate other transport properties as well then.)

britram:
I'm confused: after you have #328, why do you need this?

philsbln:
@britram I share your confusion after re-reading it… there was something mixed up.

I see the following two cases:

1. If you need a UDP like protocol #328 solves your problem:
~~~
   Preconnection := NewPreconnection(
           NewRemoteEndpoint.WithHostname("example.com").WithPort(9000),
           NewTransportProperties(unreliable-datagram)
           NewSecurityParameters())
~~~

2. If you need exactly UDP because you defined a behaviour based on the protocol headers -- I don't find an example where this is useful, but I have several for TCP -- you need #327:
~~~
   Preconnection := NewPreconnection(
           NewLocalEndpoint.WithProtocols("udp"),
           NewRemoteEndpoint.WithHostname("example.com").WithPort(9000),
           NewTransportProperties(unreliable-datagram)
           NewSecurityParameters())
~~~

draft-ietf-taps-interface.md
philsbln: There is another argument why realising this as Transport Property is a little painful: what would the following mean:
~~~
transportPorperties.require(protocol, UDP)
transportPorperties.require(protocol, TCP)
transportPorperties.require(protocol, IPv4)
transportPorperties.require(protocol, IPv6)
~~~
Does it mean what I intended (require to use TCP or UDP and prefer IPv4), or is this only satisfied by TCPoverUDPoverIPv6overIPv4 or TCPoverUDPoverIPv4overIPv6?

~~~
NewLocalEndpoint.WithProtocols("udp","tcp")
NewLocalEndpoint.WithProtocols("IPv4","IPv6"),
~~~
Is clear - I want either TCP or UDP over IPv4 or IPv6 - and does not allow to use other preferences and thus reduces complexity.

mwelzl: I think the example that you give here makes the endpoint case clear because you assume an implicit "or" between parameters of one call. In the other case, you arrive at something ambiguous because you have only one flat property called "protocol".

The distinction that creates the problem here, to me, isn't about the local endpoint or use of a transport property; it's about semantics of the defined call that you use (or assume).

GrumpyOldTroll:
It just occurred to me to wonder: Is there any reason we shouldn't do exact protocol selection as an inheriting subclass of a transport property profile?

So maybe "exact-udp" inherits from "unreliable-datagram", and app designer just picks what they need as a transport profile?

I usually don't much care for inheritance, but are there cases where this wouldn't be as good or better? 
I agree with Anna I don't much like exact protocol selection in LocalEndpoint.

philsbln:
As suggested several times, I changed the way of forcing specific protocol stacks into an ordinary Selection Property. This comes at the cost of introducing a list type for properties.

@GrumpyOldTroll I don't think inheritance solves the problem here, as one may want to give the transport system a choice between multiple sets of protocols. For example, SIP is defined to operate over SCTP, UDP or TCP – therefore, we would need a fairly generic set of Selection Properties that allows to use any of these and the newly introduced Property to force to us no other matching protocol.

britram:
Okay. I'm still not a giant fan of this, because I'm pretty sure it'll get abused. Some extra text stating that this is basically only intended for special circumstances (e.g., implementing a layer-4+-topped stack over a specific layer-4-topped stack) would address that as best we can in an abstract API, though.

mirjak:
I'm also not a fan. I think it should be a bit hard to force use of one specific protocol because the whole idea of this exercise is to have an interface that is NOT protocol-specific.

philsbln:
I thought the text in this PR is already quite clear that this property is only for wired special cases.
  
@britram please make a suggestion how to make the text even more hostile for people that want to use the property.

britram:
Per discussion on #324, we're going to do this another way.

Issue #325, opened 2019-05-15
Add support for Property Profiles


philsbln:
This issue was discussed in the May 2019 interim.

After fist implementation experiences, we feel that we should add a convenient way to express common Transport Parameter combinations as some kind of profile, e.g. "TCP like" meaning "Reliable, in-order, non-message preserving"

philsbln:
Proposal how to solve this:

Replace the TransportProperties constructor in section 5.2 with
~~~
   TransportProperties := NewTransportProperties(propertyProfile?)
~~~

@gorryfair promised this at the interim. I only open this issue to document the proposal how to solve this.

MaxF12:
This can be closed.

philsbln:
Closed by #328

PR #328, opened 2019-05-21
Add Transport Property Profiles

philsbln:
This addresses Issue #325

### DIFF
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -2267,7 +2267,8 @@ their overall expected behavior.
 ## reliable-inorder-stream
 
 This profile provides reliable, in-order transport service with
-congestion control. 
+congestion control.
+An example of a protocol that provides this service is TCP.
 It should consist of the following properties:
 
  | Property                 | Value     |
@@ -2281,6 +2282,7 @@ It should consist of the following properties:
 
 This profile provides message-preserving, reliable, in-order 
 transport service with congestion control.
+An example of a protocol that provides this service is SCTP.
 It should consist of the following properties:
 
  | Property                 | Value     |
@@ -2293,6 +2295,7 @@ It should consist of the following properties:
 ## unreliable-datagram
 
 This profile provides unreliable datagram transport service.
+An example of a protocol that provides this service is UDP.
 It should consist of the following properties:
 
  | Property                 | Value     |
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -655,7 +655,7 @@ TransportProperties := NewTransportProperties(profile?)
 
 With the creation of the TransportProperties object, an optional profile
 can be passed. These profiles consist of a list of properties that are used
-together frequently. 
+together frequently and can be overridden by adding additional TransportProperties.
 See {{property-profiles}} for a list of profiles implementations should support.
 Individual properties are then added to the TransportProperties Object:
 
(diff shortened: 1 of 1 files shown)


gorryfair:
So, some comments:
- First I suggest that we explicitly refer to these but the service they provide, not the protocol they resemble. Hence:

• The final one would be "datagram service" and UDP is an example of such. 

• I would also suggest it is useful to say that the application-supplied properties can now override these defaults, hence in the Datagram case, the property permitting/requiring partial coverage could result in UDP-Lite being selected instead of UDP to supply this service.

• Then, I have a question about the specific UDP/Datagram service. This currently says:

 | Property                 | Value     |
  |:-------------------------|:----------|
  | reliability              | avoid     |
  | preserve-order           | avoid     |
  | congestion-control       | avoid     |
  | preserve-msg-boundaries  | require   |
  | idempotent               | true      |

To me some of these are not really requirements, but to me seem like side affects. The requirement could be to need low latency, and idempotence, preserve-msg-boundaries, etc 
but I wouldn't care if the service also happened to provide the following, as long as those requirements (or whatever we agree upon) were prioritised: reliability,  preserve-order , congestion-control. (Of course UDP would not offer these, and that would also be fine).

Thoughts?

philsbln:
I agree with the first point and updated the PR accordingly.

For the second one, I also agree that avoiding reliability, order preservation, and congestion-control are side effects of latency or other optimisations and do not really belong into the profile. Therefore, I changed "avoid" to "ignore".
I am a bit hesitant about speaking about UDP lite here given your first point, tried to incorporate it as a hint in a slightly different way. 

After these changes, I would like to raise the question whether we should mention "udp like" or "tcp like" for the impatient searching the document.

gorryfair:
Something to help start thinking: we could say something like this datagram service resembles the service offered by the socket API that binds to the UDP transport?

gorryfair:
Something to help start thinking: we could say something like this datagram service resembles the service offered by the socket API that binds to the UDP transport?

mwelzl:
I checked this now and like it. I also read the dialogue and agree with all points that were raised and how they were answered.

Regarding @gorryfair 's suggestion to say: "we could say something like this datagram service resembles the service offered by the socket API that binds to the UDP transport?" - actually, when reading the update first before looking at the comments, I thought "why doesn't this say: 'an example of a protocol that provides this service is UDP' ?". I think this is almost the same thing, but I like my own phrasing better  :)   the whole "socket API that binds to" thing seems a bit clumsy to me and I don't think we need it.

draft-ietf-taps-interface.md
```diff
@@ -655,9 +655,13 @@ Selection, and Connection Properties, as well as defaults for Message Properties
 They are collected into a TransportProperties object to be passed into a Preconnection object:
 
 ~~~
-TransportProperties := NewTransportProperties()
+TransportProperties := NewTransportProperties(profile?)
```
britram: I think this is very useful in the appendix, but adding language here in the normative section might lead to some of the confusion that putting this in the appendixi is meant to avoid. Consider leaving this section unchanged, and make the modification to NewTransportProperties() in the appendix where properties are introduced.

philsbln: Hmm… I think it is a much worse idea to modify function signatures in the appendix

philsbln:
I liked the split between saying "there are profiles" in the normative part and defining examples for these in the appendix. If this is a Problem, we should rather move the profiles into the property section. I have a bad feeling about modifying function signatures defined in the main documents in the appendix.

Any other opinions? 
Leave un-merged and discuss in Montreal?

mirjak:
I actually think this is useful but more an implementation question. I don't think this need to a part of the base API, so having this in the appendix or the implementation draft seems more suitable for me.

philsbln:
@mirjak What about the Parameter to pass the profile – is re-defining it in the appendix the right way?

mirjak:
The API we define is anyway mostly a recommendation and will not be normative; so you will always she slight variations in the implementation. Proposing an optional parameter (in the appendix or implementation draft) seems okay.

Issue #331, opened 2019-05-24
9.1 defaults


mwelzl:
With the exception of read-only Properties, something should be said about the default values of the Generic Connection Properties in Section 9.1.

britram:
fixed by #335

Issue #316, opened 2019-04-29
 Section 4.2.1 - Transport Properties


gorryfair:
Section 4.2.1: “Allow different components of a TAPS implementation to pass Transport Properties, e.g., between a language frontend and a policy manager.”
- Question??? : Would it also be OK to include a suggestion that these could also be retrieved from storage, e.g. “Allow different components of a TAPS implementation to pass Transport Properties, e.g., between a language frontend and a policy manager, or to retrieve a list of properties from a file or other storage.”

 Section 4.2.1.:
Are the definitions of “should” actually “SHOULD”, I don’t otherwise understand the meaning of the lower case should. Actually, why is it SHOULD not MUST for the IETF-specified class?

philsbln:
I am in favour of adding the suggestion of adding a suggestion of using this representation on stable storage, but shouldn't it be rather "or as a representation of properties retrieved from a file or other storage."

I chose "should" over "must" here because I feared  defining protocol specific properties for all important IETF protocols may creep in as a dependency for publication of this document otherwise.
With regarded to the casing, this really must be a SHOULD.

britram:
fixed by #335

Issue #315, opened 2019-04-29
Default for Control checksum coverage


gorryfair:
In Section 5.2.7 Control checksum coverage on sending

“The recommended default is to Ignore this option.”

I argue that this default is much braver than specified in the IETF present in RFC 8085. The default for IPv6 is also that this is required by default. Again, I would argue that applications need to explicitly be designed to accept this.

—
In Section, 5.2.8. Control checksum coverage on receiving

“The recommended default is to Ignore this option.”

I argue that this default is much braver than specified in the IETF present in RFC 8085. The default for IPv6 is also that this is required by default. Again, I would argue that applications need to explicitly be designed to accept this.

britram:
this seems reasonable to me, I don't recall how we ended up at Ignore.

philsbln:
@gorryfair I think we need to re-phrase this as I guess you mis-read it.

The property only defines whether you want a protocol that provides this feature, not whether it should be enabled. The latter is controlled by the property in 7.3.6.

abrunstrom:
Yes, this seems to be a misunderstanding. Perhaps we also need to clarify that _Ignore_ just means that the property is ignored when ranking alternatives for candiadet selection and racing, not that the application can ignore it?

gorryfair:
I don't understand something about 5.2.7 and 5.2.8 ... So let me ask a more stupid question:
- IETF transports always do provide full coverage by default.
- Why do we have a property to request control of checksum coverage at all, why don't we just have a property that says you want to have (require) or would like to have (prefer), or don't care (ignore) some size of limited checksum coverage - as in 7.3.6. 

If you don't get the coverage that you ask for, then you will get more coverage. ... so "require" is in some way "prefer a protocol that gives at least an ability to provide this much?".

---
Then in 7.3.6:

I think " This property specifies the length of the section of the Message...." is actually the *minimum* length of message - the transport can always protect more.

I don't understand: " ... a special value (e.g. -1) can be used to indicate the default.  "
- the default is *full* coverage, how is this different to a zero parameter?

Gorry

britram:
per TAPS interim, let's Avoid this.

mwelzl:
PR #330 partially addresses this:
7.3.6 was re-written to include "minimum" and fix the special value related text, also using the same style as in the other Message Properties now to specify the default.

mwelzl:
PR #335 addresses the last missing bit here:

> IETF transports always do provide full coverage by default.
[...]

PR #335, opened 2019-07-02
Various minor issues - closes #314, #315 #316 and #331

### LINKED ISSUE #314: Default for Use 0-RTT session establishment
In Section 5.2.5: Use 0-RTT session establishment with an idempotent Message

“The recommended default is to Prefer this option.”

I argue that this default is much braver than specified in the IETF present. Once could argue the same for QUIC, but the TCP-basis for this feature was determined, based - as I recall - on the need for idempotent data *and* the need to provide server-side support that was not specified. The safe option is to not enable this as a default, requiring Apps to explicitly say their data is idempotent.

“potentially multiple times” - probably should be explicitly explained that “the transport does not provide idempotency, which could result in multiple copies of the message data being passed to the Receiving Endpoint.”

mwelzl:
... and some nits (capitalization, ..)

#314: 0-rtt establishment: fixed default as requested by @gorryfair and incorporated text fix suggestion
#315: last missing update to hopefully close this issue: adjusted the wording of checksum related selection properties
#316: fixed some text details regarding how transport properties are to be specified, based on a converged discussion.
#331: wrote some defaults for parameters in section 9.1

### DIFF
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -467,7 +467,7 @@ form \[\<Namespace>.\]\<PropertyName\>.
   for properties defined by an RFC which are not protocol specific.
 - Protocol Specific Properties must use the protocol acronym as
   Namespace, e.g., “tcp" for TCP specific Transport Properties.
-  For IETF protocols, property names under these namespaces should
+  For IETF protocols, property names under these namespaces SHOULD
   be defined in an RFC.
 - Vendor or implementation specific properties must use a
   a string identifying the vendor or implementation as Namespace.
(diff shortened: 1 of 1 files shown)


draft-ietf-taps-interface.md
```diff
@@ -2272,21 +2290,25 @@ Name:
 : max-send-rate / max-recv-rate
 
 Type:
-: Integer (positive)
+: Integer (positive) / Integer (positive)
```
britram: -1 is not a positive integer :P

```suggestion
: Integer / Integer
```

britram: (On second thought, do we want to restrict this beyond Numeric?)

mwelzl: Tsssssssssssk what a silly mistake!  Thanks!
Re: numeric, I absolutely agree!  I even thought about it as I made this PR and wanted to fix this (because it was an Integer before) but somehow forgot it.

PR #346, opened 2019-07-26
Fixes #313 per Gorry's comments

### LINKED ISSUE #313: API Draft author editorial comments
```
“The simplicity of this API is a key reason the Internet won the protocol wars of the 1980s. “
- Reference needed (some may be too young to remember):-0.
—
“As a first step to realizing this design, [I-D.ietf-taps-arch] describes a high-level architecture for transport services.”
- I’d much prefer to remove “As a first step to” from the start of the final para of introduction. Once that is published as a PS we do not need to be so “tentative”, and perhaps unintended but still seems to potentially devalue the arch
—
“The method for dispatching and handling Events is left as an”
- I’m unsure what “left as” intends, is it just that this is an implementation detail, or was something not done? Would the sentence read the same without /left as/?
—
“Message- as opposed to stream-orientation,”
-could this be:
“Message-orientation as opposed to stream-orientation,”
—
“An application primarily interacts with this interface through two Objects, Preconnections and Connections.”
Could we use a colon instead of a comma after objects?
—
“defined in [I-D.ietf-taps-arch].|
- could be: “defined in the TAPS Architecture [I-D.ietf-taps-arch].””
—
“following the phases described in [I-D.ietf-taps-arch].”
- could be:
“following the phases defined in the TAPS Architecture [I-D.ietf-taps-arch].”
—
“endpoint” is capitalised in the architecture Spec, when used as a TAPS term.
—
“but there is a transport protocol which provides “
- please change /which/ to /that/.
—
“this octet stream”
- do we still with in 2019 to talk about “octet” streams?
—
In 4.2 at the end of para 2. we seem to have a list delimited by inline minus symbols, could these be translated to bullets to make this easier to read?
—
In 4.2 para 2. 
- The setting of parameters is now well-described in detail in the architecture - please can we cross-ref here?
—
In 4.2 final para:
“Connection Properties specified early on may be used”
- could be better as:
“Early specification of Connection Properties allows their use”
—
“Also note that Protocol Specific Properties”
- I’d prefer to not diminish this to a note, cam we simply state:
“Protocol Specific Properties”
—
In 4.2.1.:
“for properties defined by an RFC which are not protocol specific..”
- should be:
- “for properties that are not specific to a protocol and are defined by an RFC.”
—
In Section 4.3:
“While there is no interoperability benefit to tightly defining how the interface be presented to application programmers in diverse platforms, maintaining the “shape” of the abstract interface across these platforms reduces the effort for programmers who learn the transport services interface to apply their knowledge in multiple platforms.”
- This sentence is unduly hard to parse:
“There is no interoperability benefit in tightly defining how the interface is presented to application programmers across diverse platforms. However, maintaining the “shape” of the abstract interface across these platforms reduces the effort for programmers who learn the transport services interface to then apply their knowledge across multiple platforms.”
—
“SHOULD carry the names given for them in the document, “ - could we change /carry/ to /use/ or /specify/ or /employ/ or something?
—
“, even if said implementation is a non-operation, e.g. because transport protocols implementing a given Property are not available on the platform.”
Could we break the sentence and explain as:
“. The interface should be implemented even when this results in no-operation, e.g. there is no action when the API specifies a Property that is not available in a transport protocol implemented on a specific platform.“
—
“static singleton objects” 
- probably should be explained.
—
“Pre-Establishment” is capitalised in the architecture.
—
“for the Connections they are about to make, or to query the API about potential connections they could make.”
- why is connections capitalised once and not the second time?
—
“Selection Properties which select paths”
- change /which/ to /that/.
—
“Also note that reliability is usually combined with congestion control in protocol implementations, rendering “reliable but not congestion controlled” a request that is unlikely to succeed. “
- Agree, but useful to add a parenthesis that the converse is not necessarily true.
—
“Note that marking a specific interface”
- In the next para the note is incorporated as the final line of the para in which the note applies, can we change this to also be in the same place?
—
“Implementations should provide all types that are supported on some system to all systems, in order to allow applications to write generic code.”
- I think this could be improved by:
“An implementation should provide all types that are supported on the local system to all remote systems, to allow applications to write generic code.”
- but what does /all/ imply?
—
“Note that marking a specific interface type as Required limits path selection to a small set of interfaces, and leads to less flexible and resilient connection establishment.”
- Is that ALWAYS the case? I think this is a case of “could lead to” or at best “may often lead to”
—
In 5.2.12:
“One possible side-effect is increased jitter, which may be problematic for delay-sensitive applications. The recommended default is to Prefer this option.”
- If this only leads to increase jitter, then this default may be OK. Are we sure this can not lead to duplication or reordering of data? If it did, then I’m concerned this becomes a default.
—
In 5.2.13:
“In case a unidirectional connection is requested, but unidirectional connections are not supported by the transport protocol, the system should fall back to bidirectional transport.
“
- please explain “fall back” in terms of the TAPS language.
- What is the default if not specified - I think the default should be bidirectional.
—
In section 5.3.2:
“Once configured, parameters may also be supplied during connection establishment. These are best handled as client-provided callbacks. “
- Is a callback what the TAPS architecture refers to as an Event?
```

britram:
This handles all of Gorry's editorial comments in #313, save 

> "Once configured, parameters may also be supplied during connection establishment. These are best handled as client-provided callbacks. “: Is a callback what the TAPS architecture refers to as an Event?

which should be discussed in a new issue.


britram:
unaddressed comment in #347

mwelzl:
I think that this is ready to be merged, but requires merging by someone who's less prone than me to mess something up when pressing the "Resolve conflicts" button.  @britram? Please?

Issue #312, opened 2019-04-29
Detailed author review of Arch text for -03 to prepare -04


gorryfair:
TAPS Architecture - Additional comments on rev -03 (detailed read-through).

Note: This adds to my previous set of review issues - these could be handled at the same time.

Section 1 :
- Can we make reading easier:
“This flexibility does not only enable
   faster deployment of new feature and protocols, but it can also
   support applications with racing and fallback mechanisms which
   otherwise need to be implemented in each application separately.”
- suggest:
“This flexibility enables
   faster deployment of new feature and protocols,. It can also
   support applications with racing and fallback mechanisms, which
   otherwise need to be implemented in each application separately.””
—
“Although following the Transport
   Services Architecture does of course not mean that “
- do we need /of course/ (this reads oddly).
—
“from one system to the another”
- should be:
“from one system to another”
==========
Section 1.1 :
“The model of using sockets for networking can be represented as
   follows”
- could we insert “traditional” or “existing” or something like that before “model”, since a new reader may not realise this is the description of what exists rather what this document develops?
—
“, which provides the interface to the implementations of
   UDP and TCP (typically implemented in the system's kernel), which in
   turn send data over the available network layer interfaces.”
- Two “, which” in a long sentence, could we write:
- “. This API provides the interface to the implementations of
   UDP and TCP (typically implemented in the system's kernel), which in
   turn send data over the available network layer interfaces.”
—
“The Implementation”
- Is capitalised “I” needed?
—
“for mapping the API into the various available transport”
- Is this “into” or “to”  - I think the interface maps to these?
—
“There are a few key departures that Transport Services makes from the
   sockets API:”
- a “few key” seems odd to me in a spec.
“There are key differences between the architecture of the Transport Services system and the architecture of the sockets API:” - or something like that?
==========
Section 1.2 :
“and changes to available network links”:
- insert “the” and “to” so it does not mean that it makes the change:
- “and changes in the available network links”
—
Section 1.3 :
“HTTP/1.1 uses character delimiters to segment messages
   over a stream; TLS record headers carry a version, content type, and
   length; and HTTP/2 uses”
- I think this needs text needs several references to the specs?
==========
Section 3 :
“   The following considerations were used in the design of this
   architecture.”
[...]

PR #349, opened 2019-08-20
Address review comments on architecture draft

tfpauly:
Addresses Gorry's great comments from #312!

### DIFF
--- a/draft-ietf-taps-arch.md
+++ b/draft-ietf-taps-arch.md
@@ -76,19 +76,6 @@ author:
     email: cawood@apple.com
 
 informative:
-    RFC0793:
-    RFC2119:
-    RFC3168:
-    RFC4303:
-    RFC6265:
-    RFC7413:
-    RFC8095:
-    RFC8174:
-    RFC8446:
-    I-D.ietf-taps-minset:
-    I-D.ietf-taps-interface:
-    I-D.ietf-taps-impl:
-    I-D.ietf-taps-transport-security:
     POSIX:
       title: "IEEE Std. 1003.1-2008 Standard for Information Technology -- Portable Operating System Interface (POSIX).  Open group Technical Standard: Base Specifications, Issue 7"
       url: <http://www.opengroup.org/austin>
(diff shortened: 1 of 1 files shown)


draft-ietf-taps-arch.md
```diff
@@ -196,51 +187,49 @@ Flexibility after connection establishment is also important. Transport protocol
...
+One of the key insights to come from identifying the minimal set of features provided by transport protocols {{?I-D.ietf-taps-minset}} was that features either require application interaction and guidance (referred to as Functional Features), or else can be handled automatically by a system implementing Transport Services (referred to as Automatable Features). Among the Functional Features, some were common across all or nearly all transport protocols, while others could be seen as features that, if specified, would only be useful with a subset of protocols, but would not harm the functionality of other protocols. For example, some protocols can deliver messages faster for applications that do not require messages to arrive in the order in which they were sent. However, this functionality needs to be explicitly allowed by the application, since reordering messages would be undesirable in many cases.
```
mwelzl: This is a new request - sorry - but: please replace the two occurrences of "Functional Features" with "Functional or Optimizing Features". That's because we have three categories in minset: Functional, Optimizing, Automatable (see the first 4 paragraphs of https://tools.ietf.org/html/draft-ietf-taps-minset-11#section-3 ).

draft-ietf-taps-arch.md
```diff
@@ -484,13 +473,13 @@ Clients need to ensure that security APIs are used appropriately. In cases where
...
 validated. For example, clients ought not to use PSK material created for the
-Encapsulating Security Protocol (ESP, part of IPsec) {{RFC4303}} with QUIC, and clients
+Encapsulating Security Protocol (ESP, part of IPsec) {{?RFC4303}} with QUIC, and clients
 ought not to use private keys intended for server authentication as a keys for
 client authentication.
 
 Moreover, unlike certain transport features such as TCP
-Fast Open (TFO) {{RFC7413}} or Explicit Congestion Notification (ECN)
-{{RFC3168}} which can fall back to standard configurations, Transport
+Fast Open (TFO) {{?RFC7413}} or Explicit Congestion Notification (ECN)
```
mwelzl: Another new request, sorry again to only bring this up now: the mention of TFO and ECN here seems strange, because these mechanisms are internal to a protocol (the TCP implementation can try and fall back), whereas our transport system would just use protocols as they are. For that reason, these things do also not match our definition of "Transport Features" (as defined in RFC 8095, and then repeated in RFC 8303 and minset). I would just remove all the text from "unlike ...."  until "... standard configurations".

tfpauly: Fixed!

draft-ietf-taps-arch.md
tfpauly: Fixed!

Issue #343, opened 2019-07-22
Listeners should have backpressure


tfpauly:
There should be a way to put back pressure or limits on the number of inbound connections for a listener. The two approaches are:

1. To have a one-to-one callback for listen -> received connect
2. To have a way to set a numeric limit of new inbound connections

My preference is (2). (1) has an issue of being very impractical for servers that don't know how many connections they'll receive, and may require them to call `listen` a thousand times without any other activity. (2) is more similar to how calling `listen()` on sockets work.

mwelzl:
+1 on variant 2 - indeed it's common and known to people from sockets, and just seems like the easier and more efficient solution (albeit maybe less "beautiful", in a way).

abrunstrom:
+1 on variant 2 as well.

britram:
Do variant 2.

mwelzl:
I believe that PR #350 fully addresses this; hence I'm closing it.

PR #350, opened 2019-08-27
Allow setting new connection limits for inbound connections

tfpauly:
This implements what we discussed on #343

### DIFF
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -1109,24 +1109,24 @@ After Stop() is called, the Listener can be disposed of.
 Listener -> ConnectionReceived<Connection>
 ~~~
 
-The ConnectionReceived event occurs when a Remote Endpoint has established a
+The ConnectionReceived Event occurs when a Remote Endpoint has established a
 transport-layer connection to this Listener (for Connection-oriented
 transport protocols), or when the first Message has been received from the
 Remote Endpoint (for Connectionless protocols), causing a new Connection to be
 created. The resulting Connection is contained within the ConnectionReceived
-event, and is ready to use as soon as it is passed to the application via the
+Event, and is ready to use as soon as it is passed to the application via the
 event.
 
 ~~~
 Listener.SetNewConnectionLimit(value)
 ~~~
 
-If the caller wants to rate-limit the number of inbound Connections it that will be delivered,
+If the caller wants to rate-limit the number of inbound Connections that will be delivered,
 it can set a cap using SetNewConnectionLimit(). This mechanism allows a server to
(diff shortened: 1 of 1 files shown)


draft-ietf-taps-interface.md
```diff
@@ -1109,14 +1109,26 @@ After Stop() is called, the Listener can be disposed of.
 Listener -> ConnectionReceived<Connection>
 ~~~
 
-The ConnectionReceived Event occurs when a Remote Endpoint has established a
+The ConnectionReceived event occurs when a Remote Endpoint has established a
```
mwelzl: The terminology section writes events as Events and in other places the word is capitalized too. I suggest to keep this as "Event" now to be reasonably consistent - if you want Events (and, I guess, then also Actions) to be lowercase, this can be discussed, but then let's do this for the whole document at some point.
(Personally, I couldn't care less, but I do want things to be as consistent as possible.)

draft-ietf-taps-interface.md
```diff
@@ -1109,14 +1109,26 @@ After Stop() is called, the Listener can be disposed of.
...
 transport-layer connection to this Listener (for Connection-oriented
 transport protocols), or when the first Message has been received from the
 Remote Endpoint (for Connectionless protocols), causing a new Connection to be
 created. The resulting Connection is contained within the ConnectionReceived
 event, and is ready to use as soon as it is passed to the application via the
 event.
```
mwelzl: so the "Event" vs. "event" thing applies here too.

draft-ietf-taps-interface.md
```diff
@@ -1109,14 +1109,26 @@ After Stop() is called, the Listener can be disposed of.
...
 
+~~~
+Listener.SetNewConnectionLimit(value)
+~~~
+
+If the caller wants to rate-limit the number of inbound Connections it that will be delivered,
```
mwelzl: remove "it" before "that will be"

draft-ietf-taps-interface.md
```diff
@@ -1109,14 +1109,26 @@ After Stop() is called, the Listener can be disposed of.
...
+it can set a cap using SetNewConnectionLimit(). This mechanism allows a server to
+protect itself from being drained of resources. Each time a new Connection is delivered
+by the ConnectionReceived event, the value is automatically decremented. Once the
+value reaches zero, no further Connections will be delivered until the caller sets the
+limit to a higher value. By default, this value is Infinite. The caller also is able to reset
+the value to Infinite at any point.
```
mwelzl: This may just be me (as I'm not a native speaker), but shouldn't the word order be "The caller is also able to reset..." ?

Issue #342, opened 2019-07-21
Allow error events to include reason


MaxF12:
Error events should be able to hand the application a reason that further details the circumstances of the error event being fired. This should be optional and doesn't have to be portable but an implementation should provide an enumerated list of all the possible reasons.

mwelzl:
Just to be clear, you mean "...but an application should be provided with an enumerated list of ...", right?

MaxF12:
Yes sorry, the TAPS implementation should provide a list of enumerated reasons for errors.

mwelzl:
No problem. That's an easy text fix, I think.

MaxF12:
Closed by #351

PR #351, opened 2019-09-19
Allow for error events to include an optional reason

### LINKED ISSUE #341: Make InitiateError more generic
The current reasons for an InitiateError to occur exclude several configurations that should result in one, e.g. the localAddress being a group address but the selected direction of communication being bidirectional.

MaxF12:
Closes #341 and #342

### DIFF
--- a/draft-ietf-taps-impl.md
+++ b/draft-ietf-taps-impl.md
@@ -966,37 +966,37 @@ If the application indicates a preference to avoid expensive paths, and some pat
 
 # Reasons for errors {#appendix-reasons-errors}
 
-The Transport Services API {{I-D.ietf-taps-interface}} allows for the seven generic error types to specify a more detailed reason as to why an error occurred. This appendix lists some of the possible reasons.
+The Transport Services API {{I-D.ietf-taps-interface}} allows for the several generic error types to specify a more detailed reason as to why an error occurred. This appendix lists some of the possible reasons.
 
 * BadConfiguration:
 The transport properties and endpoints provided by the application are either contradictory or incomplete. Examples include the lack of a remote endpoint on an active open or using a multicast group address while not requesting a unidirectional receive.
 
 * NoCandidates:
-None of the available transport protocols can satisfy the transport properties provided by the application.
+The configuration is valid, but none of the available transport protocols can satisfy the transport properties provided by the application.
 
-* Unresolvable:
+* ResolutionFailed:
 The remote or local specifier provided by the application can not be resolved.
 
-* CantEstablish:
+* EstablishmentFailed:
 The TAPS system was unable to establish a transport-layer connection to the remote endpoint specified by the application.
 
 * PolicyProhibited:
-The system policy prevents the TAPS system from performing the action requested by the application. 
+The system policy prevents the transport system from performing the action requested by the application. 
 
 * NotCloneable:
 The protocol stack is not capable of being cloned. 
 
-* MessageToLarge:
-The message size is too big for the taps system to handle.
+* MessageTooLarge:
+The message size is too big for the transport system to handle.
 
 * ProtocolFailure:
 The underlying protocol stack failed. 
 
-* BadMessageProperties:
-The message properties are either contradictory to the transport properties or they can not be satisfied by the TAPS system. 
+* InvalidMessageProperties:
+The message properties are either contradictory to the transport properties or they can not be satisfied by the transport system. 
 
 * BadReceive:
-The data that was received by the underlying protocol stack could not be fully retrieved by the TAPS system. 
+The data that was received by the underlying protocol stack could not be fully retrieved by the transport system. 
 
 * FailedReceive:
 The underlying protocol stack indicated that the reception of a message has failed.
--- a/draft-ietf-taps-impl.md
+++ b/draft-ietf-taps-impl.md
@@ -960,3 +960,46 @@ If the application indicates a bound on the expected Send or Receive bitrate, an
 
 * Cost Preferences:
 If the application indicates a preference to avoid expensive paths, and some paths are associated with a monetary cost, an implementation should decrease the ranking of such paths. If the application indicates that it prohibits using expensive paths, paths that are associated with a cost should be purged from the decision tree.
+
+# Reasons for errors {#appendix-reasons-errors}
+
+The Transport Services API {{I-D.ietf-taps-interface}} allows for the seven generic error types to specify a more detailed reason as to why an error occurred. This appendix lists some of the possible reasons.
+
+* BadConfiguration:
+The transport properties and endpoints provided by the application are either contradictory or incomplete. Examples include the lack of a remote endpoint on an active open or using a multicast group address while not requesting a unidirectional receive.
+
+* NoCandidates:
+None of the available transport protocols can satisfy the transport properties provided by the application.
+
+* Unresolvable:
+The remote or local specifier provided by the application can not be resolved.
+
+* CantEstablish:
+The TAPS system was unable to establish a transport-layer connection to the remote endpoint specified by the application.
+
+* PolicyProhibited:
+The system policy prevents the TAPS system from performing the action requested by the application. 
+
+* NotCloneable:
+The protocol stack is not capable of being cloned. 
+
+* MessageToLarge:
+The message size is too big for the taps system to handle.
+
+* ProtocolFailure:
+The underlying protocol stack failed. 
+
+* BadMessageProperties:
+The message properties are either contradictory to the transport properties or they can not be satisfied by the TAPS system. 
+
+* BadReceive:
+The data that was received by the underlying protocol stack could not be fully retrieved by the TAPS system. 
+
+* FailedReceive:
+The underlying protocol stack indicated that the reception of a message has failed.
+
+* ConnectionAborted:
+The connection was aborted by the remote.
+
+* Timeout:
+Delivery of a message to the remote was not possible after a timeout. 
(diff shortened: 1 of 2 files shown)


draft-ietf-taps-impl.md
```diff
@@ -963,3 +963,46 @@ If the application indicates a bound on the expected Send or Receive bitrate, an
 
 * Cost Preferences:
 If the application indicates a preference to avoid expensive paths, and some paths are associated with a monetary cost, an implementation should decrease the ranking of such paths. If the application indicates that it prohibits using expensive paths, paths that are associated with a cost should be purged from the decision tree.
+
+# Reasons for errors {#appendix-reasons-errors}
+
+The Transport Services API {{I-D.ietf-taps-interface}} allows for the seven generic error types to specify a more detailed reason as to why an error occurred. This appendix lists some of the possible reasons.
```
theri: I'm not sure we really need to count the error types here, so I suggest deleting "the seven" here.

draft-ietf-taps-impl.md
```diff
@@ -963,3 +963,46 @@ If the application indicates a bound on the expected Send or Receive bitrate, an
...
+
+* BadConfiguration:
+The transport properties and endpoints provided by the application are either contradictory or incomplete. Examples include the lack of a remote endpoint on an active open or using a multicast group address while not requesting a unidirectional receive.
+
+* NoCandidates:
+None of the available transport protocols can satisfy the transport properties provided by the application.
```
theri: Here, let's maybe make the difference from BadConfiguration explicit: "The configuration is valid, but none of the available transport protocols..."

draft-ietf-taps-impl.md
```diff
@@ -963,3 +963,46 @@ If the application indicates a bound on the expected Send or Receive bitrate, an
...
+
+* Unresolvable:
+The remote or local specifier provided by the application can not be resolved.
```
theri: Maybe change this to "ResolutionFailed" or similar - "Unresolvable" sounds like an inherent and persistent problem with the specifier, but it might just be a problem with the resolver.

draft-ietf-taps-impl.md
```diff
@@ -963,3 +963,46 @@ If the application indicates a bound on the expected Send or Receive bitrate, an
...
+The TAPS system was unable to establish a transport-layer connection to the remote endpoint specified by the application.
+
+* PolicyProhibited:
+The system policy prevents the TAPS system from performing the action requested by the application. 
+
+* NotCloneable:
+The protocol stack is not capable of being cloned. 
+
+* MessageToLarge:
+The message size is too big for the taps system to handle.
```
theri: Nits: 
"ToLarge" --> "TooLarge"
"taps system" --> "transport system"

draft-ietf-taps-impl.md
```diff
@@ -963,3 +963,46 @@ If the application indicates a bound on the expected Send or Receive bitrate, an
...
+
+* ProtocolFailure:
+The underlying protocol stack failed.
```
theri: What's the difference to "CantEstablish"?

draft-ietf-taps-impl.md
```diff
@@ -963,3 +963,46 @@ If the application indicates a bound on the expected Send or Receive bitrate, an
...
+The message properties are either contradictory to the transport properties or they can not be satisfied by the TAPS system. 
+
+* BadReceive:
+The data that was received by the underlying protocol stack could not be fully retrieved by the TAPS system. 
+
+* FailedReceive:
+The underlying protocol stack indicated that the reception of a message has failed.
+
+* ConnectionAborted:
+The connection was aborted by the remote.
```
theri: Nit: "remote endpoint"?

draft-ietf-taps-impl.md
```diff
@@ -963,3 +963,46 @@ If the application indicates a bound on the expected Send or Receive bitrate, an
...
+
+* Timeout:
+Delivery of a message to the remote was not possible after a timeout.
```
theri: Nit: "remote endpoint"?

draft-ietf-taps-impl.md
tfpauly: Agreed. Suggest using "several" or "various".

draft-ietf-taps-impl.md
tfpauly: +1

draft-ietf-taps-impl.md
```diff
@@ -963,3 +963,46 @@ If the application indicates a bound on the expected Send or Receive bitrate, an
...
+* BadConfiguration:
+The transport properties and endpoints provided by the application are either contradictory or incomplete. Examples include the lack of a remote endpoint on an active open or using a multicast group address while not requesting a unidirectional receive.
+
+* NoCandidates:
+None of the available transport protocols can satisfy the transport properties provided by the application.
+
+* Unresolvable:
+The remote or local specifier provided by the application can not be resolved.
+
+* CantEstablish:
```
tfpauly: We shouldn't use contractions in terms like these; "CannotEstablish" would be usable here, but "EstablishmentFailed" may be better?

draft-ietf-taps-impl.md
```diff
@@ -963,3 +963,46 @@ If the application indicates a bound on the expected Send or Receive bitrate, an
...
+* NotCloneable:
+The protocol stack is not capable of being cloned. 
+
+* MessageToLarge:
+The message size is too big for the taps system to handle.
+
+* ProtocolFailure:
+The underlying protocol stack failed. 
+
+* BadMessageProperties:
```
tfpauly: "Invalid" is preferable to "Bad"

draft-ietf-taps-impl.md
```diff
@@ -963,3 +963,46 @@ If the application indicates a bound on the expected Send or Receive bitrate, an
...
+The message properties are either contradictory to the transport properties or they can not be satisfied by the TAPS system. 
+
+* BadReceive:
+The data that was received by the underlying protocol stack could not be fully retrieved by the TAPS system.
```
tfpauly: Overall, please use "Transport Services" rather than "TAPS"

draft-ietf-taps-impl.md
```diff
@@ -963,3 +963,46 @@ If the application indicates a bound on the expected Send or Receive bitrate, an
...
+* MessageToLarge:
+The message size is too big for the taps system to handle.
+
+* ProtocolFailure:
+The underlying protocol stack failed. 
+
+* BadMessageProperties:
+The message properties are either contradictory to the transport properties or they can not be satisfied by the TAPS system. 
+
+* BadReceive:
```
tfpauly: What does this case mean?

tfpauly:
Thanks for working on this!

draft-ietf-taps-impl.md
MaxF12: While the underlying transport was able to receive everything properly, the transport system implementation itself was not. I wanted to keep the two distinct but that might not be necessary.

draft-ietf-taps-impl.md
MaxF12: I wanted to make it clear that this is the other side of the connection and not the remote endpoint object. There might be a better term for this than "remote".

draft-ietf-taps-impl.md
MaxF12: This one was intended to be used for protocol failures during receive and send. It might not be necessary to keep both as ProtocolFailure pretty much covers "CantEstablish".

draft-ietf-taps-impl.md
theri: If not "remote endpoint", maybe "peer"?

draft-ietf-taps-impl.md
theri: That makes sense.
In this particular case, I think "message delivery" already implies that it's delivery to the other side of the connection, so making that explicit is not necessary here.

However, reading the entire description again, I'm wondering if it makes sense to rephrase it entirely.
How about "Message delivery failed due to a timeout"?

draft-ietf-taps-impl.md
philsbln: Does it make sense to split this case between "LocalReceivFailure" und "DeframingFailure"?

theri: Another suggestion: ProtocolReceiveFail and SystemReceiveFail.
Or get rid of the distinction here (it's just a suggestion anyway) and just go with ReceiveFailed, which would be in analogy to ResolutionFailed and EstablishmentFailed.

abrunstrom: I guess we should either have a distinction or not include the example. If we have ReceiveError and the reason is ReceiveFailed I am not sure it provides much extra information? Perhaps ProtocolReceiveFailure and DeframingFailure if we want to include something for ReceiveFailed.

abrunstrom: And perhaps Failed rather than Failure to be consistent across error reasons.

draft-ietf-taps-impl.md
```diff
@@ -1006,6 +1006,46 @@ If the application indicates a bound on the expected Send or Receive bitrate, an
...
+
+* ProtocolFailed:
+The underlying protocol stack failed. 
+
+* InvalidMessageProperties:
+The message properties are either contradictory to the transport properties or they can not be satisfied by the transport system. 
+
+* DeframingFailed:
+The data that was received by the underlying protocol stack was unable to be deframed. 
+
```
abrunstrom: One small nit: "was unable to be deframed" reads a bit odd I think. Suggest to replace by "could not be deframed".

PR #352, opened 2019-10-18
Implementation: mapping back to RFC 8303

### LINKED ISSUE #337: Update implementation mapping text for SCTP
We've updated Implementation mapping details for most protocols with #333, but we didn't tackle SCTP yet.

mwelzl:
Closes #337 (SCTP mapping)
This uses a new approach for such text for TCP, UDP and SCTP.
Also: updates appendix D of interface.

### DIFF
--- a/draft-ietf-taps-impl.md
+++ b/draft-ietf-taps-impl.md
@@ -938,7 +938,7 @@ ConnectionError:
 : TIMEOUT.SCTP or ABORT-EVENT.SCTP.
 
 Listen:
-: LISTEN.SCTP. If an association with the peer already exists and stream mapping has been negotiated, `Listen` just expects to receive a new message on a new stream id (chosen in accordance with the stream number assignment procedure described above) and will return upon reception of this message.
+: LISTEN.SCTP. If an association with the peer already exists and stream mapping has been negotiated, `Listen` just expects to receive a new message on a new stream id (chosen in accordance with the stream number assignment procedure described above).
 
 ConnectionReceived:
 : LISTEN.SCTP returns without an error (a result of successful CONNECT.SCTP from the peer), or, in case of stream mapping, the first message has arrived on a new stream (in this case, `Receive` is also invoked).
--- a/draft-ietf-taps-impl.md
+++ b/draft-ietf-taps-impl.md
@@ -705,7 +705,7 @@ Clone:
 : Calling `Clone` on a TCP Connection creates a new Connection with equivalent parameters. The two Connections are otherwise independent.
 
 Send:
-: SEND.TCP. TCP does not on its own preserve Message boundaries. Calling `Send` on a TCP connection lays out the bytes on the TCP send stream without any other delineation. Any Message marked as Final will cause TCP to send a FIN once the Message has been completely written.
+: SEND.TCP. TCP does not on its own preserve Message boundaries. Calling `Send` on a TCP connection lays out the bytes on the TCP send stream without any other delineation. Any Message marked as Final will cause TCP to send a FIN once the Message has been completely written, by calling CLOSE.TCP immediately upon successful termination of SEND.TCP.
 
 Receive:
 : With RECEIVE.TCP, TCP delivers a stream of bytes without any Message delineation. All data delivered in the `Received` or `ReceivedPartial` event will be part of a single stream-wide Message that is marked Final (unless a MessageFramer is used). EndOfMessage will be delivered when the TCP Connection has received a FIN (CLOSE-EVENT.TCP or ABORT-EVENT.TCP) from the peer.
--- a/draft-ietf-taps-impl.md
+++ b/draft-ietf-taps-impl.md
@@ -651,7 +651,6 @@ The reasonable lifetime for cached performance values will vary depending on the
 # Specific Transport Protocol Considerations
 
 Each protocol that can run as part of a Transport Services implementation defines both its API mapping as well as implementation details.
-
 API mappings for a protocol apply most to Connections in which the given protocol is the "top" of the Protocol Stack. For example, the mapping of the `Send` function for TCP applies to Connections in which the application directly sends over TCP. If HTTP/2 is used on top of TCP, the HTTP/2 mappings take precendence.
 
 Each protocol has a notion of Connectedness. Possible values for Connectedness are:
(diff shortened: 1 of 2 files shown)


draft-ietf-taps-impl.md
```diff
@@ -887,29 +887,82 @@ Connection Object:
...
-a SCTP_PR_SCTP_TTL policy {{!RFC6458}}). However, since multiplexing of
-Connections onto SCTP streams may happen, and would be hidden from the application, the
-Transport System requires a per-stream receiver buffer anyway, so this potential benefit is lost
-and the "partial flag" becomes unnecessary for the system.
-
-The problem of long messages either requiring large receiver-side buffers or getting in the way of
-multiplexing is addressed by message interleaving {{!RFC8260}},
-which is yet another reason why a receivers-side transport system supporting SCTP should
-implement this mechanism.
+\[EDITOR'S NOTE: The paragraph below is placed here because, currently, only the SCTP text follows this convention. However, the suggestion is to move this paragraph up, to appear directly under "Specific Transport Protocol Considerations", and also update TCP and UDP to follow this style.]
```
tfpauly: Can we have this be an issue rather that a note in the markdown? Not a huge deal either way.

I do like the format!

mwelzl: If we're good with the format (waiting for an ACK from @britram), it's more efficient to just leave the editor comment for now and remove it with the follow-up PR (to change UDP and TCP in the same way), which I'll then do fast.

mwelzl: .... and thanks for looking at this so quickly btw!

mwelzl: woops, forgot @abrunstrom  ... adding her too.  Quick answers would be appreciated, so I can continue this method early next week if folks agree.

draft-ietf-taps-impl.md
```diff
@@ -887,29 +887,82 @@ Connection Object:
...
+: `Initiate` or `InitiateWithSend` returns without an error, i.e. SCTP's four-way handshake has completed. If an association with the peer already exists, and stream mapping has been negotiated and enough streams are available, a Connection Object instantly becomes Ready after calling `Initiate` or `InitiateWithSend`.
+
+InitiateError:
+: Failure of CONNECT.SCTP.
+
+ConnectionError:
+: TIMEOUT.SCTP or ABORT-EVENT.SCTP.
+
+Listen:
+: LISTEN.SCTP. If an association with the peer already exists and stream mapping has been negotiated, `Listen` just expects to receive a new message on a new stream id (chosen in accordance with the stream number assignment procedure described above) and will return upon reception of this message.
```
MaxF12: Wouldn't this make the Listen call blocking? Shouldn't the Listener object be returned immediately and then ConnectionReceived be called once the message arrives?

mwelzl: Good catch, thanks! Fixed. There was no need to say "will return upon reception".

JFYI: anyway these primitives are abstract - just a writing style to find the original part of the spec for the functionality. E.g., the text describing "LISTEN.SCTP" in section 4 of RFC 8303 says: "Pass 1 primitive/event: 'Initialize', followed by the 'Communication Up' or 'Restart' notification and possibly the  'Adaptation Layer' notification" - so in fact it's (supposed to be) non-blocking even with the standard SCTP socket interface, under the hood.

mwelzl:
Ah, heck. I'll just extend this now, to move the intro text to where it should be (beginning of "Specific Transport Protocol Considerations" section), removing the in-line "editor's note" comment, and update the text for TCP and UDP as planned. If someone wants me to stop (i.e., not push this to the same branch, which, I believe, would just update this PR), please say so ASAP   :-)

draft-ietf-taps-impl.md
mwelzl: Done

draft-ietf-taps-impl.md
```diff
@@ -678,22 +681,22 @@ Connection Object:
...
-: Calling `Initiate` on a TCP Connection causes it to reserve a local port, and send a SYN to the Remote Endpoint.
+: CONNECT.TCP. Calling `Initiate` on a TCP Connection causes it to reserve a local port, and send a SYN to the Remote Endpoint.
 
 InitiateWithSend:
-: Early idempotent data is sent on a TCP Connection in the SYN, as TCP Fast Open data.
+: CONNECT.TCP with parameter "user message". Early idempotent data is sent on a TCP Connection in the SYN, as TCP Fast Open data.
 
 Ready:
-: A TCP Connection is ready once the three-way handshake is complete.
+: `Initiate` or `InitiateWithSend` returns without an error, i.e. the three-way handshake is complete.
```
tfpauly: I don't quite follow this; Ready is the event that comes back after Initiate or InitiateWithSend to indicate that there was no error; the definition now is a bit tautological. The point of this was to say what the notion of being ready meant, which is just the part about the three-way handshake

draft-ietf-taps-impl.md
```diff
@@ -678,22 +681,22 @@ Connection Object:
...
 
 InitiateError:
-: TCP can throw various errors during connection setup. Specifically, it is important to handle a RST being sent by the peer during the handshake.
+: Failure of CONNECT.TCP. TCP can throw various errors during connection setup. Specifically, it is important to handle a RST being sent by the peer during the handshake.
 
 ConnectionError:
-: Once established, TCP throws errors whenever the connection is disconnected, such as due to receive a RST from the peer; or hitting a TCP retransmission timeout.
+: Once established, TCP throws errors whenever the connection is disconnected, such as due to receive a RST from the peer; or hitting a TCP retransmission timeout. This can lead to ABORT-EVENT.TCP or TIMEOUT.TCP events.
```
tfpauly: This doesn't lead to the ABORT-EVENT.TCP or TIMEOUT.TCP events I think-- rather the ABORT-EVENT.TCP or TIMEOUT.TCP events lead to the ConnectionError event!

draft-ietf-taps-impl.md
mwelzl: Agreed, I'll just remove this new bit again.

draft-ietf-taps-impl.md
mwelzl: Of course, you're right! Thanks, I'll fix this!

PR #358, opened 2019-10-29
Introduce pooled connections as optional implementation detail

### LINKED ISSUE #339: Add text about connection coalescing and connection pooling
This is a follow-up issue for #266 – As we have the ready to support connection coalescing and connection pooling, we have to describe it more closely in the Implementation.

philsbln:
closes #339


draft-ietf-taps-impl.md
```diff
@@ -556,10 +556,19 @@ not using TCP), the action should fail gracefully. The application may be inform
 The Transport Services implementation should allow protocol instances in the Protocol Stack to pass up arbitrary generic or protocol-specific
 errors that can be delivered to the application as Soft Errors. These allow the application to be informed of ICMP errors, and other similar events.
 
+## Pooled Connection {#pooled-connections}
+
+For message reordering resistant request/response protocols like HTTP, interactions can be automatically distributed across several underlying transport connections.
```
tfpauly:
```suggestion
For protocols that require in-order request/response pairs, like HTTP, interactions can be automatically distributed across several underlying transport connections. 
```

philsbln: Thank you for your suggestion – I think you got me wrong, as I wanted to stress that there is no ordering requirement across different requests, but the ordering between request/response  is certainly ordered.
I update the text to make this clearer.

draft-ietf-taps-impl.md
```diff
@@ -556,10 +556,19 @@ not using TCP), the action should fail gracefully. The application may be inform
 The Transport Services implementation should allow protocol instances in the Protocol Stack to pass up arbitrary generic or protocol-specific
 errors that can be delivered to the application as Soft Errors. These allow the application to be informed of ICMP errors, and other similar events.
 
+## Pooled Connection {#pooled-connections}
+
+For protocols that employ request/response pairs and do not require in-order delivery of the responses, like HTTP/2, interactions can be automatically distributed across several underlying transport connections.
```
mwelzl: I would prefer "For application protocols that employ" instead of just "For protocols that employ", to be even less ambiguous.

draft-ietf-taps-impl.md
```diff
@@ -556,10 +556,19 @@ not using TCP), the action should fail gracefully. The application may be inform
...
+If the Protocol Stack includes a transport protocol that supports multipath connectivity, an update to the available paths should inform the Protocol Instance of the new set of paths that are permissible based on the Selection Properties passed by the application. A multipath protocol can establish new subflows over new paths, and should tear down subflows over paths that are no longer available. Pooled Connections {{pooled-connections}} may add or remove underlaying transport connections in a similar manner. If the Protocol Stack includes a transport protocol that does not support multipath, but support migrating between paths, the update to available paths can be used as the trigger to migrating the connection. For protocols that do not support multipath or migration, the Protocol Instances may be informed of the path change, but should not be forcibly disconnected if the previously used path becomes unavailable. An exception to this case is if the System Policy changes to prohibit traffic from the Connection based on its properties, in which case the Protocol Stack should be disconnected.
```
mwelzl: I *believe* (but remember that I'm not a native speaker) that "underlaying" should be "underlying" in this paragraph.

draft-ietf-taps-impl.md
philsbln: I somewhat agree, but as HTTP has become a transport for more or less everything, I prefer being unspecific here.

draft-ietf-taps-impl.md
```diff
@@ -435,7 +435,7 @@ It is useful to process success and failure throughout the tree by child nodes r
...
+If a leaf node has successfully completed its connection, all other attempts should be made ineligible for use by the application for the original request. New connection attempts that involve transmitting data on the network should not be started after another leaf node has completed successfully, as the connection as a whole has been established. An implementation may choose to let certain handshakes and negotiations complete in order to gather metrics to influence future connections. Similarly, an implementation may choose to hold onto fully established leaf nodes that were not the first to establish for use as part of a Pooled Connection, see {{pooled-connections}}, or in future connections. The latter is not recommended since those attempts were slower to connect and may exhibit less desirable properties.
```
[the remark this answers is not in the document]
philsbln: In case of pooled connections, you may have per-message path selection, failover or other mechanisms in place that mitigate these "less desirable properties" - therefore, I don't think we need to recommend against this.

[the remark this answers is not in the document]
theri: Okay, fine with me.

draft-ietf-taps-impl.md
abrunstrom: I think you should remove "automatically". It is the implementation that distributes I guess so it is not automatic?

philsbln: I mean "automatically from perspective of the application" so yes, the automatism has to be realised by the implementation.
If that is unclear, I can add "by the transport system implementation"

abrunstrom: As this is the implementation draft, I think the perspective should be from the implementation, so perhaps "the transport implementation may choose to distribute interactions across several underlying transport connections"?

philsbln: works for me.

draft-ietf-taps-impl.md
tfpauly: I don't think that mentioning the pooled connections here is really of too much value—this sentence is about connections that lost races (using worse addresses or interfaces). Do we really want to recommend using these for pools?

draft-ietf-taps-impl.md
```diff
@@ -556,10 +556,19 @@ not using TCP), the action should fail gracefully. The application may be inform
 The Transport Services implementation should allow protocol instances in the Protocol Stack to pass up arbitrary generic or protocol-specific
 errors that can be delivered to the application as Soft Errors. These allow the application to be informed of ICMP errors, and other similar events.
 
+## Pooled Connection {#pooled-connections}
+
+For protocols that employ request/response pairs and do not require in-order delivery of the responses, like HTTP/2, the transport implementation may distribute interactions across several underlying transport connections.
```
tfpauly: HTTP/2 is not a good case to use multiple underlying TCP/TLS connections... we generally want to multiplex. Or is this just meaning that the HTTP/2 streams themselves are a pool?

draft-ietf-taps-impl.md
philsbln: Depends on the purpose of the pool:
 – If it is to improve reliability, keeping a second worse connection as backup makes sense.
 - If it is to optimise on a per-request basis, there might be one low-latency low-throughput and one high-latency high-throughput interface. The latter will loose the race, but may be more desirable for some requests.
I will update the text up make this clearer.

draft-ietf-taps-impl.md
philsbln: I only thought about the HTTP/2 Interface, not about the protocol itself. Perhaps it would be better to speak of HTTP without version. 
Yes, we want to multiplex, but maybe we want to multiplex over several connections, e.g., using different paths or interfaces (there are other reasons like getting twice the TCP share do not want to advocate for).

tfpauly: Yes, that change I think does help!

draft-ietf-taps-impl.md
```diff
@@ -435,7 +435,7 @@ It is useful to process success and failure throughout the tree by child nodes r
...
+If a leaf node has successfully completed its connection, all other attempts should be made ineligible for use by the application for the original request. New connection attempts that involve transmitting data on the network should not be started after another leaf node has completed successfully, as the connection as a whole has been established. An implementation may choose to let certain handshakes and negotiations complete in order to gather metrics to influence future connections. Similarly, an implementation may choose to hold onto fully established leaf nodes that were not the first to establish for use as part of a Pooled Connection, see {{pooled-connections}}, or in future connections. In both cases, keeping additionalconnection is generally not recommended since those attempts were slower to connect and may exhibit less desirable properties. Still, implemenations that, e.g., realize fallback using connection pools or do per-request path selection, may make use of these.
```
tfpauly:
```suggestion
If a leaf node has successfully completed its connection, all other attempts should be made ineligible for use by the application for the original request. New connection attempts that involve transmitting data on the network should not be started after another leaf node has completed successfully, as the connection as a whole has been established. An implementation may choose to let certain handshakes and negotiations complete in order to gather metrics to influence future connections. Similarly, an implementation may choose to hold onto fully established leaf nodes that were not the first to establish for use as part of a Pooled Connection, see {{pooled-connections}}, or in future connections. In both cases, keeping additional connections is generally not recommended since those attempts were slower to connect and may exhibit less desirable properties. Still, implementations that, for example, realize fallback using connection pools or do per-request path selection, may make use of these.
```

draft-ietf-taps-impl.md
tfpauly: The new text has a couple typos (see other comment), but I still don't totally get the example scenarios described in which this would be wanted. How about just saying implementations may use them, but they're not recommended in general?

philsbln: fine with me

Issue #150, opened 2018-03-21
Add Unidirectional Streams for Multicast / Source and Sink support 


britram:
We need unidirectional connections. You know, for multicast. And ICN. And unidirectional streaming protocols.

britram:
This is a dup of #170, which has more discussion

philsbln:
I am re-opening this ticket as #170 only dealt with unidirectional streams in unicast context (added with PR #191):

Current status:
 - The current model for unidirectional streams supports half-closed connections and QUIC style unidirectional streams so far.
 - We don't know yet whether it is sufficient to have can_read/can_write property on the connection level to model unidirectional connections
 - We don't have pre-connection methods to create multicast sessions yet

philsbln:
I am re-opening this ticket as #170 only dealt with unidirectional streams in unicast context (added with PR #191):

Current status:
 - The current model for unidirectional streams supports half-closed connections and QUIC style unidirectional streams so far.
 - We don't know yet whether it is sufficient to have can_read/can_write property on the connection level to model unidirectional connections
 - We don't have pre-connection methods to create multicast sessions yet

britram:
there's a selection property for this, tagged discuss, which we haven't discussed yet. I'm moving this down to the appendix as part of my restructuring in my work on #208

GrumpyOldTroll:
If help is still wanted here, I'm maybe up for an attempt or 2 at adding multicast support.
Has there been some prior discussion I can review about rough ideas on pre-connection methods for multicast sessions, or should I just try making something up?

MaxF12:
Closed by A#380

PR #360, opened 2019-11-17
udp multicast receive (#150)

GrumpyOldTroll:
https://github.com/ietf-tapswg/api-drafts/issues/150

After some discussion with Max and Theresa, we believe the interface draft does not need any changes.  This adds a UDP Multicast Receive protocol-specific section to outline the specific considerations for a UDP multicast receive connection.

### DIFF
--- a/draft-ietf-taps-impl.md
+++ b/draft-ietf-taps-impl.md
@@ -908,7 +908,7 @@ Data Unit: Datagram
 API mappings for Receiving Multicast UDP are as follows:
 
 Connection Object:
-: UDP Multicast Receive connections represent a pair of specific IP addresses and ports.  The "unidirectional receive" transport property is required, and the local endpoint must be configured with a group IP address and a port.
+: Established UDP Multicast Receive connections represent a pair of specific IP addresses and ports.  The "unidirectional receive" transport property is required, and the local endpoint must be configured with a group IP address and a port.
 
 Initiate:
 : Calling `Initiate` on a UDP Multicast Receive Connection causes an immediate InitiateError.  This is an unsupported operation.
--- a/draft-ietf-taps-impl.md
+++ b/draft-ietf-taps-impl.md
@@ -899,6 +899,53 @@ Close:
 Abort:
 : Calling `Abort` on a UDP Connection (ABORT.UDP(-Lite)) is identical to calling `Close`.
 
+## UDP Multicast Receive
+
+Connectedness: Unconnected
+
+Data Unit: Datagram
+
+API mappings for Receiving Multicast UDP are as follows:
+
+Connection Object:
+: UDP Multicast Receive connections represent a pair of specific IP addresses and ports.  The "unidirectional receive" transport property is required, and the local endpoint must be configured with a group IP address and a port.
+
+Initiate:
+: Calling `Initiate` on a UDP Multicast Receive Connection causes an immediate InitiateError.  This is an unsupported operation.
+
+InitiateWithSend:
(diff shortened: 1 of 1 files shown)


mwelzl:
Of course it's fantastic if we can keep things as simple as that!  ...but I must be missing something.
E.g., from this PR:

> SEND.UDP(-Lite). Calling `Send` on a UDP Multicast Receive connection causes an immediate SendError.  This is an unsupported operation.

So this means that I can create a connection but never use a Send on it. Actually, I can't even use Initiate. Fine... but where in the process do I learn (or, rather, specify and then expect) that this is the case?

philsbln:
You need to specify a selection property requesting a unidirectional/receive connection, therefore you are aware of it when the connection is ready.

mwelzl:
Sure, that's how I'd also do it - but that selection property doesn't exist AFAICT.  So, that's the required change to the API then.

MaxF12:
The way we are doing it is that we check if the address of the local endpoint is a mc group address. So thats where the distinction between unicast and multicast is made. If that is the case we also require the transport properties to include that the direction is unidirection receive.

mwelzl:
Ah, sure. That makes the most sense. Still, that's also not written in the API document. I'm just saying this should be there.

theri:
Well, there is a Connection Property called "Direction of communication" (Section 5.2.13 in the API draft).
Maybe this should be a Selection Property instead? Because those you have to specify before the Connection gets established.

Also, do you think it's worth adding to that section that it's not supported to call Send() on a unidirectional receive Connection or Receive() on a unidirectional send Connection?
Initiate() being unsupported is probably not something that applies to all unidirectional receive Connections, even though I haven't thought about use cases other than multicast yet.

mwelzl:
Yes and yes. "Direction of communication" should be a selection property, as it's unlikely to make sense as something to change later. Also, just because that exists, there's no relationship to multicast in the text - the word "multicast" doesn't exist in the API draft right now. I'm thinking that somewhere, almost literally the statement from @MaxF12 above should be added. Something like this:

If the address of the local endpoint is a multicast group address:

- The "Direction of communication" selection property must be set to unidirectional receive.
- Initiate, Send, ... (list them all here) are not supported.
- (whatever else should be said.)

theri:
Oh, I just realized that "Direction of communication" already is a Selection Property, so we don't need to change anything there.
But yes, it probably makes sense to add a paragraph on multicast to the "Specifying Endpoints" section 5.1 in the API draft.

philsbln:
We did not make it a selection property, because
 a) it is not of Preference type
 b) it should be queryable and probably may change to bidi after establishment if the underlaying protocol is bi-directional
c) it my be settable from bidi to recv to half-close connections.
Yes, this is one of the properties where our categories do not work so great.

draft-ietf-taps-impl.md
```diff
@@ -899,6 +899,53 @@ Close:
...
+## UDP Multicast Receive
+
+Connectedness: Unconnected
+
+Data Unit: Datagram
+
+API mappings for Receiving Multicast UDP are as follows:
+
+Connection Object:
+: Established UDP Multicast Receive connections represent a pair of specific IP addresses and ports.  The "unidirectional receive" transport property is required, and the local endpoint must be configured with a group IP address and a port.
```
philsbln:
```suggestion
: Established UDP Multicast Receive connections represent a pair of specific IP addresses and ports.  The "direction" transport property must be set to "unidirectional receive", and the local endpoint must be configured with a group IP address and a port.
```

mwelzl:
@philsbln - that's ok, then we leave the property type as it is.

PR #368, opened 2019-11-20
be explicit and careful about callbacks vs events; fix #347

### LINKED ISSUE #347: "Events" versus "Callbacks" for security-relevant asynchrony
The TAPS architecture uses Events for all asynchronous happenings, except the trust and challenge callbacks. I think this is the right way to do things but I'm not sure because the callbacks are slightly underspecified at the moment -- callbacks are distinct from events because events are fire-and-forget from the TAPS system's PoV, while callbacks might fail (and the system and transport stacks react to that failure).

This raises a few questions:

1. Should we more explicitly specify callbacks in §5.3.2? (I think yes)
2. Should we add the concept of callbacks to the architecture? (I'm not sure, maybe?)
3. Should we generalize events instead to allow them to have a boolean return? (I think no)

britram:


draft-ietf-taps-arch.md
```diff
@@ -182,12 +182,14 @@ There are key differences between the architecture of the Transport Services sys
 
 Originally, sockets presented a blocking interface for establishing connections and transferring data. However, most modern applications interact with the network asynchronously. When sockets are presented as an asynchronous interface, they generally use a try-and-fail model. If the application wants to read, but data has not yet been received from the peer, the call to read will fail. The application then waits and can try again later.
 
-All interaction with a Transport Services system is expected to be asynchronous, and use an event-driven model unlike sockets {{events}}. For example, if the application wants to read, its call to read will not fail, but will deliver an event containing the received data once it is available.
+All interaction with a Transport Services system is expected to be asynchronous, and use an event-driven model unlike sockets (see {{events}}). For example, if the application wants to read, its call to read will not return immediately, but will deliver an event containing the received data once it is available. Error handling is also asynchronous; a failure to send results in an asynchronous send error as an event.
```
tfpauly: Not sure if we want to say it won't "return immediately"... that's not quite true for async callbacks depending on the language. The function will return immediately, but have a delayed callback.

How about "its request/call to read will not complete immediately"?

draft-ietf-taps-arch.md
```diff
@@ -182,12 +182,14 @@ There are key differences between the architecture of the Transport Services sys
...
+All interaction with a Transport Services system is expected to be asynchronous, and use an event-driven model unlike sockets (see {{events}}). For example, if the application wants to read, its call to read will not return immediately, but will deliver an event containing the received data once it is available. Error handling is also asynchronous; a failure to send results in an asynchronous send error as an event. 
 
 The Transport Services API also delivers events regarding the lifetime of a connection and changes in the available network links, which were not previously made explicit in sockets.
 
 Using asynchronous events allows for a much simpler interaction model when establishing connections and transferring data. Events in time more closely reflect the nature of interactions over networks, as opposed to how sockets represent network resources as file system objects that may be temporarily unavailable.
 
+Separate from events, callbacks are also provided for asynchronous interactions with the API not directly related to events on the network or network interfaces. These are generally used in security association establishment.
```
tfpauly: I don't think callbacks are really all that specific to security associations, right? Even the send/receive events are callbacks conceptually, I think.

britram: Well... yeah, this is a little fuzzy, because from the interface standpoint, events (architectural) might turn into callbacks (implementation), but callbacks (architectural) that will always be implemented as callbacks.

I'll drop the last sentence.

draft-ietf-taps-arch.md
mwelzl: I find "will not return immediately" odd. I understand that you wanted to remove "will not fail", but "will not return immediately" doesn't seem better, as this makes it sound like a blocking call. Suggest: "For example, if the application wants to read, its call to read will schedule the invocation of an event containing the received data once it is available."

draft-ietf-taps-arch.md
philsbln: Send and Recv events can be implemented using callbacks, but can be implemented by other means (co-routines, closures, …) according to your language.
Instead of introducing a more specific concept for security association establishment, I would prefer to call these events too.

draft-ietf-taps-interface.md
```diff
@@ -191,12 +191,12 @@ itself; they are equivalent to Actions on a per-application global context.
...
-The method for dispatching and handling Events is an implementation
-detail, with the caveat that the interface for receiving Messages must require
-the application to invoke the Connection.Receive() Action once per Message to be
-received (see {{receiving}}).
+calls, for instance, and Events could be implemented via event queues, handler
+functions or classes, communicating sequential processes, or other asynchronous
+calling conventions. The method for dispatching and handling Events is an
+implementation detail, with the caveat that the interface for receiving Messages
+must require the application to invoke the Connection.Receive() Action once per
+Message to be received (see {{receiving}}).
```
mwelzl: This just got me confused. I wanted to complain about this saying "once per Message to be received" because I'm quite sure that there was a plan for @tfpauly to write that multiple message receives can be scheduled by just providing a parameter?  Yet, I don't see this in our current text and it's also not an open issue? Have we ended up deciding that it's really a 1-1 relationship after all?  Strange.....

MaxF12: It would be nice to have this, maybe add it to the convenience features section?

mwelzl: I don't know? What confuses me is that I'm quite sure that it was once agreed to change this, in general... I think @tfpauly even wanted it, and we said yes, and I even believe it happened???  Am I looking at the wrong section of the interface draft only, or what?

MaxF12: Its definitely not in any of the drafts or Ive totally overread it as well. Having to call individual receives is one of the most annoying things about the PyTAPS API.

draft-ietf-taps-arch.md
```diff
@@ -183,12 +183,14 @@ There are key differences between the architecture of the Transport Services sys
 
 Originally, sockets presented a blocking interface for establishing connections and transferring data. However, most modern applications interact with the network asynchronously. When sockets are presented as an asynchronous interface, they generally use a try-and-fail model. If the application wants to read, but data has not yet been received from the peer, the call to read will fail. The application then waits and can try again later.
 
-All interaction with a Transport Services system is expected to be asynchronous, and use an event-driven model unlike sockets {{events}}. For example, if the application wants to read, its call to read will not fail, but will deliver an event containing the received data once it is available.
+All interaction with a Transport Services system is expected to be asynchronous, and use an event-driven model unlike sockets (see {{events}}). For example, if the application wants to read, its call to read will not complete immediately, but will deliver an event containing the received data once it is available. Error handling is also asynchronous; a failure to send results in an asynchronous send error as an event.
```
abrunstrom: Suggest to reword first sentence as "Unlike sockets, all interaction with a Transport Services..." , the "unlike sockets" at the end does not read well for me.
Also wondering if "will not complete immediately" should better be "may not complete immediately" as the data may have arrived before the read.

draft-ietf-taps-interface.md
abrunstrom: @mwelzl maybe you are thinking about the similar discussion of a call to Listen returning more than one Connection? Anyway this seems to be a separate discussion.

philsbln: @mwelzl we agreed on this for listen, but not for recv. recv is a little different as it implicitly manages backpressure. Yust call recv in the event handler.

mwelzl: Thanks, @abrunstrom and @philsbln for refreshing my memory!  Then I agree, it's a separate issue - but perhaps worth opening an issue for. I'll do that (and include a reminder that this particular text would have to be adapted along with the interface update in case we make this change).

mwelzl: Sorry, I just changed my mind: I'm *not* opening an issue. This would complicate things (due to questions related to parameters), and I'm not going to add more complications. I might approve them at best  :-)   if someone else asks for it...

Issue #303, opened 2019-03-27
More expressive multipath transport property


abrunstrom:
The current multipath selection property is an on/off selection.  We will eventually need something more expressive here to distinguish for instance between handover/back-up and aggregate multipath modes.

tfpauly:
For reference, the Network.framework multi path property values are:

```
	/// Multipath services represent the modes of multipath usage that are
	/// allowed for connections.
	public enum MultipathServiceType {
		/// No multipath transport will be attempted
		case disabled
		/// Only use the expensive interface when the when the primary one is not available
		case handover
		/// Use the expensive interface more aggressively to reduce latency
		case interactive
		/// Use all available interfaces to provide the highest throughput and lowest latency
		case aggregate
	}
```

philsbln:
If I remember correctly, we had something like this that got turned into a regular preference at some point in time. I agree that this expressiveness is useful, but can't this be achieved even better by also setting CostPreference + CapacityProfile?

PR #370, opened 2019-11-20
Add multipath property enumeration

tfpauly:
Complete #303

### DIFF
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -896,16 +896,16 @@ multiple paths allows connections to migrate between interfaces as
 availability and performance properties change. Possible values are:
 
 Disabled:
-: The connection will not support using multiple paths once established
+: The connection will not attempt using multiple paths once established
 
 Handover:
-: The connection will support migrating between different paths upon interface availability changes
+: The connection should attempt to migrate between different paths upon interface availability changes
 
 Interactive:
-: The connection will support using multiple paths in response to loss or delay upon individual paths
+: The connection should attempt to use multiple paths in response to loss or delay upon individual paths
 
 Aggregate:
-: The connection will support using multiple paths in parallel in order to maximize bandwidth
+: The connection should attempt to use multiple paths in parallel in order to maximize bandwidth
 
 The default is Disabled.
 
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -907,7 +907,7 @@ Interactive:
 Aggregate:
 : The connection should attempt to use multiple paths in parallel in order to maximize bandwidth
 
-The default is Disabled.
+The default is Disabled. The other enumeration values are all interpreted as preferences.
 
 ### Direction of communication
 
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -887,12 +887,27 @@ over these options.
 Name:
 : multipath
 
-This property specifies whether an application considers it useful to
-transfer data across multiple paths between the same end hosts. Generally,
-in most cases, this will improve performance (e.g., achieve greater throughput).
-One possible side-effect is increased jitter, which may be problematic for
-delay-sensitive applications. The recommended default is to Ignore this option.
+Type:
+: Enumeration
 
+This property specifies whether an application wants to take advantage of
+transferring data across multiple paths between the same end hosts. Using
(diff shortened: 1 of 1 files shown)


draft-ietf-taps-interface.md
```diff
@@ -887,12 +887,27 @@ over these options.
...
+: The connection will not support using multiple paths once established
+
+Handover:
+: The connection will support migrating between different paths upon interface availability changes
+
+Interactive:
+: The connection will support using multiple paths in response to loss or delay upon individual paths
+
+Aggregate:
+: The connection will support using multiple paths in parallel in order to maximize bandwidth
```
mwelzl: "will" makes these choices sound like promises, when they are only expressing a wish by the application that may or may not be granted. I'd like to request changing the three instances of "The connection will support ..."  above into "The connection should support ...". The only promise we can keep is the Disabled choice, "will not support" is fine.

tfpauly: Good call, fixed! Please take a look.

mwelzl: Thanks for fixing this!

draft-ietf-taps-interface.md
```diff
@@ -887,12 +887,27 @@ over these options.
 Name:
 : multipath
 
-This property specifies whether an application considers it useful to
-transfer data across multiple paths between the same end hosts. Generally,
-in most cases, this will improve performance (e.g., achieve greater throughput).
-One possible side-effect is increased jitter, which may be problematic for
-delay-sensitive applications. The recommended default is to Ignore this option.
+Type:
+: Enumeration
```
philsbln: This removes the possibility to prefer vs require multipath functionality.
I would rather make it a set of tuples like the Interface type:
```suggestion
:  Set (Preference, Enumeration)
```

draft-ietf-taps-interface.md
```diff
@@ -887,12 +887,27 @@ over these options.
...
+
+This property specifies whether an application wants to take advantage of
+transferring data across multiple paths between the same end hosts. Using
+multiple paths allows connections to migrate between interfaces as
+availability and performance properties change. Possible values are:
+
+Disabled:
+: The connection will not attempt using multiple paths once established
```
philsbln: If we do the above, we can remove the disabled state

draft-ietf-taps-interface.md
```diff
@@ -887,12 +887,27 @@ over these options.
...
+Handover:
+: The connection should attempt to migrate between different paths upon interface availability changes
+
+Interactive:
+: The connection should attempt to use multiple paths in response to loss or delay upon individual paths
+
+Aggregate:
+: The connection should attempt to use multiple paths in parallel in order to maximize bandwidth
 
+The default is Disabled.
```
philsbln: … and need to change the default:
```suggestion
The default is to prohibit aggregate and ignore the other multipath functionalities.
```

draft-ietf-taps-interface.md
abrunstrom: Yes, I think this will also be more consistent.

draft-ietf-taps-interface.md
abrunstrom: How will I then turn it off? By setting prohibit for all other alternatives? That does not seem right. I think we would still need Disabled.

draft-ietf-taps-interface.md
```diff
@@ -887,12 +887,27 @@ over these options.
...
+
+Disabled:
+: The connection will not attempt using multiple paths once established
+
+Handover:
+: The connection should attempt to migrate between different paths upon interface availability changes
+
+Interactive:
+: The connection should attempt to use multiple paths in response to loss or delay upon individual paths
+
```
abrunstrom: This one is a bit fuzzy, particularly "delay". Does it mean "delay spikes" or the base delay is over a given delay target? But perhaps it is hard to be precise? I think this will vary from implementation to implementation.

draft-ietf-taps-interface.md
philsbln: That is a design choice we can make, but then we should be explicit abut it. Maybe "If it is set to something except disable, a taps system should prefer protocols that provide the chosen mode of operation.".

draft-ietf-taps-interface.md
britram: The proposed default here surfaces the complexity that Set { Preference, Enumeration } adds. Is this additional complexity useful?

Issue #300, opened 2019-03-25
Add property for address privacy


tfpauly:
API should allow a preconnection property to specify if a private (IPv6) address is preferred. Outbound connections should prefer this. Listeners should avoid this.

PR #372, opened 2019-11-20
Address privacy property, #300

tfpauly:
Address #300

### DIFF
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -882,6 +882,29 @@ requiring specific PvD instances or interface instances, and should be preferred
 over these options.
 
 
+### Local Address Preference
+
+Name:
+: local-address-preference
+
+Type:
+: Enumeration
+
+This property specifies whether Listeners and Connections should prefer the use of temporary addresses when possible.
+This is generally used to prevent linking connections over time when a stable address is not needed. Possible values are:
+
+Stable:
+: Prefer the use of stable (sometimes called "permanent") local addresses
+
+Temporary:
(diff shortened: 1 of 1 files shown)


draft-ietf-taps-interface.md
```diff
@@ -882,6 +882,29 @@ requiring specific PvD instances or interface instances, and should be preferred
...
+: Enumeration
+
+This property specifies whether Listeners and Connections should prefer the use of temporary addresses when possible.
+This is generally used to prevent linking connections over time when a stable address is not needed. Possible values are:
+
+Stable:
+: Prefer the use of stable (sometimes called "permanent") local addresses
+
+Temporary:
+: Prefere the use of temporary (sometimes called "privacy") addresses {{!RFC4941}}
```
mwelzl: typo s/Prefere/Prefer

draft-ietf-taps-interface.md
```diff
@@ -882,6 +882,29 @@ requiring specific PvD instances or interface instances, and should be preferred
...
+
+Default:
+: Prefer the use of stable local addresses for Listeners and Rendezvous Connections, and prefer the use of temporary addresses for other Connections.
+
+The default is Default, which adapts based on use case.
```
mwelzl: I would remove this sentence - obviously "Default" is the default. We should generally avoid this repetition whenever we have "Default" as a value choice IMO.

draft-ietf-taps-interface.md
tfpauly: Fixed!

draft-ietf-taps-interface.md
tfpauly: Removed!

PR #380, opened 2019-11-21
Add clarification on Multicast receive to API

MaxF12:


draft-ietf-taps-interface.md
```diff
@@ -623,6 +623,8 @@ The Resolve() action on Preconnection can be used by the application to force
 early binding when required, for example with some Network Address Translator
 (NAT) traversal protocols (see {{rendezvous}}).
 
+Specifying a multicast group address on the Local Endpoint will indicate to the transport system that the resulting connection will be used to receive multicast messages. This will restrict the application to establishing the Preconnection by calling Listen(). The application is unable to send messages on any resulting connections. 
+
```
mwelzl: suggest to replace:
This will restrict the application to establishing the Preconnection by calling Listen(). The application is unable to send messages on any resulting connections. 

with:
This will restrict the application to establishing the Preconnection by calling Listen(). The application will be unable to send messages on any resulting Connections, which will have the Transport Property "Direction of communication" set to "Unidirectional receive".

Also, as discussed, there should be a sentence here about the function of the specified remote endpoint as a filter.

mwelzl: Just to explain, I think @tfpauly's concern was that the previous text sounded like requiring extra work ("the application must set ...") - but I think you made this a little *too* implicit  :)  That is, I think it's fine to point at the "unidirectional" property choice, but not phrase it as something the application must take care of.

MaxF12: Fixed!

mwelzl: Just a typo left: conenction

MaxF12: Whoops, fixed thanks!

philsbln:
Maybe add the ASN source address (set source address), source filter (add remote addresses)?
Maybe also add that for sending, you specify the group as remote and the socket becomes write-only?

Should we require specifying direction? This would be slightly harder to use, but future proof in case someone implements reliable ASN on TAPS…

draft-ietf-taps-interface.md
abrunstrom: For consistency, should we perhaps add to the last sentence: "and the Transport Property "Direction of communication" will be set to "Unidirectional send".

PR #387, opened 2019-11-25
Fix #385: Resolution is key to path and stack agility. Say so.

### LINKED ISSUE #385: Clarify that more flexible resolution is an architectural feature
Especially in section 2, we should make it explicit that resolution is blended with connection establishment in a way that it is not in traditional sockets.

britram:


britram:
This is a lightweight proposal to address #385.

draft-ietf-taps-arch.md
```diff
@@ -126,32 +126,37 @@ they appear in all capitals, as shown here.
...
+        |                 |                  |
+  +------------+     +------------+    +--------------+
+  |    stub    |     | Stream API |    | Datagram API |
+  |  resolver  |     +------------+    +--------------+
+  +------------+          |                  |
+        |           +---------------------------------+
+        |           |     TCP                UDP      |
+        |           |  Kernel Protocol Implementation |
+        |           +---------------------------------+
+        |                           |
```
abrunstrom: This looks strange to me. It looks like the stub resolver bypasses the kernel protocol implementations. It should also rather run on top.

draft-ietf-taps-arch.md
```diff
@@ -160,12 +165,12 @@ The Transport Services architecture maintains this general model of interaction,
 +-----------------------------------------------------+
                           |
 +-----------------------------------------------------+
-|               Transport Services API                |
+|   Transport Services API (resolution + connection)  |
 +-----------------------------------------------------+
                           |
```
abrunstrom: Also do not like the "(resolution + connection)". What about sending and receiving data? Also looks like the API exposes resolution and connection as separate entities rather than the opposite.

draft-ietf-taps-arch.md
```diff
@@ -126,32 +126,37 @@ they appear in all capitals, as shown here.
...
+        |                           |
 +-----------------------------------------------------+
 |               Network Layer Interface               |
 +-----------------------------------------------------+
 
 ~~~~~~~~~~
 {: #fig-sockets title="socket() API Model"}
 
-The Transport Services architecture maintains this general model of interaction, but aims to both modernize the API surface exposed for transport protocols and enrich the capabilities of the transport system implementation.
+The Transport Services architecture maintains this general model of interaction, but aims to both modernize the API surface exposed for transport protocols and enrich the capabilities of the transport system implementation. It combines interfaces for multiple interaction patterns into a unified whole. By combining identifier address resoulution with connection establishment and data transfer in a single API, it allows for more flexible implementations to provide path and transport protocol agility on the application's behalf.
```
abrunstrom: Reading this again now and seeing the figures, I wonder if 

> The Transport Services architecture maintains this general model of interaction, but aims to both modernize the API surface 

may lead the reader to thinking that the Transport System Implementation runs in the kernel? Separate issue I guess, but a bit realted to the bypass comment above.

draft-ietf-taps-arch.md
britram: Yep, tweaked this.

draft-ietf-taps-arch.md
britram: Yep, have reverted this, it's clear in contrast with the previous figure.

draft-ietf-taps-arch.md
britram: I tweaked this language slightly, but would point out that nothing in the *architecture* precludes a kernel TAPS implementation.

draft-ietf-taps-arch.md
philsbln: still looks strange… is it not rather that way:
```suggestion
        |                  |                  |
  +------------+     +------------+    +--------------+
  |    stub    |     | Stream API |    | Datagram API |
  |  resolver  |     +------------+    +--------------+
  +------------+           |                  |
  |    DNS.    |           |                  |
  +------------+--------------------------------------+
  |                       TCP                UDP      |
  |              Kernel Networking Stack              |
  +---------------------------------------------------+```

Comment (tfpauly) [reply draft-ietf-taps-arch.md]: Agreed with @philsbln that the DNS resolver doesn't belong in the kernel.

How about just this:

```
        |                 |                  |
   +------------+     +------------+    +--------------+
   |    stub    |     | Stream API |    | Datagram API |
   |  resolver  |     +------------+    +--------------+
   +------------+         |                  |
                     +---------------------------------+
                     |    TCP                UDP       |
                     |    Kernel Networking Stack      |
                     +---------------------------------+
                                      |
```

File: draft-ietf-taps-arch.md [hunk 4]
```diff
@@ -126,32 +126,37 @@ they appear in all capitals, as shown here.
...
+        |                           |
 +-----------------------------------------------------+
 |               Network Layer Interface               |
 +-----------------------------------------------------+
 
 ~~~~~~~~~~
 {: #fig-sockets title="socket() API Model"}
 
-The Transport Services architecture maintains this general model of interaction, but aims to both modernize the API surface exposed for transport protocols and enrich the capabilities of the transport system implementation.
+The Transport Services architecture evolves this general model of interaction, aiming to both modernize the API surface presented to applications by the transport layer and enrich the capabilities of the transport system implementation. It combines interfaces for multiple interaction patterns into a unified whole. By combining identifier address resoulution with connection establishment and data transfer in a single API, it allows for more flexible implementations to provide path and transport protocol agility on the application's behalf.
```
tfpauly: Do we need `identifier address resolution` or can this just be `name resolution`?

Also, typo on `resoulution`.

draft-ietf-taps-arch.md
philsbln: works for me either way

draft-ietf-taps-arch.md
```diff
@@ -126,32 +126,37 @@ they appear in all capitals, as shown here.
...
+        |                 |                  |
+  +------------+     +------------+    +--------------+
+  |    stub    |     | Stream API |    | Datagram API |
+  |  resolver  |     +------------+    +--------------+
+  +------------+          |                  |
+                    +---------------------------------+
+                    |    TCP                UDP       |
+                    |    Kernel Networking Stack      |
+                    +---------------------------------+
+                                    |
```
abrunstrom: I think this is much better than the first suggestion, but is there a reason the box for the kernel networking stack does not expand also under the stub resolver? (But leaving DNS out.) This would seem most natural to me, but if the rest of you do not like that we can leave it as is.

PR #391, opened 2019-12-03
minset appendix update: section references

mwelzl:
This also addresses two inline [EDITOR's NOTE] items.

One was about the per-message override capacity profile being inconsistent with the other one; this was solved by removing details, such that the other one defines the possible choices.

Another was about selection vs. connection properties, with broken text. There was no need to add the references as the editor's note says, I thought.

### DIFF
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -943,7 +943,7 @@ Default:
 : Prefer the use of stable local addresses for Listeners and Rendezvous Connections, and prefer the use of temporary addresses for other Connections.
 
 
-### Parallel Use of Multiple Paths  {#parallel-multipath}
+### Parallel Use of Multiple Paths {#parallel-multipath}
 
 Name:
 : multipath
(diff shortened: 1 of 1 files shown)


draft-ietf-taps-interface.md
```diff
@@ -943,7 +943,7 @@ Default:
 : Prefer the use of stable local addresses for Listeners and Rendezvous Connections, and prefer the use of temporary addresses for other Connections.
 
 
-### Parallel Use of Multiple Paths
+### Parallel Use of Multiple Paths  {#parallel-multipath}
```
tfpauly:
```suggestion
### Parallel Use of Multiple Paths {#parallel-multipath}
```

draft-ietf-taps-interface.md
```diff
@@ -1671,22 +1671,6 @@ This enumerated property specifies the application's preferred tradeoffs for
 sending this Message; it is a per-Message override of the Capacity Profile
 protocol and path selection property (see {{prop-cap-profile}}).
```
abrunstrom: protocol and path selection --> connection (as this property is a connection property)

mwelzl: Fixing! I don't know how this "protocol and" slipped in there - good catch, thanks!

PR #393, opened 2019-12-06
Few very minor architecture nits

MaxF12:


draft-ietf-taps-arch.md
```diff
@@ -130,7 +130,7 @@ The traditional model of using sockets for networking can be represented as foll
   (typically implemented in the system's kernel).
 - TCP and UDP in the kernel send and receive data over the available network
   layer interfaces.
-- Sockets are bound directly to transport-layer and network-layer addresses,
+- Sockets are bound directly to transport layer and network layer addresses,
```
tfpauly: Why no hyphens here?

MaxF12: It was just for consistentcy, some times hyphend were used and sometimes they were not. I am used to the version without them, if its more common to use them I can change it to that.

tfpauly: Cases where we say "the transport layer" should not be hyphenated, which is the case. However, if you use the term as an adjective, such as "a transport-layer address", you need to add a hyphen.

The one error I see in the document by those rules is just that "available network layer interfaces" should be "available network-layer interfaces".

MaxF12: Oh of course, sorry my bad. Thanks for the explanation, fixed!

PR #394, opened 2019-12-17
Address Arch NiTS

### LINKED ISSUE #392: A surprisingly few number of NiTS in Arch
I just read Arch and have the following NiTs which I think we should address:

/Listeners by default register with multiple paths/
ought this to be:
/Listeners, by default, register with multiple paths/

/A Framer is a data translation layer that can be added to a Connection to define how application-level Messages are transmitted over a transport protocol. /
- I think this should be (because I don’t think they are protocol specific)  
/A Framer is a data translation layer that can be added to a Connection to define how application-level Messages are transmitted over the transport service. /
- 
/to retransmit in the event of packet loss,/
- better to avoid event here :-)
/to retransmit when there is packet loss,/

philsbln:
Closed #392


draft-ietf-taps-arch.md
```diff
@@ -376,7 +373,7 @@ The diagram below provides a high-level view of the actions and events during th
...
 
-* Listen: The action of marking a Listener as willing to accept incoming Connections. The Listener will then create Connection objects as incoming connections are accepted ({{events}}). Listeners by default register with multiple paths, protocols, and local endpoints, unless constrained by Selection Properties. Connections can be accepted on any of the available paths or endpoints.
+* Listen: The action of marking a Listener as willing to accept incoming Connections. The Listener will then create Connection objects as incoming connections are accepted ({{events}}). Listeners, by default, register with multiple paths, protocols, and local endpoints, unless constrained by Selection Properties. Connections can be accepted on any of the available paths or endpoints.
```
tfpauly: I don't think I agree with adding the commas here. Can we revert this change?

philsbln: I was quite sure they are optional and improve readability, but can remove them.

PR #402, opened 2020-01-01
Add some details to framers

### LINKED ISSUE #340: Comments on new framers
Just some points about the new text on framers in the API draft:

- In 10.2 it is specified in which order multiple framers run on incoming and outgoing messages, it is however never specified what the order looks like on Start and Stop events.
- In 10.6, "MessageFramer.DeliverAndAdvanceReceiveCursor" appears to be missing the "Data" argument that is present in "MessageFramer.Deliver".
- Also in 10.6, in the second paragraph it says:
> The application requests a specific amount of data it needs to have available in order to parse. 

When it is probably meant to say 

> The framer implementation requests a specific amount of data it needs to have available in order to parse. 

- While the framer implementation is able to notify the Connection that it has finished handling a Start event (by calling MessageFramer.MakeConnectionReady), there is no such way for Stop events.

MaxF12:
Closes #340

### DIFF
--- a/draft-ietf-taps-impl.md
+++ b/draft-ietf-taps-impl.md
@@ -550,7 +550,7 @@ Most Message Framers fall into one of two categories:
 - Delimiter-separated formats, such as HTTP/1.1.
 
 Common Message Framers can be provided by the Transport Services implementation,
-but an implemention ought to allow custom Message Framers to be defined by
+but an implementation ought to allow custom Message Framers to be defined by
 the application or some other piece of software. This section describes one
 possible interface for defining Message Framers as an example.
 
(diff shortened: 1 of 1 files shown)


draft-ietf-taps-impl.md
```diff
@@ -585,6 +585,20 @@ remote endpoint that can be used to parse Messages.
 MessageFramer.MakeConnectionReady(Connection)
 ~~~
 
+Alternatively, if the framer implementation deems the candidate selected during racing unsuitable, it can signal this to the Connection. 
+
+~~~
+MessageFramer.FailCandidate(Connection)
```
tfpauly: Two things here:
1. This should take an error, like `FailConnection`
2. I think this doesn't need to be distinct from `FailConnection` overall. If the place where the framer is running is during candidate racing, then its failing its view of the connection is only failing one candidate. I don't know if we should have the term "candidate" bubbling up into the API.

MaxF12: Good call, really no need to have it as a distinct call!

PR #403, opened 2020-01-13
Some small proposed fixes

mwelzl:
- Example code updates: this is to try to avoid people stumbling over, e.g., "send" being part of the "main code". I believe it's a little cleaner with these added comments.

- Lifetime Message property: mentioned -1 as a special value for infinite, else it's a bit odd to only use "infinite" for an integer

- very minor text fix about corruption protection length Message property

### DIFF
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -343,10 +343,12 @@ Connection -> Received(messageDataRequest, messageContext)
 //---- Receive event handler begin ----
 Connection.Send(messageDataResponse)
 Connection.Close()
-//---- Receive event handler end ----
 
 // Stop listening for incoming Connections
+// (this example supports only one Connection)
 Listener.Stop()
+//---- Receive event handler end ----
+
 ~~~
 
 
@@ -1525,17 +1527,17 @@ Name:
 : msg-lifetime
 
 Type:
-: Integer
+: Numeric
 
 Default:
-: infinite (a special value, e.g. -1)
+: infinite
 
 Lifetime specifies how long a particular Message can wait to be sent to the
 remote endpoint before it is irrelevant and no longer needs to be
 (re-)transmitted. This is a hint to the transport system -- it is not guaranteed
 that a Message will not be sent when its Lifetime has expired.
 
-Setting a Message's Lifetime to infinite (e.g., -1) indicates that the application does
+Setting a Message's Lifetime to infinite indicates that the application does
 not wish to apply a time constraint on the transmission of the Message, but it does not express a need for
 reliable delivery; reliability is adjustable per Message via the "Reliable Data Transfer (Message)"
 property (see {{msg-reliable-message}}). The type and units of Lifetime are implementation-specific.
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -335,14 +335,15 @@ Listener := Preconnection.Listen()
 
 Listener -> ConnectionReceived<Connection>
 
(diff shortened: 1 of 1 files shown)


draft-ietf-taps-interface.md
```diff
@@ -335,14 +335,15 @@ Listener := Preconnection.Listen()
 
 Listener -> ConnectionReceived<Connection>
 
-// Only receive complete messages
+// Only receive complete messages in a Conn.Received handler
 Connection.Receive()
 
 Connection -> Received(messageDataRequest, messageContext)
 
+//---- Receive event handler begin ----
```
tfpauly: Technically, isn't everything after this line part of the receive handler? That also includes stopping listening—or at least listen stop shouldn't be called until ConnectionReceived fires.

draft-ietf-taps-interface.md
```diff
@@ -1521,14 +1528,14 @@ Type:
 : Integer
 
 Default:
-: infinite
+: infinite (a special value, e.g. -1)
```
tfpauly: Do we need to specify an integer value like this at all?

draft-ietf-taps-interface.md
mwelzl: I thought about this and wasn't sure; I guess yes, since this code assumes only a single connection, it's easiest to put it inside the receive handler. Maybe we should change the comment line to say:
// Stop listening for incoming Connections (this example supports only one Connection)

draft-ietf-taps-interface.md
mwelzl: Ah, I agree, changing this to "Numeric" is a better solution!

draft-ietf-taps-interface.md
theri: Yes, I guess it's good to say explicitly that this example receives a single connection and then stops listening.

I think it'd make sense to adjust the text above the example ("This is an example of how an application might listen for incoming Connections […]" --> "This is an example of how an application might listen for an incoming Connection").

To make it clearer that the handler is a separate piece of code, we could put it into a separate code snippet, above which we say something like "This is an example of a receive handler which gets called when an incoming Connection is received, reads a Message, sends a response, and then stops listening for incoming Connections".

draft-ietf-taps-interface.md
```diff
@@ -379,13 +380,16 @@ Connection := Preconnection.Initiate()
 
 Connection -> Ready<>
 
+//---- Ready event handler begin ----
```
theri: Again, here we could use separate snippets of code to make it clearer that this is separate code, which gets called when the event happens.

draft-ietf-taps-interface.md
mwelzl: I thought about this too, but on the positive side, the example as it stands also reflects the sequence in which things happen - even if the code being part of the main code isn't quite right. My suggestion: I fix it as @tfpauly suggested and also incorporate your updated comment, in this PR, and once this lands, you can fine-tune it further with your suggestion to split the code out and we can see what folks think?

draft-ietf-taps-interface.md
mwelzl: Suggested approach as above

draft-ietf-taps-interface.md
theri: Fine with me :)

mwelzl:
Okay, so I think I addressed @tfpauly's comments here, and agreed with @theri that we could adjust things further later on if needed - so now I'll just merge this one, seems safe to do to me.

PR #407, opened 2020-01-14
API draft: Replace "Protocol Properties" with updated terms

theri:
When reading the API draft, I noticed that in some places it was still using the outdated term "Protocol Properties" where it actually means Connection Properties or Message Properties.
I replaced the terms in this PR.

We do still define "Specific Protocol Properties", so I kept these.

### DIFF
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -1309,8 +1309,9 @@ Calling Clone on a Connection yields a group of two Connections: the parent
 Connection on which Clone was called, and the resulting cloned Connection. These
 connections are "entangled" with each other, and become part of a Connection
 Group. Calling Clone on any of these two Connections adds a third Connection to
-the Connection Group, and so on. Connections in a Connection Group share all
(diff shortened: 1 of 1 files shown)


draft-ietf-taps-interface.md
```diff
@@ -1310,7 +1310,7 @@ Connection on which Clone was called, and the resulting cloned Connection. These
 connections are "entangled" with each other, and become part of a Connection
 Group. Calling Clone on any of these two Connections adds a third Connection to
 the Connection Group, and so on. Connections in a Connection Group share all
-Protocol Properties that are not applicable to a Message.
+Connection Properties except "Priority (Connection)", see {{conn-priority}}.
```
philsbln: I would prefer to keep this generic, as later documents may add Connection Properties.

theri: I made the text less exclusive, so now "Priority (Connection)" is an example for a non-shared Connection Property, but the text no longer implies that it is the only one.

PR #412, opened 2020-01-15
Add one sentence about fallback in sec 2.3

mirjak:
Just a proposal...


draft-ietf-taps-arch.md
```diff
@@ -214,7 +214,7 @@ Sockets, for protocols like TCP, are generally limited to connecting to a single
...
 
-Flexibility after connection establishment is also important. Transport protocols that can migrate between multiple network-layer interfaces need to be able to process and react to interface changes. Protocols that support multiple application-layer streams need to support initiating and receiving new streams using existing connections.
+Flexibility after connection establishment is also important. Transport protocols that can migrate between multiple network-layer interfaces need to be able to process and react to interface changes. Protocols that support multiple application-layer streams need to support initiating and receiving new streams using existing connections. Transport services implementations can further implement fallback mechanisms if connection establishment of one protocol fails or performance is detected to be unsatisfactory.
```
tfpauly: I like the sentence, but I'm not sure that it fits in this paragraph. This new sentence mentions a protocol failing establishment, but the paragraph begins `Flexibility after connection establishment...`.

Rather, this feels like it belongs with the previous paragraph, `Transport Services implementations are meant to be flexible at connection establishment time...`

mirjak: Done.

draft-ietf-taps-arch.md
```diff
@@ -212,7 +212,7 @@ Allowing applications to interact with messages is backwards-compatible with exi
...
 
-Transport Services implementations are meant to be flexible at connection establishment time, considering many different options and trying to select the most optimal combinations ({{gathering}} and {{racing}}). This requires applications to provide higher-level endpoints than IP addresses, such as hostnames and URLs, which are used by a Transport Services implementation for resolution, path selection, and racing.
+Transport Services implementations are meant to be flexible at connection establishment time, considering many different options and trying to select the most optimal combinations ({{gathering}} and {{racing}}). This requires applications to provide higher-level endpoints than IP addresses, such as hostnames and URLs, which are used by a Transport Services implementation for resolution, path selection, and racing. Transport services implementations can further implement fallback mechanisms if connection establishment of one protocol fails or performance is detected to be unsatisfactory.
```
abrunstrom: I find the "can further implement" a bit confusing as it sound like it is an additional thing, but I guess it is a clarification as racing gives you fallback? So maybe "In this way, Transport Services implementations can provide..." or something similar

philsbln: I see both options: Using racing in first place, but having an alt-srv like mechanism to discover more candidates or use transport level redirects.

mirjak: Do I need to change something now or not?

abrunstrom: I still think something needs to change. Indicating that "fallback mechanisms if connection establishment of one protocol fails" is something different from racing is not correct I think and not consistent with how racing is described elsewhere. But the sentence describes two things. The first to me is a clarification of what racing provides. The second "fallback if performance is detected to be unsatisfactory" is different and could perhaps be part of the general description that is given before, but could also be an add on. Perhaps just split the parts up?

PR #421, opened 2020-01-15
Proposed alternative intro text for API doc

mirjak:
I think the intro currently doesn't fit well the content of the document as any motivation for this work should be in the Arch doc. If we want to retain the text that I deleted we can maybe add it to the Intro or background section in the Arch doc...?

Note that the second paragraph I've added is stolen from the intro of the Arch doc.

### DIFF
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -122,25 +122,24 @@ and potential transport protocols to select from.
 
 # Introduction
 
-The BSD Unix Sockets API's SOCK_STREAM abstraction, by bringing network sockets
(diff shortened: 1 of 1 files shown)


mwelzl:
@theri I can't see how the existing text that @mirjak proposed to delete could reasonably be merged with the intro text of the architecture doc? and the text that's there now motivates TAPS just as nicely, IMO. I would be inclined to just delete the text and do the update exactly as @mirjak proposes here.

abrunstrom:
I agree with @mwelzl , I like the new text much better and I do not think we should add the removed text to arch.

mirjak:
I think the text did its duty and it's time to let go...

PR #424, opened 2020-01-15
Protocol Specific Properties are not well defined

mirjak:
The term Protocol Specific Properties is not well introduced. In the Arch doc we have a section about "Specialized Features" and I not added a reference and rephrased the sentence respectively. Do we need to say more than that?

### DIFF
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -463,8 +463,8 @@ earlier stages and querying them in later stages:
 Note that configuring Connection Properties and Message Properties on
 Preconnections is preferred over setting them later. Early specification of
 Connection Properties allows their use as additional input to the selection
-process. Protocol Specific Properties, that enable configuration of Specialized
-Features of a specific protocol, see Section 3.2 of {{I-D.ietf-taps-arch}}, are not
+process. Protocol Specific Properties, which enable configuration of specialized
+features of a specific protocol, see Section 3.2 of {{I-D.ietf-taps-arch}}, are not
 used as an input to the selection process but only support configuration if
 the respective prototocol has been selected.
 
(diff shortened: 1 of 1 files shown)


draft-ietf-taps-interface.md
```diff
@@ -460,11 +460,13 @@ earlier stages and querying them in later stages:
...
 - The effect of Selection Properties can be queried on Connections and Messages
 
-Note that Configuring Connection Properties and Message Properties on
+Note that configuring Connection Properties and Message Properties on
 Preconnections is preferred over setting them later. Early specification of
 Connection Properties allows their use as additional input to the selection
-process. Protocol Specific Properties, see {{property-names}}, should not be
-used as an input to the selection process.
+process. Protocol Specific Properties, that enable configuration of Specialized
+Features of a specific protocol, see sec 3.2 of {{I-D.ietf-taps-arch}}, are not
```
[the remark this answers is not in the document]

draft-ietf-taps-interface.md
```diff
@@ -460,11 +460,13 @@ earlier stages and querying them in later stages:
 - Message Properties can be set on Preconnections and Connections
 - The effect of Selection Properties can be queried on Connections and Messages
 
-Note that Configuring Connection Properties and Message Properties on
+Note that configuring Connection Properties and Message Properties on
 Preconnections is preferred over setting them later. Early specification of
 Connection Properties allows their use as additional input to the selection
-process. Protocol Specific Properties, see {{property-names}}, should not be
-used as an input to the selection process.
+process. Protocol Specific Properties, that enable configuration of Specialized
```
mwelzl: I think it should be "which enable" instead of "that enable" here, because this adds information Protocol Specific Properties (explaining what they are) rather than narrowing down the set of Protocol Specific Properties.

draft-ietf-taps-interface.md
```diff
@@ -460,11 +460,13 @@ earlier stages and querying them in later stages:
...
+Features of a specific protocol, see Section 3.2 of {{I-D.ietf-taps-arch}}, are not
```
mwelzl: s/Specialized Features/specialized features

draft-ietf-taps-interface.md

draft-ietf-taps-interface.md

PR #425, opened 2020-01-15
Normative language for property namespace

mirjak:
It looks like we should use normative language here. We could also use the keywords of the IANA protocol number registry to indicate reserved namespaces for IETF protocols (that can not be used in the vendor or implementation specific namespaces anyway). See https://www.iana.org/assignments/protocol-numbers/protocol-numbers.xhtml


mirjak:
> I agree, also with the suggestion to use the keywords of the IANA protocol number registry for IETF protocols, though I note that this PR doesn't include a pointer to this registry. Would you mind adding this?

Like this?

PR #431, opened 2020-01-15
More MAYs?

mirjak:


draft-ietf-taps-interface.md
```diff
@@ -720,8 +720,8 @@ Individual properties are then added to the TransportProperties Object:
 TransportProperties.Add(property, value)
 ~~~
 
-As preference typed selection properties may be used quite frequently, implementations should provide additional convenience functions as outlined in  {{preference-conv}}.
-In addition, implementations should provide a mechanism to create TransportProperties objects that are preconfigured for common use cases as outlined in {{property-profiles}}.
+As preference-typed Selection Properties may be used quite frequently, implementations MAY provide additional convenience functions as outlined in  {{preference-conv}}.
```
gorryfair: Starting "As" is not great English... Maybe we should say:
Preference-typed Selection Properties can be used frequently, implementations MAY therefore provide additional convenience functions, outlined in  {{preference-conv}}.

mirjak: I'm not the native speaker but the sentence doesn't make sense me without the "as". These are simply now two sentences and should have dot and not a comma. I guess we could separate by a semi-colon but that's even uglier...

gorryfair: Sorry adding "as" but not meaning the progress of time, isn't cool. Splitting into two sentences, is of course OK, how about:

Preference-typed Selection Properties can be frequently used. Implementations MAY therefore provide additional convenience functions, outlined in {{preference-conv}}.

abrunstrom: Are the conveninece functions outlined in the doc the only possible ones? It kind of sounds like it now.  So unless the second "as" was also bad English I would keep that to make it a bit more generic? So:

Preference-typed Selection Properties can be frequently used. Implementations MAY therefore provide additional convenience functions as outlined in {{preference-conv}}.

or perhaps

Preference-typed Selection Properties can be frequently used. Implementations MAY therefore provide additional convenience functions, see {{preference-conv}} for examples.

gorryfair: Yes - I think these are a sample not a definitive list. Either of the the above are fine for me:-), and the second is clearest probably.

mirjak: done

PR #432, opened 2020-01-15
"must provide sensible defaults"

mirjak:
I was about to change this must to MUST, however, MUST and "sensible" doesn't work well together, so I proposed this recommendation instead.

### DIFF
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -742,11 +742,12 @@ after establishment yields the value Required for properties of the selected
 protocol and path, Avoid for properties avoided during selection, and Ignore for
 all other properties.
 
-An implementation of this interface must provide sensible defaults for Selection
-Properties. The defaults given for each property below represent a
-configuration that can be implemented over TCP. An alternate set of default
-Protocol Selection Properties would represent a configuration that can be
-implemented over UDP.
+An implementation of this interface SHOULD use the defaults given 
+for each property below. These values represent a configuration that 
+can be implemented over TCP. If TCP is not supported or another default
+protocol is preferred, the implementation MUST specify an alternate set of
+default Protocol Selection Properties that represents a configuration that 
+can be implemented over the preferred protocol, such as UDP.
 
 
 ### Reliable Data Transfer (Connection) {#prop-reliable}
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -745,9 +745,9 @@ all other properties.
 An implementation of this interface must provide sensible defaults for Selection
 Properties. The recommended default values for each property below represent a
 configuration that can be implemented over TCP. If these default values are used
-and TCP is not supported by a TAPS implementation, then an application using the
+and TCP is not supported by a Transport Services implementation, then an application using the
 default set of Properties might not succeed in establishing a connection. Using
-the same default values for independent TAPS implementations can be beneficical
+the same default values for independent Transport Services implementations can be beneficical
 when application are ported between different implementation, even if the
 default could lead to a connection failure, as, for example, applications need
 explicitly be designed to also support a connectionless mode. In this case the
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -742,12 +742,19 @@ after establishment yields the value Required for properties of the selected
 protocol and path, Avoid for properties avoided during selection, and Ignore for
 all other properties.
 
-An implementation of this interface SHOULD use the defaults given 
-for each property below. These values represent a configuration that 
-can be implemented over TCP. If TCP is not supported or another default
-protocol is preferred, the implementation MUST specify an alternate set of
-default Protocol Selection Properties that represents a configuration that 
-can be implemented over the preferred protocol, such as UDP.
+An implementation of this interface must provide sensible defaults for Selection
+Properties. The recommended default values for each property below represent a
+configuration that can be implemented over TCP. If these default values are used
+and TCP is not supported by a TAPS implementation, then an application using the
+default set of Properties might not succeed in establishing a connection. Using
+the same default values for independent TAPS implementations can be beneficical
+when application are ported between the different implementation, even if the
+default could let to a connection failure, as, for example, applications need
+explicitly be designed to also support a connectionless mode. In this case the
+application can regonized the failure and explicitly specify a different set of
+Protocol Selection Properties that result in a usable protocol. If other default
+values than recommended below are used, it is recommended to document that well
+for application use.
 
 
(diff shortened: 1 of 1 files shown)


mirjak:
I wrote SHOULD because the next sentence explain when you can use other default values, so it seem to me that only SHOULD makes sense. There is a MUST later saying that some default MUST be used.

mwelzl:
Ah, ok!

abrunstrom:
I also like this and agree with the SHOULD. The MUST in the example that follows look odd to me though. I think that should be a MAY (or possibly a SHOULD).

mirjak:
The point about the MUST is that "default values MUST lead to a possible/implemented protocol". Maybe there is a better way to phrase that?

mirjak:
Btw. similar as in Arch I think we have to spread a couple more MUSTs, SHOULDs, and MAYs over the whole doc...

abrunstrom:
Not convinced that "default values MUST lead to an implemented protocol" is a MUST. If the TAPS default values do not lead to an implemented protocol (i.e. TCP or a protocol with similar functionality is not implemented), maybe you want applications that assume those values to get an error rather than transparently ending up with UDP? Maybe you know that your applications are aware and then you should be allowed to change them, but maybe you want to make sure that applications are aware that they get an unrelaible service and not a reliable service and then you should be allowed to keep the TAPS default values.

mirjak:
I agree with @abrunstrom but I think this means we need more text explaining this. Any proposals?

tfpauly:
Shouldn't this be:
- MUST provide some default that chooses an implemented protocol
- SHOULD use the ones we specify here

The suggestion is to invert the sentences to clarify.

tfpauly:
Or, we say that you can use the TCP defaults and raise an error if TCP isn't implemented.

abrunstrom:
I think it should be allowed to use the TAPS defaults and raise an error if no protocol that implements the TAPS defaults is available. 
In case the implementation prefers to do so. But I agree with Aaron that it is a corener case and in that sense perhaps not terribly important what we pick.

mirjak:
After thinking about this a bit more, I think that it is a must requirement that any chosen set of default values MUST NOT lead to an error, because what's otherwise the point about having default values....? However, I think we should add some big warning text that if other defaults are used that most be communicated to the users of this API others they might be really surprised...

gorryfair:
I think this has moved in a curious direction. My thoughts are still that this has been edited in the wrong direction.
I quite like the SHOULD.
Whilst the language is fine, I really do not agree at all with the direction of the "the implementation MUST specify  default Protocol Selection Properties that represents a configuration that can be implemented over the preferred protocol, such as UDP."
This is not the way I saw TAPS - I expected the exact opposite outcome, i.e.:
"the users of such an implementation MUST explicitly specify Properties that result in theimplemented over the preferred protocol, such as UDP."

draft-ietf-taps-interface.md
```diff
@@ -742,12 +742,19 @@ after establishment yields the value Required for properties of the selected
...
-protocol is preferred, the implementation MUST specify an alternate set of
-default Protocol Selection Properties that represents a configuration that 
-can be implemented over the preferred protocol, such as UDP.
+An implementation of this interface must provide sensible defaults for Selection
+Properties. The recommended default values for each property below represent a
+configuration that can be implemented over TCP. If these default values are used
+and TCP is not supported by a TAPS implementation, then an application using the
+default set of Properties might not succeed in establishing a connection. Using
+the same default values for independent TAPS implementations can be beneficical
+when application are ported between the different implementation, even if the
```
gorryfair, line 751: /the different implementation/different implementations/

draft-ietf-taps-interface.md
```diff
@@ -742,12 +742,19 @@ after establishment yields the value Required for properties of the selected
...
+default could let to a connection failure, as, for example, applications need
```
gorryfair, line 752: /let to/lead to/

draft-ietf-taps-interface.md
```diff
@@ -742,12 +742,19 @@ after establishment yields the value Required for properties of the selected
...
+explicitly be designed to also support a connectionless mode. In this case the
+application can regonized the failure and explicitly specify a different set of
+Protocol Selection Properties that result in a usable protocol. If other default
+values than recommended below are used, it is recommended to document that well
+for application use.
```
gorryfair, line 758: /for application use/ - seems awkward.

... Another aspect we did not speak about concerns the ability to provide examples of usage. If the defaults are common between systems, then common example methods and tutorials and libraries can rely on using the default Selection Properties.

mirjak: Not sure what you mean which concerns you mean in "the ability to provide examples"...? Can you further explain or provide a text proposal?

draft-ietf-taps-interface.md
```diff
@@ -743,10 +743,17 @@ protocol and path, Avoid for properties avoided during selection, and Ignore for
 all other properties.
 
 An implementation of this interface must provide sensible defaults for Selection
-Properties. The defaults given for each property below represent a
-configuration that can be implemented over TCP. An alternate set of default
-Protocol Selection Properties would represent a configuration that can be
-implemented over UDP.
+Properties. The recommended default values for each property below represent a
+configuration that can be implemented over TCP. If these default values are used
+and TCP is not supported by a TAPS implementation, then an application using the
```
tfpauly, line 748:
```suggestion
and TCP is not supported by a Transport Services implementation, then an application using the
```

draft-ietf-taps-interface.md
```diff
@@ -743,10 +743,17 @@ protocol and path, Avoid for properties avoided during selection, and Ignore for
...
+default set of Properties might not succeed in establishing a connection. Using
+the same default values for independent TAPS implementations can be beneficical
```
tfpauly, line 750:
```suggestion
the same default values for independent Transport Services implementations can be beneficical
```

draft-ietf-taps-interface.md
```diff
@@ -743,10 +743,17 @@ protocol and path, Avoid for properties avoided during selection, and Ignore for
...
-implemented over UDP.
+Properties. The recommended default values for each property below represent a
+configuration that can be implemented over TCP. If these default values are used
+and TCP is not supported by a Transport Services implementation, then an application using the
+default set of Properties might not succeed in establishing a connection. Using
+the same default values for independent Transport Services implementations can be beneficical
+when application are ported between different implementation, even if the
+default could lead to a connection failure, as, for example, applications need
+explicitly be designed to also support a connectionless mode. In this case the
+application can regonized the failure and explicitly specify a different set of
```
gorryfair, line 754: /can recognized/can recognize/

gorryfair:
Ready to merge

PR #433, opened 2020-01-16
Editorial nits to improve consistency

mirjak:


draft-ietf-taps-interface.md
```diff
@@ -873,9 +873,7 @@ that are supported on the local system to all remote systems, to allow
 applications to be written generically. For example, if a single implementation
 is used on both mobile devices and desktop devices, it should define the
 `Cellular` interface type for both systems, since an application may want to
-always `Prohibit Cellular`. Note that marking a specific interface type as
-`Required` limits path selection to a small set of interfaces, and leads to less
-flexible and resilient connection establishment.
+always `Prohibit Cellular`.
```
mirjak: Note, this sentence was redundant (see two paragraphs up).

draft-ietf-taps-interface.md
```diff
@@ -756,8 +756,8 @@ Name:
 
 This property specifies whether the application needs to use a transport
 protocol that ensures that all data is received on the other side without
-corruption. This also entails being notified when a Connection is closed or
-aborted. The default is to Require Reliable Data Transfer.
+corruption. The default is to Require Reliable Data Transfer. This also
+entails being notified when a Connection is closed or aborted.
```
mwelzl: I don't see why switching the order here would be an improvement? Since the "entails" statement gives more detail about what the property's behavior, I think this reads better in the previous version. You must have had a reason to propose this... what am I missing?

draft-ietf-taps-interface.md
```diff
@@ -768,15 +768,15 @@ This property specifies whether the application needs or prefers to use a transp
 protocol that preserves message boundaries. The default
 is to Prefer Preservation of Message Boundaries.
 
-### Configure Per-Message Reliability {#prop-partially-reliable}
+### Per-Message Reliability {#prop-partially-reliable}
```
mwelzl: I think the previous name was better: "Per-Message Reliability" reads as if "True" means that all messages will be reliably transferred. This, however, expresses a desire to be able to *configure* it per message.

draft-ietf-taps-interface.md
```diff
@@ -768,15 +768,15 @@ This property specifies whether the application needs or prefers to use a transp
...
 
 Name:
 : per-msg-reliability
 
 This property specifies whether an application considers it useful to indicate
 its reliability requirements on a per-Message basis. This property applies to
 Connections and Connection Groups. The default is to Ignore
-this option.
+this Per-Message Reliability.
```
mwelzl: I don't like this either - even if we'd go back to the previous name. I agree that "option" doesn't fit though - I think we could either repeat the name as you suggest, but without "this" before it, or we could say "Selection Property". I see that this is a general inconsistency, "this option" appears in multiple places and should be fixed uniformly, everywhere, so it's a good catch, thanks!

draft-ietf-taps-interface.md
mirjak: Maybe you are right. The point was that you only get this notification if this property is enabled. Maybe that's what we should say...

draft-ietf-taps-interface.md
mirjak: This was in order to be consistent with all other heading. Isn't you point to for other properties as well?

draft-ietf-taps-interface.md
mirjak: yes forgot to remove "this". The whole point was consistency. Feel free to take another approach.

draft-ietf-taps-interface.md
mwelzl: No - e.g. "Reliable Data Transfer (Connection)" means: I'd like to be able to send reliably. It doesn't mean "I want to be able to turn it on/off per-message". "Preservation of Message Boundaries" also just means "I want that, if possible". "Preservation of Data Ordering" means "I'd like in-order" - it doesn't mean "I care about configuring in-order per message". and so forth.

draft-ietf-taps-interface.md
mwelzl: Agreed - suggestion: switch the order back, but change the sentence - e.g.: "This also entails being notified when a Connection is closed or aborted when Reliable Data Transfer is enabled. The default is to Require Reliable Data Transfer."

draft-ietf-taps-interface.md
mwelzl: It's fine without "this", thanks!

mwelzl: (and I must remember to fix this uniformly everywhere...)

mwelzl:
Folks, all changes that I requested here are related to sentences dealing with defaults. Now, these sentences shouldn't be there anymore at all: defaults are shown in a different style, in a list. My suggestion is:
- delete this PR
- directly do an update to the master, incorporating the other changes that @mirjak suggested here (everything but the sentences talking about defaults)

Is this ok with people? If so, I'll do that.

mwelzl:
Ok, so I'll do this today.

mwelzl:
The text about defaults disappeared anyway, and two other suggestions here were incorporated with a direct commit to master.

PR #444, opened 2020-01-17
Indicate use of streams for unordered messages as example

mirjak:
Just a proposal...

### DIFF
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -1576,7 +1576,8 @@ Default:
 If true, it specifies that the receiver-side transport protocol stack may only deliver the Message to the receiving application after the previous ordered Message which was passed to the same Connection via the Send
 Action, when such a Message exists. If false, the Message may be delivered to the receiving application out of order.
 This property is used for protocols that support preservation of data ordering,
-see {{prop-ordering}}, but allow out-of-order delivery for certain messages.
+see {{prop-ordering}}, but allow out-of-order delivery for certain messages, e.g., by multiplex independent messages onto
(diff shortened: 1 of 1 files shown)


draft-ietf-taps-interface.md
```diff
@@ -1576,7 +1576,8 @@ Default:
 If true, it specifies that the receiver-side transport protocol stack may only deliver the Message to the receiving application after the previous ordered Message which was passed to the same Connection via the Send
 Action, when such a Message exists. If false, the Message may be delivered to the receiving application out of order.
 This property is used for protocols that support preservation of data ordering,
-see {{prop-ordering}}, but allow out-of-order delivery for certain messages.
+see {{prop-ordering}}, but allow out-of-order delivery for certain messages, e.g., by multiplex independent messages onto
```
mwelzl: s/multiplex/multiplexing

abrunstrom:
Well, you could do it with streams in SCTP as well so as an example it seems approriate. And good with an example I think.

mirjak:
Yes, we do have actually these kind of examples in many cases in the doc and I just thought when reading one could help here as well.

draft-ietf-taps-interface.md
mwelzl: done

mwelzl:
I don't understand why one would need streams for this in SCTP, but it doesn't matter. As I said I have no strong opinion, I'm fine with adding this example (and I just fixed the nit).

abrunstrom:
@mwelzl you do not need to, as there are multiple ways to do it in SCTP, but you could.

PR #445, opened 2020-01-17
checksum default and clarification

mirjak:
I'm not quite sure what I meant by "Only full coverage is guaranteed, any other requests are advisory". I assume we mean that even if a partial checksum is requested you get full coverage because requesting partial coverage and getting no checksum doesn't seem right...?


mwelzl:
Exactly, and you made it clearer. Thanks!

PR #460, opened 2020-01-17
editorial fix

mirjak:
Comma in the first section was wrong but better to rephrase anyway...

### DIFF
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -2100,7 +2100,7 @@ Properties will include different information:
 ## Generic Connection Properties {#connection-props}
 
 Generic Connection Properties are defined independent of the chosen protocol stack
-and therefor are available on all Connections.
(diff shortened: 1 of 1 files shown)


draft-ietf-taps-interface.md
```diff
@@ -2099,10 +2099,10 @@ Properties will include different information:
 
 ## Generic Connection Properties {#connection-props}
 
-The Connection Properties defined as independent of the chosen protocol stack,
-and available on all Connections are defined in the subsections below.
+Generic Connection Properties are defined independent of the chosen protocol stack
+and therefor are available on all Connections.
```
mwelzl: s/therefor/therefore
and: I'm really unsure about the "are" after "therefore". It might be correct, but I'm slightly more sure that the sentence is ok *without* the "are", i.e.: "Generic Connection Properties are defined independent of the chosen protocol stack and therefore available on all Connections."

PR #463, opened 2020-01-17
Mostly Editorial changes on Capacity Profile

mirjak:


draft-ietf-taps-interface.md
```diff
@@ -2244,28 +2244,28 @@ This property specifies the desired network treatment for traffic sent by the
...
   Default:
-  : The application makes no representation about its expected capacity
-  profile. No special optimizations of the tradeoff between delay, delay
-  variation, and bandwidth efficiency should be made when selecting and
-  configuring transport protocol stacks. Transport system implementations that
-  map the requested capacity profile onto per-connection DSCP signaling without
-  multiplexing SHOULD assign the DSCP Default Forwarding {{?RFC2474}} PHB; when
-  the Connection is multiplexed, the guidelines in Section 6 of {{?RFC7657}}
-  apply.
+  : The application provides now information about its expected capacity
```
theri: s/now/no/

I think it's good to explicitly say explicitly what "providing no information" means.
For example, this could be: "The application provides no information about its expected capacity profile, so it does not indicate any preference for optimizations of any tradeoffs between delay, delay variation, loss, and bandwidth efficiency."

mirjak: I can add this. However, I did cut the part about delay, delay variation, loss, and bandwidth efficiency because this is the first time something like this was mentioned and I found it  rather "unclear" at this point. I think we should rather say at the top then, that the indication of the what capacity profile the application expects can be used to optimise delay, delay variation, loss, and bandwidth efficiency. 

Btw. given we talk about delay here as well, should this rather be called traffic profile and not capacity profile.

theri: I'm also fine with adding this at the top. I just want us to make the Default clear in relation to the other options, which do talk about tradeoffs like "loss or latency", "delay or delay variation VS. bandwidth efficiency", and more.

theri: Oh, regarding "should this rather be called traffic profile": I think "traffic profile" sounds like the application is characterizing the traffic it are going to send and I vaguely remember that at some point we agreed that this is a bad idea, as we cannot predict the future. Instead, the application says that it prefers a certain way to deal with the tradeoffs of limited capacity, so Capacity Profile makes sense to me.

draft-ietf-taps-interface.md
```diff
@@ -2244,28 +2244,28 @@ This property specifies the desired network treatment for traffic sent by the
...
-  variation, and bandwidth efficiency should be made when selecting and
-  configuring transport protocol stacks. Transport system implementations that
-  map the requested capacity profile onto per-connection DSCP signaling without
-  multiplexing SHOULD assign the DSCP Default Forwarding {{?RFC2474}} PHB; when
-  the Connection is multiplexed, the guidelines in Section 6 of {{?RFC7657}}
-  apply.
+  : The application provides no information about its expected capacity
+  profile. Transport system implementations that
+  map the requested capacity profile onto per-connection DSCP signaling 
+  SHOULD assign the DSCP Default Forwarding {{?RFC2474}} PHB.
```
mwelzl: I'm fine with all other changes, but I wonder why you removed "when the Connection is multiplexed, the guidelines in Section 6 of {{?RFC7657}} apply." everywhere?

One could see this as an implementation detail, but then this part gives some hints about the expected behavior regarding DSCP values, if an implementation should choose to adapt the DSCP based on the Capacity Profile. These chosen DSCP values will be different in case of multiplexing, as Section 6 of RFC 7657 explains. So I would prefer to see this sentence stay, everywhere.

philsbln: It has not been removed, just moved from the individual items to the top, as it applies to all classes.

draft-ietf-taps-interface.md
gorryfair: There's been some probably well-intentioned editing of the DS-related text and I think this isn't now quite as it should be.

First:  "The Capacity Profile may also be used
 to set priorities on the wire for Protocol Stacks supporting prioritization. 
 Recommendations for use with DSCP are provided below for each profile; note that
 when a Connection is multiplexed, the guidelines in Section 6 of {{?RFC7657}} apply."
- that seems to somehow imply that stacks supporting prioritisation (which I see as a layer 2 QoS thing) and DSCP are similar. They are not, these are different things. Just as PvD is at still higher level.

Second, the proposal seems to assert there can be default DSCP mapping behaviours. Whether that is desirable or not for the IETF to specify, can be discussed on the TSVWG. I expect that is not a trivial discussion. I'd suggest we don't do that except for very specific cases: Scavenger; Default; Using one of the AF classes to provide more assurance than Default; etc.

Third, DSCPs should be set per flow aggregate - using multiple DSCPs within an aggregate or changing them mid-flow is generally unwise, but possible, and needs to always come with the warning in RFC7657.

So, back to the main point:  "Traffic" seems the wrong descriptor, to me in this context, it relates to the characteristics of the aggregate (rate, variance, etc); as in traffic-conditioning-agreeement. What I think this is about is the future intentions of the apps to use the network service. They likely won't have a way to characterise their traffic against a profile... but they probably can describe what class of service they expect across the path ... at it's simplest: is it AF? is it Default? is it LE?

draft-ietf-taps-interface.md
```diff
@@ -2244,28 +2244,28 @@ This property specifies the desired network treatment for traffic sent by the
 application and the tradeoffs the application is prepared to make in path and
 protocol selection to receive that desired treatment. When the capacity profile
 is set to a value other than Default, the transport system should select paths
-and profiles to optimize for the capacity profile specified. The following
-values are valid for the Capacity Profile:
+and protocols to optimize for the capacity profile specified. How this is realized
+is implementation-specific. The Capacity Profile may also be used
```
tfpauly:
```suggestion
is implementation-specific. The Capacity Profile MAY also be used
```

draft-ietf-taps-interface.md
mwelzl: From the interim: @gorryfair agrees that this is probably ok but he needs to read it again and we should wait for another ACK from him.

I think that this should be merged once we have that ACK.

mirjak:
Done

PR #464, opened 2020-01-17
Restructure TCP UTO section

mirjak:
Btw why do we have (only) properties for the TCP UTO?

### DIFF
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -2332,10 +2332,10 @@ a data transfer useful. It is given in bits per second. The special value -1 ind
 that no bound is specified.
 
 
-## TCP-specific Properties: User Timeout Option (UTO) {#tcp-uto}
+## TCP-specific Property: User Timeout Option (UTO) {#tcp-uto}
 
 These properties specifies configurations for the User Timeout Option (UTO), 
-in case TCP becomes the chosen transport protocol as an example for a protocol-specific property. 
+in case TCP becomes the chosen transport protocol. 
 Implementation is optional and of course only sensible if TCP is implemented in the transport system.
 
 All of the below parameters are optional (e.g., it is possible to specify "User Timeout Enabled" as true,
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -2332,26 +2332,58 @@ a data transfer useful. It is given in bits per second. The special value -1 ind
 that no bound is specified.
 
 
-### TCP-specific Property: User Timeout {#tcp-uto}
+## TCP-specific Properties: User Timeout Option (UTO) {#tcp-uto}
 
-This property specifies, for the case TCP becomes the chosen transport protocol:
+These properties specifies configurations for the User Timeout Option (UTO), 
+in case TCP becomes the chosen transport protocol. Implementation is optional
+and of course only sensible if TCP is implemented in the transport system.
 
-Advertised User Timeout (name: tcp.user-timeout-value, type: Integer):
-: a time value (default: the TCP default) to be advertised via the User Timeout Option (UTO) for the TCP at the remote endpoint
+All of the below parameters are optional (e.g., it is possible to specify "User Timeout Enabled" as true,
+but not specify an Advertised User Timeout value; in this case, the TCP default will be used).
+
+### Advertised User Timeout 
+
+Name: 
+: tcp.user-timeout-value
+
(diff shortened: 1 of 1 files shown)


draft-ietf-taps-interface.md
```diff
@@ -2332,27 +2332,59 @@ a data transfer useful. It is given in bits per second. The special value -1 ind
...
+
+Default:
+: false
+
+This property controls whether the UTO option is enabled for a
 connection. This applies to both sending and receiving.
 
-Changeable (name: tcp.user-timeout-recv, type: Boolean):
-: a boolean (default true) which controls whether the "Timeout for aborting Connection" (see {{conn-timeout}})
+### Timeout Changeable (
```
theri: Leftover bracket, please remove

mirjak:
I didn't get that this is only an example, so yes we should say this more explicitly. Not sure where to put that though. Maybe we can just merge this PR for now and have a separate PR to add this somewhere.

draft-ietf-taps-interface.md

mirjak:
I actually also just added "as an example for a protocol-specific property" in that section but maybe we want to say it somewhere else again...?

mwelzl:
It isn't just an example. The reasons it's there go back to minset (if you remove it, you lose this functionality forever), yet there were arguments for this being a special TCP thing only, and being too awkward to include as general functionality - much like there are now arguments that might lead to ECN becoming UDP-specific.

draft-ietf-taps-interface.md
```diff
@@ -2332,27 +2332,59 @@ a data transfer useful. It is given in bits per second. The special value -1 ind
 that no bound is specified.
 
 
-### TCP-specific Property: User Timeout {#tcp-uto}
+## TCP-specific Properties: User Timeout Option (UTO) {#tcp-uto}
```
mwelzl: This section defines only one TCP-specific property, so plural is very confusing / misleading here.

draft-ietf-taps-interface.md
```diff
@@ -2332,27 +2332,59 @@ a data transfer useful. It is given in bits per second. The special value -1 ind
...
 
-This property specifies, for the case TCP becomes the chosen transport protocol:
+These properties specifies configurations for the User Timeout Option (UTO), 
+in case TCP becomes the chosen transport protocol as an example for a protocol-specific property.
```
mwelzl: It's not an example, it's one out of many properties in the list, and we designated this as being TCP-specific. I would remove everything from (and including) "as an example" in this sentence.

draft-ietf-taps-interface.md
mirjak: Done. However, there are multiple configuration parameters...

draft-ietf-taps-interface.md
mirjak: Removed this again but I think we should say something about why this is the one and only TCP property that is included.

mwelzl:
Some more explanations:

- why is this included? From appendix C, we can see that it relates to "Suggest timeout to the peer" in minset. The argument for including this, in appendix A of the minset draft, is: "Functional because this is closely related to potentially assumed reliable data delivery." Indeed, the last paragraph of section 1 in RFC 5482 gives examples where communication might just fail for applications expecting their connections to stay up. in reality, this doesn't seem to be needed much, and so this is hardly implemented (if at all!).

- why is it TCP specific? This was discussed in issue #109: https://github.com/ietf-tapswg/api-drafts/issues/109
The key statement against having this as a general thing is probably this sentence from @tfpauly:
"If we had a timeout on overall connection establishment, from Initiate() to Ready(), then that's trying to set a timeout not just on TCP, but on DNS, racing, proxies, TLS, and who knows what else."

draft-ietf-taps-interface.md
```diff
@@ -2332,10 +2332,10 @@ a data transfer useful. It is given in bits per second. The special value -1 ind
 that no bound is specified.
 
 
-## TCP-specific Properties: User Timeout Option (UTO) {#tcp-uto}
+## TCP-specific Property: User Timeout Option (UTO) {#tcp-uto}
```
philsbln:
```suggestion

Name:
: tcp.uto

Type:
: Numeric

```

mirjak: This is just the section heading. The actually configuration properties are defined below.

philsbln: sorry, my fault

PR #465, opened 2020-01-17
New upper section for connection events

mirjak:
Btw. does this cover all events? There is an issue where I also ask if we e.g. need an event that sending failed due to lifetime expiry. is there a way to double-check if we considered all needed events? Should we also mention ConnectionErrors here? Maybe we anyway need to say more about error?

### DIFF
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -2354,7 +2354,11 @@ All of the above parameters are optional (e.g., it is possible to specify "User 
 but not specify an Advertised User Timeout value; in this case, the TCP default will be used).
 
 
-## Soft Errors
+## Connection Events
+
+During the lifetime of a connection there are events that can occure when configured.
+
+### Soft Errors
 
 Asynchronous introspection is also possible, via the SoftError Event. This event
 informs the application about the receipt of an ICMP error message related to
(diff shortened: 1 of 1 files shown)


draft-ietf-taps-interface.md
```diff
@@ -2354,7 +2354,11 @@ All of the above parameters are optional (e.g., it is possible to specify "User
 but not specify an Advertised User Timeout value; in this case, the TCP default will be used).
 
 
-## Soft Errors
+## Connection Events
+
+During the lifetime of a connection there are events that can occure when configured.
```
mwelzl: nit: s/occure/occur

mwelzl:
These are not all events (just search for the word "event" in the document), and generally they are not all lumped together - e.g. sending related events such as "Sent" appear in the "Sending Data" section - which, in fact, has its own subsection called "Send Events".

So, this would be a subsection just related to connection management... but then, you state: "During the lifetime of a connection there are events that can occur when configured." which gives  the (wrong) impression that this would be the list of all events that can happen during a connection's lifetime.

So, I wouldn't add this subsection heading and not include that statement. Or maybe add the subsection heading (which is ok but doesn't help much - as you see, there are only very few events in there), but remove that statement.

britram:
+1 to calling this "connection lifecycle events" or similar (do we refer to a connection lifecycle?)

mirjak:
Changed the heading and fixed the nit. I'm open to have another sentence there if people have proposals. We could even have no text but we usually say something after each heading.

mwelzl:
Given the new heading, I'm fine with the sentence as it is - because the context is clear now, it's no longer misleading (i.e., giving the wrong impression that these would be ALL events).

PR #474, opened 2020-01-18
Arch: Adjust Security Considerations to new Section 4.2.3

theri:
Following #389, Section 4.2.3 of the Architecture document now says that fallback between different security protocol versions might be allowed.
The Security Considerations section was out of sync. This PR fixes that.

### DIFF
--- a/draft-ietf-taps-arch.md
+++ b/draft-ietf-taps-arch.md
@@ -506,8 +506,9 @@ of TLS APIs. Transport Services APIs will expose similar functionality
 
 As described above in {{equivalence}}, if a Transport Services system races
 between two different Protocol Stacks, both SHOULD use the same security protocols
-and options unless the application explicitly specifies that it considers
-different security protocols or options equivalent.
+and options. However, a Transport Services system MAY race different security
+protocols, e.g., if the application explicitly specifies that it considers them
+equivalent.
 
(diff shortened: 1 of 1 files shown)


draft-ietf-taps-arch.md
```diff
@@ -505,8 +505,9 @@ of TLS APIs. Transport Services APIs will expose similar functionality
 {{?I-D.ietf-taps-transport-security}}.
 
 As described above in {{equivalence}}, if a Transport Services system races
-between two different Protocol Stacks, both MUST use the same security protocols
-and options.
+between two different Protocol Stacks, both SHOULD use the same security protocols
+and options unless the application explicitly specifies that it considers
+different security protocols or options equivalent.
```
mirjak: Maybe this still should be a MUST as you say "MUST ... unless". That's always a weird case for normative language...

theri: Okay, if the "unless ..." is the only exception, this should be a MUST.
But is it?
4.2.3 says "To ensure that security protocols are not incorrectly swapped, Transport Services systems SHOULD only automatically generate equivalent Protocol Stacks when the transport security protocols within the stacks are identical." and that's slightly more open.
I'm fine with either option, but we have to decide for one, so Section 4.2.3 and Section 6 have to match.

PR #481, opened 2020-01-20
Clarify Clone and Priority, closes #461

### LINKED ISSUE #461: Connection Groups, Priority, and Cloning
Section 11.1.3 of the API says about Priority:
"As noted in Section 6.4, this property is not
   entangled when Connections are cloned."
Not sure what this means. I though when you clone a connection you would actually take the priority of the cloned connection but then you can change it independently for each connection in a group. Or what's the priority after cloning? Back to the default of 100? We should say this explicitly. Or maybe you even want to provide the priority while cloning?

Actually looking at this now again, I think we would need to specify cloning a bit better anyway. I assume that calling clone() on a multiplexing capable connection would create a new stream; in case of quic is would only create local state but no on-the-wire singling. However if you call clone on a TCP connection, is that basically an initial() call that triggers the TCP handshake? We should say this explicitly.

theri:
Make clearer what "Priority is not entangled" means, following #461.
Also added a note that Clone() may result in on-the-wire signaling.

### DIFF
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -1315,7 +1315,7 @@ connections are "entangled" with each other, and become part of a Connection
 Group. Calling Clone on any of these two Connections adds a third Connection to
 the Connection Group, and so on. Connections in a Connection Group generally share
 Connection Properties. However, there may be exceptions, such as "Priority
-(Connection)", see {{conn-priority}}.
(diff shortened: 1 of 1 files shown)


draft-ietf-taps-interface.md
```diff
@@ -1337,6 +1337,9 @@ functionality to implement Clone. In that case, entangled Connections are
 multiplexed together, giving them similar treatment not only inside endpoints
 but also across the end-to-end Internet path.
 
+Note that calling Clone() may result in on-the-wire signaling depending on the
+underlying Protocol Stack.
```
mirjak: Maybe we can be even more concrete here, like "... may result in on-the-wire signalling, e.g. to open a new connection,.."?

PR #484, opened 2020-01-22
Move text removed by #421 to the arch document

philsbln:


gorryfair:
NOTES 1: If this moves, please remember to rephrase the sentences to match the Arch style, content seems OK.
NOTE 2: This probably should deal with my previous issue about defining "The Socket API (Sockets)" in such as way as to define "sockets".

philsbln:
I see value in it as it gives historic context and therefore support @theri's suggestion in #421. 
It is nothing I will really fight for, but something I wanted to try whether it is possible with little effort.

draft-ietf-taps-arch.md
```diff
@@ -97,7 +105,13 @@ This document was developed in parallel with the specification of the Transport
...
+TCP has scaled remarkably well over the past three and a half decades, but its total ubiquity has hidden an uncomfortable fact: the network is not really a	file, and stream abstractions are too simplistic for many modern application	
+programming models.	
+
+In the meantime, the nature of Internet access, and the variety of Internet	transport protocols, is evolving. The challenges that new protocols and access paradigms present to the sockets API and to programming models based on them	
+inspire the design principles of a new approach.
+Threfore, the new Transport Services architecture is based on the survey of services provided by IETF transport protocols and congestion control mechanisms {{?RFC8095}}, and the distilled minimal set of the features offered by transport protocols {{?I-D.ietf-taps-minset}}. These documents identified common features and patterns across all transport protocols developed thus far in the IETF.
```
theri: s/Threfore/Therefore/

draft-ietf-taps-arch.md
```diff
@@ -97,7 +105,13 @@ This document was developed in parallel with the specification of the Transport
 
 ## Background
 
-The Transport Services architecture is based on the survey of services provided by IETF transport protocols and congestion control mechanisms {{?RFC8095}}, and the distilled minimal set of the features offered by transport protocols {{?I-D.ietf-taps-minset}}. These documents identified common features and patterns across all transport protocols developed thus far in the IETF.
+The BSD Unix Sockets API's (Sockets) and its SOCK_STREAM abstraction was a revolution in simplicity: by bringing network sockets into the UNIX programming model, it enabled anyone who knew how to write programs that dealt with sequential-access files to also write network applications. The simplicity of this API is a key reason the	Internet won the protocol wars {{PROTOCOL-WARS}} of the 1980s. SOCK_STREAM is	tied to the Transmission Control Protocol (TCP), specified in 1981 {{?RFC0793}}.
```
theri: Why the brackets around "(Sockets)"? Without the word, the sentence wouldn't work anymore.
Also, I think we call it "Socket API", not "Sockets API" in the rest of the document.

draft-ietf-taps-arch.md
```diff
@@ -97,7 +105,13 @@ This document was developed in parallel with the specification of the Transport
...
+The BSD Unix Sockets API's (Sockets) and its SOCK_STREAM abstraction was a revolution in simplicity: by bringing network sockets into the UNIX programming model, it enabled anyone who knew how to write programs that dealt with sequential-access files to also write network applications. The simplicity of this API is a key reason the	Internet won the protocol wars {{PROTOCOL-WARS}} of the 1980s. SOCK_STREAM is	tied to the Transmission Control Protocol (TCP), specified in 1981 {{?RFC0793}}.	
+TCP has scaled remarkably well over the past three and a half decades, but its total ubiquity has hidden an uncomfortable fact: the network is not really a	file, and stream abstractions are too simplistic for many modern application	
+programming models.	
+
+In the meantime, the nature of Internet access, and the variety of Internet	transport protocols, is evolving. The challenges that new protocols and access paradigms present to the sockets API and to programming models based on them
```
theri: s/sockets API/Socket API/

draft-ietf-taps-arch.md
tfpauly: It also shouldn't have an apostrophe "API`s" -> "APIs"

tfpauly: We also do cover some of the relationship with BSD sockets in the intro.

How much do we think we really need this history here?

philsbln:
Hi, I guess I addressed all knits raised so far – so this PR is left with @tfpauly's question:
> How much do we think we really need this history here?

abrunstrom:
Reading this again, I actually think we may not need the history and keeping it short may be better. We otherwise have to deal with the overlap.

draft-ietf-taps-arch.md
```diff
@@ -89,15 +97,21 @@ This document describes an architecture for exposing transport protocol features
...
+Many application programming interfaces (APIs) to perform transport networking have been deployed, perhaps the most widely known and imitated being the BSD Socket {{POSIX}} interface (Socket API). The naming of objects and functions across these APIs is not consistent, and varies depending on the protocol being used. For example, sending and receiving streams of data is conceptually the same for both an unencrypted Transmission Control Protocol (TCP) stream and operating on an encrypted Transport Layer Security (TLS) {{?RFC8446}} stream over TCP, but applications cannot use the same socket ```send()``` and ```recv()``` calls on top of both kinds of connections. Similarly, terminology for the implementation of transport protocols varies based on the context of the protocols themselves: terms such as "flow", "stream", "message", and "connection" can take on many different meanings. This variety can lead to confusion when trying to understand the similarities and differences between protocols, and how applications can use them effectively.
```
abrunstrom: I lkie the small adjustments in the introduction. We should keep those.

draft-ietf-taps-arch.md
```diff
@@ -89,15 +97,21 @@ This document describes an architecture for exposing transport protocol features
...
+TCP has scaled remarkably well over the past three and a half decades, but its total ubiquity has hidden an uncomfortable fact: the network is not really a	file, and stream abstractions are too simplistic for many modern application	
+programming models.	
+
+In the meantime, the nature of Internet access, and the variety of Internet	transport protocols, is evolving. The challenges that new protocols and access paradigms present to the Socket API and to programming models based on them	
+inspire the design principles of a new approach.
+Therefore, the new Transport Services architecture is based on the survey of services provided by IETF transport protocols and congestion control mechanisms {{?RFC8095}}, and the distilled minimal set of the features offered by transport protocols {{?I-D.ietf-taps-minset}}. These documents identified common features and patterns across all transport protocols developed thus far in the IETF.
```
abrunstrom: I am fine leaving this extension out and keeping it simple. If we do decide to keep the text. Then I would like "new Tarnsport Services architecture" to be just "Transport Services architecture. I do not like the "new" for reasons previously commented on.

PR #487, opened 2020-01-24
Make the singular transmission text clearer

### LINKED ISSUE #367: Implementation - PMTU text is broken
This text is broken:

section 5.1
/  o  Singular Transmission: when this is true, the application requests
      to avoid transport-layer segmentation or network-layer
      fragmentation.  Some transports implement network-layer
      fragmentation avoidance (Path MTU Discovery) without exposing this
      functionality to the application; in this case, only transport-
      layer segmentation should be avoided, by fitting the message into
      a single transport-layer segment or otherwise failing.  Otherwise,
      network-layer fragmentation should be avoided--e.g. by requesting
      the IP Don't Fragment bit to be set in case of UDP(-Lite) and IPv4
      (SET_DF in [RFC8304]).	/
- we need to rewrite

adventureloop:
Closes #367.


adventureloop:
this should close #367

draft-ietf-taps-impl.md
```diff
@@ -501,7 +501,7 @@ The effect of the application sending a Message is determined by the top-level p
...
 
-- Singular Transmission: when this is true, the application requests to avoid transport-layer segmentation or network-layer fragmentation. Some transports implement network-layer fragmentation avoidance (Path MTU Discovery) without exposing this functionality to the application; in this case, only transport-layer segmentation should be avoided, by fitting the message into a single transport-layer segment or otherwise failing. Otherwise, network-layer fragmentation should be avoided---e.g. by requesting the IP Don’t Fragment bit to be set in case of UDP(-Lite) and IPv4 (SET_DF in {{!RFC8304}}).
+- Singular Transmission: When set, this property limits the message size to the Maximum Message Size Before Fragmentation or Segmentation (see Section 10.1.7 of {{I-D.ietf-taps-interface}}).  Messages larger than this size generate an error.  Setting this avoids transport-layer segmentation or network-layer fragmentation. When used with transports running over IP version 4 the Don't Fragment bit will be set.  Endpoints should avoid IP fragmentation should be avoided ({{!RFC8304}}).
```
philsbln:
```suggestion
- Singular Transmission: When set, this property limits the message size to the Maximum Message Size Before Fragmentation or Segmentation (see Section 10.1.7 of {{I-D.ietf-taps-interface}}).  Messages larger than this size generate an error.  Setting this avoids transport-layer segmentation or network-layer fragmentation. When used with transports running over IP version 4 the Don't Fragment bit will be set to avoid on-path IP fragmentation ({{!RFC8304}}).  
```

gorryfair:
Right there was a typo: Replacing the last 2 sentences, the last sentence ought to be:

" Endpoints should avoid IP fragmentation ({{!RFC8304}}) and when used with transports running over IP version 4 the Don't Fragment bit will be set."

These are two separate issues that a stack implementor has to decide to do: (1) Decide if the endpoint will fragment; (2) decide if the path can fragment the fragments. In some OS this will be bundled together, in other OS they will not.

adventureloop:
Discussed on the interim call. The name probably could be much better. We also want to model in the text the atomicity property of UDP transfers.

draft-ietf-taps-impl.md
gorryfair, line 504: This addresses the issue.

PR #497, opened 2020-01-28
Refactor Local Address Preference, fixes #436

### LINKED ISSUE #436: Local Address Preference always only preferred?
For Local Address Preference we only have the option to _prefer_ stable or temporary. However, I guess we would also need a way to e.g. specify that stable is _required_, no?

theri:
Following the suggestion in #436, made Local Address Preference a Set(Preference, Enumeration), so an application can now also Require or Prohibit stable addresses.
Added a sentence warning that, e.g., preferring stable AND temporary addresses does not make sense - but there are less obvious ways to produce a conflicting set of Selection Properties, too, and we know how to deal with them.

### DIFF
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -946,16 +946,17 @@ Name:
 : local-address-preference
 
 Type:
-: Enumeration
+: Set (Preference, Enumeration)
 
-This property specifies whether Listeners and Connections should prefer the use of temporary addresses when possible.
-This is generally used to prevent linking connections over time when a stable address is not needed. Possible values are:
+This property allows the application to control the use of stable or temporary local addresses.
+Temporary addresses are generally used to prevent linking connections over time when a stable address is not needed.
+Note that expressing preferences for both stable and temporary addresses can be redundant and/or produce a conflict. Possible values are:
 
 Stable:
-: Prefer the use of stable (sometimes called "permanent") local addresses
+: Use stable (sometimes called "permanent") local addresses
 
 Temporary:
-: Prefer the use of temporary (sometimes called "privacy") addresses {{!RFC4941}}
+: Use temporary (sometimes called "privacy") addresses {{!RFC4941}}
 
 The default is to prefer the use of stable local addresses for Listeners and
 Rendezvous Connections, and to prefer the use of temporary addresses for other
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -940,27 +940,19 @@ requiring specific PvD instances or interface instances, and should be preferred
 over these options.
 
 
-### Local Address Preference
+### Use Temporary Local Addresses
 
 Name:
-: local-address-preference
+: use-temporary-local-address
 
 Type:
-: (Preference, Enumeration)
+: Preference
 
-This property allows the application to control the use of stable or temporary local addresses.
-Temporary addresses are generally used to prevent linking connections over time when a stable address is not needed.
-Possible values are:
+Default:
+: Prefer for Active Open (Initiate), Avoid for Passive Open (Listen) and Rendezvous Connections.
 
-Stable:
-: Use stable (sometimes called "permanent") local addresses
-
-Temporary:
-: Use temporary (sometimes called "privacy") addresses {{!RFC4941}}
-
-The default is to Prefer the use of stable local addresses for Listeners and
-Rendezvous Connections, and to Prefer the use of temporary addresses for other
-Connections.
+This property allows the application to express a preference for the use of temporary local addresses, sometimes called "privacy" addresses {{!RFC4941}}.
+Temporary addresses are generally used to prevent linking connections over time when a stable address, sometimes called "permanent" address, is not needed.
 
 
 ### Parallel Use of Multiple Paths {#parallel-multipath}
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -946,11 +946,11 @@ Name:
 : local-address-preference
 
 Type:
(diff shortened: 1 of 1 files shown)


draft-ietf-taps-interface.md
```diff
@@ -946,19 +946,20 @@ Name:
...
 
 Type:
-: Enumeration
+: Set (Preference, Enumeration)
 
-This property specifies whether Listeners and Connections should prefer the use of temporary addresses when possible.
-This is generally used to prevent linking connections over time when a stable address is not needed. Possible values are:
+This property allows the application to control the use of stable or temporary local addresses.
+Temporary addresses are generally used to prevent linking connections over time when a stable address is not needed.
+Note that expressing preferences for both stable and temporary addresses can be redundant and/or produce a conflict. Possible values are:
```
mwelzl: How could such a conflict arise? If I just use Set here, once, I only have a choice between one or the other, so it looks as if a conflict wouldn't be possible.

theri: But you could call the TransportProperties.Require() twice with different values for the Enum, right? At least I always assumed that was possible and it's certainly needed for, e.g., Interface Instance or Type.

mwelzl: This confuses me. Normally, in an API, if I adjust a property with a call and make the call again with a different value, I expect that it's (only) the last call that matters. "Interface Instance or Type" has a Set of (Preference, Enum) tuples - like you're writing here. Actually, there, it makes sense to me that it's a set, but I'd assume that one call is made to hand over, e.g.: "Wifi, Prefer; Cellular, Avoid". In this case however, it's only about choosing either Stable or Temporary, and preferring it or even requiring it.

Actually, I'm also not sure that all Preference levels make sense here... e.g., what does it mean to call this and specify "Avoid Stable"? I would rather say that the enum needs one more value, e.g.: "Stable_required".

theri: Regarding what happens if you make the call again with a different value, maybe you're right, but I wouldn't expect it to always work this way. For example, how do I specify a RemoteEndpoint with multiple local addresses, other than calling RemoteSpecifier.WithIPv6Address(...) multiple times? And then I'd expect the endpoint to contain all addresses that I specified and not just the last one.
Maybe that's why I'm confused here, too.

But perhaps for Properties we should actually specify the behavior as you say.
And then, I think for the Local Address Preference, it'd be enough to have one (Preference, Enum) instead of a set. It really does only make sense to specify one of these, and if you do, you implicitly specify the other. For example, if I specify "Avoid Stable", this implicitly means I prefer temporary - right?
I don't see why another value would be required - I can just specify "Require Stable".

mwelzl: "And then, I think for the Local Address Preference, it'd be enough to have one (Preference, Enum) instead of a set. It really does only make sense to specify one of these" - does it? I'm not sure, in case of a multi-homed host, for example.  I believe we should discuss this PR at the interim.

theri: I don't see that it makes sense to, e.g., Prefer stable addresses on one local interface and Avoid them on another for the same Preconnection, and if there's case where it does make sense, the application would probably need more fine-grained control than just a list of preferences. Not sure it makes sense to specify this in our API draft.
That said, sure, we can discuss this at the interim.

philsbln: We have some properties, e.g., PvD and Interface, that make sense being called multiple times - so we can't generally say every call cancels out the earlier. 

For this preference, I would really prefer to keep it a simple enum, probably with Reuire stable / prefer stable / prefer temporary / require temporaray"

I guess it makes sense being able to specify this on a per-pvd or per-interface basis, e.g., force stable if on corporate network, otherwise prefer temp, but for me this is strictly optional.

tfpauly:
Interim discussion: let's use prohibit/avoid/prefer/require temporary addresses. Default connections to prefer temporary. Default listeners to avoid temporary.

tfpauly:
Note that it fails without temporary addresses if you require temporary; which makes IPv4 fail.

theri:
Simplified this as discussed in the interim.
Please take another look.

philsbln:
Looks good to me – But I am unsure whether we should replace the temporary address with a stable address with Preference. Rational: For IPv6 with outgoing connections, temporary is default anyway, so requiring the special case instead of avoid/prohibit the default looks much more straight forward.

theri:
> Looks good to me – But I am unsure whether we should replace the temporary address with a stable address with Preference. Rational: For IPv6 with outgoing connections, temporary is default anyway, so requiring the special case instead of avoid/prohibit the default looks much more straight forward.

I've read this a few times but I'm still not sure: What alternative solution are you proposing here?

philsbln:
> > Looks good to me – But I am unsure whether we should replace the temporary address with a stable address with Preference. Rational: For IPv6 with outgoing connections, temporary is default anyway, so requiring the special case instead of avoid/prohibit the default looks much more straight forward.
> 
> I've read this a few times but I'm still not sure: What alternative solution are you proposing here?

I am proposing to Invert the whole name and meaning of the property – instead of requesting a temporary address, request a stable address.

theri:
> > > Looks good to me – But I am unsure whether we should replace the temporary address with a stable address with Preference. Rational: For IPv6 with outgoing connections, temporary is default anyway, so requiring the special case instead of avoid/prohibit the default looks much more straight forward.
> > 
> > 
> > I've read this a few times but I'm still not sure: What alternative solution are you proposing here?
> 
> I am proposing to Invert the whole name and meaning of the property – instead of requesting a temporary address, request a stable address.

Thanks for clarifying.
I think this Property is straightforward enough as is, no need to invert.

PR #499, opened 2020-01-28
Justify why TCP UTO is included

theri:
Following the discussion in #464, explain why TCP UTO is included as a Protocol-Specific Property, while other properties are not.
Also, some minor consistency fixes.

### DIFF
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -2354,11 +2354,14 @@ that no bound is specified.
 
 ## TCP-specific Property: User Timeout Option (UTO) {#tcp-uto}
 
-These properties specifies configurations for the User Timeout Option (UTO), 
+These properties specify configurations for the User Timeout Option (UTO), 
 in case TCP becomes the chosen transport protocol. 
 Implementation is optional and of course only sensible if TCP is implemented in the transport system.
 
-All of the below parameters are optional (e.g., it is possible to specify "User Timeout Enabled" as true,
+Note that these properties are part of the minimal set of transport services
+{{I-D.ietf-taps-minset}}, therefore, they are included here.
+
+All of the below properties are optional (e.g., it is possible to specify "User Timeout Enabled" as true,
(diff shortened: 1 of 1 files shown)


draft-ietf-taps-interface.md
```diff
@@ -2354,11 +2354,14 @@ that no bound is specified.
...
 ## TCP-specific Property: User Timeout Option (UTO) {#tcp-uto}
 
-These properties specifies configurations for the User Timeout Option (UTO), 
+These properties specify configurations for the User Timeout Option (UTO), 
 in case TCP becomes the chosen transport protocol. 
 Implementation is optional and of course only sensible if TCP is implemented in the transport system.
 
-All of the below parameters are optional (e.g., it is possible to specify "User Timeout Enabled" as true,
+Note that these properties are part of the minimal set of transport services
+{{I-D.ietf-taps-minset}}, therefore, they are included here.
```
mwelzl: I would rephrase this. Here's a proposal:

While they were deemed to be TCP-specific, these properties are included here because they are part of the minimal set of transport services {{I-D.ietf-taps-minset}}, where they were categorized as "functional". This means that they cannot be offered without the application knowing about them, or else they violate assumptions that might cause the application to fail.

draft-ietf-taps-interface.md
```diff
@@ -2354,11 +2354,14 @@ that no bound is specified.
 
 ## TCP-specific Property: User Timeout Option (UTO) {#tcp-uto}
```
mwelzl: Since we're now using plural, should this be "Properties"?

draft-ietf-taps-interface.md
theri: Thanks! I was wondering what to write more and I think mentioning this makes sense.
I expanded this paragraph based on your suggestion trying to make it easier to understand - please tell me if it's correct now.

draft-ietf-taps-interface.md
theri: Yes, I guess so.

Issue #418, opened 2020-01-15
Use of normative language in Arch


mirjak:
At some point we added normative language to the arch doc but only changed cases where use of upper case keywords are obvious. If we really want to use normative language, we would need another pass and make sure that we identified everything that should be specified normatively.

Alternatively, we could just remove any normative language again and revert respective words back to lower case. I think that would be my preferred solution as a) it is less work and b) I don't think this document necessarily needs normative language.

gorryfair:
Personally, I'd prefer to keep the language - but I agree we ought to review the items.

I'd really like to use REQUIRED, RECOMMENDED, rather MUST and SHOULD - because these are not protocol operation requirements, but implementation requirements to conform.

e.g.: , these properties SHOULD apply to multiple transport protocols.
could be.: , it is RECOMMENDED that these properties apply to multiple transport protocols.

theri:
I think having normative language in there is a good idea because it makes it clearer what criteria we think an implementation has to meet or to give guidance on what to prioritize to implement.

In the current draft, there's two places in which I find the normative language odd and maybe misplaced though:
- Section 4.1: "Events MAY also provide opportunities for the application to interact with the underlying transport by querying state or updating maintenance options." - Maybe this should be rephrased or the MAY removed? Clearly events _are_ an opportunity to exchange information, which an implementation MAY or may not take.
- Section 3.2: "A Transport Services API MUST allow applications to define such requirements and constrain the system's options." - This seems to be oddly unspecific for a MUST: There MUST be some way to define some constraints, but what exact constraints is undefined? Also, this is in addition to minset, and I thought MUST should mostly cover minset to limit scope creep.

gorryfair:
I agree with bot of those Theresa! I don't think we need keywords for either of these clauses.

theri:
Following our discussion during the interim and having thought about it a bit more, I'd be okay with not having normative language in the Architecture document, as long as we make sure that it still clearly says what has to be implemented (i.e., support for minset), and as long as we still clearly warn against unsafe fallback for security protocols and such. But I guess that for me, changing the words to lower case would be enough.

However, if we do that (and maybe even if we don't), we need to make clear that the API document actually contains ALL the normative language we need. For example, I'm not seeing anything on security protocols fallback in the API document yet.

gorryfair:
I admit I misunderstood what Mirja's comment was. Gorry was Sorry. - this was the benefit of some discussion. This has made me realise:
[...]

PR #502, opened 2020-02-21
Clean up unnecessary normative language in arch

tfpauly:
Addresses #418.

Removes most of the normative language in arch, which wasn't strictly necessary. What's left is the careful text around security protocols being raced, etc, which I believe is required.

### DIFF
--- a/draft-ietf-taps-arch.md
+++ b/draft-ietf-taps-arch.md
@@ -181,7 +181,7 @@ The Transport Services architecture evolves this general model of interaction, a
 
 The Transport Services API {{!I-D.ietf-taps-interface}} defines the mechanism for an application to create network connections and transfer data. The implementation {{?I-D.ietf-taps-impl}} is responsible for mapping the API to the various available transport protocols and managing the available network interfaces and paths.
 
-There are key differences between the architecture of the Transport Services system and the architecture of the Socket API: the Transport Services API is asynchronous and event-driven; it uses messages for representing data transfer to applications, and it assumes an implementation that can use multiple IP addresses, multiple protocols, multiple paths, and provide multiple application streams.
+There are key differences between the architecture of the Transport Services system and the architecture of the Socket API: the Transport Services API is asynchronous and event-driven; it uses messages for representing data transfer to applications; and it assumes an implementation that can use multiple IP addresses, multiple protocols, multiple paths, and provide multiple application streams.
 
 ## Event-Driven API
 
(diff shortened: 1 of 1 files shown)


draft-ietf-taps-arch.md
```diff
@@ -224,27 +224,27 @@ There are several degrees in which a Transport Services system is intended to of
...
+Implementations that provide the Transport Services API {{?I-D.ietf-taps-impl}} will vary due to system-specific support and the needs of the deployment scenario. It is expected that all implementations of Transport Services will offer the entire mandatory API. All implementations are expected to offer an API that is sufficient to use the distilled minimal set of features offered by transport protocols {{?I-D.ietf-taps-minset}}, including API support for TCP and UDP transport. However, some features provided by this API will not be functional in certain implementations. For example, it is possible that some very constrained devices might not have a full TCP implementation beneath the API.
```
gorryfair: I don't agree with the change. REQUIRED seems strong - but as above - I think we at least need to RECOMMEND minset.

ALTHOUGH, stating the requirement twice is not good and one statement needs to be changed!

draft-ietf-taps-arch.md
```diff
@@ -298,15 +298,15 @@ The following diagram summarizes the top-level concepts in the architecture and
 
 Fundamentally, a Transport Services API needs to provide connection objects ({{objects}}) that allow applications to establish communication, and then send and receive data. These could be exposed as handles or referenced objects, depending on the language.
 
-Beyond the connection objects, there are several high-level groups of actions that any Transport Services API implementing this specification MUST provide:
```
gorryfair: Agree in proposal to change to needs.

draft-ietf-taps-arch.md
```diff
@@ -298,15 +298,15 @@ The following diagram summarizes the top-level concepts in the architecture and
...
+Beyond the connection objects, there are several high-level groups of actions that any Transport Services API needs to provide:
 
-* Pre-Establishment ({{preestablishment}}) encompasses the properties that an application can pass to describe its intent, requirements, prohibitions, and preferences for its networking operations. For any system that provides generic Transport Services, these properties SHOULD apply to multiple transport protocols. Properties specified during Pre-Establishment can have a large impact on the rest of the interface: they modify how establishment occurs, they influence the expectations around data transfer, and they determine the set of events that will be supported.
```
gorryfair: Disagree on this: I think this is an archtiectural requirement:

It is RECOMMENDED that any system that provides generic Transport Services,  SHOULD define properties common to multiple transport protocols.

draft-ietf-taps-arch.md
```diff
@@ -298,15 +298,15 @@ The following diagram summarizes the top-level concepts in the architecture and
...
 
 * Establishment ({{establishment}}) focuses on the actions that an application takes on the connection objects to prepare for data transfer.
 
 * Data Transfer ({{datatransfer}}) consists of how an application represents the data to be sent and received, the functions required to send and receive that data, and how the application is notified of the status of its data transfer.
 
-* Event Handling ({{events}}) defines categories of notifications which an application can receive during the lifetime of transport objects. Events MAY also provide opportunities for the application to interact with the underlying transport by querying state or updating maintenance options.
```
gorryfair: Agree

draft-ietf-taps-arch.md
```diff
@@ -337,7 +337,7 @@ The diagram below provides a high-level view of the actions and events during th
 
 ### Connections and Related Objects {#objects}
 
-* Preconnection: A Preconnection object is a representation of a potential Connection. It has state that describes parameters of a Connection that might exist in the future: the Local Endpoint from which that Connection will be established, the Remote Endpoint ({{preestablishment}}) to which it will connect, and Transport Properties that influence the paths and protocols a Connection will use. A Preconnection can be fully specified such that it represents a single possible Connection, or it can be partially specified such that it represents a family of possible Connections. The Local Endpoint ({{preestablishment}}) MUST be specified if the Preconnection is used to Listen for incoming Connections. The Local Endpoint is OPTIONAL if it is used to Initiate Connections. The Remote Endpoint MUST be specified in the Preconnection that is used to Initiate Connections. The Remote Endpoint is OPTIONAL if it is used to Listen for incoming Connections. The Local Endpoint and the Remote Endpoint MUST both be specified if a peer-to-peer Rendezvous is to occur based on the Preconnection.
```
gorryfair: I agree with the set of requirements, but these are API requirements, not architectural ones. The new text therefore seems appropriate to me.

draft-ietf-taps-arch.md
```diff
@@ -398,9 +398,9 @@ The diagram below provides a high-level view of the actions and events during th
...
+* Send: The action to transmit a Message over a Connection to the remote system. The interface to Send can accept Message Properties specific to how the Message content is to be sent. The status of the Send operation is delivered back to the sending application in an event ({{events}}).
 
-* Receive: An action that indicates that the application is ready to asynchronously accept a Message over a Connection from a remote system, while the Message content itself will be delivered in an event ({{events}}). The interface to Receive MAY include Message Properties specific to the Message that is to be delivered to the application.
+* Receive: An action that indicates that the application is ready to asynchronously accept a Message over a Connection from a remote system, while the Message content itself will be delivered in an event ({{events}}). The interface to Receive can include Message Properties specific to the Message that is to be delivered to the application.
```
gorryfair: Don't see this change as sufficient.

The idea of adding Send and Receive properties seem close to  encapsulating important TAPS architecture, but as written here seem more appropriate to the API.

I do think this IS something we ought to call-out as a fundamental aspect of any TAPS system.
A TAPS API is REQUIRED to accept Message Properties specific to how a Message content is to be sent, but the details of how these messages are used can be system-dependent. A TAPS API MAY SUPPLY Message Properties specific to how a Message content is to be received, but the details of how these messages are used can be system-dependent.

draft-ietf-taps-arch.md
```diff
@@ -410,21 +410,21 @@ The following categories of events can be delivered to an application:
...
+* Close: The action an application takes on a Connection to indicate that it no longer intends to send data, is no longer willing to receive data, and that the protocol should signal this state to the remote system if the transport protocol allows this. (Note that this is distinct from the concept of "half-closing" a bidirectional connection, such as when a FIN is sent in one direction of a TCP connection. Indicating the end of a stream in the Transport Services architecture is possible using Message Properties when sending.)
 
-* Abort: The action the application takes on a Connection to indicate a Close and also indicate that the Transport Services system SHOULD NOT attempt to deliver any outstanding data. This is intended for immediate termination of a connection, without cleaning up state.
+* Abort: The action the application takes on a Connection to indicate a Close and also indicate that the Transport Services system should not attempt to deliver any outstanding data. This is intended for immediate termination of a connection, without cleaning up state.
```
gorryfair: OK with this proposed change - seems like an API detail.

draft-ietf-taps-arch.md
```diff
@@ -487,9 +487,9 @@ There are several reasons, however, that an application might want to explicitly
 - Privacy concerns about allowing Connections to multiplex together, which can tell a Remote Endpoint that all of the Connections are coming from the same application (for example, when Connections are multiplexed HTTP/2 or QUIC streams).
 - Performance concerns about Connections introducing head-of-line blocking due to multiplexing or needing to share state on a single thread.
 
-The Transport Services API SHOULD allow applications to explicitly define Connection Groups that force separation of Cached State and Protocol Stacks. For example, a web browser application might use Connection Groups with separate caches for different tabs in the browser to decrease linkability.
```
gorryfair: I actually prefer the change proposed to use /can/ .... although others may protest? -  I think that whether a TAPS system /implements/ this is not fundamental to TAPS.

draft-ietf-taps-arch.md
```diff
@@ -487,9 +487,9 @@ There are several reasons, however, that an application might want to explicitly
...
+The Transport Services API can allow applications to explicitly define Connection Groups that force separation of Cached State and Protocol Stacks. For example, a web browser application might use Connection Groups with separate caches for different tabs in the browser to decrease linkability.
 
-The interface to specify a Connection Group MAY expose fine-grained tuning for which properties and cached state is allowed to be shared with other Connections. For example, an application might want to allow sharing TCP Fast Open cookies across groups, but not TLS session state.
+The interface to specify a Connection Group can expose fine-grained tuning for which properties and cached state is allowed to be shared with other Connections. For example, an application might want to allow sharing TCP Fast Open cookies across groups, but not TLS session state.
 
 # IANA Considerations
```
gorryfair: I think the change is OK. Because this is a detail of the API, not an architectural aspect.

draft-ietf-taps-arch.md
```diff
@@ -410,21 +410,21 @@ The following categories of events can be delivered to an application:
...
+* Message Received: Delivers received Message content to the application, based on a Receive action. This can include an error if the Receive action cannot be satisfied due to the Connection being closed.
 
 * Message Sent: Notifies the application of the status of its Send action. This might indicate a failure if the Message cannot be sent, or an indication that the Message has been processed by the protocol stack.
 
 * Path Properties Changed: Notifies the application that some property of the Connection has changed that might influence how and where data is sent and/or received.
 
 ### Termination Actions {#termination}
 
-* Close: The action an application takes on a Connection to indicate that it no longer intends to send data, is no longer willing to receive data, and that the protocol SHOULD signal this state to the remote system if the transport protocol allows this. (Note that this is distinct from the concept of "half-closing" a bidirectional connection, such as when a FIN is sent in one direction of a TCP connection. Indicating the end of a stream in the Transport Services architecture is possible using Message Properties when sending.)
```
gorryfair: Is this architectural - it could be seen as explaining something from Minset and referencing section 6.1 about close could be helpful.
I think we should usee SHOULD here, because I think helps explaining the requirements for one-sided operation: A protocol SHOULD signal the close state to the remote system if the transport protocol allows this.

draft-ietf-taps-arch.md
```diff
@@ -181,7 +181,7 @@ The Transport Services architecture evolves this general model of interaction, a
 
 The Transport Services API {{!I-D.ietf-taps-interface}} defines the mechanism for an application to create network connections and transfer data. The implementation {{?I-D.ietf-taps-impl}} is responsible for mapping the API to the various available transport protocols and managing the available network interfaces and paths.
 
-There are key differences between the architecture of the Transport Services system and the architecture of the Socket API: the Transport Services API is asynchronous and event-driven; it uses messages for representing data transfer to applications, and it assumes an implementation that can use multiple IP addresses, multiple protocols, multiple paths, and provide multiple application streams.
+There are key differences between the architecture of the Transport Services system and the architecture of the Socket API: the Transport Services API is asynchronous and event-driven; it uses messages for representing data transfer to applications; and it assumes an implementation that can use multiple IP addresses, multiple protocols, multiple paths, and provide multiple application streams.
```
gorryfair: I think the word /assumes/ is not a good choice of word. Can we change this to something like:

/and it can provide benefit when an implementation can use multiple IP addresses, multiple protocols, multiple paths, and provide multiple application streams./

gorryfair:
I would still be in favour of using RFC2119 language to describe what is required architecturally to be a TAPS system. In my review (above) the number of requirements is now reduced.

I really do wonder if we could (please) group all these requirements on architecture into a single subsection, so they are easy to read - and we can they say that these are the architectural requirements, explcitly refererencing those in section XX relating to racing.  I suggest we insert this either as short text blocks, or simply a set of bullets in section 1.3 (although I am open to other places where we can call-out the architectural requirements). 

Perhaps we should also check we still have captured something along the lines of: "The System should permit evolution of the system without affecting programs using it. In normal use, TAPS systems are RECOMMENDED to result in consistent selection for the same network conditions when they have the same set of preferences, but applications using the TAPS API are REQUIRED to be robust to the automated selection, including the possibility to express requirements that constrain the choices that can be made by the system."

I would then suggest to use lower case within the document and perhaps refer to this subsection, where necessary to get the RFC2119 sentences.

draft-ietf-taps-arch.md
```diff
@@ -224,27 +224,27 @@ There are several degrees in which a Transport Services system is intended to of
 
 ## Common APIs for Common Features
 
-Functionality that is common across multiple transport protocols SHOULD be accessible through a unified set of API calls. An application ought to be able to implement logic for its basic use of transport networking (establishing the transport, and sending and receiving data) once, and expect that implementation to continue to function as the transports change.
+Functionality that is common across multiple transport protocols ought to be accessible through a unified set of API calls. An application using a Transport Services API can implement logic for its basic use of transport networking (establishing the transport, and sending and receiving data) once, and expect that implementation to continue to function as the transports change.
 
-Any Transport Services API is REQUIRED to allow access to the distilled minimal set of features offered by transport protocols {{?I-D.ietf-taps-minset}}.
```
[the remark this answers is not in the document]
britram: This one seems like a point for wider discussion at 107 or a future interim.

draft-ietf-taps-arch.md
britram: As above.

draft-ietf-taps-arch.md
britram: this also seems tied to the minset discussion, suggest we pick it up at 107 or after.

draft-ietf-taps-arch.md
britram: let's pick this up in/after 107.

draft-ietf-taps-arch.md
britram: let's pick this up at/after 107.

Issue #416, opened 2020-01-15
add example for "information about the received Message"?


mirjak:
arch doc sec 4.1.4:
"When receiving Messages, Message Properties can contain
      information about the received Message, such as metadata generated
      at the receiver and information signalled by the remote endpoint."
Maybe provide an example here...?

mwelzl:
Suggestion: ECN, which is information given from UDP in the form of such metadata.

tfpauly:
QUIC stream identifier example works

tfpauly:
Final message is in API, just use that

britram:
closed by #505

PR #505, opened 2020-02-21
Address #416, add message property example

tfpauly:


draft-ietf-taps-arch.md
```diff
@@ -396,7 +396,7 @@ The diagram below provides a high-level view of the actions and events during th
...
+* Message Properties: Message Properties are used to specify details about Message transmission. They can be specified directly on individual Messages, or can be set on a Preconnection or Connection as defaults. These properties might only apply to how a Message is sent (such as how the transport will treat prioritization and reliability), but can also include properties that specific protocols encode and communicate to the Remote Endpoint. When receiving Messages, Message Properties can contain information about the received Message, such as metadata generated at the receiver and information signalled by the remote endpoint. For example, a Message can be marked with a Message Property indicating that it is the final message on a connection if the peer sent a TCP FIN.
```
mwelzl: "TCP FIN" feels to protocol specific here; couldn't we say "... if the peer has requested to Close the Connection." ?

Issue #429, opened 2020-01-15
selection algorithm (text in API doc)


mirjak:
In section 5.2. in the API doc:
"Internally, the transport system will first exclude all protocols and
   paths that match a Prohibit, then exclude all protocols and paths
   that do not match a Require, then sort candidates according to
   Preferred properties, and then use Avoided properties as a
   tiebreaker.  Selection Properties that select paths take preference
   over those that select protocols.  For example, if an application
   indicates a preference for a specific path by specifying an
   interface, but also a preference for a protocol not available on this
   path, the transport system will try the path first, ignoring the
   preference."

This sounds rather like an example algorithm that should maybe go into the implementation doc instead? Or is there any part of this that we need to specify normatively in order to ensure that all systems implement the same?

theri:
This goes back to issue #149 on "conflicting policies". At the time, we decided that there must be a deterministic outcome, therefore, text was added to both the API and the Implementation document, see #169. 
If we keep it that way, perhaps the text here should point out that the goal is to resolve conflicting policies, so we achieve a deterministic outcome.

mirjak:
Ah okay, thanks for the background info; missed that. I think we need to use normative language here then. Maybe also a separate subsection makes this more prominent.

gorryfair:
This reads to me like the rules for processing polices rather than an example algorithm. I think it is part of the API... Normative language would be good and a stronger opening that explains why the rules exist.

mwelzl:
I don't think a separate subsection would make this a better read - this fits fine as it stands, as an intro to the transport property specification IMO. Regarding normative language, that's a separate discussion. So, can we close this issue?

theri:
Agree that we don't need a separate subsection. I can try to tweak the beginning of this part a bit to explain why the rules exist as @gorryfair suggested.
Regarding normative language - we definitely have it in the API draft, right? So why not have the discussion on it here as well?
if we're serious about deterministic outcomes, I guess the whole thing needs to be a MUST, right?

mwelzl:
Sorry, true, that normative language debate was about the arch draft, and we definitely have it in the API draft. So yes, let's discuss it here. I also think a MUST would make sense.

philsbln:
So the order of filtering out require/prohibit is not important (outcome is the same) - making the order a MUST does not make sense. Also, first sorting by preference and then filtering has the same outcome as doing it the other way around. the only thing we need to say is that filtering MUST happen and sorting SHOULD happen.

[...]

PR #511, opened 2020-03-02
Move algorithm text to Implementation draft, make Interface draft more flexible

theri:
In the [February 2020 interim](http://etherpad.tools.ietf.org:9000/p/notes-ietf-interim-2020-feb), people found the normative language in #496 too restrictive and agreed with the original suggestion in #429.
Therefore, this PR moves the selection algorithm text to the implementation draft and removes some normative language in the interface draft.
We still want to provide consistency to applications when possible, so there's a sentence in the implementation draft on this now as well.

### DIFF
--- a/draft-ietf-taps-impl.md
+++ b/draft-ietf-taps-impl.md
@@ -174,8 +174,8 @@ Aggregate [Endpoint: www.example.com:80] [Interface: Any]   [Protocol: TCP]
 Any one of these sub-entries on the aggregate connection attempt would satisfy the original application intent. The concern of this section is the algorithm defining which of these options to try, when, and in what order.
 
 During Candidate Gathering, an implementation first excludes all protocols and
-paths that match a Prohibit and all protocols and paths that do not match a
-Require. Then, the implementation will sort branches according to Preferred
+paths that match a Prohibit or do not match all Require properties.
+Then, the implementation will sort branches according to Preferred
 properties, Avoided properties, and possibly other criteria.
 
 
--- a/draft-ietf-taps-impl.md
+++ b/draft-ietf-taps-impl.md
@@ -173,6 +173,12 @@ Aggregate [Endpoint: www.example.com:80] [Interface: Any]   [Protocol: TCP]
 
 Any one of these sub-entries on the aggregate connection attempt would satisfy the original application intent. The concern of this section is the algorithm defining which of these options to try, when, and in what order.
 
+During Candidate Gathering, an implementation first excludes all protocols and
+paths that match a Prohibit and all protocols and paths that do not match a
+Require. Then, the implementation will sort branches according to Preferred
+properties, Avoided properties, and possibly other criteria.
+
+
 ## Candidate Gathering {#gathering}
 
 The step of gathering candidates involves identifying which paths, protocols, and endpoints may be used for a given Connection. This list is determined by the requirements, prohibitions, and preferences of the application as specified in the Selection Properties.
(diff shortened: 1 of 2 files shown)


draft-ietf-taps-impl.md
```diff
@@ -173,6 +173,12 @@ Aggregate [Endpoint: www.example.com:80] [Interface: Any]   [Protocol: TCP]
 
 Any one of these sub-entries on the aggregate connection attempt would satisfy the original application intent. The concern of this section is the algorithm defining which of these options to try, when, and in what order.
 
+During Candidate Gathering, an implementation first excludes all protocols and
+paths that match a Prohibit and all protocols and paths that do not match a
+Require. Then, the implementation will sort branches according to Preferred
```
philsbln: I would try to avoid having three "and"s in the sentence: 
```suggestion
During Candidate Gathering, an implementation first excludes all protocols and
paths that match a Prohibit or do not match all Require properties.
Then, the implementation will sort branches according to Preferred
```

theri: Good suggestion, thanks!

draft-ietf-taps-impl.md
```diff
@@ -372,7 +378,9 @@ An implementation may use the Capacity Profile to prefer paths optimized for the
 
 Implementations should process properties in the following order: Prohibit, Require, Prefer, Avoid.
 If Selection Properties contain any prohibited properties, the implementation should first purge branches containing nodes with these properties. For required properties, it should only keep branches that satisfy these requirements. Finally, it should order branches according to preferred properties, and finally use avoided properties as a tiebreaker.
+When ordering branches, an implementation may give more weight to properties that the application has explicitly set than to properties that are default.
 
+As the available protocols and paths on a specific system and in a specific context may vary, the result of sorting and the outcome of racing may vary even given the same Selection and Connection Properties. However, an implementation may aim to provide a consistent outcome to applications, e.g., by preferring protocols and paths that existing Connections with similar Properties are already using.
```
gorryfair: I'd still love to say "ought to aim" 

and then (maybe in there implementation) say  are RECOMMENDED to aim...
- I thin k this is a very useful goal, the SHOULD/RECOMMENDED is admittedly a little weak, in that it's hard t precisely conform, but I still think the intent is important to the usefulness of the TAPS system.

draft-ietf-taps-interface.md
```diff
@@ -731,21 +731,20 @@ In addition, the pseudo-level ``Default`` can be used to reset the property to t
...
 example, if an application indicates a preference for a specific path by
 specifying an interface, but also a preference for a protocol, a situation
 might occur in which the preferred protocol is not available on the preferred
-path. In such cases, implementations MUST ensure a deterministic outcome by
-prioritizing Selection Properties that select paths over those that select
-protocols. Therefore, the transport system MUST try the path first, ignoring
-the protocol preference if the protocol does not work on the path.
+path. In such cases, implementations SHOULD prioritize Selection Properties
+that select paths over those that select protocols. Therefore, the transport
+system SHOULD try the path first, ignoring the protocol preference if the
```
gorryfair: /try the path/ seems a tad loose, do you mean to race the path first, or evaluate the path or something else?
... I don't know the best words:-)

draft-ietf-taps-impl.md
theri: Sure, "ought to aim" sounds good to me as well.
We don't have normative language in the implementation document, but I'll put a sentence in the API document saying RECOMMENDED.

draft-ietf-taps-interface.md
theri: I'd say "race" works, because it implies that branches with the preferred path are sorted to the top, thus, raced first. Will change, thanks for the suggestion!

PR #512, opened 2020-03-03
Update Multipath properties - closes #437

### LINKED ISSUE #437: Configuring MP active/passive
For MP use we have the following option: Disable, Handover, Interactive, Aggregate. 

I think the option as not well defined (except Disable). Maybe the following is a bit better:

Handover: : The connection should attempt to migrate to a new paths when interface availability changes and performance of the currently used path decreased below a certain threshold. This threshold is implementation-specific.

Interactive: : The connection should attempt to use multiple paths in parallel in order to always select the best path available or at least achieve about the same aggregated bandwidth as provided by the best of the available paths. This can be achieved by using coupled congestion control as proposed in {{?RFC6356}} for MP-TCP.

Aggregate: : The connection should attempt to use multiple paths in parallel in order to maximize bandwidth. That usually means that independent congestion control instances are used on each path.

I also proposed another option:

Redundant: : The connection should send the same data on two of the available paths. A message property could be used to indicate which messages should be send redundantly.

And further I think we also need a configuration option to indicate when/if a new subpath should be opened. I think there are four options: 

Never-passive (rely on other ending opening new paths and never send new available IPs),
Never-active (rely on other ending opening new paths but signal send new available IPs), 
Immediate (open new path when IP because available locally or provided by other end), 
On-Request (application needs to indicate).

Note that there is also rfc6897 specifying a rather different interface...

philsbln:
Closes #437.
* allow to control active/passive opertaion
* clairify functionality
* allow to control exposure of alternative addresses

### DIFF
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -1039,7 +1039,7 @@ Note that if an application Requires the use of temporary addresses, the
 resulting Connection cannot use IPv4, as temporary addresses do not exist in
 IPv4.
 
-### Parallel Use of Multiple Paths {#multipath-mode}
+### Multi-Paths Transport {#multipath-mode}
 
 Name:
 : multipath
@@ -1060,20 +1060,20 @@ Disabled:
 : The connection will not use multiple paths once established, even if the chosen transport supports using multiple paths.
 
 Active:
-: The connection will negotiate the use of multiple paths if the chosen transport supports this. It will actively bring up paths based on the multipath policy ({{multipath-objective}}) selected by the application, as limited by the Interface Instance and Provisioning Domain Instance properties (see {{prop-interface}} and {{prop-pvd}}).
+: The connection will negotiate the use of multiple paths if the chosen transport supports this. It will actively bring up paths based on the multipath policy ({{multipath-policy}}) selected by the application, as limited by the Interface Instance and Provisioning Domain Instance properties (see {{prop-interface}} and {{prop-pvd}}).
 
 Passive:
 : The connection will support the use of multiple paths if the remote endpoint requests it. The connection will not actively initiate the use of any new paths, but will accept paths initiated by the remote endpoint.
 
-The objective for using multiple paths is specified using the separate ```multipath-mode property```, see {{multipath-objective}} below.
+The policy for using multiple paths is specified using the separate ```multipath-policy``` property, see {{multipath-policy}} below.
 To enable the peer endpoint to initiate additional paths towards a local address other than the one initially used, it is necessary to set the Alternative Addresses property (see {{altaddr}} below).
 
-Setting this property to "Active", may have privacy implications: It enables the transport to establish connectivity using alternate paths that may make users linkable across multiple paths, even if the Exposure of Alternative Addresses property (see {{altaddr}} below) is set to false.
+Setting this property to "Active", may have privacy implications: It enables the transport to establish connectivity using alternate paths that may make users linkable across multiple paths, even if the Advertisement of Alternative Addresses property (see {{altaddr}} below) is set to false.
 
 Enumeration values other than "Disabled" are interpreted as a preference for choosing protocols that can make use of multiple paths.
-The "Disabled" value implies a requirement not to use multiple paths in parallel but does not prevent choosing a protocol that is capable of using multiple paths.
+The "Disabled" value implies a requirement not to use multiple paths in parallel but does not prevent choosing a protocol that is capable of using multiple paths, e.g., it does not prevent choosing TCP, but prevents sending the ```MP_CAPABLE``` option in the TCP handshake.
 
-### Exposure of Alternative Addresses {#altaddr}
+### Advertisement of Alternative Addresses {#altaddr}
 
 Name:
 : advertises-altaddr
@@ -2408,10 +2408,10 @@ The Capacity Profile for a selected protocol stack may be modified on a
 per-Message basis using the Transmission Profile Message Property; see
 {{send-profile}}.
 
-### Ojective of Using Multiple Paths {#multipath-objective}
+### Policy for using Multi-Path Transports {#multipath-policy}
 
 Name:
-: multipath-objective
+: multipath-policy
 
 Type:
 : Enumeration
@@ -2419,22 +2419,18 @@ Type:
 Default:
 : Handover
 
-This property specifies the local objective of transferring data across multiple paths between the same end hosts if Parallel Use of Multiple Paths not set to Disabled (see {{multipath-mode}}). Possible values are:
-
-None:
-: Do not use other paths as long as the initial path is available, but allow the peer endpoint to make use it to achieve its objective.
+This property specifies the local policy of transferring data across multiple paths between the same end hosts if Parallel Use of Multiple Paths not set to Disabled (see {{multipath-mode}}). Possible values are:
 
 Handover:
-: The connection should only attempt to migrate between different paths when the original path is lost or becomes unreliable.
-The actual thresholds to declare a path unreliable are implementation specific.
+: The connection should only attempt to migrate between different paths when the original path is lost or becomes unusable. The actual thresholds to declare a path unusable are implementation specific.
 
 Interactive:
-: The connection should attempt to use multiple paths in parallel in order to minimize loss and delay. The actual strategy is implementation specific and may depend on the multipath protocol used, but should not aim to exceed the capacity provided by the best of the available paths.
+: The connection should attempt to use multiple paths in parallel in order to minimize loss and delay. The actual strategy is implementation specific, but should not exceed the capacity provided by the best of the available paths.
 
 Aggregate:
-: The connection should attempt to use multiple paths in parallel in order to maximize bandwidth, possibly trading delay for bandwidth. The actual strategy is implementation specific.
+: The connection should attempt to use multiple paths in parallel in order to maximize bandwidth and possibly overcome bandwidth limitations of the individual paths. The actual strategy is implementation specific.
 
-Note that the peer endpoint can choose a different objective.
+Note that this is a local choice – the peer endpoint can choose a different policy.
 
 ### Bounds on Send or Receive Rate
 
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -1063,7 +1063,7 @@ Active:
 : The connection will negotiate the use of multiple paths if the chosen transport supports this. It will actively bring up paths based on the multipath policy ({{multipath-objective}}) selected by the application, as limited by the Interface Instance and Provisioning Domain Instance properties (see {{prop-interface}} and {{prop-pvd}}).
 
 Passive:
-: The transport system will not actively try to establish additional paths once the inital connection is established, but allows the peer endpoint to initiate additional paths and migrate existing connections to other paths.
+: The connection will support the use of multiple paths if the remote endpoint requests it. The connection will not actively initiate the use of any new paths, but will accept paths initiated by the remote endpoint.
 
 The objective for using multiple paths is specified using the separate ```multipath-mode property```, see {{multipath-objective}} below.
 To enable the peer endpoint to initiate additional paths towards a local address other than the one initially used, it is necessary to set the Alternative Addresses property (see {{altaddr}} below).
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -1032,8 +1032,7 @@ Note that if an application Requires the use of temporary addresses, the
 resulting Connection cannot use IPv4, as temporary addresses do not exist in
 IPv4.
 
-
-### Parallel Use of Multiple Paths {#parallel-multipath}
+### Parallel Use of Multiple Paths {#multipath-mode}
 
 Name:
 : multipath
@@ -1045,36 +1044,27 @@ Default:
 : Disabled for connections created through initiate and rendezvous, Passive for listeners
 
 This property specifies whether and how applications want to take advantage of
-transferring data across multiple paths between the same end hosts. Using
-multiple paths allows connections to migrate between interfaces or aggregate bandwidth
-as availability and performance properties change. Possible values are:
+transferring data across multiple paths between the same end hosts.
+Using multiple paths allows connections to migrate between interfaces or aggregate bandwidth
+as availability and performance properties change.
+Possible values are:
 
 Disabled:
-: The connection will not allow using multiple paths once established
+: The connection will not allow using multiple paths once established and will disable respective functionality in case the chosen transport supports it.
+
+Active:
+: The transport system will actively tries, if a to establish additional paths once the inital connection is established if advisable by the {{multipath-objective}} chosen by the application. The desireable paths to established are controlled by the Interface Instance and Provisioning Domain Instance properties (see {{prop-interface}} and {{prop-pvd}}).
 
 Passive:
-: The connection will not actively try using multiple paths once established, but allow the peer endpoint to initiate using multiple paths and migrate existing connections to other paths.
+: The transport system will not actively try to establish additional paths once the inital connection is established, but allows the peer endpoint to initiate additional paths and migrate existing connections to other paths.
 
-Handover:
-: The connection should only attempt to migrate between different paths when the original path is lost or becomes unreliable.
-The actual thresholds to declare a path unreliable are implementation specific.
+The objective for using multiple paths is specified using the separate ```multipath-mode property```, see {{multipath-objective}} below.
+To enable the peer endpoint to initiate additional paths towards a local address other than the one initially used, it is necessary to set the Alternative Addresses property (see {{altaddr}} below).
 
-Interactive:
-: The connection should attempt to use multiple paths in parallel in order to minimize loss and delay. The actual strategy is implementation specific and may depend on the multipath protocol used, but should not aim to exceed the capacity provided by the best of the available paths.
+Setting this property to "Active", may have privacy implications: It enables the transport to establish connectivity using alternate paths that may make users linkable across multiple paths, even if the Exposure of Alternative Addresses property (see {{altaddr}} below) is set to false.
 
-Passive-Interactive:
-: As Interactive, but only using paths established by the peer endpoint.
-
-Aggregate:
-: The connection should attempt to use multiple paths in parallel in order to maximize bandwidth, possibly trading delay for bandwidth. The actual strategy is implementation specific.
-
-Passive-Aggregate:
-: As Aggregate, but only using paths established by the peer endpoint.
-
-Enumeration values other than "Disabled" are interpreted as a preference for choosing protocols that can make use of multiple paths. 
+Enumeration values other than "Disabled" are interpreted as a preference for choosing protocols that can make use of multiple paths.
 The "Disabled" value implies a requirement not to use multiple paths in parallel but does not prevent choosing a protocol that is capable of using multiple paths.
-
-Setting this property to "Handover", "Interactive" or "Aggregate" may have privacy implications: It enables the transport to establish connectivity using alternate paths that may make users linkable across multiple paths, even if the Exposure of Alternative Addresses property (see {{altaddr}} below) is set to false.
 
 ### Exposure of Alternative Addresses {#altaddr}
 
@@ -1091,7 +1081,7 @@ This property specifies whether alternative addresses, e.g., of other interfaces
 peer endpoint by the protocol stack. Advertising these addresses enables the peer-endpoint to establish additional connectivity, e.g., for connection migration or using multiple paths.
 
 Note that this may have privacy implications because it may make users linkable across multiple paths.
-Also, note that setting this to false does not prevent the local transport system from _establishing_ connectivity using alternate paths (see {{parallel-multipath}} above); it only prevents _procative advertisement_ of addresses.
+Also, note that setting this to false does not prevent the local transport system from _establishing_ connectivity using alternate paths (see {{multipath-mode}} above); it only prevents _procative advertisement_ of addresses.
 
 ### Direction of communication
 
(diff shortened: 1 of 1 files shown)


draft-ietf-taps-interface.md
```diff
@@ -1045,26 +1045,53 @@ Type:
...
-: The connection will not attempt using multiple paths once established
+: The connection will not allow using multiple paths once established
+
+Passive:
+: The connection will not actively try using multiple paths once established, but allow the peer endpoint to initiate using multiple paths and migrate existing connections to other paths.
 
 Handover:
-: The connection should attempt to migrate between different paths upon interface availability changes
+: The connection should only attempt to migrate between different paths when the original path is lost or becomes unreliable.
+The actual thresholds to declare a path unreliable are implementation specific.
```
mwelzl: In #437, @tfpauly writes, as a conclusion from the interim discussion: "Exposing alternate addresses to the peer and actively migrating is likely a separate property that we should specify."  I interpret this as meaning that we don't want "Handover" here. If we do want to keep it, I agree with @gorryfair who, in #437, suggests to call this "Failover" instead.

draft-ietf-taps-interface.md
```diff
@@ -1045,26 +1045,53 @@ Type:
...
 
 Interactive:
-: The connection should attempt to use multiple paths in response to loss or delay upon individual paths
+: The connection should attempt to use multiple paths in parallel in order to minimize loss and delay. The actual strategy is implementation specific and  my depend on the multipath protocol used, but should not aim to exceed the bandwidth provided by the best of the available paths.
+
+Passive-Interactive:
+: As Interactive, but only if using multiple paths was initiated by the peer endpoint.
```
mwelzl: I agree with @gorryfair and @abrunstrom who call this "fuzzy" in #437 (this applies to both "Interactive" and "Passive-Interactive". Do we need this now, given that we don't seem to have any RFC explaining how to best implement this?  I think it's cool functionality, but I question whether this should be inside a Transport System - would an application be able to trust this to implement just the right trade-off for it, between e.g. capacity usage and delay minimization?

Side note: if we do keep it, I think the text of this update is better.

draft-ietf-taps-interface.md
philsbln: No, these are still two separate things:
 - Exposure of Alternative Addresses is defined as a separate property below
 - You don't need to expose an alternative addresses prior initiating handover.
 - Servers will most likely never advertise alternative addresses but still want to support the client in migrating.

Maybe make this more explicit that the Multipath property really defines path management behaviour while the exposure property defines advertisement.

draft-ietf-taps-interface.md
philsbln: This is a slightly extended version of Apple's implementation and can be easily implemented, e.g., for MPTCP by choosing the right path manager. It is a matter of choosing "Just use the lowest RTT path" vs. "use all paths"

mirjak:
I think having one parameter for all interfaces is not enough as you might want to decide differently on a per interface basis. You can also have a look at section 5.3 of rfc6897 for what was recommended for MPTCP. Maybe the Apple people have also more input based in what is currently implemented for MPTCP in their stack?

draft-ietf-taps-interface.md
mwelzl: So if you have a long-RTT path with potentially a lot of extra capacity as well as a short-RTT path, this is always going to pick the short-RTT one?  Even if the capacity on the short-RTT one is quite tiny and the RTT difference isn't very large? Or else, how do you choose these thresholds?  => yes there CAN be a simple implementation, but there are also trade-offs in the design.

draft-ietf-taps-interface.md
mwelzl: I saw the Exposure of Alternative Addresses below, and I agree that this is (or can be) separate from "actively migrating". My understanding of the sentence was that both of these things should be a separate property (or are somehow combined into a single separate property).

draft-ietf-taps-interface.md
philsbln: I assume this is a weird corner case, but yes there are trade-offs the implementation must make. Do we need text on that?

mwelzl: I don't think it's a weird corner case; I just don't think that many applications would want to say "give me the lower latency path, at all cost, no matter how miniscule the difference, no matter how much capacity I lose". I'm not proposing to address this with some discussion in text - I'm proposing not to delegate the algorithm that chooses paths on this basis to the transport system, but leaving it inside the application instead.

draft-ietf-taps-interface.md
philsbln: @tfpauly Do you have a preference?

philsbln:
@mirjak I think this would be a cool extension, but I would prefer to have a basic version first that leaves interface specific decisions to the implementation / system policy.

mirjak:
I don't see that as an extension but an alternative. I just believe a generic config to enable this for all interfaces will not be useful at all because you always want to make a decision on a per interface basis.

britram:
Discussion seems ongoing enough that we shouldn't merge for the 107 draft deadline. Keeping this one open.

draft-ietf-taps-interface.md
```diff
@@ -1045,26 +1045,53 @@ Type:
...
 
 Aggregate:
-: The connection should attempt to use multiple paths in parallel in order to maximize bandwidth
+: The connection should attempt to use multiple paths in parallel in order to maximize bandwidth, possibly trading delay for bandwidth. The actual strategy is implementation specific.
+
+Passive-Aggregate:
+: As Aggregate, but only if using multiple paths was initiated by the peer endpoint.
+
+Enumeration values other than "Disabled" and "Passive" are interpreted as preferences.
+
```
abrunstrom: Should this not be other than Disabled. Passive still requires a protocol with multipath support I guess. And should perhaps also say that Disabled is a requirement?

draft-ietf-taps-interface.md
abrunstrom: Leaving it to the application may seem attractive, but is not really feasible in practise as it quickly gets very complicated. For instance, if you ask for a packet to be sent on one path and this packet is lost, are you allowed to retransmit it on another path for reliability? What if the path is lost before you manged to send the packet at all?
I think the updated text is an improvement.

mwelzl: Ok, glad to hear that you like this - I don't personally have a strong opinion, and I agree that it quickly gets complicated!  (but that's also why I felt quite insecure about simply delegating this to the transport system). I would like to hear @tfpauly's opinion too, though.

draft-ietf-taps-interface.md
philsbln: It makes sense for a client to pick active, as ist allows to initiate additional subflows and, thus, implicitly expose some addresses, but not expose other addresses.
I consider exposing addresses without actively establishing sub-flows rather as a corner-case for things like GPT over SCPT or to bootstrap connection migration through NAT.
Note that I use MPTCP terminology here, but I think this is not limited to MPTCP.

draft-ietf-taps-interface.md
philsbln: Ack; I'll fix that

draft-ietf-taps-interface.md
mirjak: You can also be active and only establish connections if a new address is advertised by the peer.

draft-ietf-taps-interface.md
mirjak: I think having "passive" here mixes up two things, one is the question of who will establish paths and the other one is how to use paths. These are separate points and should not be mixed together.

draft-ietf-taps-interface.md
```diff
@@ -1045,26 +1045,54 @@ Type:
...
 
 Aggregate:
-: The connection should attempt to use multiple paths in parallel in order to maximize bandwidth
+: The connection should attempt to use multiple paths in parallel in order to maximize bandwidth, possibly trading delay for bandwidth. The actual strategy is implementation specific.
+
+Passive-Aggregate:
+: As Aggregate, but only if using paths established by the peer endpoint.
+
+Enumeration values other than "Disabled" are interpreted as a preference for choosing protocols that can make use of multiple paths. 
+The "Disabled" value implies a requirement not to use multiple paths in parallel but does not prevent choosing a protocol that is principally capable of using multiple paths.
```
mwelzl, line 1078: I think that "principally" here is a German-ism ... I guess you mean "in principle". However, even this doesn't really add anything useful to the statement, IMO. Can we just remove the word "principally"?

draft-ietf-taps-interface.md
```diff
@@ -1045,26 +1045,54 @@ Type:
...
+: The connection will not actively try using multiple paths once established, but allow the peer endpoint to initiate using multiple paths and migrate existing connections to other paths.
 
 Handover:
-: The connection should attempt to migrate between different paths upon interface availability changes
+: The connection should only attempt to migrate between different paths when the original path is lost or becomes unreliable.
+The actual thresholds to declare a path unreliable are implementation specific.
 
 Interactive:
-: The connection should attempt to use multiple paths in response to loss or delay upon individual paths
+: The connection should attempt to use multiple paths in parallel in order to minimize loss and delay. The actual strategy is implementation specific and  my depend on the multipath protocol used, but should not aim to exceed the bandwidth provided by the best of the available paths.
```
mwelzl, line 1066: s/my depend/may depend
and, please: s/bandwidth provided/capacity provided

draft-ietf-taps-interface.md
```diff
@@ -1045,26 +1045,54 @@ Type:
...
+
+Passive-Interactive:
+: As Interactive, but only using paths established by the peer endpoint.
 
 Aggregate:
-: The connection should attempt to use multiple paths in parallel in order to maximize bandwidth
+: The connection should attempt to use multiple paths in parallel in order to maximize bandwidth, possibly trading delay for bandwidth. The actual strategy is implementation specific.
+
+Passive-Aggregate:
+: As Aggregate, but only if using paths established by the peer endpoint.
```
mwelzl, line 1075: s/only if using/only using

draft-ietf-taps-interface.md
```diff
@@ -1045,26 +1045,54 @@ Type:
...
+Name:
+: advertises-altaddr
+
+Type:
+: Boolean
+
+Default:
+: False
+
+This property specifies whether addresses of other interfaces should be advertises to the
```
mwelzl, line 1092: s/advertises/advertised

philsbln:
@abrunstrom I clarified the relationship between the two multi-path properties. Please review whether this works for you now, as I don't think merging them would make things clearer.

With regards to the per-interface choice, I would prefer to use Interface Instance or Type and add a few notes how these effect multi-path in a separate PR

tfpauly:
Interim discussion:

Possible knobs:
- Who opens the paths? (active/passive/disabled)
- Should we announce our other IP addresses? (allowed or not)
- How are packets scheduled on the subpaths? (handover, interactive, aggregate)

philsbln:
I made a new version based on the decisions we made in the interim. Please review.

draft-ietf-taps-interface.md
```diff
@@ -1049,26 +1048,47 @@ Type:
...
-availability and performance properties change. Possible values are:
+This property specifies whether and how applications want to take advantage of
+transferring data across multiple paths between the same end hosts.
+Using multiple paths allows connections to migrate between interfaces or aggregate bandwidth
+as availability and performance properties change.
+Possible values are:
 
 Disabled:
-: The connection will not attempt using multiple paths once established
+: The connection will not allow using multiple paths once established and will disable respective functionality in case the chosen transport supports it.
```
tfpauly, line 1060:
```suggestion
: The connection will not use multiple paths once established, even if the chosen transport supports using multiple paths.
```

draft-ietf-taps-interface.md
```diff
@@ -1049,26 +1048,47 @@ Type:
...
 
-Handover:
-: The connection should attempt to migrate between different paths upon interface availability changes
+Active:
+: The transport system will actively tries, if a to establish additional paths once the inital connection is established if advisable by the {{multipath-objective}} chosen by the application. The desireable paths to established are controlled by the Interface Instance and Provisioning Domain Instance properties (see {{prop-interface}} and {{prop-pvd}}).
```
tfpauly, line 1063:
```suggestion
: The connection will negotiate the use of multiple paths if the chosen transport supports this. It will actively bring up paths based on the multipath policy ({{multipath-objective}}) selected by the application, as limited by the Interface Instance and Provisioning Domain Instance properties (see {{prop-interface}} and {{prop-pvd}}).
```

draft-ietf-taps-interface.md
```diff
@@ -1049,26 +1048,47 @@ Type:
...
 
-Interactive:
-: The connection should attempt to use multiple paths in response to loss or delay upon individual paths
+Passive:
+: The transport system will not actively try to establish additional paths once the inital connection is established, but allows the peer endpoint to initiate additional paths and migrate existing connections to other paths.
```
tfpauly, line 1066:
```suggestion
: The connection will support the use of multiple paths if the remote endpoint requests it. The connection will not actively initiate the use of any new paths, but will accept paths initiated by the remote endpoint.
```

draft-ietf-taps-interface.md
```diff
@@ -1039,8 +1039,7 @@ Note that if an application Requires the use of temporary addresses, the
 resulting Connection cannot use IPv4, as temporary addresses do not exist in
 IPv4.
 
-
-### Parallel Use of Multiple Paths {#parallel-multipath}
+### Parallel Use of Multiple Paths {#multipath-mode}
```
tfpauly, line 1042: How about "Support for Multiple Parallel Paths"?

draft-ietf-taps-interface.md
```diff
@@ -1049,26 +1048,47 @@ Type:
...
-: The connection should attempt to use multiple paths in parallel in order to maximize throughput
+The objective for using multiple paths is specified using the separate ```multipath-mode property```, see {{multipath-objective}} below.
+To enable the peer endpoint to initiate additional paths towards a local address other than the one initially used, it is necessary to set the Alternative Addresses property (see {{altaddr}} below).
+
+Setting this property to "Active", may have privacy implications: It enables the transport to establish connectivity using alternate paths that may make users linkable across multiple paths, even if the Exposure of Alternative Addresses property (see {{altaddr}} below) is set to false.
+
+Enumeration values other than "Disabled" are interpreted as a preference for choosing protocols that can make use of multiple paths.
+The "Disabled" value implies a requirement not to use multiple paths in parallel but does not prevent choosing a protocol that is capable of using multiple paths.
+
+### Exposure of Alternative Addresses {#altaddr}
```
tfpauly, line 1076: Advertisement of Alternative Addresses?

draft-ietf-taps-interface.md
```diff
@@ -2388,6 +2408,33 @@ The Capacity Profile for a selected protocol stack may be modified on a
 per-Message basis using the Transmission Profile Message Property; see
 {{send-profile}}.
 
+### Ojective of Using Multiple Paths {#multipath-objective}
```
tfpauly, line 2411: Typo on Ojective -> Objective. Beyond that, I'd call this "Multipath Policy" or something similar. Objective doesn't seem like the right word.

draft-ietf-taps-interface.md
```diff
@@ -2388,6 +2408,33 @@ The Capacity Profile for a selected protocol stack may be modified on a
...
+Type:
+: Enumeration
+
+Default:
+: Handover
+
+This property specifies the local objective of transferring data across multiple paths between the same end hosts if Parallel Use of Multiple Paths not set to Disabled (see {{multipath-mode}}). Possible values are:
+
+None:
+: Do not use other paths as long as the initial path is available, but allow the peer endpoint to make use it to achieve its objective.
```
tfpauly, line 2425: Does this value make sense? It seems like it's a duplicate of Passive; why not just say Passive doesn't use the multi path policy?

draft-ietf-taps-interface.md
```diff
@@ -2388,6 +2408,33 @@ The Capacity Profile for a selected protocol stack may be modified on a
...
+
+Handover:
+: The connection should only attempt to migrate between different paths when the original path is lost or becomes unreliable.
```
tfpauly, line 2428: Rather than unreliable, let's say "unusable". Unreliable could just mean it has some loss—many links aren't reliable, but are good enough to use.

draft-ietf-taps-interface.md
```diff
@@ -2388,6 +2408,33 @@ The Capacity Profile for a selected protocol stack may be modified on a
...
+The actual thresholds to declare a path unreliable are implementation specific.
+
+Interactive:
+: The connection should attempt to use multiple paths in parallel in order to minimize loss and delay. The actual strategy is implementation specific and may depend on the multipath protocol used, but should not aim to exceed the capacity provided by the best of the available paths.
+
+Aggregate:
+: The connection should attempt to use multiple paths in parallel in order to maximize bandwidth, possibly trading delay for bandwidth. The actual strategy is implementation specific.
+
+Note that the peer endpoint can choose a different objective.
```
tfpauly, line 2437: Again, policy rather than objective?

draft-ietf-taps-interface.md
mwelzl: Okay, sorry, but I must ask: how do you people do this? I can't find how to make such direct text change proposals

draft-ietf-taps-interface.md
philsbln: This had been my first idea too, but calling it policy does not really tell you whether it is the policy with regards to allowing multipath or the policy how to perform multipath. Objective makes that clear. Any better suggejstions?

gorryfair: Sorry Phil, I don't understand "objective" in this context, if we need a word to describe the mode, we need another word.

draft-ietf-taps-interface.md
philsbln: This property only effects the packet scheduler, not to the path manager as above. 
As a server, you may choose passive above and "Aggregate" or "Interactive" here.

I am unsure whether we need it, as the difference i see between "None" and "Handover" is really small. In iOS, I would expect it turns off WiFi Assist and does hand-over once a path breaks.

draft-ietf-taps-interface.md
philsbln: This depends on whether to include "None" or not – if we drop "None", we should change it to unusable, but otherwise, I see more freedom to declare a path "bad enough to hand over to something else".

draft-ietf-taps-interface.md
```diff
@@ -1049,26 +1048,47 @@ Type:
 : Enumeration
 
 Default:
-: Disabled
+: Disabled for connections created through initiate and rendezvous, Passive for listeners
```
mirjak, line 1051: I afraid I don't fully understand this. In case of MPTCP does disabled mean I will not open any new sub flows, reject all sub flows (SYNs) from the other end, or not even include the MPTCP option in the handshake of the initial connection?

draft-ietf-taps-interface.md
mirjak: yes objective doesn't seem right. I think policy is fine here.

draft-ietf-taps-interface.md
mirjak: I would remove this and only have "handover"

draft-ietf-taps-interface.md
mirjak: +1 to unusable or another term. unreliable doesn't seem right

draft-ietf-taps-interface.md
```diff
@@ -2388,6 +2408,33 @@ The Capacity Profile for a selected protocol stack may be modified on a
...
+
+None:
+: Do not use other paths as long as the initial path is available, but allow the peer endpoint to make use it to achieve its objective.
+
+Handover:
+: The connection should only attempt to migrate between different paths when the original path is lost or becomes unreliable.
+The actual thresholds to declare a path unreliable are implementation specific.
+
+Interactive:
+: The connection should attempt to use multiple paths in parallel in order to minimize loss and delay. The actual strategy is implementation specific and may depend on the multipath protocol used, but should not aim to exceed the capacity provided by the best of the available paths.
```
mirjak, line 2432: Actually you don't use multiple that _in parallel_ You always just use one but decide to switch over based on latency. I'm also not sure I understand the second sentence. Is that needed?

draft-ietf-taps-interface.md
```diff
@@ -2388,6 +2408,33 @@ The Capacity Profile for a selected protocol stack may be modified on a
...
+
+Aggregate:
+: The connection should attempt to use multiple paths in parallel in order to maximize bandwidth, possibly trading delay for bandwidth. The actual strategy is implementation specific.
```
mirjak, line 2435: Not sure we need this half sentence about "trading delay for bandwidth". If the "needed" capacity is larger than what a single path can provide, there is no other chance than to use both. If you don't want that choose the "interactive" option instead

draft-ietf-taps-interface.md
mirjak: Or "management of multiple parallel paths"...?

draft-ietf-taps-interface.md
mirjak: Or let me put this differently: if multi path support is enabled and a MP option sent in the initial handshake, I think we only need "active" and "passive" as modes because otherwise if I reject all your SYNs why should I even include the MPTCP option in the initial handshake?

philsbln: We need Disabled here:
- Disabled results in not sending the MPTCP option, 
- Passive results in sending the MPTCP option it but not opening additional subflows.
- Active results in sending the MPTCP option and opening additional subflows

draft-ietf-taps-interface.md
philsbln: You may, e.g., to probe/monitor the other paths latency to have a decision criteria when to switch.

draft-ietf-taps-interface.md
mirjak: Still not sure here because the decision to open a new subflows really depends on the path/interface. Not sure you even want to say: whenever you see a new IP address, immediately open a new subflow. So I think you only want disable and enable here and leave the decision to open new subflows on the interface level.

philsbln: I agree that we should have some control on the Interface level, but relying on this also for the active/passive path establishment looks to over-complicate things for me.
This Property just controls whether the local side accepts additional incoming sub-flows or whether it also actively tries to establish them. I suspect that this one most applications can easily set:
- Passive in the Server-Case
- Active or Disabled in the Client case
The per-interface one is a major hassle and I believe most programmer really want the implementation to take care here.

What we should discuss is whether we want to make "Passive" also the default for connections created through initiate and rendezvous…

draft-ietf-taps-interface.md
```diff
@@ -2388,6 +2408,29 @@ The Capacity Profile for a selected protocol stack may be modified on a
...
+Default:
+: Handover
+
+This property specifies the local policy of transferring data across multiple paths between the same end hosts if Parallel Use of Multiple Paths not set to Disabled (see {{multipath-mode}}). Possible values are:
+
+Handover:
+: The connection should only attempt to migrate between different paths when the original path is lost or becomes unusable. The actual thresholds to declare a path unusable are implementation specific.
+
+Interactive:
+: The connection should attempt to use multiple paths in parallel in order to minimize loss and delay. The actual strategy is implementation specific, but should not exceed the capacity provided by the best of the available paths.
```
tfpauly, line 2428:
```suggestion
: The connection should attempt to use multiple paths in parallel in order to minimize loss and delay. The actual strategy is implementation specific.
```

PR #527, opened 2020-04-02
Closes #515.  + addresses a few more of the easy things from Kyle's review.

### LINKED ISSUE #515: Bandwidth Efficiency as a term in API
I am going to query  /bandwidth efficiency/ - since this sounds like it should be measured in b/Hz, which isn't at all what this draft are talking about. In fact, I did not see anything in the text about the efficiency or the use of bandwith ... as I read, it is all about capacity sharing tradeoffs.

I suspect the tyerm should be "capacity-seeking" or we just say latency-seeking rather than capacity-seeking? 

--
And while I am digging here ... how about /to use multiple paths in parallel in order to maximize bandwidth/to use multiple paths in parallel in  maximize use of available capacity/.

mwelzl:
... such as: using CamlCase instead of dashes for property names.
Also: fixes consistency problem of apostrophs / quotation marks  (` vs `` vs " )

### DIFF
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -508,7 +508,7 @@ form \[\<Namespace>.\]\<PropertyName\>.
 - The Namespace part MUST be empty for well-known, generic properties, i.e., for
   properties that are not specific to a protocol and are defined in an RFC.
 - Protocol Specific Properties MUST use the protocol acronym as Namespace, e.g.,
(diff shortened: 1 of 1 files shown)


mwelzl:
I'll just merge this now, to avoid conflicts with a new PR I'm writing (I'd like to continue using camlCase for property names). Moving to camlCase has been discussed on the list, and the other things here - ensuring uniform appearance of quotation marks, and replacing "-1" with special values - seem quite non-controversial.

PR #535, opened 2020-04-06
Adjust level of branching related subsections

mirjak:
I think this is just an minor oversight that needs fixing but I would like someone to double-check before merged...


tfpauly:
Yes, I think that this is more correct also. Thanks!

Issue #508, opened 2020-02-21
Discuss interaction between Listen() and ListenClone()


tfpauly:
- Should ListenClone() fail for non-multiplexed transports
- How to specify if a top-level listener can accept cloned (multiplexed) streams
- Should ListenClone() be called ListenForClone()...?

mwelzl:
Just a note to say that issue #507 waits for this issue to be addressed.

tfpauly:
- Rename function
- Trying to do this without a multiplexing transport should fail

mwelzl:
ping: issue #507 is waiting for this...

Issue #507, opened 2020-02-21
More text on clone


mwelzl:
Currently the "Clone" part of, e.g., TCP (section 10.1) only says:
"Calling Clone on a TCP Connection creates a new Connection with equivalent parameters. The two Connections are otherwise independent."

This text should reflect that equivalent behavior must be maintained upon later Connection configuration calls. And, there should be more text in general on cloning (see issue #441).

mwelzl:
This text must refer to a new "do I want a Connection that supports initiate followed by listen?" property. This property will be added to address #508 - so I'll address this issue when the PR addressing #508 has landed.

PR #541, opened 2020-04-07
Added a selection property for a special communication pattern

### LINKED ISSUE #441: Questions on Connection Groups
A couple of questions we might want to address regarding Connection Groups.

1) Can the application query the interface to figure if two connection in a group share the same transport connection but on different streams or not. Is that useful? I thinks we should explicitly say that if a multiplex protocol is used, calling Clone() MUST create a new stream (and not a new connection).

2) If a new connection is created by a Listener it is automatically part of a group, right? Can I detach that connection from the group? Otherwise how can I change a property only for that connection? Isn't is actually confusing that these connection are grouped?

2) Should we add a pointer to coupled congestion control (rmcat) in that section?

mwelzl:
.. that gets in the way of using streams when cloning.
Closes #441.

### DIFF
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -124,17 +124,6 @@ informative:
         name: Stein Gjessing
     date: 2018-04-15
 
-  NEAT-flow-mapping:
-    title: Transparent Flow Mapping for NEAT
-    seriesinfo:
-        Workshop on Future of Internet Transport (FIT 2017)
-    authors:
-      -
-        ins: F. Weinrank
-     -
-        ins: M. Tuexen
-    date: 2017
-
 
 --- abstract
 
@@ -1157,7 +1146,7 @@ Requiring this property limits the choice of mappings to underlying protocols,
 which can reduce
 efficiency. For example, it prevents the transport system from mapping
 Connections to SCTP streams, where
-the first transmitted data takes the role of active open {{NEAT-flow-mapping}}.
+the first transmitted data takes the role of an active open signal {{I-D.ietf-taps-impl}}.
 
 
 ## Specifying Security Parameters and Callbacks {#security-parameters}
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -1125,7 +1125,7 @@ not all ICMP errors will necessarily be delivered, so applications cannot rely
 on receiving them.
 
 
-### Active side is not the first to write {#active-read-before-send}
+### Initiating side is not the first to write {#active-read-before-send}
 
 Name:
 : activeReadBeforeSend
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -1140,8 +1140,9 @@ The most common client-server communication pattern involves the
 client actively opening a connection, then sending data to the server. The
 server listens (passive open), reads, and then answers. This property
 specifies whether an application wants to diverge from this pattern -- either
-by actively opening, immediately followed by reading, or passively opening,
-immediately followed by writing.
+by actively opening with Initiate(), immediately followed by reading, or passively opening with Listen(),
+immediately followed by writing. This property is irrelevant for connection
+establishment using Rendezvous().
 Requiring this property limits the choice of mappings to underlying protocols,
 which can reduce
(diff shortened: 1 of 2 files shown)


draft-ietf-taps-interface.md
```diff
@@ -1124,6 +1125,30 @@ not all ICMP errors will necessarily be delivered, so applications cannot rely
...
+
+Default:
+: Ignore
+
+The most common client-server communication pattern involves the
+client actively opening a connection, then sending data to the server. The
+server listens (passive open), reads, and then answers. This property
+specifies whether an application wants to diverge from this pattern -- either
+by actively opening, immediately followed by reading, or passively opening,
+immediately followed by writing.
```
philsbln, line 1144: I get the problem for read-after-active-open, but what is the problem with passive open and immediately writing – this is fairly common in TCP based protocols and you can easily put the write in the connection ready.

mwelzl: It's the same thing with multi-streaming: a stream only comes into being when the first message is sent on it. So, "passive open and immediately writing" doesn't map onto anything with streams - there is no such thing as a stream becoming available without data arriving on it.

draft-ietf-taps-interface.md
```diff
@@ -1124,6 +1125,30 @@ not all ICMP errors will necessarily be delivered, so applications cannot rely
 on receiving them.
 
 
+### Active side is not the first to write {#active-read-before-send}
```
philsbln, line 1128:
```suggestion
### Initiating side is not the first to write {#active-read-before-send}
```

csperkins:
How does this interact with `Rendezvous()`?

draft-ietf-taps-interface.md
```diff
@@ -1124,6 +1125,30 @@ not all ICMP errors will necessarily be delivered, so applications cannot rely
...
+
+Name:
+: activeReadBeforeSend
+
+Type:
+: Preference
+
+Default:
+: Ignore
+
```
csperkins, line 1138: I wonder if this should be `Type: Enumeration`, with values `InitiatorSendsFirst` (the default) and `ListenerSendsFirst`?

mwelzl: Why make it so limiting? Preference allows for values such as Prefer, Avoid, Require, ...

mwelzl:
@csperkins - this now states that it is irrelevant for Rendezvous(), and explicitly mentions the Initiate() and Listen() calls when talking about actively and passively opening a Connection.

Why is it irrelevant?

Remember, the problem is that you actively open a Connection that could be mapped to a stream, in which case an active Open doesn't do anything on the wire - and then, right away, you want to read. Then the peer never heard from you, you haven't even tried sending anything to it.

Enabling this property will allow the TAPS system to carry out such a Connection = stream mapping. When calling rendezvous(), however, just using a stream is out of the question - you need signaling, and chances are that you're going to actively send stuff in any case. In other words, I don't think there's a mapping that would work with rendezvous() that enabling this property would allow, and hence the property is irrelevant for rendezvous().

mwelzl:
Merging this now: I think I have addressed the comment by @csperkins. We can always fix things later again if we must - but for now, this closes #441, and it also relates to #508, which gates #507.... so it's probably better to move forward at this point.

draft-ietf-taps-interface.md
csperkins: How does a preference work here? If the application protocol requires the listener to send first, then it need to do this or fail.

draft-ietf-taps-interface.md
```diff
@@ -1124,6 +1125,31 @@ not all ICMP errors will necessarily be delivered, so applications cannot rely
...
+The most common client-server communication pattern involves the
+client actively opening a connection, then sending data to the server. The
+server listens (passive open), reads, and then answers. This property
+specifies whether an application wants to diverge from this pattern -- either
+by actively opening with Initiate(), immediately followed by reading, or passively opening with Listen(),
+immediately followed by writing. This property is irrelevant for connection
+establishment using Rendezvous().
```
csperkins, line 1145: "irrelevant" isn't right here - maybe "is ignored"?

mwelzl: ACK, fixed with a commit to the master

PR #544, opened 2020-04-07
Editorial edits on fast open

mirjak:


mirjak:
I would further recommend to move this section into 4 somewhere as this section is more about connection establishment than sending messages

draft-ietf-taps-impl.md
```diff
@@ -531,13 +531,13 @@ If a Connection becomes finished before a requested Receive action can be satisf
...
+If the application has 0-RTT data to send in any protocol handshakes, it needs to provide this data before any handshake is initiated. When racing, this means that the data should be provided before the process of connection establishment has begun. If the application wants to send 0-RTT data, it indicates this to the implementation by setting the Idempotent flag in the message context to true and providing initial data to send at the connection initiation. In general, 0-RTT data may be replayed (for example, if a TCP SYN contains data, and the SYN is retransmitted, the data will be retransmitted as well but may be considered as a new connection instead of a retransmission). Also when racing connections, different leaf nodes may send the same data independently. If data is truly idempotent, this should be permissible.
```
tfpauly, line 534:
```suggestion
If the application has 0-RTT data to send in any protocol handshakes, it needs to provide this data before any handshake is initiated. When racing, this means that the data should be provided before the process of connection establishment has begun. If the application wants to send 0-RTT data, it indicates this to the implementation by setting the Idempotent flag in the message context to true and providing initial data to send at the connection initiation. In general, 0-RTT data may be replayed (for example, if a TCP SYN contains data, and the SYN is retransmitted, the data will be retransmitted as well but may be considered as a new connection instead of a retransmission). Also, when racing connections, different leaf nodes may send the same data independently. If data is truly idempotent, this should be permissible.
```

draft-ietf-taps-impl.md
```diff
@@ -531,13 +531,13 @@ If a Connection becomes finished before a requested Receive action can be satisf
...
+0-RTT handshakes often rely on previous state, such as TCP Fast Open cookies, previously established TLS tickets, or out-of-band distributed pre-shared keys (PSKs). Implementations should be aware of security concerns around using these tokens across multiple addresses or paths when racing. In the case of TLS, any given ticket or PSK should only be used on one leaf node as servers likely will not accept dublicate tickets to prevent replays (see section-8.1 {{?RFC8446}}). If implementations have multiple tickets available from a previous connection, each leaf node attempt can use a different ticket. In effect, each leaf node will send the same early application data, yet encoded (encrypted) differently on the wire.
```
tfpauly, line 540:
```suggestion
0-RTT handshakes often rely on previous state, such as TCP Fast Open cookies, previously established TLS tickets, or out-of-band distributed pre-shared keys (PSKs). Implementations should be aware of security concerns around using these tokens across multiple addresses or paths when racing. In the case of TLS, any given ticket or PSK should only be used on one leaf node, since servers will likely reject duplicate tickets in order to prevent replays (see section-8.1 {{?RFC8446}}). If implementations have multiple tickets available from a previous connection, each leaf node attempt can use a different ticket. In effect, each leaf node will send the same early application data, yet encoded (encrypted) differently on the wire.
```

PR #547, opened 2020-04-07
Add reference to draft-ietf-tcpm-2140bis

mirjak:


draft-ietf-taps-impl.md
```diff
@@ -793,6 +794,8 @@ An implementation should use this information, when possible, to determine prefe
 
 The reasonable lifetime for cached performance values will vary depending on the nature of the value. Certain information, like the connection establishment success rate to a Remote Endpoint using a given protocol stack, can be stored for a long period of time (hours or longer), since it is expected that the capabilities of the Remote Endpoint are not changing very quickly. On the other hand, Round Trip Time observed by TCP over a particular network path may vary over a relatively short time interval. For such values, the implementation should remove them from the cache more quickly, or treat older values with less confidence/weight.
 
+As one example, {{?I-D.ietf-tcpm-2140bis}} provides guidance about sharing of TCP Control Block information between connections on initialization.
```
tfpauly, line 797: This says it is "one example"; what's it an example of? The text above doesn't discuss sharing the cached information with peers. If anything, this is a new suggestion that could start "In addition, cached information may be used to share information with peers..."

mwelzl: @tfpauly, I don't understand what you say about "peers" here. Obviously, everything here is about caching things locally, but per peer, right?

Regarding the reference, I agree, "As one example," should simply be removed.

PR #554, opened 2020-04-08
preamble text in security considerations section

mirjak:
Mainly adding references to arch and API docs...

### DIFF
--- a/draft-ietf-taps-impl.md
+++ b/draft-ietf-taps-impl.md
@@ -1161,6 +1161,8 @@ This document has no actions for IANA.
 
 # Security Considerations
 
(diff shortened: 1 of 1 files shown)


draft-ietf-taps-impl.md
```diff
@@ -1161,6 +1161,8 @@ This document has no actions for IANA.
 
 # Security Considerations
 
+{{I-D.ietf-taps-arch}} outlines general security consideration and requirements for any system that implements the TAPS archtecture. {{I-D.ietf-taps-interface}} provides further discussion on security and privacy implications of the TAPS API. This document provides additional guidance on implementation specifics for the TAPS API and as such the security considerations in both of these documents apply. The next two subsections discuss further considerations that are specific to mechanisms specified in this document.
```
gorryfair, line 1164: /This document provides/ after describing something,  is often trouble to work which /this/ is. Is it the former or the ID itself? - I think it's the former, can we replace by ", which provides" or something?

PR #556, opened 2020-04-14
API: state "Ephemeral port used when no Local Endpoint specified"

### LINKED ISSUE #539: Listeners and ports
Impl draft section 4.6 (Implementing listeners):
"If no Local Endpoint is specified, the implementation should either use an ephemeral port or generate an error."
Do we need an interface for the application to indicate that an ephemeral port should be used?

mwelzl:
Closes #539.


mwelzl:
Done.

PR #558, opened 2020-04-14
Clarifications about per-connection vs per-message defaults

### LINKED ISSUE #520: Reliable Data Transfer (Message) default - from Kyle Rose's review
List discussion:

**Kyle:**
> 7.4.7. Reliable Data Transfer (Message)
> 
> * The default isn't "true", it's whatever the underlying connection had, right?
> 

**@philsbln**:
>  Ack

**@mwelzl**:

> I agree too, but doesn’t this warrant some more discussion?
> I think our defaults all translate into things that can work in some way when TCP is used - with “Reliable Data Transfer” being an exception, as it just can’t work in a UDP-only system. Are there others?  I think I’ll open an issue for this one.
> 

**@gorryfair**:

> I don't agree at all - To me this was just the default behaviour of the API.
> 
> I think the default needs to be "reliable" and an app will have to over-ride that to get datagram/unreliable operation. That doesn't mean if you have only UDP (for instance) that you can't offer TAPS, to me it simply means you always need to say that the app wants this service.

mwelzl:
Closes #520. I inserted the text that I quoted in this issue, with a minor tweak.


draft-ietf-taps-interface.md
```diff
@@ -1659,6 +1659,8 @@ a Connection must provide reliability to allow setting an infinite value for the
 lifetime property of a Message. Sending a Message with Message Properties
 inconsistent with the Selection Properties of the Connection yields an error.
 
+Connection Properties describe the default behavior for all Messages on a Connection. If a Message Property contradicts a Connection Property, and if this per-Message behavior can be supported, it overrides the Connection Property for the specific Message. For example, if `Reliable Data Transfer (Connection)` is set to `Avoid` and a protocol with configurable per-Message reliability is used, setting `Reliable Data Transfer (Message)` to `Prefer` for a particular Message will ensure that this Message is reliably delivered. If the underlying protocol can not support this per-Message behavior, this Message Property choice will yield an error.
+
```
abrunstrom, line 1663: Reliable Data Transfer (Message) is a Boolean so the 'Prefer' does not make sense. Also in the description of the property is says 

>  Changing the Reliable Data Transfer property on Messages is only possible for Connections that were established with the Selection Property Configure Per-Message Reliability enabled. When this is not the case, changing it will generate an error.

So maybe the text about the error should also be aligned.

mwelzl: Good catch, thanks!  Fixing the "type should be Boolean" problem, and regarding the error, I resorted to repeating the statement that we already have at the end of this paragraph ("Note that...").

abrunstrom: Thanks! Reading it again I am wondering if is is not better to swap the example though? Setting the connection property to required and the per message property to false? I think it will read better.

mwelzl: Done - better now?

abrunstrom: Suggested change: cause --> allow
Also I think the part about the error can actually be removed. I think the error should rather come if the Per-Message Reliability property is not set to enabled. At least this is what the description later in the text says.

gorryfair: At least for me, the last sentence would need to be more like: 
"If the underlying protocol can not support a required per-Message behavior, this Message Property choice will yield an error." - since prefer would not yield an error.

mwelzl: @abrunstrom done; I did remove the part about the error as you suggested, so I think that also addresses @gorryfair's suggestion in some sense (he suggested to rephrase the sentence, but I removed it now  :)  )

PR #561, opened 2020-04-16
Replace "idempotent" with "safely replayable" everywhere

### LINKED ISSUE #522: Idempotency vs replay-safe  (from Kyle Rose's review)
_From Kyle:_
idempotency != replay-safe. DELETE is idempotent, but not safe for replay because someone might have done a PUT or POST in the meantime.

What I mean is maybe best illustrated by 0RTT in TLS or QUIC. If a user sends a 0RTT DELETE /object, followed by a PUT /object, then within the replay window and without any replay prevention at the application layer, an attacker could replay the 0RTT DELETE /object and cause the object to be deleted again. DELETE is idempotent but it's not safe to replay.

mwelzl:
Closes #522.
This replaces a previous PR for the same issue, which I deleted: in this PR, I just added the definition of idempotency from RFC 7231, but this isn't good enough. We can't guarantee that these Messages would only *consecutively* arrive as duplicates, so these should be called "safely replayable" instead. See the discussion in #522.

### DIFF
--- a/draft-ietf-taps-impl.md
+++ b/draft-ietf-taps-impl.md
@@ -503,7 +503,7 @@ The effect of the application sending a Message is determined by the top-level p
 
 - Ordered: when this is false, it disables the requirement of in-order-delivery for protocols that support configurable ordering.
 
-- Idempotent: when this is true, it means that the Message can be used by mechanisms that might transfer it multiple times -- e.g., as a result of racing multiple transports or as part of TCP Fast Open.
+- Safely Replayable: when this is true, it means that the Message can be used by mechanisms that might transfer it multiple times -- e.g., as a result of racing multiple transports or as part of TCP Fast Open.
 
 - Final: when this is true, it means that a transport connection can be closed immediately after its transmission.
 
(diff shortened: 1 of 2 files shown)


draft-ietf-taps-impl.md
```diff
@@ -968,7 +968,7 @@ Initiate:
 : Calling `Initiate` on a TLS Connection causes it to first initiate a TCP connection. Once the TCP protocol is Ready, the TLS handshake will be performed as a client (starting by sending a `client_hello`, and so on).
 
 InitiateWithSend:
-: Early idempotent data is supported by TLS 1.3, and sends encrypted application data in the first TLS message when performing session resumption. For older versions of TLS, or if a session is not being resumed, the initial data will be delayed until the TLS handshake is complete. TCP Fast Option can also be enabled automatically.
+: Early safely replayable data is supported by TLS 1.3, and sends encrypted application data in the first TLS message when performing session resumption. For older versions of TLS, or if a session is not being resumed, the initial data will be delayed until the TLS handshake is complete. TCP Fast Option can also be enabled automatically.
```
abrunstrom, line 972: TCP Fast Option should be TCP Fast Open I think.

mirjak:
Sorry, I don't understand the point here. Why is it important that messages arrive "consecutively as duplicates"?

mwelzl:
See the discussion in #522; the issue is explained in the last message there from @squarooticus

draft-ietf-taps-impl.md
mwelzl: Fixed; I almost hesitated to fix it, we could have left it in there as an Easter Egg of some kind. An April 1 (2021) RFC could then have followed up and fully spec'ed the TCP Fast Option...

Issue #519, opened 2020-03-27
Architecture: normative arch language


gorryfair:
Overall review: I think we need to keep the normative language for the key features that differentiate the API. I'd really like to use REQUIRED, RECOMMENDED, rather MUST and SHOULD - because these are not protocol operation requirements, but are implementation requirements to conform. This includes requirements based on Minset - where the WG dervied this API: This is after all why the WG worked on Minset as the basis of the API.

I suggest we group all the requirements on architecture as short one or two sentence blocks (preferably as a bullet or definition list) collected into a single subsection, so they are easy to find and read and we say that these are the key architectural requirements. 

We insert this new requirements text either their own section, or in section 1.3 (although I am open to other places where we can call-out the architectural requirements).

----

The following are key requirements, that define the TAPS architecture:

* It is RECOMMENDED that functionality that is common across multiple transport protocols is made accessible through a unified set of TAPS interface calls. As a baseline, a Transport Services system interface is REQUIRED to allow access to the minimal set of features offered by transport protocols {{?I-D.ietf-taps-minset}}.

* A Transport Services system automates the selection of protocols and features and network, based on a set of Properties supplied through the TAPS interface. It is RECOMMENDED that a Transport Services system offers Properties that are common to multiple transport protocols. Specifying common Properties enables a system to appropriately select between protocols that offer equivalent features. This design permits evolution of the transport protocols and functions without affecting the programs using the system.

* Specifying common Properties enables a Transport Services system to appropriately select an appropriate network interface. It is RECOMMENDED that a system offers Properties that are common to a variety network layer interfaces and paths. This design permits racing of different network paths without affecting the programs using the system.

* Applications using the Transport Services system interface are REQUIRED to be robust to the automated selection provided by the system, including the possibility to express requirements and preferences that constrain the choices that can be made by the system. 

* If two different Protocol Stacks can be safely swapped, or raced in parallel (see Section 4.2.2), by the Transport Services system then they are considered to be "equivalent". Equivalent Protocol Stacks need to meet the requirements for racing  specified  in  section XX.

[...]

Issue #516, opened 2020-03-09
More protocol-specific properties?


mwelzl:
I suspect that we need more protocol-specific properties to be able to talk to various kinds of native protocol peers (other than TCP and UDP: these should already work fine with what we have, I believe).

A concrete example: we're hiding stream numbers. In NEAT, we needed to make them configurable though, because we wanted a NEAT client to be able to talk to a native SCTP server, where an application may expect certain type of data to arrive via a certain stream number. I suspect that we may need something similar for a QUIC peer.

- Should we address this?  (I'd say yes)
- How .. is it a protocol-specific (per-Message) property?   (I'd say yes... and perhaps a per-Connection property too, with text explaining that the per-Message one automatically becomes the Connection default unless otherwise specified)
- What other such things are we missing... what other expectations may e.g. a native QUIC or SCTP application have that we just cannot fulfil with our current system?

PR #564, opened 2020-04-16
Consolidate normative language in architecture

tfpauly:
This proposal consolidates normative language into the section on design principles (now requirements), using the suggestions in #519.


draft-ietf-taps-arch.md
```diff
@@ -216,35 +216,53 @@ Transport Services implementations are meant to be flexible at connection establ
...
+To ensure that security protocols are not incorrectly swapped, Transport Services systems MUST only select Protocol Stacks when the transport security protocols within the stacks are identical. Transport Services systems MUST NOT automatically fall back from secure protocols to insecure protocols, or to weaker versions of secure protocols. A Transport Services system MAY allow applications to explicitly specify that fallback to a specific other version of a protocol is allowed, e.g., to allow fallback to TLS 1.2 if TLS 1.3 is not available.
```
mwelzl, line 259: The first sentence lays out a very strict requirement.
The second is in line with the first, but it describes something broader nevertheless - i.e., it makes me think: "if you say that they must be *identical*, why tell me that falling back to a weaker version is disallowed?
The third sentence, then, is an outright contradiction to the first.

I'm not sure how to best solve this... maybe precede the third sentence with "As an exception, ...", and swap order: move the second sentence to the end.

draft-ietf-taps-arch.md
```diff
@@ -216,35 +216,53 @@ Transport Services implementations are meant to be flexible at connection establ
...
+It is important to note that neither the Transport Services API {{!I-D.ietf-taps-interface}} nor the Implementation document {{?I-D.ietf-taps-impl}} define new protocols or protocol capabilities that affect what is communicated across the network. Use of a Transport Services system MUST NOT require that a peer on the other side of a connection uses the same API or implementation. A Transport Services system acting as a connection initiator can communicate with any existing system that implements the transport protocol(s) selected by the Transport Services system. Similarly, a Transport Services system acting as a listener can receive connections for any protocol that is supported by the system from existing initiators that implement the protocol, independent of whether the initiator uses a Transport Services system or not.
```
mwelzl, line 263: Regarding the ability to communicate with "any existing system that implements the transport protocol(s) selected by the transport system" (or have been selected by the peer, in the passive case): this is sadly not the case, see issue #516.

We could say "With the exception of cases where applications make protocol-specific assumptions, ..."   ... but that sounds clumsy, and it also loses a benefit that we have: I think the statement IS correct for TCP. Perhaps for UDP too? Beyond that... harder. I'm not sure how to best phrase this then, sorry :(

draft-ietf-taps-arch.md
```diff
@@ -216,35 +216,53 @@ Transport Services implementations are meant to be flexible at connection establ
...
 
-There are applications that will need to control fine-grained details of transport protocols to optimize their behavior and ensure compatibility with remote systems. A Transport Services system therefore ought to also permit more specialized protocol features to be used. The interface for these specialized options ought to be exposed differently from the common options to ensure flexibility.
+Properties defaults SHOULD be selected to ensure correctness for the widest set of applications, while providing the widest set of options for selection. For example, since both applications that require data reliability and those that do not require data reliability can function correctly when a protocol provides reliability, reliability ought to be required by default. In contrast, the default value for a Property regarding the selection of network interfaces ought to permit as many interfaces as possible.
```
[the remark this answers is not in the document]
gorryfair: I'd suggest: "The default for Properties". 
I'd suggest: "The default for this property is to require reliability". 
.... because we need to be careful in how we use the word 'require' in TAPS

draft-ietf-taps-arch.md
gorryfair: Or...
I think my thought was more along the lines of:
"A Transport Services system acting as a connection initiator is able to communicate with any existing system that implements the transport protocol(s) selected by the Transport Services system." 
Do I understand your comment as additionally thinking also when an endpoint asks for additional property X then if that's not supported we still don't then have a connection? 
If so... then "A Transport Services system acting as a connection initiator is able to communicate with any existing system that implements the transport protocol(s) and all required properties selected by the Transport Services system."

mwelzl: Yes - as you say, "when an endpoint asks for additional property X then if that's not supported we still don't then have a connection". The concrete problem on my mind is stream numbers: if we don't expose them (and I believe we don't), a TAPS Transport System cannot talk to an SCTP or QUIC peer where the application makes assumptions about stream numbers.

Stream numbers specifically should probably be included as a protocol-specific property, to avoid disappointing people implementing QUIC, which is why I filed issue #516.

Your proposed sentence at the end here, with "...and all the required properties selected by the Transport Services system"  sounds like a good way out to me!

draft-ietf-taps-arch.md
theri: Agree with @mwelzl that there's a contradiction here.
In previous text (now to be deleted), we start out by saying that the Transpor Services system MUST offer security features as requested by the application. Maybe we want to say this here, too, and then continue with SHOULD be identical and MUST NOT race or automatically fall back?

draft-ietf-taps-arch.md
```diff
@@ -216,35 +216,53 @@ Transport Services implementations are meant to be flexible at connection establ
...
+
+Applications using a Transport Services system interface are REQUIRED to be robust to the automated selection provided by the system, given the ability to express requirements and preferences that constrain the choices that can be made by the system.
```
theri, line 235: I don't really understand the second part.
They are REQUIRED to be robust, given [that the have or are being given] the ability?
Or they are REQUIRED to be given the ability?

I feel like there may be a verb missing here.

draft-ietf-taps-arch.md
theri, line 233: I'm wondering why the path selection (or "selection of network interfaces") sentence starts with "In contrast" - where's the contrast? Aren't both examples about having the default settings be as broad as possible?

draft-ietf-taps-arch.md
```diff
@@ -216,35 +216,53 @@ Transport Services implementations are meant to be flexible at connection establ
...
+
+## Allow Access to Specialized Features
+
+There are applications that will need to control fine-grained details of transport protocols to optimize their behavior and ensure compatibility with remote systems. A Transport Services system therefore SHOULD permit more specialized protocol features to be used. The interface for these specialized options ought to be exposed differently from the common options to ensure flexibility.
```
theri, line 239: I don't quite understand this yet:
What does "exposed differently" mean here? Is this about us not standardizing specific protocol selection and/or other specific features? Basically "make this harder to use so it doesn't become default and, thus, unnecessarily constrains selection"?

However: Couldn't one argue that some of the Selection Properties we have in the API draft are pretty specialized, too?
E.g., the User Timeout Option, which we give as an example below, is a Selection Property just like the others.

draft-ietf-taps-arch.md
```diff
@@ -216,35 +216,53 @@ Transport Services implementations are meant to be flexible at connection establ
...
+
+The following two examples show non-equivalent Protocol Stacks:
+
+ - If the application requires preservation of message boundaries, a Protocol Stack that runs UDP as the top-level interface to the application is not equivalent to a Protocol Stack that runs TCP as the top-level interface. A UDP stack would allow an application to read out message boundaries based on datagrams sent from the remote system, whereas TCP does not preserve message boundaries on its own, but needs a framing protocol on top to determine message boundaries.
+
+- If the application specifies that it requires reliable transmission of data, then a Protocol Stack using UDP without any reliability layer on top would not be allowed to replace a Protocol Stack using TCP.
 
-## Scope for API and Implementation Definitions
+The following example shows an equivalent Protocol Stack:
```
theri, line 255: nit: Isn't this "The following example shows equivalent Protocol Stacks:"?
After all, it's two stacks that could be regarded equivalent.

draft-ietf-taps-arch.md
abrunstrom: I like Gorry's second sentence as well but I am confused in relation to issue #516 . Isn't the problem there rather that an application that relies on a protocol specific property can not use a TAPS system that does not expose this property? This is different from what is used by the TAPS system when setting up a connection.

draft-ietf-taps-arch.md
```diff
@@ -216,35 +216,53 @@ Transport Services implementations are meant to be flexible at connection establ
...
+There are several degrees in which a Transport Services system is intended to offer flexibility to an application: it can provide access to multiple sets of protocols and protocol features; it can use these protocols across multiple paths that could have different performance and functional characteristics; and it can communicate with different remote systems to optimize performance, robustness to failure, or some other metric. Beyond these, if the API for the system remains the same over time, new protocols and features can be added to the system's implementation without requiring changes in applications for adoption.
 
-## Common APIs for Common Features
+In order to provide this functionality without causing incompatibility or introducing security vulnerabilities, Transport Services APIs and Implementations should follow the set of requirements described here.
```
tfpauly, line 225: Interim: Add a comment that this is the only normative section of the document

gorryfair: The subsection with requirements could also be called out in the overview as one sentence in 1.2?

draft-ietf-taps-arch.md
mwelzl: For this PR, we can then conclude that we agree that this is just an editorial fix that can be solved by using Gorry's second sentence.

@abrunstrom: regarding #516, I'm not sure I fully get what you say here because (to me) you make it sound like a local problem. It is related to setting up a connection though. I had the concrete case from NEAT in mind, where Michael Tuexen wanted stream numbers so that he could write a NEAT client that can talk to a non-NEAT SCTP server which makes assumptions about stream numbers. So, my concern is the peer relying on a protocol specific property.

Considering this example, the sentence: "A Transport Services system acting as a connection initiator can communicate with any existing system that implements the transport protocol(s) selected by the Transport Services system." is just not true unless we expose stream numbers (and whatever else a peer may expect to be used). That's how #516 and the text above are related.

gorryfair: In the sense that any transport can actually define unique features, the API is never generic, however, can we just keep a general claim with respect to the architecture? What word(s) would you change?

draft-ietf-taps-arch.md
gorryfair, line 259: /MUST only select Protocol Stacks when the transport security protocols within the stacks are identical./
should be something like:
/MUST only race Protocol Stacks when the transport security protocols within the stacks are identical.

draft-ietf-taps-arch.md
abrunstrom: @gorryfair I like your second sentence and do not want to change anything. 
I just wanted to point out to @mwelzl that it does not solve his issue #516. That is a related but slightly different issue. But I do not think we need to solve that here.

mwelzl: Yes, let's end this discussion and agree on @gorryfair's second sentence.

draft-ietf-taps-arch.md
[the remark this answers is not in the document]
tfpauly: Fixed

draft-ietf-taps-arch.md
```diff
@@ -109,7 +109,7 @@ This document describes the Transport Services architecture in three sections:
 
 - {{model}} describes how the API model of Transport Services differs from traditional socket-based APIs. Specifically, it offers asynchronous event-driven interaction, the use of messages for data transfer, and the flexibility to use different transport protocols and paths without requiring major changes to the application.
 
-- {{design}} explains the design principles behind the Transport Services API. These principles are intended to make sure that transport protocols can continue to be enhanced and evolve without requiring too many changes by application developers.
+- {{requirements}} explains the fundamental requirements for a Transport Services API. These principles are intended to make sure that transport protocols can continue to be enhanced and evolve without requiring too many changes by application developers.
```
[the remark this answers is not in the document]
tfpauly: Updated

draft-ietf-taps-arch.md
```diff
@@ -216,35 +216,53 @@ Transport Services implementations are meant to be flexible at connection establ
...
+## Provide Common APIs for Common Features
 
-As a baseline, any Transport Services API needs to allow access to the distilled minimal set of features offered by transport protocols {{?I-D.ietf-taps-minset}}.
+Any functionality that is common across multiple transport protocols SHOULD be made accessible through a unified set of Transport Services API calls. As a baseline, any Transport Services API MUST allow access to the minimal set of features offered by transport protocols {{?I-D.ietf-taps-minset}}.
 
-## Access to Specialized Features
+An application can specify constraints and preferences for the protocols, features, and network interfaces it will use via Properties. A Transport Services API SHOULD offer Properties that are common to multiple transport protocols, in order to enable the system to appropriately select between protocols that offer equivalent features. Similarly, a Transport Services API SHOULD offer Properties that are applicable to a variety network layer interfaces and paths, in order to permit racing of different network paths without affecting the applications using the system. Each Property is expected to have a default value.
```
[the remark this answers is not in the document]
tfpauly: Fixed

draft-ietf-taps-arch.md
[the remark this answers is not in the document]
tfpauly: Updated

draft-ietf-taps-arch.md
tfpauly: Fixed up. Please check!

draft-ietf-taps-arch.md
tfpauly: Took second sentence here.

draft-ietf-taps-arch.md
tfpauly: Reworded!

draft-ietf-taps-arch.md
tfpauly: Made this "as another example..."

draft-ietf-taps-arch.md
tfpauly: Removed the sentence, as I agree it doesn't add much.

With regards to User Timeout Option, I think this argues that if an app prefers UTO, it should do a soft prefer or express "if you use TCP, you must use UTO" as opposed to simply requiring UTO (which would lead to just using TCP).

draft-ietf-taps-arch.md
tfpauly: Fixed

draft-ietf-taps-arch.md
tfpauly: Added

draft-ietf-taps-arch.md
tfpauly: Fixed

tfpauly:
Please review the latest updates!

draft-ietf-taps-arch.md
mwelzl: LGTM; thanks for doing this!

PR #565, opened 2020-04-17
Sentence about minIncompleteLength when rcv. on streams

### LINKED ISSUE #543: minIncompleteLength when receiving data on streams
Impl draft section 5.2 Receiving Messages:
"if, on the other hand, the top-level protocol only supports a byte-stream and no deframers were supported, the application must specify the minimum number of bytes of Message content it wants to receive (which may be just a single byte) to control the flow of received data."
This text says the application MUST specify minIncompleteLength, however, the API draft says that this values is optional. This should be further clarified and more guidance given in case the value is not set.

mwelzl:
As described in #543, I removed the condition that requires the application to specify the length upon receiving from this example.
Closes #543.


tfpauly:
I don't think removing this is the right thing to do. Rather, we can soften the requirement—it shouldn't be a "must", it's just descriptive. How about:

If the top-level protocol only supports a byte-stream and no framers were supported, the application can control the flow of received data by specifying the minimum number of bytes of Message content it wants to receive at one time.

draft-ietf-taps-impl.md
```diff
@@ -523,7 +523,7 @@ Since sending a Message may involve a context switch between the application and
...
+Similar to sending, Receiving a Message is determined by the top-level protocol in the established Protocol Stack. The main difference with Receiving is that the size and boundaries of the Message are not known beforehand. The application can communicate in its Receive action the parameters for the Message, which can help the implementation know how much data to deliver and when. For example, if the application only wants to receive a complete Message, the implementation should wait until an entire Message (datagram, stream, or frame) is read before delivering any Message content to the application. This requires the implementation to understand where messages end, either via a supplied deframer or because the top-level protocol in the established Protocol Stack preserves message boundaries.
```
tfpauly, line 526:
```suggestion
Similar to sending, Receiving a Message is determined by the top-level protocol in the established Protocol Stack. The main difference with Receiving is that the size and boundaries of the Message are not known beforehand. The application can communicate in its Receive action the parameters for the Message, which can help the implementation know how much data to deliver and when. For example, if the application only wants to receive a complete Message, the implementation should wait until an entire Message (datagram, stream, or frame) is read before delivering any Message content to the application. This requires the implementation to understand where messages end, either via a supplied deframer or because the top-level protocol in the established Protocol Stack preserves message boundaries. If the top-level protocol only supports a byte-stream and no framers were supported, the application can control the flow of received data by specifying the minimum number of bytes of Message content it wants to receive at one time.
```

draft-ietf-taps-impl.md
```diff
@@ -523,7 +523,7 @@ Since sending a Message may involve a context switch between the application and
...
+Similar to sending, Receiving a Message is determined by the top-level protocol in the established Protocol Stack. The main difference with Receiving is that the size and boundaries of the Message are not known beforehand. The application can communicate in its Receive action the parameters for the Message, which can help the implementation know how much data to deliver and when. For example, if the application only wants to receive a complete Message, the implementation should wait until an entire Message (datagram, stream, or frame) is read before delivering any Message content to the application. This requires the implementation to understand where messages end, either via a supplied deframer or because the top-level protocol in the established Protocol Stack preserves message boundaries. If the top-level protocol only supports a byte-stream and no framers were supported, the application can control the flow of received data by specifying the minimum number of bytes of Message content it wants to receive at one time.
```
gorryfair, line 526: /can communicate in its Receive action the parameters for the Message/can communicate the parameters for the Message in its Receive action/

gorryfair: /the minimum number of bytes of Message content it wants to receive at one time./the minimum number of bytes of Message content that it wants to receive at one time.

gorryfair: /The main difference with Receiving is that the size and boundaries of the Message are not known beforehand. /The main difference in Receiving a Message is that the size and boundaries of the Message are not known beforehand. /
- or something, this one was hard for me to parse

PR #570, opened 2020-04-17
clarify that taps interface types are implementation specific, fix #523

### LINKED ISSUE #523: Registry for "interface instance or type"  (from Kyle Rose's review)
_from Kyle_:
> These type symbols really deserve an actual registry, or at least the start of one. Otherwise, we are likely to end up with a mess.

**Philipp Tiesel**:
> There is already one, but that one was not useful for our matters.

britram:


draft-ietf-taps-interface.md
```diff
@@ -976,7 +976,8 @@ is used on both mobile devices and desktop devices, it should define the
 always `Prohibit Cellular`.
 
 The set of interface types is expected to change over time as new access
-technologies become available.
+technologies become available. The taxonomy of interface types on a given
+TAPS system is implementation-specific.
```
tfpauly, line 980: We generally don't say TAPS system

tfpauly, line 980:
```suggestion
Transport Services system is implementation-specific.
```

Issue #567, opened 2020-04-17
Three oddities in the abstract?


gorryfair:
Odd English:
/where endpoints have multiple interfaces and potential transport protocols to select from/where endpoints could select from multiple interfaces and potential transport protocols./

We use the words /abstract programming interface/ and the acronym /API/ but never tie these two together.

I am not huge fan of: /lowest common denominator interface/ ... as an engineer, this sounds like a good thing, although /lowest/ isn't that clear, so I wonder if we can find a better term?

gorryfair:
Resolved in PR

PR #572, opened 2020-04-18
Fix three oddities in the abstract of draft-ietf-taps-interface

gorryfair:
Fixing three oddities in the abstract, issue #567

### DIFF
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -126,13 +126,13 @@ informative:
 
 --- abstract
 
-This document describes an abstract programming interface to the transport
+This document describes an abstract programming interface, API, to the transport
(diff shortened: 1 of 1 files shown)


draft-ietf-taps-interface.md
```diff
@@ -126,13 +126,13 @@ informative:
 
 --- abstract
 
-This document describes an abstract programming interface to the transport
+This document describes an abstract programming interface, API, to the transport
```
mwelzl, line 129: This now looks as if API was an abbreviation for "abstract programming interface", which it really isn't  :)   so I suggest to write instead: "This document describes an abstract application programming interface, API, to the transport..."

gorryfair: Yes, I would think that would be good.

Issue #569, opened 2020-04-17
NiTs in section 1


gorryfair:
Section 1:
This section contains a RFC2119 keyword that precedes the definition!

We use the words /abstract programming interface/ and the acronym /API/ but never tie these together.

NiTs:
/provide applications a way to/provide a way for applications to/
/as it comes at the cost of reduced portability/because it can reduce the portability./
/How these abstract concepts/The way these abstract concepts/
/is largely dependent on/mostly depends on/

gorryfair:
A PR for this issue has already been merged.

PR #573, opened 2020-04-18
NiTs in section 1

gorryfair:
Resolving Editorial NiTs in section 1 #569

### DIFF
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -138,7 +138,7 @@ and potential transport protocols to select from.
 
 # Introduction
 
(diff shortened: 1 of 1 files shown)


draft-ietf-taps-interface.md
```diff
@@ -138,39 +138,38 @@ and potential transport protocols to select from.
 
 # Introduction
 
-This document specifies a modern abstract programming interface atop the
+This document specifies a modern abstract programming interface (API) atop the
```
gorryfair, line 141: Should be /abstract Application/ if this to define API, as per MW's comment to abstract.

Issue #366, opened 2019-11-19
Some editorial (I hope) comments on implementation


gorryfair:
I did not manage to review it all, but I did read much of the ID, and have minor corrections-mainly comments below

/such as mobile client/
- as
/e.g., a mobile client/
---
/Like with derived endpoints/
as
/Similar to a derived endpoint/
---
/Protocol options are checked next in order./
as
/Protocol options are next checked in order./
---
Whys is there no section on immediate - it would be more complete and also the subsection no. would match the list of items in sect 4.4.
---

Sect:Derived Endpoints
----
/The connection library should order the derived endpoints/
- what does /should/ mean here? - I thought the next stage took an ordered list (highest preference first), so I'd expect it does order? and it would seem hugely sensible to put most preferred first, aka RFC8305.
---
/ as a branch which can be attempted independently from other hostnames. Each of these hostnames may also resolve to one or more addresses, thus creating multiple layers of branching./
- In above change /which/that/
- In above /may also resolve... thus creating/ ... I can see this /might resolve to one or more addresses, that would create/
... but I could also read multiple other ways to view the sentence.
---
/connection over either interface. /
- why 2? - should this be:
/ connection over any interface. /
---
/A single connection may be allowed to use an HTTP-based proxy, a SOCKS-based proxy, or connect directly. /
- what does /may be allowed to use / actually mean?
- is it allowed to use, or if not, what controls whether it is or not?
---
/that may provide preferable characteristics when available. /
- could this be:
/that, when available, could provide preferable characteristics./
- is it /characteristics/, did the WG previously call these features?
---
/Implementations that support racing protocols and protocol options should maintain a history of which protocols and protocol options successfully established, on a per-network basis /
- is it only on a per-network; could it be also per-endpoint.
---
/ since usable DNS results on one network may not necessarily be the same as DNS results on another network due to local network entities, supported address families, or enterprise network configurations. /
- Personally, I'd go further than /may not necessarily be/ towards /there are situations when the results will be different/?
---
I could be overly sensitive, but didn't like:
/has cached the information that SCTP is not available on the path over WiFi,/
- can this be:
/has current cache information that SCTP is not available on the path over WiFi,/
- I wanted to avoid the impression this was something in any way permanent.
---
Sorting branches:
/ in order of their preference rank/
- increasing or decreasing preference?
---
/Implementations should process properties in the following order: Prohibit, Require, Prefer, Avoid. /
- the words should process... seem to chime wrongly to me
[...]

PR #574, opened 2020-05-26
Address editorial comments on Implementation

tfpauly:
Addresses comments from #366

### DIFF
--- a/draft-ietf-taps-impl.md
+++ b/draft-ietf-taps-impl.md
@@ -270,7 +270,7 @@ DNS hostname-to-address resolution is the most common method of endpoint derivat
   1.4 [2001:DB8::3.80, Wi-Fi, TCP]
 ~~~~~~~~~~
 
-DNS-Based Service Discovery {{?RFC6763}} can also provide an endpoint derivation step. When trying to connect to a named service, the client may discover one or more hostname and port pairs on the local network using multicast DNS {{?RFC6762}}. These hostnames should each be treated as a branch that can be attempted independently from other hostnames. Each of these hostnames might resolve to one or more addresses, whcih would create multiple layers of branching.
+DNS-Based Service Discovery {{?RFC6763}} can also provide an endpoint derivation step. When trying to connect to a named service, the client may discover one or more hostname and port pairs on the local network using multicast DNS {{?RFC6762}}. These hostnames should each be treated as a branch that can be attempted independently from other hostnames. Each of these hostnames might resolve to one or more addresses, which would create multiple layers of branching.
 
 ~~~~~~~~~~
 1 [term-printer._ipp._tcp.meeting.ietf.org, Wi-Fi, TCP]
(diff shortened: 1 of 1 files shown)


draft-ietf-taps-impl.md
```diff
@@ -270,7 +270,7 @@ DNS hostname-to-address resolution is the most common method of endpoint derivat
   1.4 [2001:DB8::3.80, Wi-Fi, TCP]
 ~~~~~~~~~~
 
-DNS-Based Service Discovery {{?RFC6763}} can also provide an endpoint derivation step. When trying to connect to a named service, the client may discover one or more hostname and port pairs on the local network using multicast DNS {{?RFC6762}}. These hostnames should each be treated as a branch which can be attempted independently from other hostnames. Each of these hostnames may also resolve to one or more addresses, thus creating multiple layers of branching.
+DNS-Based Service Discovery {{?RFC6763}} can also provide an endpoint derivation step. When trying to connect to a named service, the client may discover one or more hostname and port pairs on the local network using multicast DNS {{?RFC6762}}. These hostnames should each be treated as a branch that can be attempted independently from other hostnames. Each of these hostnames might resolve to one or more addresses, whcih would create multiple layers of branching.
```
mwelzl, line 274: typo "whcih"

draft-ietf-taps-impl.md
```diff
@@ -424,7 +427,7 @@ An example in which failover is recommended is a race between a protocol stack t
...
+Successes and failures of given attempt should be reported up to parent nodes (towards the trunk of the tree). For example, in the following case, if 1.1.1 fails to connect, it reports the failure to 1.1. Since 1.1 has no other child nodes, it also has failed and reports that failure to 1. Because 1.2 has not yet failed, 1 is not considered to have failed. Since 1.2 has not yet started, it is started and the process continues. Similarly, if 1.1.1 successfully connects, then it marks 1.1 as connected, which propagates to the trunk node 1. At this point, the connection as a whole is considered to be successfully connected and ready to process application data
```
mwelzl, line 430: I think there's a missing article in the first sentence ("Success and failures of _a_ given attempt...")

draft-ietf-taps-impl.md
```diff
@@ -270,7 +270,7 @@ DNS hostname-to-address resolution is the most common method of endpoint derivat
   1.4 [2001:DB8::3.80, Wi-Fi, TCP]
 ~~~~~~~~~~
 
-DNS-Based Service Discovery {{?RFC6763}} can also provide an endpoint derivation step. When trying to connect to a named service, the client may discover one or more hostname and port pairs on the local network using multicast DNS {{?RFC6762}}. These hostnames should each be treated as a branch which can be attempted independently from other hostnames. Each of these hostnames may also resolve to one or more addresses, thus creating multiple layers of branching.
+DNS-Based Service Discovery {{?RFC6763}} can also provide an endpoint derivation step. When trying to connect to a named service, the client may discover one or more hostname and port pairs on the local network using multicast DNS {{?RFC6762}}. These hostnames should each be treated as a branch that can be attempted independently from other hostnames. Each of these hostnames might resolve to one or more addresses, whcih would create multiple layers of branching.
```
tfpauly, line 273:
```suggestion
DNS-Based Service Discovery {{?RFC6763}} can also provide an endpoint derivation step. When trying to connect to a named service, the client may discover one or more hostname and port pairs on the local network using multicast DNS {{?RFC6762}}. These hostnames should each be treated as a branch that can be attempted independently from other hostnames. Each of these hostnames might resolve to one or more addresses, which would create multiple layers of branching.
```

draft-ietf-taps-impl.md
tfpauly, line 430:
```suggestion
Successes and failures of a given attempt should be reported up to parent nodes (towards the trunk of the tree). For example, in the following case, if 1.1.1 fails to connect, it reports the failure to 1.1. Since 1.1 has no other child nodes, it also has failed and reports that failure to 1. Because 1.2 has not yet failed, 1 is not considered to have failed. Since 1.2 has not yet started, it is started and the process continues. Similarly, if 1.1.1 successfully connects, then it marks 1.1 as connected, which propagates to the trunk node 1. At this point, the connection as a whole is considered to be successfully connected and ready to process application data
```

draft-ietf-taps-impl.md
tfpauly: Fixed

draft-ietf-taps-impl.md
tfpauly: Fixed!

PR #581, opened 2020-05-29
Remove specification that message is bytes-like

### LINKED ISSUE #477: Is format of message an implementation detail?
In section 7.1 of API it says:

> Message data is transferred as an array of bytes, and the resulting object contains both the byte array and the length of the array.
`messageData := "hello".bytes()
 Connection.Send(messageData)`

For some languages that are dynamic typed (i.e. python) this limits the usability of the API as it adds an unnecessary extra step to each send.
To me this seems like it should be an implementation detail and not specified in the API.

MaxF12:
Closes #477


mirjak:
Looks fine to me. I wonder if we want to add some text and given some implementation hints. Not sure if that is needed but it could be useful...?

csperkins:
Looks good.

Issue #579, opened 2020-05-29
5.1.1. Message Properties Editorial NiTs


gorryfair:
5.1.1. Message Properties 

This isn’t really about /removing/ we need to express this differently, not sure what you wish to say: 

For example, TCP cannot remove bytes from its send buffer, while in case of SCTP, such control over the SCTP send buffer can be exercised using the partial reliability extension [RFC8303]. 

and:

/Final: when this is true, it means that a transport connection can be closed immediately after its transmission./
I think this is:
/Final: when this is true, it means that a transport connection can be closed immediately after transmission of the message./

gorryfair:
See PR Update draft-ietf-taps-impl.md #582

PR #582, opened 2020-05-30
Editorial update to message properties in draft-ietf-taps-impl.md 

gorryfair:
Changes to address Issue #579


draft-ietf-taps-impl.md
```diff
@@ -500,15 +500,15 @@ The effect of the application sending a Message is determined by the top-level p
...
 
-- Ordered: when this is false, it disables the requirement of in-order-delivery for protocols that support configurable ordering.
+- Ordered: when this is false, this disables the requirement of in-order-delivery for protocols that support configurable ordering.
 
-- Safely Replayable: when this is true, it means that the Message can be used by mechanisms that might transfer it multiple times -- e.g., as a result of racing multiple transports or as part of TCP Fast Open.
+- Safely Replayable: when this is true, this means that the Message can be used by mechanisms that might transfer it multiple times -- e.g., as a result of racing multiple transports or as part of TCP Fast Open.
 
-- Final: when this is true, it means that a transport connection can be closed immediately after its transmission.
+- Final: when this is true, this means that a transport connection can be closed immediately after transmission of the message..
```
tfpauly, line 511:
```suggestion
- Final: when this is true, this means that a transport connection can be closed immediately after transmission of the message.
```

draft-ietf-taps-impl.md
```diff
@@ -500,15 +500,15 @@ The effect of the application sending a Message is determined by the top-level p
...
+- Lifetime: this should be implemented by removing the Message from its queue of pending Messages after the Lifetime has expired. A queue of pending Messages within the transport system implementation that have yet to be handed to the Protocol Stack can always support this property, but once a Message has been sent into the send buffer of a protocol, only certain protocols may support de-queueing a message. For example, TCP cannot remove bytes from its send buffer, while in case of SCTP, such control over the SCTP send buffer can be exercised using the partial reliability extension {{!RFC8303}}. When there is no standing queue of Messages within the system, and the Protocol Stack does not support the removal of a Message from the stack's send buffer, this property may be ignored.
```
tfpauly, line 503:
```suggestion
- Lifetime: this should be implemented by removing the Message from the queue of pending Messages after the Lifetime has expired. A queue of pending Messages within the transport system implementation that have yet to be handed to the Protocol Stack can always support this property, but once a Message has been sent into the send buffer of a protocol, only certain protocols may support removing a message. For example, an implementation cannot cannot remove bytes from a TCP send buffer, while it can remove data from a SCTP send buffer using the partial reliability extension {{!RFC8303}}. When there is no standing queue of Messages within the system, and the Protocol Stack does not support the removal of a Message from the stack's send buffer, this property may be ignored.
```

Issue #576, opened 2020-05-29
Section 4.1.5 - minor wording NiTs?


gorryfair:
Scavenger: This reads wrong to me, this is unutilised *capacity*
/Highest available bandwidth/available capacity/

Why is /satisfy./ used for CR streaming? To me this implies admission-based QoS. Do we really mean this? Or should we say /expected/ here also?

/Implementations process properties/
- It would really have helped me to read this if we insert /the/ before properties.

/As the available protocols and paths on a specific system and in a specific context may vary,/
would seem better as:
/The available protocols and paths on a specific system and in a specific context can vary, therefore/

/, by preferring protocols and paths that existing Connections with similar Properties are already using./
should be something like:
/, by preferring protocols and paths that are already used by existing Connections that specified similar Properties./

gorryfair:
See PR #583 (which includes text polishing).

PR #583, opened 2020-05-30
Rewording of Section 4.1.5 on Properties

gorryfair:
Not intended to change meaning, but intended to improve reading, see issue #576.

### DIFF
--- a/draft-ietf-taps-impl.md
+++ b/draft-ietf-taps-impl.md
@@ -382,7 +382,7 @@ Implementations process the Properties in the following order: Prohibit, Require
 If Selection Properties contain any prohibited properties, the implementation should first purge branches containing nodes with these properties. For required properties, it should only keep branches that satisfy these requirements. Finally, it should order the branches according to the preferred properties, and finally use any avoided properties as a tiebreaker.
 When ordering branches, an implementation can give more weight to properties that the application has explicitly set, than to the properties that are default.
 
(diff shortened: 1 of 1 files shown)


draft-ietf-taps-impl.md
```diff
@@ -366,23 +366,23 @@ In addition to the properties provided by the application, an implementation may
...
+When ordering branches, an implementation can give more weight to properties that the application has explicitly set, than to the properties that are default.
 
-As the available protocols and paths on a specific system and in a specific context may vary, the result of sorting and the outcome of racing may vary even given the same Selection and Connection Properties. However, an implementation ought to aim to provide a consistent outcome to applications, e.g., by preferring protocols and paths that existing Connections with similar Properties are already using.
+The available protocols and paths on a specific system and in a specific context can chnage, therefore, the result of sorting and the outcome of racing may vary, even when using the same Selection and Connection Properties. However, an implementation ought to provide a consistent outcome to applications, e.g., by preferring protocols and paths that are already used by existing Connections that specified similar Properties.
```
tfpauly, line 385:
```suggestion
The available protocols and paths on a specific system and in a specific context can change; therefore, the result of sorting and the outcome of racing may vary, even when using the same Selection and Connection Properties. However, an implementation ought to provide a consistent outcome to applications, e.g., by preferring protocols and paths that are already used by existing Connections that specified similar Properties.
```

PR #587, opened 2020-06-19
Clarify immediate racing

### LINKED ISSUE #577: 4.2 Racing
In Section 4.2.1., I didn't really understand what "immediate" means from the text. I think it just needs one sentence.

tfpauly:
Closes #577


gorryfair:
I actually really like: "Simultaneous" and "Staggered" as descriptive terms that avoid the question of "delaying".

PR #601, opened 2020-07-13
Clarify naming convention variance

squarooticus:
* Updates PR #596 to remove reference to lower-case in namespaces
   corresponding to IANA protocol names.

 * Express that strict adherence to the doc's naming convention is not
   required, but that symbols in different languages should be
   visually similar.


squarooticus:
@mwelzl @britram 

Last minute PR ⏰

PR #604, opened 2020-07-16
Add diagram and more explanation for message framers in API doc

### LINKED ISSUE #457: Framers again...
I know we moved some text about framers into the implementation doc but that really doesn't seem right when I look at it now.

The section about framers in the implementation doc is the only section that defines an interface which probably is already an indication that just moving this text there was not right. However, also the text in the API doc seems now incomplete. 

I think we should move it back and rather define some real simple examples framers in the implementation document (just a prefixed length field (as many UDP mapping use) or maybe even an example implementation for STARTTLS). Further after moving it back we could figure out if really all parts of the interface are necessary or if there would be a more minimal interface that we could use. E.g. I'm not sure about the FailConnection() call or at least I think this could need more explanation.

tfpauly:
Closes #457

### DIFF
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -2155,7 +2155,7 @@ Message Framers are loaded into Connections when specified in a Preconnection.
 When Message Framers are being used, they can interact with the setup and
 teardown of a Connection. For example, a Framer can start sending data
 before the application sends data if the framing protocol requires a prefix
-or handshake (see {{RFC8229}} for an example of such a framing protocol).
+or handshake (see {{?RFC8229}} for an example of such a framing protocol).
 Most importantly, Message Framers intercept all calls to Send() or Receive(),
 and can modify application data.
 
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -2151,13 +2151,12 @@ does not provide message boundary preservation, and since many of these
 protocols require message boundaries to function, each application layer
 protocol has defined its own framing.
 
-Message Framers are loaded into Connections when specified in a Preconnection.
-When Message Framers are being used, they can interact with the setup and
-teardown of a Connection. For example, a Framer can start sending data
+To use a Message Framer, the application adds it to its Preconnection object.
+Then, the Message Framer can intercept all calls to Send() or Receive()
+on a Connection to add Message semantics, in addition to interacting with
+the setup and teardown of the Connection. A Framer can start sending data
 before the application sends data if the framing protocol requires a prefix
 or handshake (see {{?RFC8229}} for an example of such a framing protocol).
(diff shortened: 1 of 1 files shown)


csperkins:
I think this helps, but one of the confusions is the distinction between a framing to provide a message-oriented API, and the use of framers underneath that API to change the representation of messages on the wire. I wonder if some of the difficulties we've had with framers is that some of us have been using the term to represent one of these concepts, and some the other?

One is the concept of a framer that accumulates data and returns an `HttpResponse` object from a `recv()` call, rather than returning a byte array containing a (partial) HTTP response. That's the API-level framer.

The other is the sort of framer that performs TLS encryption on the HTTP request, after `send()` has been called. This transforms data, below the API. 

Do we need two different terms?

mwelzl:
@csperkins - personally, I always thought that Framers can do both of these things; if this wasn't clear enough before, then I think your part 1 (returning an HTTPResponse object) was clear from before, and now this PR makes it clear that they also intercept the send call to transform data afterwards. So IMHO that's just the clarification that was needed.

mwelzl:
Sorry! Obviously I only wanted to comment, but hit the wrong button  :)

draft-ietf-taps-interface.md
```diff
@@ -2151,6 +2151,32 @@ does not provide message boundary preservation, and since many of these
...
+
+~~~~~~~~~~
+
+  Initiate()   Send()   Receive()   Close()
+      |          |         ^          |
+      |          |         |          |
+ +----v----------v---------+----------v-----+
+ |                Connection                |
+ +----+----------+---------^----------+-----+
+      |          |         |          |
```
tfpauly, line 2170: Denote that what a framer vends in receive is a well-defined message object (as opposed to a byte stream below)

draft-ietf-taps-interface.md
```diff
@@ -2151,6 +2151,32 @@ does not provide message boundary preservation, and since many of these
 protocols require message boundaries to function, each application layer
 protocol has defined its own framing.
 
+Message Framers are loaded into Connections when specified in a Preconnection.
+When Message Framers are being used, they can interact with the setup and
+teardown of a Connection. For example, a Framer can start sending data
```
theri, line 2156: To me this sounds a bit like Framers are only coming into effect when/after the Preconnection becomes a Connection. Also, this makes me wonder what "load into Connections" means.

Maybe let's rather say something like:
"To use a Message Framer, an application has to add it to a Preconnection.
Then, the Message Framer can interact with the setup and teardown of the Connection. ..."

tfpauly: Reworked this text, let me know if that works!

csperkins: Better, I think

PR #606, opened 2020-07-16
Clarify text on staggered racing, reference happy eyeballs and ICE

### LINKED ISSUE #537: Recommendations for Delayed Racing
Section 4.2.1  on delayed racing:
"For example, if the first child represents an IP address with a known route, and the second child represents another IP address, the delay between starting the first and second IP addresses can be based on the expected retransmission cadence for the first child's connection (derived from historical round-trip-time). Alternatively, if the first child represents a branch on a Wi-Fi interface, and the second child represents a branch on an LTE interface, the delay should be based on the expected time in which the branch for the first interface would be able to establish a connection, based on link quality and historical round-trip-time."

What the reasoning for these two recommendations? I'm actually not sure what exactly is meant. Is the first case 3xRTT (-> initial RTO) and the second 1xRTT? Why?

Also in the next paragraph:
"Generally, branches between paths and protocols should have longer delays than branches between derived endpoints."

What? Why?

And then 
"The maximum delay should be considered with regards to how long a user is expected to wait for the connection to complete."

How do I know this?

tfpauly:
Closes #537


draft-ietf-taps-impl.md
```diff
@@ -413,11 +413,13 @@ Staggered racing can be used whenever a single node of the tree has multiple chi
 
 Staggered racing attempts occur in parallel. Implementations should not terminate an earlier child connection attempt upon starting a secondary child.
```
abrunstrom, line 414: I think this needs to be "Staggered racing attempts may occur in parallel" considering that the part that now follows right after in the text is the case when they do not occur in parallel.

draft-ietf-taps-impl.md
```diff
@@ -413,11 +413,13 @@ Staggered racing can be used whenever a single node of the tree has multiple chi
...
+Generally, the delay before starting a given child node should be based on the length of time the previously started child node is expected to take before succeeding or making progress in connection establishment. Algorithms like Happy Eyeballs choose a delay based on how long the transport connection handshake takes. When performing staggered races in multiple layers (such as racing between network interfaces, and then racing between IP addresses), a longer delay may be given for the higher layers of racing. For example, when racing between network interfaces, the delay should also take into account the amount of time it takes to prepare the network interface (such as radio association) and name resolution over that interface, in addition to the delay that would be added for a single transport connection handshake.
+
+Any staggered delay should have a defined minimum and maximum value, which may vary depending on the layer of racing. The maximum delay should be considered with regards to how long a user is expected to wait for the connection to establish.
```
abrunstrom, line 422: I do not understand this part. How does the minimum and maximum come in to play? And who sets these values?

draft-ietf-taps-impl.md
```diff
@@ -413,11 +413,13 @@ Staggered racing can be used whenever a single node of the tree has multiple chi
...
 
-If a child node fails to connect before the delay timer has fired for the next child, the next child should be started immediately.
+Generally, the delay before starting a given child node should be based on the length of time the previously started child node is expected to take before succeeding or making progress in connection establishment. Algorithms like Happy Eyeballs choose a delay based on how long the transport connection handshake takes. When performing staggered races in multiple layers (such as racing between network interfaces, and then racing between IP addresses), a longer delay may be given for the higher layers of racing. For example, when racing between network interfaces, the delay should also take into account the amount of time it takes to prepare the network interface (such as radio association) and name resolution over that interface, in addition to the delay that would be added for a single transport connection handshake.
```
tfpauly, line 420: layers -> branching levels

theri: In the Impl draft we actually call it Branch Types. Section 4.1.3 says:
"There are three types of branching from a parent node into one or more child nodes. […] Derived Endpoints […] Alternate Paths […] Protocol Options"
So maybe we can go with branch types here, too?

draft-ietf-taps-impl.md
```diff
@@ -413,11 +413,13 @@ Staggered racing can be used whenever a single node of the tree has multiple chi
...
+
+Any staggered delay should have a defined minimum and maximum value, which may differ depending on the layer of racing. The maximum delay should be considered with regards to how long a user is expected to wait for the connection to establish.
```
tfpauly, line 422: Explain that because implementations may choose the delay based on dynamic information, they should have min/max bounds in the implementation (and can choose to ignore a branch).

draft-ietf-taps-impl.md
gorryfair: Is this clearer: ""Staggered racing attempts can proceed in parallel"?

draft-ietf-taps-impl.md
tfpauly: Done!

draft-ietf-taps-impl.md
tfpauly: Updated!

draft-ietf-taps-impl.md
tfpauly: See update!

tfpauly:
@abrunstrom @theri please take a look at the updated text!

draft-ietf-taps-impl.md
```diff
@@ -411,13 +411,15 @@ Simultaneous racing is when multiple alternate branches are started without wait
...
 
-If a child node fails to connect before the delay timer has fired for the next child, the next child should be started immediately.
+Generally, the delay before starting a given child node ought to be based on the length of time the previously started child node is expected to take before it succeeds or makes progress in connection establishment. Algorithms like Happy Eyeballs choose a delay based on how long the transport connection handshake is expected to take. When performing staggered races in multiple branch types (such as racing between network interfaces, and then racing between IP addresses), a longer delay may be chose for some branch types. For example, when racing between network interfaces, the delay should also take into account the amount of time it takes to prepare the network interface (such as radio association) and name resolution over that interface, in addition to the delay that would be added for a single transport connection handshake.
+
```
abrunstrom, line 421: be chose --> be chosen

draft-ietf-taps-impl.md
```diff
@@ -411,13 +411,15 @@ Simultaneous racing is when multiple alternate branches are started without wait
...
 
-If a child node fails to connect before the delay timer has fired for the next child, the next child should be started immediately.
+Generally, the delay before starting a given child node ought to be based on the length of time the previously started child node is expected to take before it succeeds or makes progress in connection establishment. Algorithms like Happy Eyeballs choose a delay based on how long the transport connection handshake is expected to take. When performing staggered races in multiple branch types (such as racing between network interfaces, and then racing between IP addresses), a longer delay may be chose for some branch types. For example, when racing between network interfaces, the delay should also take into account the amount of time it takes to prepare the network interface (such as radio association) and name resolution over that interface, in addition to the delay that would be added for a single transport connection handshake.
```
tfpauly, line 420:
```suggestion
Generally, the delay before starting a given child node ought to be based on the length of time the previously started child node is expected to take before it succeeds or makes progress in connection establishment. Algorithms like Happy Eyeballs choose a delay based on how long the transport connection handshake is expected to take. When performing staggered races in multiple branch types (such as racing between network interfaces, and then racing between IP addresses), a longer delay may be chosen for some branch types. For example, when racing between network interfaces, the delay should also take into account the amount of time it takes to prepare the network interface (such as radio association) and name resolution over that interface, in addition to the delay that would be added for a single transport connection handshake.
```

draft-ietf-taps-impl.md
tfpauly: Done!

PR #607, opened 2020-07-16
Rename ListenClone

### LINKED ISSUE #508: Discuss interaction between Listen() and ListenClone()

tfpauly:
Closes #508

### DIFF
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -1481,12 +1481,16 @@ In addition, incoming entangled Connections can be received by creating a
 Listener on an existing connection:
 
 ~~~
-Listener := Connection.ListenClone()
+Listener := Connection.ListenWithConnection()
 ~~~
 
-ListenClone() creates a Listener that listens on the same LocalEndpoint as the one the
-cloned Connection is using. Any new Connection received by this Listener will be
-entangled with the cloned Connection.
+ListenWithConnection() creates a Listener that can accept new streams, given
+a Connection that uses a multi-streaming protocol. Any new Connection received by
+this Listener will be "entangled" with the cloned Connection, and be part of
+the same Connection Group. If ListenWithConnection() is called on a Connection
+that does not support multi-streaming, the operation will return an error instead
+of a Listener.
+
 Changing one of the Connection Properties on one Connection in the group
 changes it for all others. Message Properties, however, are not
 entangled. For example, changing `Timeout for aborting Connection` (see


draft-ietf-taps-interface.md
```diff
@@ -1481,12 +1481,16 @@ In addition, incoming entangled Connections can be received by creating a
...
 
-ListenClone() creates a Listener that listens on the same LocalEndpoint as the one the
-cloned Connection is using. Any new Connection received by this Listener will be
-entangled with the cloned Connection.
+ListenWithConnection() creates a Listener that can accept new streams, given
+a Connection that uses a multi-streaming protocol. Any new Connection received by
+this Listener will be "entangled" with the cloned Connection, and be part of
+the same Connection Group. If ListenWithConnection() is called on a Connection
+that does not support multi-streaming, the operation will return an error instead
+of a Listener.
```
mwelzl, line 1492: I'm very sorry for only bringing this up here - I should have said it at the last interim, or at least write it in the issue ... somehow I overlooked this:

I'm strongly *against* writing that this should fail in case the Connection doesn't support multi-streaming.

First, I just don't think that should happen - we have produced lots of research on heuristics to detect common bottlenecks, and methods to combine congestion controls; we have shown that, with multiple TCP connections, it's quite possible to offer something VERY similar to multi-streaming (and this gets much easier when one can *assume* a common bottleneck, e.g. when all is tunneled over the same five-tuple, e.g. with a VPN; if you're interested, please get in touch  :)  ). So, at least, we shouldn't prohibit that.

Second, there are no guarantees offered regarding mux'ing, and e.g. priorities making a difference - so, why fail anyway? One just has to make sure that, if e.g. multiple TCP connections are below, calls that configure something on TCP connection 1 lead to a similar configuration for TCP connection 2.

Third, it's an implementation detail; we shouldn't be so prescriptive here anyway.

Fourth, a (capital-c) "Connection" is what we expose, not what the underlying transport uses. FWIW, there could in fact be multiple underlying connections that we're multiplexing a single transfer on!

This is all easily fixed by removing the last sentence   :)

abrunstrom: I missed the interim discussion on this, but I agree with @mwelzl that ListenWithClone should not fail for transports without multi-streaming. That would break the Connection abstraction.

tfpauly:
Proposal: replace this with a way to query for a connection what the other connections in a group are (which will only happen passively for multi streamed protocols today).

draft-ietf-taps-interface.md
mwelzl: To make progress on this one: at the interim, I have been convinced that, at least with today's protocols (hmm hmm... what if connections connections arrive via the same tunnel, e.g. in a VPN case?), supporting connection coupling on the server side will not be possible.

However, there's also no need to be so restrictive in the text in the interface draft, IMO. The final statement regarding multi-streaming could safely be removed here, and included in the implementation draft. That would be my proposal.

tfpauly:
Implemented #610; @mwelzl does this work?

draft-ietf-taps-interface.md
```diff
@@ -1477,16 +1477,18 @@ the Connection Group, and so on. Connections in a Connection Group generally sha
...
-Listener := Connection.ListenClone()
+[]Connection := Connection.GroupedConnections()
 ~~~
 
-ListenClone() creates a Listener that listens on the same LocalEndpoint as the one the
-cloned Connection is using. Any new Connection received by this Listener will be
-entangled with the cloned Connection.
+Connections will be in the same group if the application previously called Clone.
+Passive Connections can also be added to the same group when a Listener receives
+a new Connection that is one stream of a multi-streaming protocol instance.
```
mwelzl, line 1490: The last sentence again  :)
I'd much prefer it if this could be written as an example.
```suggestion
Passive Connections can also be added to the same group -- e.g., when a Listener receives a new Connection that is just a new stream of an already active multi-streaming protocol instance.
```

mwelzl:
I'm merging this now so I can proceed with re-structuring. It reflects what we discussed at the interim, so I don't think there should be more concerns.

PR #611, opened 2020-07-22
Transport Properties and Security Parameters: Set() not Add()

### LINKED ISSUE #609: Rename "add" properties to "set" if they override
We should rename these if they are really setting and not adding.

theri:
If I understand correctly, we're staying with the current master.
Therefore, here's my PR to fix #609 - I think we always want to override, and if we override, calling it Set() makes this clearer than Add().

For Security Parameters, perhaps it should be possible to add multiple identities, ciphersuites, etc.
But that's out of scope for this PR.

### DIFF
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -756,10 +756,11 @@ object:
 TransportProperties := NewTransportProperties()
 ~~~
 
-Individual properties are then added to the TransportProperties Object:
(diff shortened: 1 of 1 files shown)


mwelzl:
@philsbn - I agree with the response by @theri in #612, which concludes: "If I misunderstood, I think the best approach is to indeed allow specifying multiple (Preference, Enumeration) in a single Set() call."   KISS, I say :)

Issue #545, opened 2020-04-07
More guidance needed on Handling Path Changes


mirjak:
Section 7.2 on Handling Path Changes in the implementation draft needs to be better aligned to the interface we are currently still discussing for multipath protocols. Currently it reads like an taps implementation can freely decide which paths to use when. That's probable not the case but in any case needs also more specific guidance. However, we probably need to leave this issue as a reminder for later when we have decided on the final interface for MP protocols. 

I would eventually like to separate sections for handover (two parallel paths), migration (move from one to the other but keep same protocol instance) and fallback (move form one protocol instance to another).

PR #614, opened 2020-08-05
Edits on Section on Handling Path Changes

mirjak:
This is a first attempt to work on the section "Handling Path Changes" (see issue #545), however, I think more work is needed.

To be honest I don't fully understand this part and it would be could if someone else could provide more input about the System Policy part:
"For protocols that do not support multipath or migration, the Protocol Instances should be informed of the path change, but should not be forcibly disconnected if the previously used path becomes unavailable. An exception to this case is if the System Policy changes to prohibit traffic from the Connection based on its properties, in which case the Protocol Stack should be disconnected."

Also the last part about pooled connection should say more but I'm not sure what. We currently have now only this sentence left:
"Pooled Connections {{pooled-connections}} may add or remove underlying transport connections in a similar manner."

And finally the originally issue #545 was also mentioning fallback which I didn't address yet but I have this feeling that that might need a separate section. Should I add a separate issue for that or is that already addressed somewhere and I missed it?


philsbln:
II like your change - makes the text much clearer.

draft-ietf-taps-impl.md
```diff
@@ -722,8 +722,13 @@ This enables implementations to realize transparent connection coalescing, conne
...
+
+Pooled Connections {{pooled-connections}} may add or remove underlying transport connections in a similar manner.
```
philsbln, line 731:
```suggestion
In case of Pooled Connections {{pooled-connections}}, the transport system may add connections over new paths or different protocols to the pool if permissible based on the multipath policy and Selection Properties. In case a previously used path becomes unavailable, the transport system may disconnect all connections that require this path, but should not disconnect the pooled connection object exposed to the application. The strategy how is implementation specific, but should me consistent with the behaviour of multipath transports. 
```

draft-ietf-taps-impl.md
```diff
@@ -722,9 +722,13 @@ This enables implementations to realize transparent connection coalescing, conne
...
+When a path change occurs, e.g. when the IP address of an interface changes, this will usually break connectivity on that path for any active connection. Still, the Transport Services implementation is responsible for notifying Protocol Instances in the Protocol Stack if path changes are detected to enable them to quickly react to path failures before a time-out.
```
tfpauly, line 725: Changing one IP address is a somewhat degenerate case—the more interesting case is a new interface/link coming up or down, and that's what drives more significant multipath and migration events. I also think that focusing first on the case that make break connectivity is not a great way to start out.

draft-ietf-taps-impl.md
```diff
@@ -722,9 +722,13 @@ This enables implementations to realize transparent connection coalescing, conne
...
 
+For protocols that do not support multipath or migration, the Protocol Instances should be informed of the path change, but should not be forcibly disconnected if the previously used path becomes unavailable. An exception to this case is if the System Policy changes to prohibit traffic from the Connection based on its properties, in which case the Protocol Stack should be disconnected.
```
tfpauly, line 727: This is trying to explain the case where an address goes away due to an interface going down, but one shouldn't necessarily tear down the transport immediately. For example, if someone walks into an elevator, drops off of Wi-Fi, and then exits the elevator and comes back onto Wi-Fi, a non-migrating connection would do better to just wait for the network to come back, rather than tear down and interrupt a download.

mirjak: My question was more about when you should every even try to "forcibly disconnect" rather the just let the connection/protocol instance decide to close the connect...? And what the relevant policy bit here?

draft-ietf-taps-impl.md
mirjak: I was trying to separate this in the cases where a) the protocol does not support multipath migration in any more and b) where is does (however also realising that full multiparty support like MPTCP and path migration ala QUIC is maybe not that similar). So for a) the only interesting event is know that a path somehow is confirmed to be failed and therefore the protocol instance can decide to close the connection immediately rather than waiting for a time out, no?

gorryfair: Is this is about interfaces that fail, and what happens after. ... SCTP-PR, MP-TCP, QUIC, etc. I'm unsure I see why this is important except in corner cases, isn't this action like a reset (from the local network layer)?

draft-ietf-taps-impl.md
gorryfair: I agree with Tommy - if the protocol implements fail-over in some way, then it can use this. Otherwise, the transport ought to treat this as a soft error. It should not alter its state (to disconnect) because this makes it think a part of the path might not work.

mirjak: I also agree. I was asking if there every is a case where the protocol instance should be forcibly disconnect. If not we should rephrase this.

gorryfair: I think if the case exists, it is likely a special design - or a function of a specific transport, in general I'd say leave the transport to make the correct action.

draft-ietf-taps-impl.md
tfpauly, line 725:
```suggestion
When a path change occurs, e.g., when the IP address of an interface changes or a new interface becomes available, the Transport Services implementation is responsible for notifying the application of the change. The path change may interrupt connectivity on a path for an active connection or provide an opportunity for a transport that supports multipath or migration to adapt to the new paths.
```

draft-ietf-taps-impl.md
```diff
@@ -722,9 +722,13 @@ This enables implementations to realize transparent connection coalescing, conne
...
+
+If the Protocol Stack includes a transport protocol that also supports multipath connectivity with migration support, the Transport Services implementation should also inform the Protocol Instance of potentially new paths that become permissible based on the Selection Properties passed by the application. A protocol can then establish new subflows over new paths while an active path is still available or, if migration is supported, also after a break has been detected, and should attempt to tear down subflows over paths that are no longer used. The taps API provides an interface to set a multipath policy that indicates when and how different paths should be used. However, detailed handling of these policies is still implementation-specific. The decision about when to create a new path or to announce a new path or set of paths to the remote endpoint, e.g., in the form of additional IP addresses, is implementation-specific or could be be supported by future API extensions. If the Protocol Stack includes a transport protocol that does not support multipath, but does support migrating between paths, the update to the set of available paths can trigger the connection to be migrated.
```
tfpauly, line 729:
```suggestion
If the Protocol Stack includes a transport protocol that also supports multipath connectivity with migration support, the Transport Services implementation should also inform the Protocol Instance of potentially new paths that become permissible based on the Selection Properties passed by the application. A protocol can then establish new subflows over new paths while an active path is still available or, if migration is supported, also after a break has been detected, and should attempt to tear down subflows over paths that are no longer used. The Transport Services API provides an interface to set a multipath policy that indicates when and how different paths should be used. However, detailed handling of these policies is still implementation-specific. The decision about when to create a new path or to announce a new path or set of paths to the remote endpoint, e.g., in the form of additional IP addresses, is implementation-specific or could be be supported by future API extensions. If the Protocol Stack includes a transport protocol that does not support multipath, but does support migrating between paths, the update to the set of available paths can trigger the connection to be migrated. 
```

draft-ietf-taps-impl.md
```diff
@@ -722,9 +722,13 @@ This enables implementations to realize transparent connection coalescing, conne
...
+
+In case of Pooled Connections {{pooled-connections}}, the transport system may add connections over new paths or different protocols to the pool if permissible based on the multipath policy and Selection Properties. In case a previously used path becomes unavailable, the transport system may disconnect all connections that require this path, but should not disconnect the pooled connection object exposed to the application. The strategy how is implementation specific, but should me consistent with the behaviour of multipath transports.
```
tfpauly, line 731:
```suggestion
In case of Pooled Connections {{pooled-connections}}, the transport system may add connections over new paths or different protocols to the pool if permissible based on the multipath policy and Selection Properties. In case a previously used path becomes unavailable, the transport system may disconnect all connections that require this path, but should not disconnect the pooled connection object exposed to the application. The strategy how is implementation-specific, but should be consistent with the behavior of multipath transports. 
```

draft-ietf-taps-impl.md
tfpauly, line 727:
```suggestion
For protocols that do not support multipath or migration, the Protocol Instances should be informed of the path change, but should not be forcibly disconnected if the previously used path becomes unavailable.
```

tfpauly (approved):

gorryfair:
@mirjak : Is the path change problem resolved here - there's been stuff added on paths and policy since this.

PR #633, opened 2020-08-10
Clarify that amount of 0-RTT data varies by protocol, closes #580

### LINKED ISSUE #580: 5.3 Is  "Fast Open" just one "packet"?
The opening para reads to me like there is just one first packet. ... or is this actually more than one packets before the handshake complete?

theri:
Removed phrasing that implies 0-RTT data is just one packet and added a reference to the Maximum Message Size with Connection Establishment property.
Added a sentence on how/when an implementation can set this property.

### DIFF
--- a/draft-ietf-taps-impl.md
+++ b/draft-ietf-taps-impl.md
@@ -542,7 +542,9 @@ If a Connection becomes finished before a requested Receive action can be satisf
 
 ## Handling of data for fast-open protocols {#fastopen}
 
-Several protocols allow sending higher-level protocol or application data within the first packet of their protocol establishment, such as TCP Fast Open {{!RFC7413}} and TLS 1.3 {{!RFC8446}}. This approach is referred to as sending Zero-RTT (0-RTT) data. This is a desirable property, but poses challenges to an implementation that uses racing during connection establishment.
+Several protocols allow sending higher-level protocol or application data during their protocol establishment, such as TCP Fast Open {{!RFC7413}} and TLS 1.3 {{!RFC8446}}. This approach is referred to as sending Zero-RTT (0-RTT) data. This is a desirable property, but poses challenges to an implementation that uses racing during connection establishment.
+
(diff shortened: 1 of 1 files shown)


draft-ietf-taps-impl.md
```diff
@@ -542,7 +542,9 @@ If a Connection becomes finished before a requested Receive action can be satisf
 
 ## Handling of data for fast-open protocols {#fastopen}
 
-Several protocols allow sending higher-level protocol or application data within the first packet of their protocol establishment, such as TCP Fast Open {{!RFC7413}} and TLS 1.3 {{!RFC8446}}. This approach is referred to as sending Zero-RTT (0-RTT) data. This is a desirable property, but poses challenges to an implementation that uses racing during connection establishment.
+Several protocols allow sending higher-level protocol or application data during their protocol establishment, such as TCP Fast Open {{!RFC7413}} and TLS 1.3 {{!RFC8446}}. This approach is referred to as sending Zero-RTT (0-RTT) data. This is a desirable property, but poses challenges to an implementation that uses racing during connection establishment.
+
+The amount of data that can be sent as 0-RTT data varies by protocol and can be queried by the application using the `Maximum Message Size Concurrent with Connection Establishment` Connection Property. An implementation can set this property according to the protocols that it will race based on the given Selection Properties when the application requests to establish a connection.
```
abrunstrom, line 548: So this means the implementation would set it to the minimum size supported by all the protocols that may be raced? I think it is ok to leave the text as is, just want to check that I understand it correctly.

theri: I don't think we actually specify that anywhere, so I kept vague just saying it's based on the protocols, not what exactly the size should be. 
I guess that, unless the application sets 0-RTT as Required, there's no guarantee that the message will indeed be sent using 0-RTT. But if a message is sent, it should be the entire message and not just part of it, so setting the minimum probably makes sense.
Anyone have an opinion on whether I should add this to the PR?

abrunstrom: I think the current text is fine. As you say I think this is a property that you would set as Preferred rather than Required anyway. If it needs to be the whole message or not I think depends on the implementation and the protocol, for TCP this does not hold. I have merged the pull request.

Issue #571, opened 2020-04-17
How do you add multiple IP addresses/interfaces to an endpoint?


mirjak:
what happened if you add multiple addresses or interfaces to an endpoint: Does it overwrite the existing one or not?

MaxF12:
I would say that it does not overwrite but simply add to a list of addresses. This is mostly because for a Listen() the remote endpoint can be used to filter incoming connections and I don't think it makes much sense to only be able to constrain it to a single address.

philsbln:
I agree with @MaxF12 – but we should really make this explicit

mirjak:
Discussion at interim: Clarification is needed. add() during preconnection will overwrite, and add() cannot be used in connection state. Also add a note in the text that e.g. specifying a specific IP address might restrict mobility or use of multiple paths and could cause the connection to fails when the IP address goes way rather than enabling some handover. Still open: Do we need a way to add of multiple addresses in the first place? This is used for filtering remote IP addresses for listener for multicast.

tfpauly:
The current API text doesn't use "add", but rather "with...":

```
RemoteSpecifier := NewRemoteEndpoint()
RemoteSpecifier.WithIPv4Address(192.0.2.21)
RemoteSpecifier.WithPort(443)
```

It also implies that multiple is already allowed:

```
Multiple endpoint identifiers can be specified for each Local Endpoint and
Remote Endpoint.  For example, a Local Endpoint could be configured with two
interface names, or a Remote Endpoint could be specified via both IPv4 and
IPv6 addresses. These multiple identifiers refer to the same transport
endpoint.
```

How does that line up with the comment about `add()`, and having that replace?

theri:
As far as I can see, in the API draft we currently have add() for Properties and Security Parameters where each new add() overwrites the value for the old one.
So if we change this to add(), probably IP addresses should work the same.
But can't we just allow add() with a list of IP addresses instead of just with a single IP address?

tfpauly:
- Allow adding an alias endpoint to an endpoint
- Any one endpoint has at most one hostname
- Any one endpoint has at most one port/service (if you need more, add an alias)
- Any one endpoint has at most one address

If you need multiple source-specific multicast filters, use multiple connection objects

mwelzl:
I'm adding a point from @squarooticus here (and closing a separate issue that he opened, because this tackles the same problem space):

[...]

PR #637, opened 2020-08-14
Clarify behaviour of Rendezvous()

### LINKED ISSUE #584: Rendezvous() text should be a bit more detailed
See this email thread:  https://mailarchive.ietf.org/arch/msg/taps/B5iPOvreJ49aSINAqrM04jNNOVk/

Copy + pasting a relevant part from an email from @csperkins:

```
This is one of the areas where the drafts likely need expanding, but I think the model is:

1) application configures STUN (and maybe also TURN) local endpoints

2) application calls Resolve() on these, to find server reflexive candidates

3) application shares these candidates with its peer, via out-of-band means, and receives remote endpoint candidates from the peer

4) application adds remote endpoints for the peer’s candidates

5) application calls Rendezvous(), which performs the ICE-style probing to find a working path, based on the candidates, then returns s Connection

This relies on the Endpoint abstraction supporting STUN, which is maybe implicit in the drafts. 

```

To me at least, it wasn't clear from the text that rendezvous() would kick off ICE-style probing. It's kind of obvious (when else would it happen?), but I think it would be good for the draft to spell this out.

csperkins:
Closes #584

### DIFF
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -1413,9 +1413,42 @@ Remote Endpoint, and also the transport properties and security parameters
 needed for Protocol Stack selection.
 
 The Rendezvous() Action causes the Preconnection to listen on the Local
-Endpoint for an incoming Connection from the Remote Endpoint, while
-simultaneously trying to establish a Connection from the Local Endpoint to the
-Remote Endpoint. This corresponds to a TCP simultaneous open, for example.
+Endpoint for an incoming Connection from the Remote Endpoint, while also
+simultaneously trying to establish a Connection from the Local Endpoint to
+the Remote Endpoint. 
+
+If there are multiple local or remote endpoints configured, then a rendezvous
+action will systematically probe reachability of those endpoints following the
+approach used in Interactive Connectivity Establishment (ICE) {{?RFC5245}}.
+
+If the endpoints are suspected to be behind a NAT, Rendezvous() can be
+initiated using Local and Remote Endpoints that support a method of
+discovering NAT bindings such as Session Traversal Utilities for NAT (STUN)
+{{?RFC8489}} or Traversal Using Relays around NAT (TURN) {{?RFC5766}}.
+In this case, the Local Endpoint will resolve to a mixture of local and
+server reflexive addresses. The Resolve() action on the Preconnection can
+be used to discover these bindings:
+
+~~~
+[]Preconnection := Preconnection.Resolve()
+~~~
+
+The Resolve() call returns a list of Preconnection Objects, that represent the
+concrete addresses, local and server reflexive, on which a Rendezvous() for
+the Preconnection will listen for incoming Connections. These resolved
+Preconnections will share all other Properties with the Preconnection from
+which they are derived, though some Properties may be made more-specific by
+the resolution process. 
+
+An application that uses Rendezvous() to establish a peer-to-peer connection
+in the presence of NATs will configure the Preconnection object with a Local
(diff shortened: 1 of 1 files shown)


draft-ietf-taps-interface.md
```diff
@@ -1413,9 +1413,42 @@ Remote Endpoint, and also the transport properties and security parameters
...
-Endpoint for an incoming Connection from the Remote Endpoint, while
-simultaneously trying to establish a Connection from the Local Endpoint to the
-Remote Endpoint. This corresponds to a TCP simultaneous open, for example.
+Endpoint for an incoming Connection from the Remote Endpoint, while also
+simultaneously trying to establish a Connection from the Local Endpoint to
+the Remote Endpoint. 
+
+If there are multiple local or remote endpoints configured, then a rendezvous
+action will systematically probe reachability of those endpoints following the
+approach used in Interactive Connectivity Establishment (ICE) {{?RFC5245}}.
```
britram, line 1422: Do we need to specify this this tightly? (I think it's safe to do so, ICE is an approach, not a protocol. But we've pointedly avoided to make tight bindings to a Single Way Of Doing Things, so we should make this choice explicitly)

britram: also, if we're specifying ICE, this needs to be {{!RFC5245}}.

csperkins: Will rephrase to be slightly more general: "If there are multiple local or remote endpoints configured, then a rendezvous action will need to systematically probe reachability of those endpoints following an approach such as that used in Interactive Connectivity Establishment (ICE) {{?RFC5245}}."

draft-ietf-taps-interface.md
```diff
@@ -1413,9 +1413,42 @@ Remote Endpoint, and also the transport properties and security parameters
...
 
 The Rendezvous() Action causes the Preconnection to listen on the Local
-Endpoint for an incoming Connection from the Remote Endpoint, while
-simultaneously trying to establish a Connection from the Local Endpoint to the
-Remote Endpoint. This corresponds to a TCP simultaneous open, for example.
+Endpoint for an incoming Connection from the Remote Endpoint, while also
+simultaneously trying to establish a Connection from the Local Endpoint to
+the Remote Endpoint. 
+
+If there are multiple local or remote endpoints configured, then a rendezvous
```
gorryfair, line 1420: Should Local and Remote Endpoint be capitalised?

mwelzl (changes requested):
Thanks a lot for doing this, this is what I was hoping to see!

So, this LGTM except that I also request what @britram already said - maybe you could just do a quick check of the whole text (not just the sentence @britram quoted) to see if it's too prescriptive.

csperkins:
@mwelzl 7df2d9c should address this

csperkins:
```
   The Preconnection Object must be specified with both a Local Endpoint
   and a Remote Endpoint, and also the transport properties and security
   parameters needed for Protocol Stack selection.
```

To `Rendezvous()` we need multiple local and remote candidates. Should this be represented as either:
1) `Rendezvous()` taking two lists of `Preconnection` objects; or
2) as two `Preconnection` objects each with multiple addresses, ports, etc.

I suggest option 1, since this makes it easier to resolve a `Preconnection` to multiple preconnections, for NAT binding discovery.

Relates to #571 but with also different types of endpoint.

mwelzl:
I like option 1 because it contains complexity.

mwelzl (approved):

PR #639, opened 2020-08-15
Clarify priority framework

### LINKED ISSUE #619: Priority Framework is confusing
I'm not sure the priority framework really fits together.

Sec 6.4 says we 'should' do something that sounds like WFQ, but 7.1.5 allows the application to decide what happens. Meanwhile message priority suggests a strict priority-based scheduler.

Meanwhile, in sec 8.1.3.2 "Note that this property is not a per-message override of the  connection Priority - see Section 7.1.3.  Both Priority properties
   may interact, but can be used independently and be realized by
   different mechanisms." 

At the risk of overindexing on a partly-baked draft in httpbis, how would I implement strict priority of HTTP/3 streams over a taps API that implements QUIC? Would I need to request strict priority in *both* connection priority and message priority? Or I guess connection priority would be sufficient?

britram:
fixes #619 

(well, the priority framework is still confusing, this PR makes it clear that that is intentional, and why)

### DIFF
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -1521,16 +1521,10 @@ Connection -> CloneError<reason?>
 The Connection Priority Connection Property operates on entangled Connections 
 using the same approach as in {{msg-priority}}: when allocating available network
 capacity among Connections in a Connection Group, sends on Connections with
-higher Priority values will be prioritized over sends on Connections with
-lower Priority values. In keeping with the principle of least surprise,
-transport system implementation should, in the absense of other requirements, 
-assign each Connection the capacity share (M-N) x C / M, where N is the 
-Connection's Priority value, M is the maximum Priority value used by all 
-Connections in the group and C is the total available capacity. However, 
-the Priority setting is purely advisory, and no guarantees are given 
-about the way capacity is shared. Each implementation is free to implement a 
-way to share capacity that it sees fit. See {{priority-in-taps}}.
-
+lower Priority values will be prioritized over sends on Connections with
+higher Priority values. Capacity will be shared among these Connections according to
+the Connection Group Transmission Scheduler property ({{conn-scheduler}}). 
+See {{priority-in-taps}} for more.
 
 # Managing Connections {#introspection}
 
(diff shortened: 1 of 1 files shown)


draft-ietf-taps-interface.md
```diff
@@ -1516,17 +1519,18 @@ Attempts to clone a Connection can result in a CloneError:
 Connection -> CloneError<reason?>
 ~~~
 
-The Connection Property "Priority" operates on entangled Connections as in {{msg-priority}}:
-when allocating available network
+The Connection Priority Connection Property operates on entangled Connections 
+using the same approach in {{msg-priority}}: when allocating available network
```
mwelzl, line 1523:
```suggestion
using the same approach as in {{msg-priority}}: when allocating available network
```

draft-ietf-taps-interface.md
```diff
@@ -1516,17 +1519,18 @@ Attempts to clone a Connection can result in a CloneError:
...
-Each implementation is free to implement a way to share
-capacity that it sees fit.
+lower Priority values. In keeping with the principle of least surprise,
+transport system implementation should, in the absense of other requirements, 
+assign each Connection the capacity share (M-N) x C / M, where N is the 
+Connection's Priority value, M is the maximum Priority value used by all 
+Connections in the group and C is the total available capacity. However, 
+the Priority setting is purely advisory, and no guarantees are given 
+about the way capacity is shared. Each implementation is free to implement a 
+way to sharecapacity that it sees fit. See {{priority-in-taps}}.
```
mwelzl, line 1533:
```suggestion
way to share capacity that it sees fit. See {{priority-in-taps}}.
```

draft-ietf-taps-interface.md
```diff
@@ -2498,6 +2505,25 @@ the Connection could not be established will not result in a
 SendError separate from the InitiateError signaling the failure of Connection
 establishment.
 
+### Priority in TAPS {#priority-in-taps}
+
+The Transport Services interface provides two properties to allow a sender 
+to signal the relative priority of data transmission: the Priority Message 
+Property {{msg-priority}}, and the Connection Priority Connection Property 
+{{conn-priority}}. These properties are designed to allow the expression 
+and implementation a wide variety of approaches to transmission priority in
```
mwelzl, line 2514:
```suggestion
and implementation of a wide variety of approaches to transmission priority in 
```

draft-ietf-taps-interface.md
```diff
@@ -2498,6 +2505,25 @@ the Connection could not be established will not result in a
...
+to signal the relative priority of data transmission: the Priority Message 
+Property {{msg-priority}}, and the Connection Priority Connection Property 
+{{conn-priority}}. These properties are designed to allow the expression 
+and implementation a wide variety of approaches to transmission priority in 
+the transport and application layer, including those which do not appear on
+the wire (affecting only sender-side transmission scheduling) as well as those
+that do (e.g. {{?I-D.ietf-httpbis-priority}}.
+
+Given this variety, the specific behaviors of the prioritization framework 
+are leftto the transport system implementation and/or the implementation of
```
mwelzl, line 2520:
```suggestion
are left to the transport system implementation and/or the implementation of 
```

mwelzl:
Martin (I think) said that he'd check this offline after the Sep 11 interim.

draft-ietf-taps-interface.md
```diff
@@ -1515,18 +1518,13 @@ Attempts to clone a Connection can result in a CloneError:
 Connection -> CloneError<reason?>
 ~~~
 
-The Connection Property "Priority" operates on entangled Connections as in {{msg-priority}}:
-when allocating available network
+The Connection Priority Connection Property operates on entangled Connections
```
martinduke, line 1521: Bikeshed alert: can we just call it the "Priority Connection Property" to go with the "Priority Message Property"?

martinduke (approved):
Some nits, but I believe this addresses my concerns in the issue.

Issue #634, opened 2020-08-12
In Arch.... this sentence reads like a paper, rather than an Arch spec.


gorryfair:
OLD:
Among the Functional and Optimizing Features, some were common across all or nearly all transport protocols, while others could be seen as features that, if specified, would only be useful with a subset of protocols, but would not harm the functionality of other protocols.

Suggest NEW:
Some Functional and Optimizing Features are common across all or nearly all transport protocols. Other features might only be useful with a subset of protocols, but would not harm the functionality of other protocols.

gorryfair:
Re-reading this in context, I proposed a simpler one-word addition as a part of PR #641.

PR #641, opened 2020-08-17
Editorial draft-ietf-taps-arch.md

gorryfair:
This is a set of proposed editorial changes, following a complete pass through the ID. It also offers a one-word fix for Issue #634

This is mainly minor editorial work - except for a proposal to move the definition of "endpoint" upwards, because I found discussion of endpoints in preconception relied upon the endpoint concept. I also think we should have made our meaning of endpoint more explicit, because it is a term that we use a lot.

I also suggested we replaced the word "options" within the selection properties para, because a long time ago I think we changed the use of that word in this context  to "properties".

### DIFF
--- a/draft-ietf-taps-arch.md
+++ b/draft-ietf-taps-arch.md
@@ -354,7 +354,7 @@ The diagram below provides a high-level view of the actions and events during th
 ~~~~~~~~~~
 {: #fig-lifetime title="The lifetime of a Connection object"}
 
-### Endpoints
+### Endpoint Objects
 
 * Endpoint: An Endpoint represents an identifier for one side of a transport connection.
   Endpoints can be Local Endpoints or Remote Endpoints, and respectively represent an identity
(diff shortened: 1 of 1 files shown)


draft-ietf-taps-arch.md
```diff
@@ -206,7 +206,7 @@ The Transport Services API represents data as messages, so that it more closely
...
 * the ability to automatically assign messages and connections to underlying transport connections to utilize multi-streaming and pooled connections.
 
-Allowing applications to interact with messages is backwards-compatible with existings protocols and APIs, as it does not change the wire format of any protocol. Instead, it gives the protocol stack additional information to allow it to make better use of modern transport services, while simplifying the application's role in parsing data. For protocols which natively use a streaming abstraction, framers ({{datatransfer}}) bridge the gap between the two abstractions.
+Allowing applications to interact with messages is backwards-compatible with existings protocols and APIs because it does not change the wire format of any protocol. Instead, it gives the protocol stack additional information to allow it to make better use of modern transport services, while simplifying the application's role in parsing data. For protocols which natively use a streaming abstraction, framers ({{datatransfer}}) bridge the gap between the two abstractions.
```
tfpauly, line 209:
```suggestion
Allowing applications to interact with messages is backwards-compatible with existing protocols and APIs because it does not change the wire format of any protocol. Instead, it gives the protocol stack additional information to allow it to make better use of modern transport services, while simplifying the application's role in parsing data. For protocols which natively use a streaming abstraction, framers ({{datatransfer}}) bridge the gap between the two abstractions.
```

Issue #628, opened 2020-08-07
Clarify idle timeout, specify keep alive enablement


martinduke:
sec 7.1.4
It would be helpful to clarify the relevance of control vs application data in the idle timeout -- for various reasons, the connection could stay alive regardless of how active the connection is (in fact, perhaps "keepalive" is a generally useful notion that can be exposed in the API?)

tfpauly:
The timeout in 7.1.4 should include control data (like ACKs, etc). This should be clarified: "data" in this context is any transport data, not just application data.

We can have a simple boolean to enable keepalives as well.

mwelzl:
Aha, this should have been closed by PR #643, but "responds to issue #628" wasn't auto-detected by git. Manually closing now.

PR #643, opened 2020-08-17
Add new sections on keep alives

gorryfair:
Responds to issue #628.

### DIFF
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -1718,7 +1718,7 @@ Name:
 : keepaliveCount
 
 Type:
-: Numeric, with special value `Disabled`
+: Numeric`
 
 Default:
 : 5
@@ -1731,9 +1731,6 @@ will only take effect when the underlying stack both supports sending keep-alive
 and the `keepaliveTimeout` is enabled. 
 
 A value greater than 1 is desired to provide robustness to loss of a keep-alive packet or its response.
-A larger value can increase the time to detect a path failure or to perform multipath fail-over {{multipath-policy}}. 
-The special value `Disabled` means that the Local Endpoint does not need to monitor responses to keep-alive
-messages. 
 
 ### Connection Group Transmission Scheduler {#conn-scheduler}
 
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -1680,15 +1680,15 @@ Name:
 : connTimeout
 
 Type:
-: Numeric, with special value `Default`
+: Numeric, with special value `Disabled`
 
 Default:
-: Default
+: Disabled
 
 This property specifies how long to wait before deciding that an active Connection has
 failed when trying to reliably deliver data to the Remote Endpoint. Adjusting this Property
 will only take effect when the underlying stack supports reliability. The special value
-`Default` means that this timeout will use the default for the selected transport. 
+`Disabled` means that this timeout is not scheduled to happen.
 
 ### Timeout for keep alive packets {#keep-alive-timeout}
 
@@ -1696,10 +1696,10 @@ Name:
 : keepaliveTimeout
 
 Type:
-: Numeric, with special value `Disabled`
+: Numeric, with special value `Default`
 
 Default:
-: Disabled
+: Default
 
 A transport system can request a protocol that supports sending keep alive packets {{keep-alive}}.
 This property specifies the maximum time an idle connection (one for which no transport
@@ -1708,8 +1708,7 @@ the Local Endpoint sends a keep-alive packet to the Remote Endpoint. Adjusting t
 will only take effect when the underlying stack supports sending keep-alive packet. 
 Guidance on setting this value for datagram transports is 
 provided in {{!RFC8085}}. The special value
-`Disabled` means that this timeout is not scheduled to happen, and does not request the
-transport system to send keep-alive packets. 
+`Default` means that this timeout will use the default for the selected transport. 
 A value greater than {{conn-timeout}} will disable sending of keep-alive packets.
 
 ### Connection Group Transmission Scheduler {#conn-scheduler}
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -962,8 +962,8 @@ Default:
 This property specifies whether the application would like the Connection to send
 keep-alive packets or not. Note that if a Connection determines that keep-alive
 packets are being sent, the applicaton should itself avoid generating additional keep alive
-messages. Note that when supported, the system also needs to
-configure the generation and use of the keep alive-packets.
+messages. Note that when supported, the system will use the default periodic
+for generation of the keep alive-packets. (See also {{keep-alive-timeout}}).
 
 ### Interface Instance or Type {#prop-interface}
 
@@ -1680,17 +1680,15 @@ Name:
 : connTimeout
 
 Type:
-: Numeric, with special value `Disabled`
+: Numeric, with special value `Default`
 
 Default:
-: Disabled
+: Default
 
 This property specifies how long to wait before deciding that an active Connection has
 failed when trying to reliably deliver data to the Remote Endpoint. Adjusting this Property
 will only take effect when the underlying stack supports reliability. The special value
-`Disabled` means that this timeout is not scheduled to happen. This can be a valid
-choice with unreliable data transfer (e.g., when UDP is the underlying transport protocol).
-See also {{keep-alive-timeout}}.
+`Default` means that this timeout will use the default for the selected transport. 
 
 ### Timeout for keep alive packets {#keep-alive-timeout}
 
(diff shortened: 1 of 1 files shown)


draft-ietf-taps-interface.md
```diff
@@ -948,6 +948,22 @@ is usually combined with congestion control in protocol implementations,
...
+Type:
+: Preference
+
+Default:
+: Ignore
+
+This property specifies whether the application would like the Connection to send
+keep-alive packets or not. Note that if a Connection determines that keep-alive
+packets are being sent, the applicaton should itself avoid generating additional keep alive
+message {{!BCP145}}. Note that when supported, the system also needs to
```
mirjak, line 965: Where does BCP145 actually say that? Usually if the application sends keeps-lives with a higher rate that just mean the transport would never do as there is never a large enough idle time. So that's no problem. BCP145 says "Keep-alive messages are NOT RECOMMENDED for general use." however this is specific recommendation for UDP as time-out usually have to be low(er) for UDP (than TCP).

draft-ietf-taps-interface.md
```diff
@@ -1669,12 +1685,55 @@ Type:
...
+### Timeout for keep alive packets {#keep-alive-timeout}
+
+Name:
+: keepaliveTimeout
+
+Type:
+: Numeric, with special value `Disabled`
+
+Default:
+: Disabled
```
mirjak, line 1704: If it's possible to set this to disable we actually don't need the boolean on/of switch at all. But don't we usually set it to 0 or -1 and say that's a special value (didn't check but should do this uniformly).

draft-ietf-taps-interface.md
```diff
@@ -1669,12 +1685,55 @@ Type:
...
+
+A transport system can be requested a protcol that supports sending keep alive packets {{keep-alive}}.
+This property specifies the maximum time an idle connection (one for which no transport
+packets have been sent) should wait before 
+the Local Endpoint sends a keep-alive packet to the Remote Endpoint. Adjusting this Property
+will only take effect when the underlying stack supports sending keep-alive packet. The special value
+`Disabled` means that this timeout is not scheduled to happen, and does not request the
+transport system to send keep-alive packets. Guidance on setting this value is 
+provided in {{!BCP145}}.
```
mirjak, line 1713: Actually BCP145 is only for UDP. The default value for TCP is much larger. I guess we need a way to just config please use the default of the transport you select...?

draft-ietf-taps-interface.md
```diff
@@ -1669,12 +1685,55 @@ Type:
...
+
+### Limit for Keepalives {#keep-alive-count}
 
+Name:
+: keepaliveCount
```
mirjak, line 1718: Not sure we need this as this might usually be the same as the usual retry count in other cases of loss...?

draft-ietf-taps-interface.md
```diff
@@ -1669,12 +1685,55 @@ Type:
...
+
+Type:
+: Numeric, with special value `Disabled`
+
+Default:
+: 5
```
mirjak, line 1724: I guess this should be disabled as well if the other one is disabled by default.

draft-ietf-taps-interface.md
```diff
@@ -1669,12 +1685,55 @@ Type:
...
+
+Some protocols track responses to the sent keep-alive packets that are returned by the remote
```
mirjak, line 1726: This sentence makes me convince even more that we should leave this one to protocol specific parameters.

draft-ietf-taps-interface.md
```diff
@@ -1669,12 +1685,55 @@ Type:
...
+peer. When this function is supported, the Local Endpoint counts the number of consecutive 
+keep-alive packets for which no corresponding response is received. If this count
+equals the `keepaliveTimeout` then the Connection is aborted. This Property
+will only take effect when the underlying stack both supports sending keep-alive packets
+and the `keepaliveTimeout` is enabled. 
+
+A value greater than 1 is desired to provide robustness to loss of a keep-alive packet or its response.
+A larger value can increase the time to detect a path failure or to perform multipath fail-over {{multipath-policy}}.
```
mirjak, line 1734: Not sure if this is an issue as you only send keep-alives when your are idle otherwise, so taking more time to detect failure is not a huge issue. However, not sure if you even should start the failover when you are idle anyway or just wait until new data is coming. But that might depend on the application (and protocol e.g. if 0-RTT is supported or not). I guess  that is a good discussion to have somewhere in some draft but maybe not here but somewhere in the implementation draft?

draft-ietf-taps-interface.md
```diff
@@ -1669,12 +1685,55 @@ Type:
...
+peer. When this function is supported, the Local Endpoint counts the number of consecutive 
+keep-alive packets for which no corresponding response is received. If this count
+equals the `keepaliveTimeout` then the Connection is aborted. This Property
+will only take effect when the underlying stack both supports sending keep-alive packets
+and the `keepaliveTimeout` is enabled. 
+
+A value greater than 1 is desired to provide robustness to loss of a keep-alive packet or its response.
+A larger value can increase the time to detect a path failure or to perform multipath fail-over {{multipath-policy}}. 
+The special value `Disabled` means that the Local Endpoint does not need to monitor responses to keep-alive
+messages.
```
mirjak, line 1736: Hm, just reading this now. That's not what I would have read as disabled. But it's that really a protocol-specific thing but then usually something that just there or not and can not be configured?

draft-ietf-taps-interface.md
gorryfair: I was trying to find a REF that said don't waste bandwidth on excessive keep-alives. BCP145 probably wasn't the best, I'm removing that - because it's not the focus at all - so do we need a REF?

draft-ietf-taps-interface.md
gorryfair: Sure, added that this was only datagrams. I think the RFC 5382 as I recall only says that with TCP you need to tune per app.

draft-ietf-taps-interface.md
gorryfair: I don't think it is the same number. I'm also not sure all Apps need the same number.

draft-ietf-taps-interface.md
gorryfair: That might be OK, but then it means the App needs to say if you use SCTP, do this, if you use TCP do that, if you use STUN do this, etc. I suspect this becomes messy very quickly.

draft-ietf-taps-interface.md
gorryfair: I'd be happy to remove line 1734...  I'll remove "A larger value can increase the time to detect a path failure or to perform multipath fail-over {{multipath-policy}}. "

draft-ietf-taps-interface.md
gorryfair: You could also say "infinite" has the same semantics, since it carries on trying to probe for ever.... so maybe we don't need an explicit disable. I'll remove that also and see what is left.

draft-ietf-taps-interface.md
gorryfair: Not so. I think this is a real number, that you can use a safe robustness factor. The default when the method is enabled it should be this.

draft-ietf-taps-interface.md
mirjak: Don't waste bandwidth is a good advise and BCP145 might be a good reference, however, it not relevant on switch layer you waste the bandwidth :-)

draft-ietf-taps-interface.md
mirjak: What I'm saying is that this is rather a protocol decision than an application one. Why do I as an application that is anyway idle need to control when the connection will be declared failed?

draft-ietf-taps-interface.md
mwelzl: I agree that this eliminates the need for the on/off switch (unless I'm mising something?). And no, special values 0 or -1 aren't used anymore, they have been replaced with more appropriate types.

draft-ietf-taps-interface.md
```diff
@@ -1669,12 +1685,52 @@ Type:
...
+: Numeric, with special value `Disabled`
+
+Default:
+: Disabled
+
+A transport system can be requested a protocol that supports sending keep alive packets {{keep-alive}}.
+This property specifies the maximum time an idle connection (one for which no transport
+packets have been sent) should wait before 
+the Local Endpoint sends a keep-alive packet to the Remote Endpoint. Adjusting this Property
+will only take effect when the underlying stack supports sending keep-alive packet. The special value
```
mwelzl, line 1710:
```suggestion
will only take effect when the underlying stack supports sending keep-alive packets. The special value
```

mwelzl: Shouldn't the "when" in this sentence be an "if" ?   Both choices can be justified, I guess, as things can change over time, but I think that "if" matches the more common case.

draft-ietf-taps-interface.md
```diff
@@ -1669,12 +1685,52 @@ Type:
...
+transport system to send keep-alive packets. Guidance on setting this value for datagram transports is 
+provided in {{!BCP145}}.
+
+### Limit for Keepalives {#keep-alive-count}
+
+Name:
+: keepaliveCount
+
+Type:
+: Numeric`
```
mwelzl, line 1721:
```suggestion
: Numeric
```

draft-ietf-taps-interface.md
mwelzl: Yes, I think the intention is to design a generic, protocol-independent type of keep-alive here?  I certainly don't think "if you use SCTP do this, if you use TCP do that" etc. is how this should become.

mwelzl (approved):
Works for me... I'm a bit uneasy about this but maybe it's because my thinking of this isn't grounded in a use case...

draft-ietf-taps-interface.md
mwelzl: .... and this is another case of a BCP reference that makes the compilation fail. Please replace the BCP reference with {{!RFC8085}}.

mwelzl:
Interim: to be turned into a single knob

draft-ietf-taps-interface.md
gorryfair: The other text surrounding uses /when/, however, I';d be happy with /if/.

draft-ietf-taps-interface.md
gorryfair: I am OK with /if/, however the surrounding text says /when/.

britram:
Merging because GitHub UX is confusing. @gorryfair to go over these comments again and add a second PR if necessary.

PR #644, opened 2020-08-28
Clairified collections - closes #612

### LINKED ISSUE #612: Allow setting multiple values for some Security Parameters
If calling Set() for Security Parameters overrides (#609), then the API should allow setting a Security Parameter to multiple values, e.g., using a list. This would be needed, e.g., to set multiple cipher suites.

I don't think we need to make this overly formal. Adding one sentence to the Security Parameters section should be enough. 
Right?

philsbln:


draft-ietf-taps-interface.md
```diff
@@ -531,6 +531,12 @@ Transport Properties can have one of a set of data types:
   selection; see {{selection-props}}. When querying, a Preference is of
   type Boolean, with `true` indicating that the Selection Property
   has been applied.
+- Tuple: takes values composed from types defined above.
+  In this this document, it is written as a list in brackets, e.g., ```(Enumeration, Preference)``` 
+  The compsition of types and the order depends on the property and is fix for the property.
+  The actual representation is implementation-dependen.
+- Collection: takes a valiable number of a single of the above types.
+  The actual representation, e.g. as set or array, is implementation-dependen.
```
tfpauly, line 539:
```suggestion
  The actual representation, e.g. as a set or an array, is implementation-dependent.
```

draft-ietf-taps-interface.md
```diff
@@ -531,6 +531,12 @@ Transport Properties can have one of a set of data types:
   selection; see {{selection-props}}. When querying, a Preference is of
   type Boolean, with `true` indicating that the Selection Property
   has been applied.
+- Tuple: takes values composed from types defined above.
+  In this this document, it is written as a list in brackets, e.g., ```(Enumeration, Preference)``` 
+  The compsition of types and the order depends on the property and is fix for the property.
+  The actual representation is implementation-dependen.
+- Collection: takes a valiable number of a single of the above types.
```
tfpauly, line 538:
```suggestion
- Collection: A group of one or more value types.
```

draft-ietf-taps-interface.md
```diff
@@ -531,6 +531,12 @@ Transport Properties can have one of a set of data types:
   selection; see {{selection-props}}. When querying, a Preference is of
   type Boolean, with `true` indicating that the Selection Property
   has been applied.
+- Tuple: takes values composed from types defined above.
+  In this this document, it is written as a list in brackets, e.g., ```(Enumeration, Preference)``` 
+  The compsition of types and the order depends on the property and is fix for the property.
```
tfpauly, line 536:
```suggestion
  The composition of types and the order depends on the property and is fix for the property.
```

draft-ietf-taps-interface.md
```diff
@@ -531,6 +531,12 @@ Transport Properties can have one of a set of data types:
   selection; see {{selection-props}}. When querying, a Preference is of
   type Boolean, with `true` indicating that the Selection Property
   has been applied.
+- Tuple: takes values composed from types defined above.
```
tfpauly, line 534: A specific grouping of multiple value types.

draft-ietf-taps-interface.md
philsbln: I like your wording but would prefer to stress that these have to be of the same type.

draft-ietf-taps-interface.md
```diff
@@ -531,6 +531,12 @@ Transport Properties can have one of a set of data types:
   selection; see {{selection-props}}. When querying, a Preference is of
   type Boolean, with `true` indicating that the Selection Property
   has been applied.
+- Tuple: A specific grouping of multiple value types.
```
theri, line 534:
```suggestion
- Tuple: An ordered grouping of multiple value types.
```
Unless "specific" really means something here? But I wouldn't know what it means.

draft-ietf-taps-interface.md
```diff
@@ -531,6 +531,12 @@ Transport Properties can have one of a set of data types:
...
+  In this this document, it is written as a list in brackets, e.g., ```(Enumeration, Preference)``` 
+  The composition of types and the order depends on the property and is fixed for the property.
+  The actual representation is implementation-dependen.
```
theri, line 537:
```suggestion
  The actual representation is implementation-dependent.
```

draft-ietf-taps-interface.md
```diff
@@ -531,6 +531,12 @@ Transport Properties can have one of a set of data types:
   selection; see {{selection-props}}. When querying, a Preference is of
   type Boolean, with `true` indicating that the Selection Property
   has been applied.
+- Tuple: A specific grouping of multiple value types.
+  In this this document, it is written as a list in brackets, e.g., ```(Enumeration, Preference)``` 
+  The composition of types and the order depends on the property and is fixed for the property.
```
theri, line 536:
```suggestion
  The composition of types and their order depends on the property and is fixed for the property.
```

draft-ietf-taps-interface.md
```diff
@@ -531,6 +531,12 @@ Transport Properties can have one of a set of data types:
...
+  The actual representation is implementation-dependen.
+- Collection: A group of one or more value of the same type.
```
theri, line 538:
```suggestion
- Collection: An unordered grouping of one or more values of the same type.
```
This way, it's consistent with Tuple being a "grouping" as well.

draft-ietf-taps-interface.md
```diff
@@ -531,6 +531,12 @@ Transport Properties can have one of a set of data types:
...
+  The actual representation, e.g. as a set or an array, is implementation-dependent.
```
theri, line 539: Not sure we need this sentence. It basically says "The implementation is implementation-dependent", doesn't it? 
Also, a set is not necessarily ordered, while an array is usually ordered, I would say. So I think removing this sentence would make the entire thing less confusing.

mwelzl (approved):

mwelzl:
I should have said: I approve this, modulo the changes requested by @theri.

draft-ietf-taps-interface.md
philsbln: I still see value in the sentence – especially stressing that the grouping is by design unordered, but might be implemented with some ordered data structure.

theri: I'm not sure why it'd be necessary to stress how it could be implemented in the interface draft.
But I'm fine with merging this PR as it is.

theri (approved):
Thanks for incorporating my suggested changes. I think this can be merged now.

PR #650, opened 2020-09-08
Change Excessive Retransmit to Excessive Loss

### LINKED ISSUE #647: retransmission vs. loss?
We have an Excessive retransmissions event (section 7.3.2) which says this is only available for reliable transport. However, QUIC with datagram would also be able to generate a loss notification (even though data would not be retransmitted as there are ACKs anyway). Shouldn't this be an Excessive loss event instead?

philsbln:
This PR is a try to apply the change discussed in Issue #647 "retransmission vs. loss?"

closes #647 if merged

### DIFF
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -1139,7 +1139,7 @@ Default:
 : Ignore
 
 This property specifies whether an application considers it useful to be
-informed in case sent data was lost more often than a certain
+informed in case packets of the underlying transport protocol have been lost more often than a certain
 threshold (see {{conn-excss-loss}} for configuration of this threshold).
 
 
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -1127,7 +1127,7 @@ unidirectional communication may cause a transport stack that supports
 bidirectional communication to be selected.
 
 
-### Notification of Excessive Loss {#prop-establish-loss-notify}
+### Notification of Excessive Loss {#prop-establish-excss-loss-notify}
 
 Name:
 : excessLossNotify
@@ -2938,7 +2938,7 @@ coverage, see {{prop-checksum-control-send}} and {{prop-checksum-control-receive
 `TCP-specific Property: User Timeout` ({{tcp-uto}}).
 
 * Notification of Excessive Retransmissions (early warning below abortion threshold):
-`Notification of excessive retransmissions` property ({{prop-establish-retrans-notify}}).
+`Notification of excessive retransmissions` property ({{prop-establish-excss-loss-notify}}).
 
 * Notification of ICMP error message arrival:
 `Notification of ICMP soft error message arrival` property ({{prop-soft-error}}).
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -1127,10 +1127,10 @@ unidirectional communication may cause a transport stack that supports
 bidirectional communication to be selected.
 
 
-### Notification of excessive retransmissions {#prop-establish-retrans-notify}
+### Notification of Excessive Loss {#prop-establish-loss-notify}
 
 Name:
-: retransmitNotify
+: excessLossNotify
 
 Type:
 : Preference
@@ -1139,8 +1139,8 @@ Default:
 : Ignore
 
 This property specifies whether an application considers it useful to be
-informed in case sent data was retransmitted more often than a certain
-threshold (see {{conn-excss-retransmit}} for configuration of this threshold).
+informed in case sent data was lost more often than a certain
+threshold (see {{conn-excss-loss}} for configuration of this threshold).
 
 
 ### Notification of ICMP soft error message arrival {#prop-soft-error}
@@ -1612,19 +1612,19 @@ Many Connection Properties have a corresponding Selection Property that
 enables applications to express their preference for protocols providing a supporting transport feature.
 
 
-### Retransmission Threshold Before Excessive Retransmission Notification {#conn-excss-retransmit}
+### Loss Threshold before Excessive Loss Notification {#conn-excss-loss}
 
 Name:
-: retransmitNotifyThreshold
+: excessLossNotifyThreshold
 
 Type:
-: Integer, with special value `Disabled`
+: Numeric, with special value `Disabled`
 
 Default:
 : Disabled
 
-This property specifies the number of retransmissions required to cause the application
-to be notified of `Excessive Retransmissions`.
+This property specifies the number of lost packtes per second required to cause the application
+to be notified of `Excessive Loss`.
 
 ### Required Minimum Corruption Protection Coverage for Receiving {#conn-recv-checksum}
 
@@ -1936,14 +1936,15 @@ is no guarantee that a soft error will be signaled.
 Connection -> SoftError<>
 ~~~
 
-### Excessive retransmissions {#conn-retrans-notify}
+### Excessive Loss {#conn-excss-loss-notify}
 
-This event notifies the application of excessive retransmissions, based on a
-configured threshold (see {{conn-excss-retransmit}}). This will only happen if
-the underlying protocol stack supports reliability and, with it, such notifications.
+This event notifies the application of excessive loss events,
+e.g., indicated by non-acked packets or exessive retransmissions,
+based on a configured threshold (see {{conn-excss-loss}}).
+This will only happen if the underlying protocol stack supports loss detection and, with it, such notifications.
 
 ~~~
-Connection -> ExcessiveRetransmission<>
+Connection -> ExcessiveLoss<>
 ~~~
 
 # Data Transfer {#datatransfer}


mirjak (approved):

draft-ietf-taps-interface.md
```diff
@@ -1612,19 +1612,19 @@ Many Connection Properties have a corresponding Selection Property that
...
 Type:
-: Integer, with special value `Disabled`
+: Numeric, with special value `Disabled`
 
 Default:
 : Disabled
 
-This property specifies the number of retransmissions required to cause the application
-to be notified of `Excessive Retransmissions`.
+This property specifies the number of lost packtes per second required to cause the application
```
mwelzl, line 1626:
```suggestion
For reliable data transfer, this property specifies the number of retransmissions required to cause the application
```

draft-ietf-taps-interface.md
```diff
@@ -1612,19 +1612,19 @@ Many Connection Properties have a corresponding Selection Property that
...
+to be notified of `Excessive Loss`.
```
mwelzl, line 1627:
```suggestion
to be notified of `Excessive Retransmission`. In case of unreliable data transfer, configuring this threshold will not have any effect.
```

draft-ietf-taps-interface.md
```diff
@@ -1127,10 +1127,10 @@ unidirectional communication may cause a transport stack that supports
 bidirectional communication to be selected.
 
 
-### Notification of excessive retransmissions {#prop-establish-retrans-notify}
+### Notification of Excessive Loss {#prop-establish-excss-loss-notify}
```
mwelzl, line 1130:
```suggestion
### Notification of Loss {#prop-establish-excss-loss-notify}
```

draft-ietf-taps-interface.md
```diff
@@ -1936,14 +1936,15 @@ is no guarantee that a soft error will be signaled.
 Connection -> SoftError<>
 ~~~
 
-### Excessive retransmissions {#conn-retrans-notify}
+### Excessive Loss {#conn-excss-loss-notify}
```
mwelzl, line 1939:
```suggestion
### Loss {#conn-excss-loss-notify}
```

draft-ietf-taps-interface.md
```diff
@@ -1936,14 +1936,15 @@ is no guarantee that a soft error will be signaled.
...
 
-This event notifies the application of excessive retransmissions, based on a
-configured threshold (see {{conn-excss-retransmit}}). This will only happen if
-the underlying protocol stack supports reliability and, with it, such notifications.
+This event notifies the application of excessive loss events,
```
mwelzl, line 1941:
```suggestion
This event notifies the application of loss events,
```

draft-ietf-taps-interface.md
```diff
@@ -1936,14 +1936,15 @@ is no guarantee that a soft error will be signaled.
...
+e.g., indicated by non-acked packets or exessive retransmissions,
```
mwelzl, line 1942:
```suggestion
e.g., indicated by non-acked packets or (in case of reliable data transfer) retransmissions,
```
(note: if you don't accept this suggestion, still fix the typo: s/exessive/excessive)

draft-ietf-taps-interface.md
```diff
@@ -1936,14 +1936,15 @@ is no guarantee that a soft error will be signaled.
...
+based on a configured threshold (see {{conn-excss-loss}}).
+This will only happen if the underlying protocol stack supports loss detection and, with it, such notifications.
 
 ~~~
-Connection -> ExcessiveRetransmission<>
+Connection -> ExcessiveLoss<>
```
mwelzl, line 1947:
```suggestion
Connection -> Loss<>
```

draft-ietf-taps-interface.md
```diff
@@ -2937,7 +2938,7 @@ coverage, see {{prop-checksum-control-send}} and {{prop-checksum-control-receive
 `TCP-specific Property: User Timeout` ({{tcp-uto}}).
 
 * Notification of Excessive Retransmissions (early warning below abortion threshold):
-`Notification of excessive retransmissions` property ({{prop-establish-retrans-notify}}).
+`Notification of excessive retransmissions` property ({{prop-establish-excss-loss-notify}}).
```
mwelzl, line 2941:
```suggestion
`Notification of loss` property ({{prop-establish-excss-loss-notify}}).
```

mwelzl (changes requested):
I suggested many changes. They build upon the following premise:

- for unreliable data transfer, we shouldn't assume that the underlying transport can allow a threshold to be configured (losing so many packets / sec), but just inform the application every time. The application can ignore these notifications, or decide to act depending on how many notifications have arrived, possibly per time interval, but these decisions should be application logic.
- for reliable transfer, we need to apply a threshold based on retransmissions, or else we're not compatible with TCP.

If you agree with these two points, please accept all my suggestions, and then this should be good to go  :-)

draft-ietf-taps-interface.md
```diff
@@ -1139,8 +1139,8 @@ Default:
 : Ignore
 
 This property specifies whether an application considers it useful to be
-informed in case sent data was retransmitted more often than a certain
-threshold (see {{conn-excss-retransmit}} for configuration of this threshold).
+informed in case sent data was lost more often than a certain
+threshold (see {{conn-excss-loss}} for configuration of this threshold).
```
philsbln, line 1143: Okay… I agree on that one – In case we want to keep the semantics of this PR (see comment below) packets of the PR, it hast to say 
```suggestion
informed in case packets of the underlying transport protocol have been lost more often than a certain
threshold (see {{conn-excss-loss}} for configuration of this threshold).
```
or it makes no sense.

draft-ietf-taps-interface.md
```diff
@@ -1612,19 +1612,19 @@ Many Connection Properties have a corresponding Selection Property that
 enables applications to express their preference for protocols providing a supporting transport feature.
 
 
-### Retransmission Threshold Before Excessive Retransmission Notification {#conn-excss-retransmit}
+### Loss Threshold before Excessive Loss Notification {#conn-excss-loss}
```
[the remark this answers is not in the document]
philsbln: This makes sense for TCP, but not for QUIC with datagrams as @mirjak stated in issue #647 – QUIC still detects the loss, but does not necessarily retransmit the data. Each retransmit is usually also a loss event (ignoring re-ordering problems), but not the other way around.

philsbln:
I guess @mirjak's idea behind #647 was to challenge the hard separation between reliable and unreliable transport (as QUIC datagrams enables mixing both), while @mwelzl's two points, exactly refer to Minset and state that there is a need for a hard separation. – This gap is exactly why I marked this PR as discuss and we need to agree on.

- If we want to accompany QUIC datagram in a more general framework, we have to talk about loss (and set aside whether it was compensated by retransmitting the packet), which slightly diverges from TCP semantics in a deterministic way.
  We can still expose in a TCP specific properties/events – what this PR tries to do.
- If we want to exactly match TCP as described in Minset, we need to separate retransmit and loss and either need to build ugly QUIC specific stuff or ignore QUIC datagrams.

draft-ietf-taps-interface.md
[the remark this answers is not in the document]
gorryfair: The transport data was not "lost" the IP packets were. I've changed my position to argue against having this metric at all. It is very transport specific. (see my comment in #647).

abrunstrom:
I do not see why a separation of the events for reliable and unreliable transport is not aligned with QUIC datagrams? I think it makes sense to separate the two events as different information makes sense for reliable and unreliable transport. You will not ask to send something reliably and get it sent in a QUIC datagram? The underlying QUIC connection may carry both a reliable and an unrelaible TAPS connection, but it will still provide different semantics to the two.

draft-ietf-taps-interface.md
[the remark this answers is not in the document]
mwelzl: Right - generally, the threshold doesn't make sense for QUIC, but my other proposed text changes eliminate the threshold's meaning for unreliable data transfer.

draft-ietf-taps-interface.md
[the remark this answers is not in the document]
mwelzl: The IP packets are not known to the application. This should really refer to Messages.

draft-ietf-taps-interface.md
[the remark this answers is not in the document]
mirjak: My point was that having a loss notification does make sense for unreliable transports like datagram in QUIC and therefore it should not only apply for reliable protocols.

mwelzl:
See my suggestions in #647 - I think there's a pretty easy way to address both worlds, and this is what my text changes proposed (but they would all need to be applied together).

draft-ietf-taps-interface.md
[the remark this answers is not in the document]
mwelzl: Yes! but the threshold makes no sense, and calling it "Excessive" doesn't, either. Keep the loss notification, but use the threshold for reliable transfers only, I say.

mwelzl:
TODO after interim: Close this, make another PR to remove "Excessive Retransm." (TCP, for Michael)

PR #652, opened 2020-09-09
Add a Group Connection Limit, extend Client example with clone.

### LINKED ISSUE #618: Streams/Connections Relationship is unclear
I found the relationship between streams and connections to be extremely murky throughout the draft, and left it not really sure how I would control a streamed transport.

Section 6.4 hints suggests that opening additional streams is an alternative implementation of Clone() for transports that can do so. Does this imply that if the peer stream limit does not allow new streams, the transport should open a connection instead? How does SetNewConnectionLimit() in Sec 6.2 relate to streamed transport -- are there two levels of listener: a stream listener and a connection listener? How would opening "stream listeners" relate to the maximum number of streams the peer could open (i.e. do I have to call listen() 8 times to allow the client to open 8 streams)? Or are there no stream listeners and these "connections" just sort of appear?

Skimming the -impl draft, I can see there's some text here but it is not fitting together for me. There are at least three streamed transports and it would be helpful to explain how the API conceptualizes this, perhaps in a new section.

mwelzl:
Closes #618.

### DIFF
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -324,7 +324,7 @@ byte stream into a sequence of Messages ({{framing}}).
 ### Server Example
 
 This is an example of how an application might listen for incoming Connections
-using the Transport Services Interface, receive a request, and send a response.
+using the Transport Services Interface, and receive a request, and send a response.
 
 ~~~
 LocalSpecifier := NewLocalEndpoint()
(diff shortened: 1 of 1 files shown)


draft-ietf-taps-interface.md
```diff
@@ -1926,6 +1930,24 @@ based on a UTO option received from the remote peer. This boolean becomes false
...
+Name:
+: groupConnLimit
+
+Type:
+: Numeric (with special value `Unlimited`)
+
+Default:
+: Unlimited
+
+This property controls the number Connections that may arrive from
```
britram, line 1944:
```suggestion
This property controls the number of Connections that may arrive from
```

britram (approved):

PR #653, opened 2020-09-10
Text on clone for TCP, and a little bit for SCTP. Closes #507.

### LINKED ISSUE #507: More text on clone

mwelzl:
Closes #507. 
This provides more detail on how to implement Clone with TCP, and clarifies that SCTP's "Initiate" should not yield a stream of an already existing SCTP Association in case "activeReadBeforeSend" is Preferred or Required.


draft-ietf-taps-impl.md
```diff
@@ -867,7 +891,7 @@ ConnectionReceived:
 : TCP Listeners will deliver new connections once they have replied to an inbound SYN with a SYN-ACK.
 
 Clone:
-: Calling `Clone` on a TCP Connection creates a new Connection with equivalent parameters. The two Connections are otherwise independent.
+: Calling `Clone` on a TCP Connection creates a new Connection with equivalent parameters. These Connections, and Connections generated via later calls to `Clone` on one of them, form a Connection Group. To realize `entanglement` for these Connections, with the exception of `Priority(Connection)`, changing a Connection Property on one of them must affect the Connection Properties of the others too. The Connection Property `Priority(Connection)` can be ignored; when it is reasonable to assume that Connections traverse the same path (e.g., when they share the same encapsulation), support for it can also experimentally be implemented using a congestion control coupling mechanism (see for example {{TCP-COUPLING}} or {{?RFC3124}}).
```
abrunstrom, line 895: I do not understand the part about the Priority being ignored if connections traverse the same path?

abrunstrom (commented):
Added a question for @mwelzl

draft-ietf-taps-impl.md
mwelzl: These are two separate things, the semicolon is misleading. My point was that the property can be ignored, as no guarantees are given to the application.

draft-ietf-taps-impl.md
```diff
@@ -867,7 +891,7 @@ ConnectionReceived:
 : TCP Listeners will deliver new connections once they have replied to an inbound SYN with a SYN-ACK.
 
 Clone:
-: Calling `Clone` on a TCP Connection creates a new Connection with equivalent parameters. The two Connections are otherwise independent.
+: Calling `Clone` on a TCP Connection creates a new Connection with equivalent parameters. These Connections, and Connections generated via later calls to `Clone` on one of them, form a Connection Group. To realize `entanglement` for these Connections, with the exception of `Connection Priority`, changing a Connection Property on one of them must affect the Connection Properties of the others too. No guarantees of honoring the Connection Property `Connection Priority` are given, and thus it is safe for an implementation of a transport system to ignore this property. When it is reasonable to assume that Connections traverse the same path (e.g., when they share the same encapsulation), support for it can also experimentally be implemented using a congestion control coupling mechanism (see for example {{TCP-COUPLING}} or {{?RFC3124}}).
```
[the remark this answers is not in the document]
theri: I think this would have helped my confusion

PR #654, opened 2020-09-10
Various nits and wordsmithing

squarooticus:
Feel free to cherry-pick. Some of these are self-explanatory; in other cases, I'll add a comment to explain the change.

### DIFF
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -321,7 +321,7 @@ If none of the available transport protocols provides Preservation of Message
 Boundaries, but there is a transport protocol that provides a reliable ordered
 byte stream, an application could receive this byte stream as partial
 Messages and transform it into application-layer Messages.  Alternatively,
-an application might provide a Message Framer, which transforms a
+an application might provide a Message Framer, which can transform a
 sequence of Messages into a byte stream and vice versa ({{framing}}).
 
 ### Server Example
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -322,7 +322,7 @@ Boundaries, but there is a transport protocol that provides a reliable ordered
 byte stream, an application could receive this byte stream as partial
 Messages and transform it into application-layer Messages.  Alternatively,
 an application might provide a Message Framer, which transforms a
-byte stream to or from a sequence of Messages ({{framing}}).
+sequence of Messages into a byte stream and vice versa ({{framing}}).
 
 ### Server Example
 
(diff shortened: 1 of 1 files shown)


draft-ietf-taps-interface.md
```diff
@@ -318,8 +321,8 @@ If none of the available transport protocols provides Preservation of Message
 Boundaries, but there is a transport protocol that provides a reliable ordered
 byte stream, an application could receive this byte stream as partial
 Messages and transform it into application-layer Messages.  Alternatively,
-an application might provide a Message Framer, which can transform a
-byte stream into a sequence of Messages ({{framing}}).
+an application might provide a Message Framer, which transforms a
+byte stream to or from a sequence of Messages ({{framing}}).
```
mwelzl, line 325:
```suggestion
sequence of Messages into a byte stream and vice versa ({{framing}}).
```

I like your suggestion better than what we have (the original just maps better to why this is really necessary in our own minds IMO), but now it seems awkward to begin with the byte stream... I just think that it reads a bit nicer with the change above.

draft-ietf-taps-interface.md
```diff
@@ -318,8 +321,8 @@ If none of the available transport protocols provides Preservation of Message
 Boundaries, but there is a transport protocol that provides a reliable ordered
 byte stream, an application could receive this byte stream as partial
 Messages and transform it into application-layer Messages.  Alternatively,
-an application might provide a Message Framer, which can transform a
-byte stream into a sequence of Messages ({{framing}}).
+an application might provide a Message Framer, which transforms a
```
mwelzl, line 324:
```suggestion
an application might provide a Message Framer, which can transform a
```
I suggest to keep the "can transform" because I don't think Message Framers *always* do this transformation.

draft-ietf-taps-interface.md
```diff
@@ -1242,7 +1246,7 @@ SecurityParameters.Set('pre-shared-key', key, identity)
 ~~~
 
 - Session cache management: Used to tune cache capacity, lifetime, re-use,
-and eviction policies, e.g., LRU or FIFO.may also me changed, but are implementation-specific.
+and eviction policies, e.g., LRU or FIFO.may also be changed, but are implementation-specific.
```
mwelzl, line 1249:
```suggestion
and eviction policies, e.g., LRU or FIFO.
```

I agree with your proposed change, but this missed something more - the sentence was broken ("FIFO.may ..."). ISTM that this extra sentence is not needed here anyway.

mwelzl (changes requested):
Thanks a lot for doing this, several great catches there!  I made a few minor suggestions

PR #674, opened 2020-09-11
move text on early name binding to privacy consideration

### LINKED ISSUE #657: Early binding  -> privacy considerations
I would reword this text in §5.1:

> The API explicitly does not require the application to resolve names, though there is a tradeoff between early and late binding of addresses to names. Early binding allows the API implementation to reduce connection setup latency, at the cost of potentially limited scope for alternate path discovery during Connection establishment, as well as potential additional information leakage about application interest when used with a resolution method (such as DNS without TLS) which does not protect query confidentiality.

such that the tradeoff is listed in a (new) privacy considerations section, leaving the description of the reduced connection setup latency in early binding here as a functional justification.

britram:
closes #657

### DIFF
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -691,7 +691,7 @@ endpoint.
 
 The transport services API resolves names internally, when the Initiate(),
 Listen(), or Rendezvous() method is called to establish a Connection. Privacy
-considerations for the timing of this resolution are given in {{privacy-security}}
+considerations for the timing of this resolution are given in {{privacy-security}}.
 
 The Resolve() action on Preconnection can be used by the application to force
 early binding when required, for example with some Network Address Translator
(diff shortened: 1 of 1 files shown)


draft-ietf-taps-interface.md
```diff
@@ -690,14 +690,8 @@ IPv6 addresses. These multiple identifiers refer to the same transport
...
-Listen(), or Rendezvous() method is called to establish a Connection. The API
-explicitly does not require the application to resolve names, though there is
-a tradeoff between early and late binding of addresses to names. Early binding
-allows the API implementation to reduce connection setup latency, at the cost
-of potentially limited scope for alternate path discovery during Connection
-establishment, as well as potential additional information leakage about
-application interest when used with a resolution method (such as DNS without
-TLS) which does not protect query confidentiality.
+Listen(), or Rendezvous() method is called to establish a Connection. Privacy
+considerations for the timing of this resolution are given in {{privacy-security}}
```
mwelzl, line 694:
```suggestion
considerations for the timing of this resolution are given in {{privacy-security}}.
```

mwelzl (approved):
LGTM  (note: accept 1 nit fix before pressing Merge)

PR #680, opened 2020-09-18
Inherit some msg defaults from the connection, and divide fragmentation into layers 3 and 4.

### LINKED ISSUE #666: Message properties
Shouldn't the default value for every message property basically be "inherit", i.e., from the connection?

§8.1.3.3:

> If true, it specifies that the receiver-side transport protocol stack only delivers the Message to the receiving application after any existing previous ordered Message that was passed to the same Connection via the Send Action.

This is a bit awkwardly worded. If this is supported for both reliable and unreliable transports (unclear?), then:

* For reliable transports, this means that this message will not be delivered to the application until all previous ordered messages on the connection are delivered.
* For unreliable transports, this means that any message prior to this one in ordering but received subsequently will not be delivered to the application.

### LINKED ISSUE #667: Fragmentation
I'm not 100% sure, but I believe the DF bit applies only to intermediate hops, not to the originating machine. E.g., if you have DONTFRAG set on your socket (or at the system level via pmtud) and shove 60K into a sendto() call, the kernel will still happily chop your datagram into multiple <=1500 octet packets. So, the text in §8.1.3.9 is probably misleading.

mwelzl:
Closes #666. Closes #667.


squarooticus (approved):

draft-ietf-taps-interface.md
```diff
@@ -2380,11 +2383,27 @@ Default:
 
 This property specifies that a message should be sent and received as a single
 packet without network-layer fragmentation, if possible.
-Attempts to send a message with this property set to a size greater than the
+This only takes effect when used with transports supporting this functionality
+and running over IP version 4. When it does take effect, setting this property to
```
gorryfair, line 2387: Why only IPv4? IPv4 and IPv6 stacks can perform host fragmentation.
I'd expected: "When it does take effect, setting this property to true will inhibit network layer fragmentation by the endpoint, and will cause the IPv4 to set the Don't Fragment bis in the IP header, ...

draft-ietf-taps-interface.md
```diff
@@ -3054,8 +3073,11 @@ coverage, see {{prop-checksum-control-send}} and {{prop-checksum-control-receive
 * "Specify minimum checksum coverage required by receiver" and "Disable checksum requirement when receiving":
 `Required Minimum Corruption Protection Coverage for Receiving` property ({{conn-recv-checksum}}) and `Full Checksum Coverage on Receiving` property ({{prop-checksum-control-receive}}).
 
-* "Specify DF" field and "Request not to bundle messages":
-the `No Fragmentation` Message Property combines both of these requests, i.e. if a request not to bundle messages is made, this also turns off fragmentation (i.e., sets DF=1) in the case of a protocol that allows this (only UDP and UDP-Lite, which cannot bundle messages anyway) ({{send-singular}}).
+* "Specify DF field":
+`No Network-Layer Fragmentation` property ({{send-singular}}).
```
gorryfair, line 3077: Still slightly unhappy with the "v4-centric" words of "DF field" and would prefer something different and definitely to omit "field"

draft-ietf-taps-interface.md
```diff
@@ -2380,11 +2383,27 @@ Default:
...
+true will cause the Don't Fragment bit to be set in the IP header, and
+attempts to send a message with this property set to a size greater than the
 transport's current estimate of its maximum packet size (`singularTransmissionMsgMaxLen`)
-will
-result in a `SendError`. When used with transports supporting this functionality
-and running over IP version 4, the Don't Fragment bit will be set.
+will result in a `SendError`.
+
+#### No Transport-Layer Fragmentation {#no-transport-fragmentation}
+
```
gorryfair, line 2394: We could call this "segmentation"?
- At least I would use "segmentation" here, to avoid the ever-returning argument about fragments being good/bad.

draft-ietf-taps-interface.md
```diff
@@ -2380,11 +2383,27 @@ Default:
 
 This property specifies that a message should be sent and received as a single
 packet without network-layer fragmentation, if possible.
-Attempts to send a message with this property set to a size greater than the
+This only takes effect when used with transports supporting this functionality
```
gorryfair, line 2386: This isn't really about the feature of /transports/ ?????
- Actually this is the transport service when the transport uses a network layer that supports....

gorryfair (changes requested):
I think this heads in the correct direction, but I still have issues that I think need thought, and probably changes

draft-ietf-taps-interface.md
mwelzl: I think the concern was about making a false promise for the IPv6 case. As long as we say that the system will attempt to do this, that's okay.

draft-ietf-taps-interface.md
mwelzl: Here I actually disagree: I don't think that "segment" is the word used by all transports? Seems rather TCP-specific to me...   IMO, there's a slight risk that "segmentation" is more cryptic to some readers than "transport-layer fragmentation", which is rather explicit about what exactly it is. Don't you think?

draft-ietf-taps-interface.md
mwelzl: Oh, this shouldn't change, it's a reference to minset, where this is the exact wording!

draft-ietf-taps-interface.md
gorryfair: I don't agree - but I also don't care on which words we use, as long as we separate the definition of terms.

mwelzl: Ok - if I couldn't convince you, I feel that going with your view is the better choice. I'll adjust the text.

draft-ietf-taps-interface.md
gorryfair: So, suggest: /This only takes effect when the transport uses a network layer that supports this functionality/

draft-ietf-taps-interface.md
gorryfair: Or... call it transport fragmentation, and explicitly say that it is about how a transport segments the data into network-layer packets.

draft-ietf-taps-interface.md
```diff
@@ -2380,11 +2383,27 @@ Default:
...
+
+Type:
+: Boolean
+
+Default:
+: false
+
+This property specifies that a message should be sent and received as a single
+network-layer datagram without transport-layer fragmentation, if possible.
+This only takes effect when used with transports supporting this functionality.
```
gorryfair, line 2407: I suggest something like: 

When set to true, this property requests the network layer at the 
sending endpoint to not fragment the packets generated by the
transport layer.  When running over IPv4, setting this property to
true will also cause the Don't Fragment bit to be set in the IP header
When this property is set, an
attempt to send a message size greater than the
transport's current estimate of its maximum packet size (`singularTransmissionMsgMaxLen`)
will result in a `SendError`.
This only takes effect when the transport and network layer 
support this functionality.

draft-ietf-taps-interface.md
mwelzl: I'm following your suggestion above, which eliminates the IPv4 part of the sentence too, so I think this is settled.

draft-ietf-taps-interface.md
mwelzl: Thanks!  I now used this text; because it says very clearly what is going on, it seemed ok to me to also call the heading "No Segmentation".

draft-ietf-taps-interface.md
mwelzl: I went with "No Segmentation".

gorryfair (approved):
Looks like these issues have been addressed in this PR

PR #686, opened 2020-10-19
Connection properties: clarify that no guarantees are given for priority,…

mwelzl:
… and move the group limit connection property to the right place  (it was misplaced).

Note that section 8.2.7 ("Priority in TAPS") already clarifies that a Connection Priority choice doesn't guarantee a certain outcome; this just adds another sentence in the paragraph describing the Connection property to make it clearer (less easy to miss).

### DIFF
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -1873,11 +1873,11 @@ Type:
 Default:
 : Unlimited
 
-This property controls the number of Connections that may arrive from
+This property controls the number of Connections that can be accepted from
 a peer as new members of the Connection's group. Similar to SetNewConnectionLimit(),
 this limits the number of ConnectionReceived Events that will occur, but constrained
-to the group of the Connection associated with this property. In case of a multi-streaming
-transport, this takes the role of a limit on the number of allowed streams.
+to the group of the Connection associated with this property. For a multi-streaming transport,
+this limits the number of allowed streams.
 
 ### Read-only Connection Properties {#read-only-conn-prop}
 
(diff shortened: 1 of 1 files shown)


draft-ietf-taps-interface.md
```diff
@@ -1861,6 +1862,23 @@ lower-bound rate below which the application does not deem
...
+Name:
+: groupConnLimit
+
+Type:
+: Numeric (with special value `Unlimited`)
+
+Default:
+: Unlimited
+
+This property controls the number of Connections that may arrive from
```
gorryfair, line 1876: Sorry - but being picky - you can't control what "may arrive", you can control what might be "accepted"

gorryfair (commented):
2 minor comments

draft-ietf-taps-interface.md
```diff
@@ -1861,6 +1862,23 @@ lower-bound rate below which the application does not deem
...
+a peer as new members of the Connection's group. Similar to SetNewConnectionLimit(),
+this limits the number of ConnectionReceived Events that will occur, but constrained
+to the group of the Connection associated with this property. In case of a multi-streaming
+transport, this takes the role of a limit on the number of allowed streams.
```
gorryfair, line 1880: NiT: /in case of/in the case of/
However: /this take the role of a/
... why so obscure about a role? ... or is there more to be said? ... I don't understand whether there needs to be more said.

Instead, would it be fair just to say:
 /For a multistreaming transport, this limits the numbering of allowed streams./

draft-ietf-taps-interface.md
mwelzl: Fixed; I wrote "...that can be accepted from ..."

draft-ietf-taps-interface.md
mwelzl: Done, except that I wrote "number" instead of "numbering"  (I don't understand the use of "numbering" here, but I believe that "number" should be ok anyway)

gorryfair:
Looks ready to merge to me.

Issue #658, opened 2020-09-10
To abstract or not to abstract? That is the question


squarooticus:
Selection property querying is better than it was, but querying still feels odd:

* Using "has been applied" to mean "on" or "in use" or "active" necessitates a bunch of explanatory text in §5.2 that would otherwise be unnecessary.

* Reporting "true" for properties that are "on" or "in use" or "active" is very UNIX-y, but less abstract than I'd like to see here. At least we're not encoding -1 as a special return value anymore. A perhaps suitably abstract interface here would be something like:

```
tp := Connection.GetTransportProperties()
if tp.has('reliability'): ...
```

mwelzl:
Welcome, author!

squarooticus:
Ugh. Can I insert a GIF of a pile of books falling on me?

tfpauly:
Also, make SecurityParameters more abstract (don't do "strings" that look like Python).

squarooticus:
> Also, make SecurityParameters more abstract (don't do "strings" that look like Python).

I made a change in the third commit of #658. LMK if that matches your expectations.

squarooticus:
Er, I mean #687. @tfpauly

PR #687, opened 2020-10-20
Improve API abstraction #658

squarooticus:
This is probably not done, but it's getting closer to separating the abstract API from concrete API details. Edits, comments, and criticism are welcome, of course.

### DIFF
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -876,35 +876,10 @@ Setting a Transport Property to a value overrides the previous value of this Tra
 TransportProperties.Set(property, value)
 ~~~
 
-Transport Properties may also be restored to implementation defaults:
-
-~~~
-TransportProperties.Unset(property)
-~~~
-
 To aid readability, implementations MAY provide additional convenience functions to simplify use of Selection Properties: see {{preference-conv}} for examples.
 In addition, implementations MAY provide a mechanism to create TransportProperties objects that are preconfigured for common use cases as outlined in {{property-profiles}}.
 
-For an existing Connection, the Transport Properties can be queried any time
-by using the following call on the Connection Object:
-
-~~~
-TransportProperties := Connection.GetTransportProperties()
-~~~
-
-An individual property is then queried via the following call on the
-`TransportProperties` object:
-
-~~~
-value := TransportProperties.Get(property)
-~~~
-
-If the given Transport Property is of type Boolean, `Has` may be provided as an
-alias for `Get`, useful for readability in such constructions as the following:
-
-~~~
-if TransportProperties.Has(boolean_property) then ...
-~~~
+Transport Properties for an established connection can be queried via the Connection object, as outlined in {{introspection}}.
 
 A Connection gets its Transport Properties either by being explicitly configured
 via a Preconnection, by configuration after establishment, or by inheriting them
@@ -1734,6 +1709,8 @@ At any point, the application can query Connection Properties.
 
 ~~~
 ConnectionProperties := Connection.GetProperties()
+value := ConnectionProperties.Get(property)
+if ConnectionProperties.Has(boolean_or_preference_property) then ...
 ~~~
 
 Depending on the status of the connection, the queried Connection
@@ -1745,17 +1722,19 @@ Properties will include different information:
 * Whether the connection can be used to send data. A connection can not be used
   for sending if the connection was created with the Selection Property
   `Direction of Communication` set to `unidirectional receive` or if a Message
-  marked as `Final` was sent over this connection, see {{msg-final}}.
+  marked as `Final` was sent over this connection. See {{msg-final}}.
 
 * Whether the connection can be used to receive data. A connection cannot be
   used for reading if the connection was created with the Selection Property
   `Direction of Communication` set to `unidirectional send` or if a Message
-  marked as `Final` was received, see {{receiving-final-messages}}. The latter
+  marked as `Final` was received. See {{receiving-final-messages}}. The latter
   is only supported by certain transport protocols, e.g., by TCP as half-closed
   connection.
 
-* For Connections that are Establishing: Transport Properties that the
-  application specified on the Preconnection, see {{selection-props}}.
+* For Connections that are Established or Closing: Transport Properties that the
+  application specified on the Preconnection. See {{selection-props}}.
+  Selection properties of type `Preference` will be exposed as boolean values
+  indicating whether or not the property applies to the selected transport.
 
 * For Connections that are Established, Closing, or Closed: Selection
   ({{selection-props}}) and Connection Properties ({{connection-props}}) of the
@@ -1831,7 +1810,7 @@ Type:
 Default:
 : Disabled
 
-This property specifies how many seconds to wait before deciding that an active Connection has
+This property specifies how long to wait before deciding that an active Connection has
 failed when trying to reliably deliver data to the Remote Endpoint. Adjusting this Property
 will only take effect when the underlying stack supports reliability. The special value
 `Disabled` means that no timeout is scheduled.
@@ -1839,22 +1818,22 @@ will only take effect when the underlying stack supports reliability. The specia
 ### Timeout for keep alive packets {#keep-alive-timeout}
 
 Name:
-: keepaliveTimeout
+: keepAliveTimeout
 
 Type:
-: Numeric
+: Numeric, with special value `Disabled`
 
 Default:
 : Implementation-defined
 
 A Transport Services system can request a protocol that supports sending keep alive packets {{keep-alive}}.
-This property specifies the maximum number of seconds an idle connection (one for which no transport
+This property specifies the maximum length of time an idle connection (one for which no transport
 packets have been sent) should wait before 
 the Local Endpoint sends a keep-alive packet to the Remote Endpoint. Adjusting this Property
 will only take effect when the underlying stack supports sending keep-alive packets. 
 Guidance on setting this value for datagram transports is 
 provided in {{!RFC8085}}.
-A value greater than the connection timeout ({{conn-timeout}}) will disable the sending of keep-alive packets.
+A value greater than the connection timeout ({{conn-timeout}}), or the special value `Disabled`, will disable the sending of keep-alive packets.
 
 ### Connection Group Transmission Scheduler {#conn-scheduler}
 
@@ -3156,6 +3135,11 @@ platform, where they can be immediately detected, such as by throwing an
 exception when attempting to initiate a connection with inconsistent
 Transport Properties. An error can provide an optional reason to the
 application with further details about why the error occurred.
+
+## Time Duration
+
+Time duration types are implementation-specific.
+For instance, it could be a number of seconds, number of milliseconds, or a `struct timeval` in C or a user-defined `Duration` class in C++.
 
 # Convenience Functions
 
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -539,13 +539,11 @@ for Protocol Specific Properties and MUST NOT be used for vendor or implementati
 Transport Properties each have a type, which can be:
 
 - One of the basic types described in {{notation}}; or
-- Preference, which on Preconnection is an Enumeration with six possible
-  values: Prohibit, Avoid, Ignore, Prefer, Require, or Default. Each of the
-  first five denotes a level of preference of a given property during protocol
-  selection, while Default indicates deference to an implementation's preference.
-  (See {{selection-props}}.) When querying the properties of a Connection
-  following protocol selection, a Preference is of type Boolean, with `true`
-  indicating that the Selection Property applies to the chosen transport.
+- Preference, which is an Enumeration with five possible
+  values: Prohibit, Avoid, Ignore, Prefer, or Require. Each of these
+  denotes a level of preference of a given property during protocol
+  selection.
+  (See {{selection-props}}.) The Preference type is used only on Preconnections, and only for Selection Properties.
 
 ## Scope of the Interface Definition
 
@@ -833,7 +831,7 @@ between multiple local interfaces that are connected to different access
 networks.
 
 Most Selection Properties are represented as Preferences, which can
-take one of six values:
+take one of five values:
 
    | Preference | Effect                                                                 |
    |------------|------------------------------------------------------------------------|
@@ -842,7 +840,6 @@ take one of six values:
    | Ignore     | No preference                                                          |
    | Avoid      | Prefer protocols/paths not providing the property, proceed otherwise   |
    | Prohibit   | Select only protocols/paths not providing the property, fail otherwise |
-   | Default    | Use the implementation's default preference level for this property    |
 {: #tab-pref-levels title="Selection Property Preference Levels"}
 
 The implementation MUST ensure an outcome that is consistent with all application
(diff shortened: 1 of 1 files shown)


draft-ietf-taps-interface.md
```diff
@@ -520,31 +538,16 @@ for Protocol Specific Properties and MUST NOT be used for vendor or implementati
...
-per property; it is up to implementations how these special values are
-represented (e.g., by using -1 for an otherwise non-negative value).
+- One of the basic types described in {{notation}}; or
+- Preference, which on Preconnection is an Enumeration with six possible
+  values: Prohibit, Avoid, Ignore, Prefer, Require, or Default. Each of the
+  first five denotes a level of preference of a given property during protocol
+  selection, while Default indicates deference to an implementation's preference.
+  (See {{selection-props}}.) When querying the properties of a Connection
+  following protocol selection, a Preference is of type Boolean, with `true`
+  indicating that the Selection Property applies to the chosen transport.
```
MaxF12, line 550: I think this can be a bit confusing. It might be better to just keep the first two sentences here and move the last one to after the table (line 732).

draft-ietf-taps-interface.md
```diff
@@ -777,6 +777,20 @@ by using the following call on the Connection Object:
...
+value := TransportProperties.Get(property)
+~~~
+
+If the given Transport Property is of type Boolean, `Has` may be provided as an
+alias for `Get`, useful for readability in such constructions as the following:
+
+~~~
+if TransportProperties.Has(boolean_property) then ...
+~~~
+
```
MaxF12, line 793: Isn't this redundant (in most languages)? Wouldn't `if TransportProperties.Get(boolean_property) then` deliver the same result? In any case, this should be moved to Appendix A (convenience functions).

draft-ietf-taps-interface.md
```diff
@@ -2951,6 +2965,44 @@ and for contributing text, e.g., on multicast.
...
+  a C function that returns a non-negative `int` on success may return -1 on
+  failure. In Python, such a function might return `None` or raise an
+  exception.
+- In C, a Tuple may be represented as a `struct` with one member for each of
+  the value types in the ordered grouping. In Python, by contrast, a Tuple may
+  be represented natively as a `tuple`, a sequence of dynamically-typed
+  elements.
+- A Collection may be represented as a `std::set` in C++ or as a `set` in
+  Python. In C, it may be represented as an array or as a higher-level data
+  structure with appropriate accessors defined.
```
MaxF12, line 2994: This might be too implementation specific for the API draft and would maybe better fit into the implementation draft.

MaxF12 (commented):
Thanks for doings this, I think it's a good change. I added some comments, mostly about reordering things a bit.

draft-ietf-taps-interface.md
mwelzl: I'm not sure... it's about the interface, and explains the abstract notation - it can help the reader mentally translate the text into their own language of choice.

For instance, Kyle himself probably didn't like the use of "-1" as an error because that's not his programming mindset...  and he probably thought that such a section would have helped him, as a reader. I think that applies to others in the same way - so I believe that this fits here.

draft-ietf-taps-interface.md
```diff
@@ -2951,6 +2965,44 @@ and for contributing text, e.g., on multicast.
...
+
+## Events and Errors
+
+This specification treats Events and Errors similarly. Errors, just as any
+other Events, may occur asynchronously in network applications. However, it is
+recommended that implementations of this interface also return Errors
+immediately, according to the error handling idioms of the implementation
+platform, for errors that can be immediately detected, such as inconsistency
+in Transport Properties. An error can provide an optional reason to the
+application with further details about why the error occurred.
```
mwelzl, line 3004: If anything here would rather fit in the implementation section, then it's perhaps this paragraph - because it has the words: "it is recommended that implementations of this interface ..."  - this is implementation guidance, then. Maybe re-formulate, or move it there.

mwelzl (changes requested):
I agree, this is very useful, thanks!  I inserted a few comments.

draft-ietf-taps-interface.md
squarooticus: The purpose of this PR is to make the abstract API actually abstract. -1 is a C-ism, and C is about as inabstract as languages get. :-)

FWIW, I absolutely think the guidance in this appendix is appropriate for the interface draft because it's describing via example appropriate patterns for translating an abstract interface into an API in a real language, whereas the implementation draft is about how to implement the functionality that TAPS provides in a language-agnostic way.

draft-ietf-taps-interface.md
MaxF12: True, it might not be best to have different behaviour depending on wether the connection has been established or not. Perhaps it would be an option to be able to query the transport properties on a Connection object (i.e. Connection.get_property(property)) which will return what the property actually turned out to be while queries on the TransportProperties will always return the preference level. 
I am also unsure what the current PR text would mean for reusability of TransportProperties, if the preference level is overwritten with a bool after connection establishment I guess the same TransportProperties object could not be used in a second Preconnection.

squarooticus: Ooh... this might actually make the case for Get != Has: Get returns the original preference, while Has is defined only after protocol selection (e.g., throws an exception prior) and always flattens it to a boolean based on the selected protocol.

draft-ietf-taps-interface.md
```diff
@@ -727,12 +730,9 @@ have one of five preference levels:
    | Ignore     | No preference                                                          |
    | Avoid      | Prefer protocols/paths not providing the property, proceed otherwise   |
    | Prohibit   | Select only protocols/paths not providing the property, fail otherwise |
+   | Default    | Use the implementation's default preference level for this property    |
```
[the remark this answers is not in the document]
mwelzl: I agree that unsetting a variable could be a good concept, but I'd be afraid of the number of possible changes to the whole document that this approach might incur... is this worth it?

Anyway, either way is fine with me; even just keeping "Default" as a preference level, I could live with, it's not a big deal IMO.

draft-ietf-taps-interface.md
mwelzl: So this is resolved; sorry for mentioning "-1", I realize it was a bad example  :)

draft-ietf-taps-interface.md
MaxF12: I still think that would cause some problems if you reuse transport properties. What if you use the same TransportProperties object in two different Preconnections that lead to two Connections with different selected protocols?

britram (approved):
This is a very good start; I'd actually be happy merging it and tweaking in follow-up PRs at this point.

draft-ietf-taps-interface.md
```diff
@@ -520,31 +538,16 @@ for Protocol Specific Properties and MUST NOT be used for vendor or implementati
...
-  The actual representation is implementation-dependent.
-- Collection: An unordered grouping of one or more values of the same type.
-  The actual representation, e.g. as a set or an array, is implementation-dependent.
+Transport Properties each have a type, which can be:
 
-For types Integer and Numeric, special values can be defined
-per property; it is up to implementations how these special values are
-represented (e.g., by using -1 for an otherwise non-negative value).
+- One of the basic types described in {{notation}}; or
+- Preference, which on Preconnection is an Enumeration with six possible
```
tfpauly, line 544:
```suggestion
- Preference, which is an Enumeration with six possible
```

draft-ietf-taps-interface.md
philsbln: That's solvable: Just make ```has``` a method on the connection

mwelzl:
@britram - we could do this, and take notes of the suggested changes (which should be easy to apply) in an issue - but then again, how is this more efficient than updating this PR?

draft-ietf-taps-interface.md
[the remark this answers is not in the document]
philsbln: I don't think we have many places where this would require a change – just in the examples and the description of Preconnection / TransportProperties. Default values stay default values, unset/reset just restores it.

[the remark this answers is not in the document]
mirjak: When would you want to "reset" anything?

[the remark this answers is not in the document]
gorryfair: Just agree/do anything ... this isn't super important!

draft-ietf-taps-interface.md
```diff
@@ -2951,6 +2965,44 @@ and for contributing text, e.g., on multicast.
...
+  be represented natively as a `tuple`, a sequence of dynamically-typed
+  elements.
+- A Collection may be represented as a `std::set` in C++ or as a `set` in
+  Python. In C, it may be represented as an array or as a higher-level data
+  structure with appropriate accessors defined.
+
+## Events and Errors
+
+This specification treats Events and Errors similarly. Errors, just as any
+other Events, may occur asynchronously in network applications. However, it is
```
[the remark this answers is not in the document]
squarooticus: I just moved this; I didn't change the wording. FWIW, I like "may occur" both because some errors are synchronous, and because sometimes there aren't errors.

draft-ietf-taps-interface.md
squarooticus: I'll just soften the language.

draft-ietf-taps-interface.md
squarooticus: I think we need to discuss this one.

draft-ietf-taps-interface.md
```diff
@@ -1812,30 +1829,29 @@ Type:
 Default:
 : Disabled
 
-This property specifies how long to wait before deciding that an active Connection has
+This property specifies how many seconds to wait before deciding that an active Connection has
```
squarooticus, line 1832: Here and below I propose that these times are in seconds. Is that right, or do we want the unit of duration to be implementation-specific?

draft-ietf-taps-interface.md
```diff
@@ -1812,30 +1829,29 @@ Type:
...
-`Disabled` means that this timeout is not scheduled to happen.
+`Disabled` means that no timeout is scheduled.
 
 ### Timeout for keep alive packets {#keep-alive-timeout}
 
 Name:
 : keepaliveTimeout
 
 Type:
-: Numeric, with special value `Default`
```
squarooticus, line 1826: I removed `Default` here because of the new `Unset` method. Does there need to be a `Disabled` value for when keepalive functionality is off?

mwelzl (approved):
LGTM now modulo open discussion. FWIW, I don't have a strong opinion regarding spec'ing seconds or leaving it implementation-dependent - there are pro's and con's...

draft-ietf-taps-interface.md
```diff
@@ -879,7 +876,13 @@ Setting a Transport Property to a value overrides the previous value of this Tra
 TransportProperties.Set(property, value)
 ~~~
 
-Selection Properties of type `Preference` might often be frequently used. Implementations MAY therefore provide additional convenience functions to simplify use, see {{preference-conv}} for examples.
+Transport Properties may also be restored to implementation defaults:
+
+~~~
+TransportProperties.Unset(property)
+~~~
```
tfpauly, line 883: I'd suggest removing this

draft-ietf-taps-interface.md
```diff
@@ -889,6 +892,20 @@ by using the following call on the Connection Object:
...
+
+~~~
+value := TransportProperties.Get(property)
+~~~
+
+If the given Transport Property is of type Boolean, `Has` may be provided as an
+alias for `Get`, useful for readability in such constructions as the following:
+
+~~~
+if TransportProperties.Has(boolean_property) then ...
```
tfpauly, line 906: I'd suggest removing get and has here, and just specifying a bit more detail on Connection.GetProperties()

draft-ietf-taps-interface.md
theri: I suggest adding a reference to Section 7 here, where it explains how to query the "result" of your Selection Properties on a Connection.

squarooticus: * Remove getters: Get, Has, Unset are not a required part of the API for a preconnection-phase Property object.
* Clean up getting preferences in 4.2.2.

draft-ietf-taps-interface.md
tfpauly: Let's instead say that the representation of duration is implementation specific?

squarooticus: Say something about the representation of duration being implementation-specific.

draft-ietf-taps-interface.md
```diff
@@ -1812,30 +1831,29 @@ Type:
...
-: Numeric, with special value `Default`
+: Numeric
 
 Default:
-: Default
+: Implementation-defined
 
 A Transport Services system can request a protocol that supports sending keep alive packets {{keep-alive}}.
-This property specifies the maximum time an idle connection (one for which no transport
+This property specifies the maximum number of seconds an idle connection (one for which no transport
```
tfpauly, line 1851: Same here

draft-ietf-taps-interface.md
squarooticus: Resolve in accordance with removing Unset

draft-ietf-taps-interface.md
```diff
@@ -3070,6 +3088,45 @@ and for contributing text, e.g., on multicast.
...
+- An Integer can typically be represented in C by an `int` or `long`, subject
+  to the underlying platform's ranges for each. To accommodate special values,
+  a C function that returns a non-negative `int` on success may return -1 on
+  failure. In Python, such a function might return `None` or raise an
+  exception.
+- In C, a Tuple may be represented as a `struct` with one member for each of
+  the value types in the ordered grouping. In Python, by contrast, a Tuple may
+  be represented natively as a `tuple`, a sequence of dynamically-typed
+  elements.
+- A Collection may be represented as a `std::set` in C++ or as a `set` in
```
tfpauly, line 3115: We could mention time duration representation here (can be a strong type in C++, in C it could be milliseconds as an integer).

squarooticus:
Add something about time durations to appendix? In C++ it could be a time duration object, while in C it might be seconds.

draft-ietf-taps-interface.md
squarooticus: Not sure if I got this right. Check the corresponding text in the updated PR.

mwelzl (approved):
lgtm

draft-ietf-taps-interface.md
```diff
@@ -1726,17 +1722,19 @@ Properties will include different information:
...
+  marked as `Final` was received. See {{receiving-final-messages}}. The latter
   is only supported by certain transport protocols, e.g., by TCP as half-closed
   connection.
 
-* For Connections that are Establishing: Transport Properties that the
-  application specified on the Preconnection, see {{selection-props}}.
+* For Connections that are Established or Closing: Transport Properties that the
+  application specified on the Preconnection. See {{selection-props}}.
+  Selection properties of type `Preference` will be exposed as boolean values
+  indicating whether or not the property applies to the selected transport.
```
philsbln, line 1737: I think we should find a better solution for that – will try to make a PR once this one jas landed.

philsbln (approved):
Works for me.

Still, I think we should re-iterate the case of querying how to query selection results in a new PR.

PR #697, opened 2020-11-02
Clarify endpoint objects

### LINKED ISSUE #571: How do you add multiple IP addresses/interfaces to an endpoint?

### LINKED ISSUE #623: List of endpoint attributes
sec 5.1
I found it odd that there didn't appear to be any sort of formal list of attributes of an endpoint, just some examples.

tfpauly:
Closes #571, closes #623

### DIFF
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -613,10 +613,111 @@ Preconnection during pre-establishment.
 ## Specifying Endpoints {#endpointspec}
 
 The transport services API uses the Local Endpoint and Remote Endpoint Objects
-to refer to the endpoints of a transport connection.
-Actions on these Objects can be used to represent various different types of
-endpoint identifiers, such as IP addresses, DNS names, and interface names,
-as well as port numbers and service names.
+to refer to the endpoints of a transport connection. Endpoints can be created
+as either Remote or Local:
+
+~~~
+RemoteSpecifier := NewRemoteEndpoint()
+LocalSpecifier := NewLocalEndpoint()
+~~~
+
+A single Endpoint Object represents the identity of a network host. That endpoint
+can be more or less specific depending on which identifiers are set. For example,
+an Endpoint that only specifies a hostname may in fact end up corresponding
+to several different IP addresses on different hosts.
+
+An Endpoint Object can be configured with the following identifiers:
+
+- Hostname (string)
+
+~~~
+RemoteSpecifier.WithHostname("example.com")
+~~~
+
+- Port (a 16-bit integer) or a Service (string) that maps to a port
+
+~~~
+RemoteSpecifier.WithService(443)
+~~~
+
+~~~
+RemoteSpecifier.WithService("https")
+~~~
+
+- IP address (IPv4 or IPv6 address)
+
+~~~
+RemoteSpecifier.WithIPv4Address(192.0.2.21)
+~~~
+
+~~~
+RemoteSpecifier.WithIPv6Address(2001:db8:4920:e29d:a420:7461:7073:0a)
+~~~
+
+- Interface (string name)
+
+~~~
+LocalSpecifier.WithInterface("en0")
+~~~
+
+- Protocol (implementation-specific enumeration)
+
+~~~
+LocalSpecifier.WithProtocol(TCP)
+~~~
+
+An Endpoint cannot have multiple identifiers of a same type set. That is,
+an endpoint cannot have two IP addresses specified. Two separate IP addresses
+are represented as two Endpoint Objects. If a Preconnection specifies a Remote
+Endpoint with a specific IP address set, it will only establish Connections to
+that IP address. If, on the other hand, the Remote Endpoint specifies a hostname
+but no addresses, the Connection can perform name resolution and attempt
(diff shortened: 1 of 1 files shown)


draft-ietf-taps-interface.md
```diff
@@ -613,10 +613,111 @@ Preconnection during pre-establishment.
...
+- Hostname (string)
+
+~~~
+RemoteSpecifier.WithHostname("example.com")
+~~~
+
+- Port (a 16-bit integer) or a Service (string) that maps to a port
+
+~~~
+RemoteSpecifier.WithService(443)
```
MaxF12, line 640: Shouldn't this be WithPort(443)?

draft-ietf-taps-interface.md
```diff
@@ -613,10 +613,111 @@ Preconnection during pre-establishment.
...
+- Interface (string name)
+
+~~~
+LocalSpecifier.WithInterface("en0")
+~~~
+
+- Protocol (implementation-specific enumeration)
+
+~~~
+LocalSpecifier.WithProtocol(TCP)
```
MaxF12, line 666: What exactly does this do?

draft-ietf-taps-interface.md
```diff
@@ -613,10 +613,111 @@ Preconnection during pre-establishment.
...
+~~~
+
+An Endpoint cannot have multiple identifiers of a same type set. That is,
+an endpoint cannot have two IP addresses specified. Two separate IP addresses
+are represented as two Endpoint Objects. If a Preconnection specifies a Remote
+Endpoint with a specific IP address set, it will only establish Connections to
+that IP address. If, on the other hand, the Remote Endpoint specifies a hostname
+but no addresses, the Connection can perform name resolution and attempt
+using any address derived from the original hostname Remote Endpoint.
```
mwelzl, line 675: I stumbled over "original hostname Remote Endpoint". Is there a missing "for the" before "Remote"?

draft-ietf-taps-interface.md
```diff
@@ -613,10 +613,111 @@ Preconnection during pre-establishment.
...
+
+The Transport Services API resolves names internally, when the Initiate(),
+Listen(), or Rendezvous() method is called to establish a Connection. Privacy
+considerations for the timing of this resolution are given in {{privacy-security}}.
+
+The Resolve() action on Preconnection can be used by the application to force
```
mwelzl, line 681:
```suggestion
The Resolve() action on a Preconnection can be used by the application to force
```

draft-ietf-taps-interface.md
```diff
@@ -613,10 +613,111 @@ Preconnection during pre-establishment.
...
+Remote Endpoint can be used to filter incoming multicast from specific senders. Such
+a Preconnection will only support calling Listen(), not Initiate(). The accepted Connections
+are receive-only.
+
+Similarly, specifying a multicast group address on the Remote Endpoint will indicate that the
+resulting connection will be used to send multicast messages.
+
+### Endpoint Aliases
+
+An Endpoint can have an alternative definition when using different protocols.
```
mwelzl, line 698: I'm just wondering: is this limited to different protocols? Or would aliases also be a way to, e.g., bundle different IP addresses together to one "alias" ?

IF it is strictly about protocols, maybe the wording could be a bit more explicit about this being the only use case. Maybe insert a limiting sentence here... like "Aliases cannot be used to specify ... " ?  These are just vague thoughts, do whatever you want with them. Totally fine to ignore all of this.

mwelzl (approved):
I left a few comments, but nothing very important.

draft-ietf-taps-interface.md
martinduke: Yes this fits uncomfortably with the transport-agnostic principle of TAPS.

draft-ietf-taps-interface.md
tfpauly: Fixed!

draft-ietf-taps-interface.md
tfpauly: Good call, fixed

draft-ietf-taps-interface.md
tfpauly: Whoops, indeed!

draft-ietf-taps-interface.md
tfpauly: Moved this down to the aliasing section to clarify

draft-ietf-taps-interface.md
tfpauly: I'm not sure if we want to limit the alias to this use case only, so let's leave it a bit loose for now, and discuss more?

tfpauly:
@MaxF12 does the update about transport aliasing help?

MaxF12 (approved):

draft-ietf-taps-interface.md
mwelzl: OK!  I like loose. But then, why not say something like "can have an alternative definition with different identifiers." ?   Else, "when using different protocols" really sounds like the only possible case.  Just a thought - if you *want* it half-loose as it is, that's also ok with me.

PR #710, opened 2020-12-10
Update Rendezvous() description

### LINKED ISSUE #676: Preconnection.Resolve() needs work
In §6.3 we have:

>    If the endpoints are suspected to be behind a NAT, Rendezvous() can
>    be initiated using Local and Remote Endpoints that support a method
>    of discovering NAT bindings such as Session Traversal Utilities for
>    NAT (STUN) [RFC8489] or Traversal Using Relays around NAT (TURN)
>    [RFC5766].  In this case, the Local Endpoint will resolve to a
>    mixture of local and server reflexive addresses.  The Resolve()
>    action on the Preconnection can be used to discover these bindings:
>    ```[]Preconnection := Preconnection.Resolve()```
>    The Resolve() call returns a list of Preconnection Objects, that
>    represent the concrete addresses, local and server reflexive, on
>    which a Rendezvous() for the Preconnection will listen for incoming
>    Connections.  These resolved Preconnections will share all other
>    Properties with the Preconnection from which they are derived, though
>    some Properties may be made more-specific by the resolution process.

This doesn't work, because `Preconnection.Rendezvous()` takes only a single `Preconnection`. The `Resolve()` action on a `Preconnection` should rather be on a `LocalEndpoint` or `RemoteEndpoint`.

csperkins:
Closes #676

### DIFF
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -1508,53 +1508,58 @@ Rendezvous() Action:
 Preconnection.Rendezvous()
 ~~~
 
-The Preconnection Object must be specified with both a Local Endpoint and a
-Remote Endpoint, and also the transport properties and security parameters
-needed for Protocol Stack selection.
+A Preconnection Object used in a Rendezvous() MUST have both the
+Local Endpoint candidates and the Remote Endpoint candidates specified,
+along with the transport properties and security parameters needed for
+Protocol Stack selection, before the Rendezvous() Action is initiated.
 
-The Rendezvous() Action causes the Preconnection to listen on the Local
-Endpoint for an incoming Connection from the Remote Endpoint, while also
-simultaneously trying to establish a Connection from the Local Endpoint to
-the Remote Endpoint. 
+The Rendezvous() Action causes the caller to listen on the Local Endpoint
+candidates for an incoming Connection from the Remote Endpoint candidates,
+while also simultaneously trying to establish a Connection from the Local
+Endpoint candidates to the Remote Endpoint candidates.
 
-If there are multiple Local Endpoints or Remote Endpoints configured, then 
-initiating a rendezvous action will systematically probe the reachability
-of those endpoints following an approach such as that used in Interactive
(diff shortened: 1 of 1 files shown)


draft-ietf-taps-interface.md
```diff
@@ -1508,53 +1508,58 @@ Rendezvous() Action:
...
-are then configured on the Preconnection, allowing the Rendezvous() Action to be
-initiated.
-
-The Rendezvous() Action returns a Connection object. Once Rendezvous() has been
-called, any changes to the Preconnection MUST NOT have any effect on the
-Connection. However, the Preconnection can be reused, e.g., for Rendezvous of
-another Connection.
+in the presence of NATs will configure the Preconnection object with at least
+one a Local Endpoint that supports NAT binding discovery. It will then Resolve()
+the Preconnection, and pass the resulting list of Local Endpoints candidates to
```
mwelzl, line 1546:
```suggestion
the Preconnection, and pass the resulting list of Local Endpoint candidates to
```

draft-ietf-taps-interface.md
```diff
@@ -1566,35 +1571,21 @@ transport-layer connection is established; for Connectionless transports,
 it occurs when the first Message is received from the Remote Endpoint. The
 resulting Connection is contained within the RendezvousDone<> Event, and is
 ready to use as soon as it is passed to the application via the Event.
+Changes made to a Preconnection made after Rendezvous() has been called do
```
mwelzl, line 1574:
```suggestion
Changes made to a Preconnection after Rendezvous() has been called do
```

mwelzl (commented):
LGTM, I made two tiny text suggestions; and thanks!

draft-ietf-taps-interface.md
csperkins: Fixed.

draft-ietf-taps-interface.md
csperkins: Fixed

mwelzl (approved):

tfpauly (approved):

PR #723, opened 2021-01-12
Allow disuse of identifiers in a connection group

### LINKED ISSUE #706: Flushing protocol state is not in the interface
Section 9.1 of taps-impl loosely refers to the need for an interface to "flush protocol state" to preserve anonymity, but such a capability is not in taps-interface.

martinduke:
There are many ways to do this, but this one mimics the "private browsing" paradigm" as I understand it.

Resolves #706.


mwelzl (approved):
A very nice approach IMO, thanks for doing this

philsbln (approved):
I like this approach for the simplicity, but it still solves only half of the problem we have with connection groups vs. connection contexts. I am fine with merging this, but IMHO it does not fully resolve #706.

draft-ietf-taps-impl.md
```diff
@@ -820,8 +820,10 @@ Cached protocol state is primarily used during Connection establishment for a si
 implementation's preference between several candidate Protocol Stacks. For example, if two IP address Endpoints are otherwise
 equally preferred, an implementation may choose to attempt a connection to an address for which it has a TCP Fast Open cookie.
 
-Applications must have a way to flush protocol cache state if desired. This may be necessary, for example, if
-application-layer identifiers rotate and clients wish to avoid linkability via trackable TLS tickets or TFO cookies.
+Applications must have a way to not use protocol cache state for certain
+connections, if desired. This may be necessary, for example, if application-
```
gorryfair, line 824: I like the intent, however
(1) This wasn't clear: /if desired/  ... if who desires? The application, the TAPS system, the implementor. 
(2) I'm not a huge fan of /if desired/... /must/ even when it is lower case.
Could this be: Applications can request that the TAPS system ought not to use the cached protocol state for certain connections?

abrunstrom: One question on the phrase "not use cached protocol state". I read that as not using the cached state that is there when initiating a connection, but the taps system should also not cache state about this connection for use in later connections.  Not sure if that is captured by the wording or should be made explicit? It is clear in the update to the API section.
Otherwise looks good to me as well!

martinduke:
> 
> 
> I like this approach for the simplicity, but it still solves only half of the problem we have with connection groups vs. connection contexts. I am fine with merging this, but IMHO it does not fully resolve #706.

Can you say more? I read #706 as specifically about flushing state, not Connection Group vs. Connection issues generally. This is not literally "flushing" the state but I've update taps-interface to reflect this design while, I think, preserving the spirit of the original text.

abrunstrom (approved):
Thanks @martinduke, looks good to me now.

draft-ietf-taps-impl.md
```diff
@@ -820,8 +820,12 @@ Cached protocol state is primarily used during Connection establishment for a si
 implementation's preference between several candidate Protocol Stacks. For example, if two IP address Endpoints are otherwise
 equally preferred, an implementation may choose to attempt a connection to an address for which it has a TCP Fast Open cookie.
 
-Applications must have a way to flush protocol cache state if desired. This may be necessary, for example, if
-application-layer identifiers rotate and clients wish to avoid linkability via trackable TLS tickets or TFO cookies.
+Applications can request that a Connection Group have a separate cache for
+protocol state. Connections in the connection group will not use cached state
```
tfpauly, line 824:
```suggestion
protocol state. Connections in the group will not use cached state
```

draft-ietf-taps-impl.md
```diff
@@ -820,8 +820,12 @@ Cached protocol state is primarily used during Connection establishment for a si
 implementation's preference between several candidate Protocol Stacks. For example, if two IP address Endpoints are otherwise
 equally preferred, an implementation may choose to attempt a connection to an address for which it has a TCP Fast Open cookie.
 
-Applications must have a way to flush protocol cache state if desired. This may be necessary, for example, if
-application-layer identifiers rotate and clients wish to avoid linkability via trackable TLS tickets or TFO cookies.
+Applications can request that a Connection Group have a separate cache for
```
tfpauly, line 823:
```suggestion
Applications can request that a Connection Group maintain a separate cache for
```

draft-ietf-taps-interface.md
```diff
@@ -1987,6 +1987,22 @@ this limits the number of ConnectionReceived Events that will occur, but constra
 to the group of the Connection associated with this property. For a multi-streaming transport,
 this limits the number of allowed streams.
 
+### Private Session (#private-session}
```
tfpauly, line 1990:
```suggestion
### Private Session (#private-session}

```

draft-ietf-taps-interface.md
```diff
@@ -1987,6 +1987,22 @@ this limits the number of ConnectionReceived Events that will occur, but constra
...
+
+Type:
+: Boolean
+
+Default:
+: false
+
+When set to true, this property will initiate new connections without using
+any cached information (such as session tickets or cookies) from previous
+connections that are not entangled with it. The residue of this connection will
```
tfpauly, line 2002:
```suggestion
connections that are not entangled with it. Any state generated by this connection will
```

draft-ietf-taps-interface.md
```diff
@@ -1987,6 +1987,22 @@ this limits the number of ConnectionReceived Events that will occur, but constra
...
+Name:
+: privateSession
+
+Type:
+: Boolean
+
+Default:
+: false
+
+When set to true, this property will initiate new connections without using
```
tfpauly, line 2000: I think "connections" should be capitalized throughout?

draft-ietf-taps-interface.md
```diff
@@ -1987,6 +1987,22 @@ this limits the number of ConnectionReceived Events that will occur, but constra
 to the group of the Connection associated with this property. For a multi-streaming transport,
 this limits the number of allowed streams.
 
+### Private Session {#private-session}
```
tfpauly, line 1990: Interim discussion: rename to "isolate" instead of "private".

draft-ietf-taps-interface.md
```diff
@@ -1987,6 +1987,22 @@ this limits the number of ConnectionReceived Events that will occur, but constra
...
+: Boolean
+
+Default:
+: false
+
+When set to true, this property will initiate new Connections without using
+any cached information (such as session tickets or cookies) from previous
+connections that are not entangled with it. Any state generated by this
+Connection will only be shared with entangled connections. Cloned Connections
+will use saved state from within the Connection Group.
```
tfpauly, line 2004: Note that this not guarantee no leakage of information, since things like RTT estimates, etc, may be fully isolated.

theri, line 2004: For me, this sentence implies that the state generated by this new Connection will **not** be used by other Connections that are not in the same Connection Group. But maybe it's worth making this explicit.

tfpauly (approved):

PR #724, opened 2021-01-12
Half-closed connections and such. Resolves #707

### LINKED ISSUE #707: Issues with TCP-specific considerations (taps-impl 10.1)
- I don't think you mean that TAPS should throw a ConnectionError on a TCP Retransmission Timeout!

- I'm somewhat concerned about half-close, though this might just be editorial. An HTTP/3 implementation needs to send FIN when it's done requesting, for instance (I realize that HTTP/3 is not TCP, but I imagine the semantics are the same). So I think it would send its request with a end-of-message, According to the "Send" description, TAPS will send CLOSE.TCP automatically after the FIN.

A regular "Close" will trigger a "Closed" event which tells the application it isn't getting more data, so I _guess_ this just barely works if TAPS sends CLOSE.TCP but not Closed, and waits for the Close command from the app? Like I said, I'm not sure this is wrong, but it's not clearly stated how an app that needs half-close functionality should implement that.

mwelzl:
Closing: be more explicit about half-closed connections, allow it in the API too, fix errneous ConnectionError text.
Resolves #707.


mwelzl:
Same build error. I don't know what to do about this missing circleci file thing  :(

mirjak (approved):

draft-ietf-taps-impl.md
```diff
@@ -897,13 +897,13 @@ Clone:
...
+: With RECEIVE.TCP, TCP delivers a stream of bytes without any Message delineation. All data delivered in the `Received` or `ReceivedPartial` event will be part of a single stream-wide Message that is marked Final (unless a Message Framer is used). EndOfMessage will be delivered when the TCP Connection has received a FIN (CLOSE-EVENT.TCP or ABORT-EVENT.TCP) from the peer. Note that reception of a FIN should not cause the `Closed` event to be delivered to the application, as it will still be possible for the application to send data.
 
 Close:
-: Calling `Close` on a TCP Connection indicates that the Connection should be gracefully closed (CLOSE.TCP) by sending a FIN to the peer and waiting for a FIN-ACK before delivering the `Closed` event.
+: Calling `Close` on a TCP Connection indicates that the Connection should be gracefully closed (CLOSE.TCP) by sending a FIN to the peer. It will then still be possible to receive data until the peer closes or aborts the TCP connection. The `Closed` event will be issued upon reception of a FIN-ACK.
```
martinduke, line 906:
```suggestion
: Calling `Close` on a TCP Connection indicates that the Connection should be gracefully closed (CLOSE.TCP) by sending a FIN to the peer. It will then still be possible to receive data until the peer closes or aborts the TCP connection. The `Closed` event will be issued upon reception of a FIN.
```

martinduke (approved):
LGTM, except for one thing.

mwelzl:
@martinduke good catch, thanks, incorporated

draft-ietf-taps-impl.md
```diff
@@ -1091,7 +1091,7 @@ Receive:
...
+If this is the only Connection object that is assigned to the SCTP association, CLOSE.SCTP is called, and the `Closed` event is immediately delivered to the application (because SCTP does not support half-closed connections). Else, the Connection object is one out of several Connection objects that are assigned to the same SCTP assocation, and RESET_STREAM.SCTP must be called, which informs the peer that the stream will no longer be used for mapping and can be used by future `Initiate`, `InitiateWithSend` or `Listen` calls. At the peer, the event RESET_STREAM-EVENT.SCTP will fire, which the peer must answer by issuing RESET_STREAM.SCTP too. The resulting local RESET_STREAM-EVENT.SCTP informs the transport system that the stream number can now be re-used by the next `Initiate`, `InitiateWithSend` or `Listen` calls.
```
abrunstrom, line 1095: "the `Closed` event is immediately delivered to the application" may sound a bit misleading as data sent from either end but not yet delivered is reliably delivered before this event fires I think. So you can still receive data after issuing the Close if it was already in the peers send queue, right?

Do we not need to also specify when the 'Closed' event should fire in the multistreamed case. After getting the RESET_STREAM-EVENT.SCTP back I guess?

draft-ietf-taps-impl.md
```diff
@@ -897,13 +897,13 @@ Clone:
...
 
 Receive:
-: With RECEIVE.TCP, TCP delivers a stream of bytes without any Message delineation. All data delivered in the `Received` or `ReceivedPartial` event will be part of a single stream-wide Message that is marked Final (unless a Message Framer is used). EndOfMessage will be delivered when the TCP Connection has received a FIN (CLOSE-EVENT.TCP or ABORT-EVENT.TCP) from the peer.
+: With RECEIVE.TCP, TCP delivers a stream of bytes without any Message delineation. All data delivered in the `Received` or `ReceivedPartial` event will be part of a single stream-wide Message that is marked Final (unless a Message Framer is used). EndOfMessage will be delivered when the TCP Connection has received a FIN (CLOSE-EVENT.TCP or ABORT-EVENT.TCP) from the peer. Note that reception of a FIN should not cause the `Closed` event to be delivered to the application, as it will still be possible for the application to send data.
```
abrunstrom, line 904: This is in the old text, but why is ABORT-EVENT.TCP listed for receiving a FIN?

abrunstrom (changes requested):
Thanks @mwelzl , this mainly look good. Few details that need fixing.

tfpauly (approved):

draft-ietf-taps-impl.md
mwelzl: Good catch! Sorry, I don't know how that happened, it's obviously wrong. I'll fix it!

draft-ietf-taps-impl.md
mwelzl: 1) Right! in the peers send queue, or in flight, or in the local host's receive buffer, I believe. Would it be ok to fix this as follows:
If this is the only Connection object that is assigned to the SCTP association, CLOSE.SCTP `Closed` event is delivered to the application when this call returns.
[no need for us to go into these specifics here? Close semantics are given in the interface document anyway.]

2) Right! Would it be ok to fix this as follows:
"The resulting local RESET_STREAM-EVENT.SCTP informs the transport system that the stream number can now be re-used by the next `Initiate`, `InitiateWithSend` or `Listen` calls, and invokes a `Closed` event towards the application.
[I suppose that "after delivering any data from the local receive buffer" could be added, but I also don't think that this adds clarity, and so I would just formulate it as above).

abrunstrom: For 1. perhaps "... when the call is completed". I guess the call will return immediately, but the close event is delivered to the application when the shutdown complete event is received.

2. looks good!

mwelzl: About 1) also true, thanks - silly me, it should be upon "CLOSE-EVENT.SCTP" from RFC 8303. I once wrote that, after all  :-)

abrunstrom (approved):
Looks good.

PR #725, opened 2021-01-12
Fix "confirmed route". Resolves #703

### LINKED ISSUE #703: What is a "confirmed" route?
In Sec 4.5 of taps-impl:

"An implementation should consider such a protocol stack to be established as soon as a local route to the peer endpoint is confirmed."

What does this mean for an unconnected protocol? That there is an entry in the route table, or that it's received a packet?

mwelzl:
My first try to make a PR from the web interface. Maybe this compiles without problems?
Resolves #703.

### DIFF
--- a/draft-ietf-taps-impl.md
+++ b/draft-ietf-taps-impl.md
@@ -489,7 +489,7 @@ Instead, calling the ConnectionReceived event may be delayed until the first Mes
 
 ## Handling "unconnected" protocols {#unconnected-racing}
 
-While protocols that use an explicit handshake to validate a Connection to a peer can be used for racing multiple establishment attempts in parallel, "unconnected" protocols such as raw UDP do not offer a way to validate the presence of a peer or the usability of a Connection without application feedback. An implementation should consider such a protocol stack to be established as soon as connectivity with the peer endpoint is established.
+While protocols that use an explicit handshake to validate a Connection to a peer can be used for racing multiple establishment attempts in parallel, "unconnected" protocols such as raw UDP do not offer a way to validate the presence of a peer or the usability of a Connection without application feedback. An implementation should consider such a protocol stack to be established as soon as the Transport Services system has selected a path on which to send data.
 
 However, if a peer is not reachable over the network using the unconnected protocol, or data cannot be exchanged for any other reason, the application may want to attempt using another candidate Protocol Stack. The implementation should maintain the list of other candidate Protocol Stacks that were eligible to use.
 
--- a/draft-ietf-taps-impl.md
+++ b/draft-ietf-taps-impl.md
@@ -487,7 +487,7 @@ handed over, it cannot be guaranteed that the other endpoint will have any way t
 a passive endpoint's ConnectionReceived event may not be called upon an active endpoint's Inititate.
 Instead, calling the ConnectionReceived event may be delayed until the first Message arrives.
 
-## Handling racing with "unconnected" protocols {#unconnected-racing}
+## Handling "unconnected" protocols {#unconnected-racing}
 
 While protocols that use an explicit handshake to validate a Connection to a peer can be used for racing multiple establishment attempts in parallel, "unconnected" protocols such as raw UDP do not offer a way to validate the presence of a peer or the usability of a Connection without application feedback. An implementation should consider such a protocol stack to be established as soon as connectivity with the peer endpoint is established.
 
--- a/draft-ietf-taps-impl.md
+++ b/draft-ietf-taps-impl.md
@@ -489,7 +489,7 @@ Instead, calling the ConnectionReceived event may be delayed until the first Mes
 
 ## Handling racing with "unconnected" protocols {#unconnected-racing}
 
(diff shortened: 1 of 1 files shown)


mwelzl:
Aha, the error appears even when doing it over the web interface.

martinduke (approved):

draft-ietf-taps-impl.md
```diff
@@ -489,7 +489,7 @@ Instead, calling the ConnectionReceived event may be delayed until the first Mes
...
+While protocols that use an explicit handshake to validate a Connection to a peer can be used for racing multiple establishment attempts in parallel, "unconnected" protocols such as raw UDP do not offer a way to validate the presence of a peer or the usability of a Connection without application feedback. An implementation should consider such a protocol stack to be established as soon as connectivity with the peer endpoint is established.
 
 However, if a peer is not reachable over the network using the unconnected protocol, or data cannot be exchanged for any other reason, the application may want to attempt using another candidate Protocol Stack. The implementation should maintain the list of other candidate Protocol Stacks that were eligible to use.
```
abrunstrom, line 494: I do not think this update helped. In fact, I think it made it worse. What does "connectivity with the peer endpoint is established" mean for an unconnected protocol?

abrunstrom (commented):
This update made things less clear to me.

draft-ietf-taps-impl.md
tfpauly: Agreed. How about "as soon as a route on which to send and receive packets is present, and any necessary state like local port reservations has been established."

tfpauly (changes requested):

gorryfair:
I think I am missing something ... how do endpoints know a route exists (is present)? My thinking from NEAT was that endpoints only knew this was established when there was some "confirmation" from the remote endpoint. Do you mean something else?

abrunstrom:
@gorryfair this is for UDP, nothing is sent to the other end until the application sends data. it was the same in neat, our open call in neat just returned immediately for UDP.

gorryfair:
True - for UDP, as you say, you basically know the Interface is up (and often not much more), but the words "established" sounds more like what happened when you get the first packet back... Maybe I could live with this, but it doesn't immediately inspire the right sense...

abrunstrom:
Yes, "connectivity is established" is not good at all, this is why I said the proposed update made it worse :) 
Or you mean that "local port reservations has been established" also triggers the wrong idea because of "established"?
How about "... any necessary state like local port reservations is available/has been allocated."

draft-ietf-taps-impl.md
```diff
@@ -489,7 +489,7 @@ Instead, calling the ConnectionReceived event may be delayed until the first Mes
 
 ## Handling racing with "unconnected" protocols {#unconnected-racing}
 
-While protocols that use an explicit handshake to validate a Connection to a peer can be used for racing multiple establishment attempts in parallel, "unconnected" protocols such as raw UDP do not offer a way to validate the presence of a peer or the usability of a Connection without application feedback. An implementation should consider such a protocol stack to be established as soon as a local route to the peer endpoint is confirmed.
+While protocols that use an explicit handshake to validate a Connection to a peer can be used for racing multiple establishment attempts in parallel, "unconnected" protocols such as raw UDP do not offer a way to validate the presence of a peer or the usability of a Connection without application feedback. An implementation should consider such a protocol stack to be established as soon as connectivity with the peer endpoint is established.
```
tfpauly, line 492:
```suggestion
While protocols that use an explicit handshake to validate a Connection to a peer can be used for racing multiple establishment attempts in parallel, "unconnected" protocols such as raw UDP do not offer a way to validate the presence of a peer or the usability of a Connection without application feedback. An implementation should consider such a protocol stack to be established as soon as the Transport Services system has selected a valid path.
```

draft-ietf-taps-impl.md
```diff
@@ -489,7 +489,7 @@ Instead, calling the ConnectionReceived event may be delayed until the first Mes
 
 ## Handling racing with "unconnected" protocols {#unconnected-racing}
```
tfpauly, line 490:
```suggestion
## Handling "unconnected" protocols {#unconnected-racing}
```

draft-ietf-taps-impl.md
tfpauly, line 494: We can recommend that applications other than a "fire-and-forget" case make their connections stateful, such as with a Message Framer on top of UDP that can verify when the peer is reachable.

draft-ietf-taps-impl.md
martinduke: Tommy's text is an improvement, but I wonder if we should be more brutally obvious. Perhaps "the Transport Services system has selected a destination IP address to which it has a route."

That text is unambiguous, with no wiggle room about what is "valid", etc.

mwelzl: I understand the problem now. I like Tommy's text, and I like your suggestion even better. Let's see how many thumbs up this gets...  if we get the critical mass, maybe we can just apply this fix and be done with this.

gorryfair: Close: I don't that much like "route" - I'd prefer to not mention the routing system. So I suggest:
"the Transport Services system has selected a  a valid path and the corresponding destination IP address it will use."

abrunstrom: I think @martinduke was trying to get rid of "valid path", I agree that term is not very clear.

gorryfair: ah - but valid "route" is much worse. "established" connection is yucky. Maybe "A path on which it is valid to send" ...not implying the path is valid end-to-end?

PR #730, opened 2021-01-29
Describe reasons to not tear connections down during path changes

### LINKED ISSUE #688: Provide examples for path changes
Provide examples of why we don't tear down path changes

tfpauly:
Closes #688


mwelzl (approved):

draft-ietf-taps-impl.md
```diff
@@ -753,6 +753,8 @@ When a path change occurs, e.g., when the IP address of an interface changes or
 
 For protocols that do not support multipath or migration, the Protocol Instances should be informed of the path change, but should not be forcibly disconnected if the previously used path becomes unavailable.
 
+There are many common user scenarios that can lead to a path becoming temporarily unavailable, and then recovering before the transport protocol hits a timeout error. These are particularly common with mobile devices. Examples include: an Ethernet cable becoming unplugged and then plugged back in; a device losing Wi-Fi signal while a user in an elevator, and reattaching when the user leaves the elevator; and a user losing radio signal while riding a train through a tunnel. If the device is able to rejoin a network with the same IP address, a stateful transport connection can generally resume. Thus, while it is useful for an application to be aware of a temporary loss of connectivity, the Transport Services implementation should not aggressively close connections in these scenarios.
+
```
abrunstrom, line 757: user in --> user is in

"Thus, while it is useful for an application" - I think the context is informing Protocol Instances, so this should be "Protocol Instance" rather than "application" I think?

abrunstrom (commented):
Two nits, otherwise looks good, thanks!

draft-ietf-taps-impl.md
abrunstrom: Seems the first nit was fixed while i was typing :)

tfpauly: That was @gorryfair =)

draft-ietf-taps-impl.md
```diff
@@ -753,6 +753,8 @@ When a path change occurs, e.g., when the IP address of an interface changes or
 
 For protocols that do not support multipath or migration, the Protocol Instances should be informed of the path change, but should not be forcibly disconnected if the previously used path becomes unavailable.
 
+There are many common user scenarios that can lead to a path becoming temporarily unavailable, and then recovering before the transport protocol reaches a timeout error. These are particularly common using mobile devices. Examples include: an Ethernet cable becoming unplugged and then plugged back in; a device losing a Wi-Fi signal while a user is in an elevator, and reattaching when the user leaves the elevator; and a user losing the radio signal while riding a train through a tunnel. If the device is able to rejoin a network with the same IP address, a stateful transport connection can generally resume. Thus, while it is useful for an application to be aware of a temporary loss of connectivity, the Transport Services implementation should not aggressively close connections in these scenarios.
```
tfpauly, line 756:
```suggestion
There are many common user scenarios that can lead to a path becoming temporarily unavailable, and then recovering before the transport protocol reaches a timeout error. These are particularly common using mobile devices. Examples include: an Ethernet cable becoming unplugged and then plugged back in; a device losing a Wi-Fi signal while a user is in an elevator, and reattaching when the user leaves the elevator; and a user losing the radio signal while riding a train through a tunnel. If the device is able to rejoin a network with the same IP address, a stateful transport connection can generally resume. Thus, while it is useful for a Protocol Instance to be aware of a temporary loss of connectivity, the Transport Services implementation should not aggressively close connections in these scenarios.
```

PR #736, opened 2021-02-26
Change Unconnected to Connectionless

### LINKED ISSUE #716: Unconnected -> Connectionless
UDP is best described as connectionless, not unconnected: it has no notion of a connection, where "unconnected" might imply a particular usage pattern for a protocol that also supports connections.

britram:
fixes #716


tfpauly (approved):

gorryfair:
Just check "connected" - in case we mean connection-oriented; and that usage of "connect" is sane, we need care.

PR #739, opened 2021-03-19
Add CloseGroup

### LINKED ISSUE #732: Closing multistreaming transports
Inspired by the discussion in #722, I don't think we adequately addressed #705.

Abstracting away from current designs, in a connection group there are four possible termination actions:
- Clean close a connection
- Abort a connection
- Clean close a connection group
- Abort a connection group

TAPS has all but the third.

QUIC has three of these; SCTP has two (three with RFC 6525); HTTP/2, IIUC, in principle has all four, though I believe RFC 7540 doesn't envision HTTP/2 ever initiating TCP RST. There is an added complication that in some protocols these are half-closes or half-aborts.

I think we're trying to be too cute by not offering all four of these. Let's just have all of them (which means adding CloseGroup) and then figure out how the protocol mappings respond to actions they don't actually support. For example
1. Non-streaming protocols simply apply the close or abort to each of their connections.
2. When QUIC gets CloseGroup, it sends clean close on all connections and then sends group abort. (Note that apps with mandatory streams [like HTTP3] can't use this functionality)
3. When SCTP with 7525 gets ConnectionAbort, it does ConnectionClose instead
4. When SCTP without 7525 gets ConnectionClose or ConnectionAbort, it's a NOP.
5. When HTTP/2 gets AbortGroup, it sends GOAWAY followed by a CloseGroup.

tfpauly:
Closes #732

### DIFF
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -2928,17 +2928,25 @@ signaled.
 Connection -> Closed<>
 ~~~
 
-Abort terminates a Connection without delivering any remaining data. This action does
+Abort terminates a Connection without delivering any remaining Messages. This action does
 not affect any other Connection that is entangled with this one in a Connection Group.
 
 ~~~
 Connection.Abort()
 ~~~
 
-AbortGroup terminates a Connection and also terminates any other Connections that are
+CloseGroup gracefully terminates a Connection and any other Connections that are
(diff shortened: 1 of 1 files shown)


martinduke (approved):

mwelzl (approved):
Looks good, thanks for doing this!

PR #741, opened 2021-03-25
Reference to arch section 4.2.2 for "racing". Closes #714.

### LINKED ISSUE #714: "Racing"
Is "racing" really the right term? It might be a term of art that applies even in cases like this, but many of the relationships between different branches in the tree are based on preference rather than lowest latency. I don't feel strongly about this, but I don't know if this has been brought up yet, so I figured I'd mention it.

mwelzl:
Closes #714.

Oh yes, and it IS editorial !  :-D


abrunstrom (approved):
Thanks for fixing this!

PR #742, opened 2021-03-25
Bring the multipath text in line with the API

### LINKED ISSUE #704: Define multipath terms
Before distinguishing between protocols that do "multipath" and "migrating between paths", it might help to define these terms. This is tripping is up in QUIC, and IIRC the original "Multipath TCP" was pretty much the migration case.

Something like, "In this section we distinguish between 'multipath' protocols and protocols that allow migration between paths. The first means <blah blah>..."

mwelzl:
Closes #704.

Note, the change to the API draft here is a purely editorial detail that I fixed here in passing: the multipath selection property was renamed and this fixes the text referring to it.


martinduke (approved):

draft-ietf-taps-impl.md
```diff
@@ -757,15 +757,15 @@ This enables implementations to realize transparent connection coalescing, conne
...
 
-In case of Pooled Connections {{pooled-connections}}, the transport system may add connections over new paths or different protocols to the pool if permissible based on the multipath policy and Selection Properties. In case a previously used path becomes unavailable, the transport system may disconnect all connections that require this path, but should not disconnect the pooled connection object exposed to the application. The strategy to do so is implementation-specific, but should be consistent with the behavior of multipath transports. 
+In case of Pooled Connections {{pooled-connections}}, the Transport Services implementation may add connections over new paths or different protocols to the pool if permissible based on the multipath policy and Selection Properties. In case a previously used path becomes unavailable, the transport system may disconnect all connections that require this path, but should not disconnect the pooled connection object exposed to the application. The strategy to do so is implementation-specific, but should be consistent with the behavior of multipath transports.
```
abrunstrom, line 768: "may add connections over new paths or different protocols", I think the "or different protocols" part has nothing to do with multipath and should be removed.

draft-ietf-taps-impl.md
```diff
@@ -757,15 +757,15 @@ This enables implementations to realize transparent connection coalescing, conne
 
 ## Handling Path Changes
 
-When a path change occurs, e.g., when the IP address of an interface changes or a new interface becomes available, the Transport Services implementation is responsible for notifying the application of the change. The path change may interrupt connectivity on a path for an active connection or provide an opportunity for a transport that supports multipath or migration to adapt to the new paths.
+When a path change occurs, e.g., when the IP address of an interface changes or a new interface becomes available, the Transport Services implementation is responsible for notifying the application of the change. The path change may interrupt connectivity on a path for an active connection or provide an opportunity for a transport that supports multipath or migration to adapt to the new paths. Note that, from the Transport Services API point of view, migration is considered a part of multipath connectivity; it is just a limiting policy on multipath usage. If the `multipath` Selection Property is set to `Disabled`, migration is disallowed.
```
abrunstrom, line 760: notifying the application of the change. --> notifying the Protocol Instance of the change.

abrunstrom (changes requested):
Not sure this is what we discussed, but I like it better so I am happy with the update, thanks! Noticed two small things to update on the "old" text.

mwelzl:
"what we discussed" - you mean in #704, where I say that I stumbled over a sentence? This PR does update that sentence (simply removing "with migration support" from it because any multipath protocol can do that).

tfpauly (approved):
Looks good to me!

abrunstrom:
> "what we discussed" - you mean in #704, where I say that I stumbled over a sentence? 

no, i meant what we discussed at the interim about using the handover, interactive, aggregate terminology. but i saw that this was not so easily done and we are all happy with your solution, so all good :)

abrunstrom (approved):
Thanks for fixing the nits!

PR #745, opened 2021-04-06
Editorial rollup

britram:


draft-ietf-taps-impl.md
```diff
@@ -739,7 +739,7 @@ Connection Properties. A Connection can also generate events in the form of Soft
 
 The set of Connection Properties that are supported for setting and getting on a Connection are described in {{I-D.ietf-taps-interface}}. For
 any properties that are generic, and thus could apply to all protocols being used by a Connection, the Transport System should store the properties
-in a generic storage, and notify all protocol instances in the Protocol Stack whenever the properties have been modified by the application.
+in an object common to all protocols, and notify all protocol instances in the Protocol Stack whenever the properties have been modified by the application.
```
abrunstrom, line 742: "an object" sounds odd to me here. I think it needs to store it in "the Connection Object" if we want to express it at this abstraction level.

draft-ietf-taps-impl.md
```diff
@@ -519,7 +519,7 @@ The most basic mapping for sending a Message is an abstraction of datagrams, in
 
 For protocols that expose byte-streams, the only delineation provided by the protocol is the end of the stream in a given direction. Each Message in this case corresponds to the entire stream of bytes in a direction. These Messages may be quite long, in which case they can be sent in multiple parts.
 
-Protocols that provide the framing (such as length-value protocols, or protocols that use delimiters) provide data boundaries that may be longer than a traditional packet datagram. Each Message for framing protocols corresponds to a single frame, which may be sent either as a complete Message, or in multiple parts.
+Protocols that provide the framing (such as length-value protocols, or protocols that use delimiters) provide data boundaries that may be longer than a traditional packet. Each Message for framing protocols corresponds to a single frame, which may be sent either as a complete Message, or in multiple parts.
```
abrunstrom, line 522: I think keeping "datagram" is more aligned with the surrounding text here.

abrunstrom (commented):
Thanks! Two comments on the updates.

draft-ietf-taps-impl.md
britram:
```suggestion
Protocols that provide the framing (such as length-value protocols, or protocols that use delimiters) provide data boundaries that may be longer than the typical datagram. Each Message for framing protocols corresponds to a single frame, which may be sent either as a complete Message, or in multiple parts.
```

draft-ietf-taps-impl.md
britram:
```suggestion
in storage common to all protocols, and notify all protocol instances in the Protocol Stack whenever the properties have been modified by the application.
```

draft-ietf-taps-impl.md
```diff
@@ -519,7 +519,7 @@ The most basic mapping for sending a Message is an abstraction of datagrams, in
 
 For protocols that expose byte-streams, the only delineation provided by the protocol is the end of the stream in a given direction. Each Message in this case corresponds to the entire stream of bytes in a direction. These Messages may be quite long, in which case they can be sent in multiple parts.
 
-Protocols that provide the framing (such as length-value protocols, or protocols that use delimiters) provide data boundaries that may be longer than a traditional packet datagram. Each Message for framing protocols corresponds to a single frame, which may be sent either as a complete Message, or in multiple parts.
+Protocols that provide the framing (such as length-value protocols, or protocols that use delimiters) provide data boundaries that may be longer than the typical datagram. Each Message for framing protocols corresponds to a single frame, which may be sent either as a complete Message, or in multiple parts.
```
squarooticus, line 522: "...that may not fit within a single datagram."

PR #748, opened 2021-04-06
Clarify "consumed" and "failed to send" timing.

### LINKED ISSUE #747: "meaning of the Message" in 5.1.2 Send Completion is unclear
I stared at this sentence for a while before I thought I'd made sense of it. PR forthcoming but I think this might need a little discussion.

britram:
Fixes #747

### DIFF
--- a/draft-ietf-taps-impl.md
+++ b/draft-ietf-taps-impl.md
@@ -553,7 +553,7 @@ transport protocol session or DCCP connection).
 
 ### Send Completion
 
-The application should be notified whenever a Message or partial Message has been consumed by the Protocol Stack, or has failed to send. The meaning of the Message being consumed by the stack may vary depending on the protocol. For a basic datagram protocol like UDP, this may correspond to the time when the packet is sent into the interface driver. For a protocol that buffers data in queues, like TCP, this may correspond to when the data has entered the send buffer.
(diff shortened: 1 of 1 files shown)


mwelzl (approved):
lgtm

draft-ietf-taps-impl.md
```diff
@@ -553,7 +553,7 @@ transport protocol session or DCCP connection).
...
+The application should be notified whenever a Message or partial Message has been consumed by the Protocol Stack, or has finally failed to send. The time at which a Message is considered to have been consumed by the Protocol Stack may vary depending on the protocol. For example, for a basic datagram protocol like UDP, this may correspond to the time when the packet is sent into the interface driver. For a protocol that buffers data in queues, like TCP, this may correspond to when the data has entered the send buffer. The time at which a message has finally failed to send is after the Protocol Stack or the Transport Services implementation itself has not successfully sent the entire Message content or partial Message content on any open candidate connection; this may depend on protocol-specific timeouts.
```
abrunstrom, line 556: Not sure about the "finally failed to send" rather than just "failed to send". The send may also fail if the set of Message Properties are not consistent with the Connection's transport properties, so it may also fail immediately. Possibly we should add this a second example?

britram:
```suggestion
The application should be notified whenever a Message or partial Message has been consumed by the Protocol Stack, or has failed to send. The time at which a Message is considered to have been consumed by the Protocol Stack may vary depending on the protocol. For example, for a basic datagram protocol like UDP, this may correspond to the time when the packet is sent into the interface driver. For a protocol that buffers data in queues, like TCP, this may correspond to when the data has entered the send buffer. The time at which a message has failed to send is after the Protocol Stack or the Transport Services implementation itself has not successfully sent the entire Message content or partial Message content on any open candidate connection; this may depend on protocol-specific timeouts.
```

tfpauly (approved):

PR #766, opened 2021-04-09
Update draft-ietf-taps-impl.md

zaheduzzaman:
explained the "deep-copy" term.


draft-ietf-taps-impl.md
```diff
@@ -146,7 +146,7 @@ The connection objects that are exposed to applications for Transport Services a
 - the Connection, the basic object that represents a flow of data as Messages in either direction between the Local and Remote Endpoints;
 - and the Listener, a passive waiting object that delivers new Connections.
 
-Preconnection objects should be implemented as bundles of properties that an application can both read and write. Once a Preconnection has been used to create an outbound Connection or a Listener, the implementation should ensure that the copy of the properties held by the Connection or Listener is immutable. This may involve performing a deep-copy if the application is still able to modify properties on the original Preconnection object.
+Preconnection objects should be implemented as bundles of properties that an application can both read and write. Once a Preconnection has been used to create an outbound Connection or a Listener, the implementation should ensure that the copy of the properties held by the Connection or Listener is immutable. This may involve performing a deep-copy, copying the object with all the objects it refers, if the application is still able to modify properties on the original Preconnection object.
```
mwelzl, line 149:
```suggestion
Preconnection objects should be implemented as bundles of properties that an application can both read and write. Once a Preconnection has been used to create an outbound Connection or a Listener, the implementation should ensure that the copy of the properties held by the Connection or Listener is immutable. This may involve performing a deep-copy, copying the object with all the objects it references, if the application is still able to modify properties on the original Preconnection object.
```

(I'm not a native English speaker, but I *believe* that "references" or "refers _to_" are correct, not just "refers)

tfpauly (approved):

PR #767, opened 2021-04-09
Editorial rollup, bis.

### LINKED ISSUE #746: Editorial issues on 5.1.1. Message Properties
I believe these are all editorial.

First, this list needs some front matter: "protocols should implement each of the following Message Properties on send as follows" or similar.

- "Ordered" should state that protocols that do not support configurable ordering may ignore the property.
- "Final" should note that it is not mandatory to close the connection after a Final message.
- "Corruption Protection Length" should note that it can be ignored basically everywhere but UDP-Lite. 

One open question: is it an error to send a Message after a Final message, and does the protocol implementation need to handle this itself?

### LINKED ISSUE #749: What if there's no context switch penalty for sending messages? (Sec 5.1.2)
Section 5.1.2: "the implementation should hold off on processing Messages until the batch is complete" assumes that there is indeed a context-switch penalty. Batch semantics are only for hints to the implementation/protocol stack, and this advice only applies when that hint allows the stack to reduce context switch penalty. Correct?

(If so, this should be noted explicitly)

britram:
fixes #749. fixes #746.

### DIFF
--- a/draft-ietf-taps-impl.md
+++ b/draft-ietf-taps-impl.md
@@ -557,7 +557,7 @@ The application should be notified whenever a Message or partial Message has bee
 
 ### Batching Sends
 
-Since sending a Message may involve a context switch between the application and the transport system, sending patterns that involve multiple small Messages can incur high overhead if each needs to be enqueued separately. To avoid this, the application can indicate a batch of Send actions through the API. When this is used, the implementation can defer the processing Messages until the batch is complete.
+Since sending a Message may involve a context switch between the application and the transport system, sending patterns that involve multiple small Messages can incur high overhead if each needs to be enqueued separately. To avoid this, the application can indicate a batch of Send actions through the API. When this is used, the implementation can defer the processing of Messages until the batch is complete.
 
 ## Receiving Messages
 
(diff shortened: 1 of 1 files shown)


tfpauly (approved):

draft-ietf-taps-impl.md
```diff
@@ -557,7 +557,7 @@ The application should be notified whenever a Message or partial Message has bee
 
 ### Batching Sends
 
-Since sending a Message may involve a context switch between the application and the transport system, sending patterns that involve multiple small Messages can incur high overhead if each needs to be enqueued separately. To avoid this, the application can indicate a batch of Send actions through the API. When this is used, the implementation should hold off on processing Messages until the batch is complete.
+Since sending a Message may involve a context switch between the application and the transport system, sending patterns that involve multiple small Messages can incur high overhead if each needs to be enqueued separately. To avoid this, the application can indicate a batch of Send actions through the API. When this is used, the implementation can defer the processing Messages until the batch is complete.
```
tfpauly, line 560:
```suggestion
Since sending a Message may involve a context switch between the application and the transport system, sending patterns that involve multiple small Messages can incur high overhead if each needs to be enqueued separately. To avoid this, the application can indicate a batch of Send actions through the API. When this is used, the implementation can defer the processing of Messages until the batch is complete.
```

PR #768, opened 2021-04-09
Restructure tree structure

### LINKED ISSUE #743: Is the "candidate tree" concept actually normative?
Section 4 paragraph 3: "This document structures candidates for racing as a tree". This statement is true, but it doesn't say whether this structure is a necessary outcome of the architecture and the design of the interface, or merely a terminological convention that makes it much easier to discuss. I think it's the latter.  

At the least, this sentence should be clarified to make it clear that the tree structure in section 4 (and subsequent) is terminological, not normative. To make this clearer, I'd also create a new "operations on candidate trees" sections, containing sections 4.1.2, 4.1.3, 4.1.4, and 4.1.5.

### LINKED ISSUE #756: Structuring Options as a Tree?: what is an option?
“4.1.2. Structuring Options as a Tree”
- I don’t like “options” on its own, I think this needs some other word with it!
- what is an option v. an connection attempt? 
“When an implementation responsible for connection establishment needs to consider multiple options,”
…or is something optional here (v. what is meant by 4.1.1.3.3)?
… or actually is this “Alternatives for Structuring as a Tree”? 
 (help me please!)

britram:
🎶 

Fixes #743. Fixes #756.


britram:
Best reviewed by reading the full rendered document [here](https://ietf-tapswg.github.io/api-drafts/743-retree/draft-ietf-taps-impl.html#name-implementing-connection-est)

draft-ietf-taps-impl.md
```diff
@@ -185,7 +185,9 @@ It is expected that the database of system policies and the method of looking up
...
 
-Connection establishment is divided into two top-level steps: Candidate Gathering, to identify the paths, protocols, and endpoints to use, and Candidate Racing (see Section 4.2.2 of {{I-D.ietf-taps-arch}}), in which the necessary protocol handshakes are conducted so that the transport system can select which set to use. This document structures candidates for racing as a tree.
+Connection establishment is divided into two top-level steps: Candidate Gathering, to identify the paths, protocols, and endpoints to use, and Candidate Racing (see Section 4.2.2 of {{I-D.ietf-taps-arch}}), in which the necessary protocol handshakes are conducted so that the transport system can select which set to use.
+
+This document structures the candidates for racing as a tree as terminological convention; it is not necessary for an implementation to do so, but it does ease the illustration of how racing works.
```
gorryfair, line 191: /it is not necessary for an implementation to do so, but it/a tree structure is not the only way in which bus can be implemented, but this does/ ... might seem to be clearer.

draft-ietf-taps-impl.md
```diff
@@ -185,7 +185,9 @@ It is expected that the database of system policies and the method of looking up
...
+
+This document structures the candidates for racing as a tree as terminological convention; it is not necessary for an implementation to do so, but it does ease the illustration of how racing works. 
 
 The most simple example of this process might involve identifying the single IP address to which the implementation wishes to connect, using the system's current default interface or path, and starting a TCP handshake to establish a stream to the specified IP address. However, each step may also vary depending on the requirements of the connection: if the endpoint is defined as a hostname and port, then there may be multiple resolved addresses that are available; there may also be multiple interfaces or paths available, other than the default system interface; and some protocols may not need any transport handshake to be considered "established" (such as UDP), while other connections may utilize layered protocol handshakes, such as TLS over TCP.
```
gorryfair, line 192: Is this /vary/ or /differ/ ?

draft-ietf-taps-impl.md
```diff
@@ -410,6 +388,30 @@ When ordering branches, an implementation can give more weight to properties tha
...
 
 
+## Candidate Gathering {#gathering}
+
+The step of gathering candidates involves identifying which paths, protocols, and endpoints may be used for a given Connection. This list is determined by the requirements, prohibitions, and preferences of the application as specified in the Selection Properties.
+
+### Gathering Endpoint Candidates
+
+Both Local and Remote Endpoint Candidates must be discovered during connection establishment.  To support Interactive Connectivity Establishment (ICE) {{?RFC8445}}, or similar protocols that involve out-of-band indirect signalling to exchange candidates with the Remote Endpoint, it’s important to be able to query the set of candidate Local Endpoints, and give the protocol stack a set of candidate Remote Endpoints, before it attempts to establish connections.
```
gorryfair, line 397: /it's/it is/
is /be able to/ needed?
/give the protocol stack/provide the protocol stack with/
/it attempts/the Local Endpoint attempts/

draft-ietf-taps-impl.md
```diff
@@ -410,6 +388,30 @@ When ordering branches, an implementation can give more weight to properties tha
...
+
+#### Local Endpoint candidates
+
+The set of possible Local Endpoints is gathered.  In the simple case, this merely enumerates the local interfaces and protocols, allocates ephemeral source ports.  For example, a system that has WiFi and Ethernet and supports IPv4 and IPv6 might gather four candidate locals (IPv4 on Ethernet, IPv6 on Ethernet, IPv4 on WiFi, and IPv6 on WiFi) that can form the source for a transient.
```
gorryfair, line 401: /allocates ephemeral source ports/and allocates ephemeral source ports/
/locals/Local Endpoints/
/the source/a source/
/transient/????

draft-ietf-taps-impl.md
```diff
@@ -410,6 +388,30 @@ When ordering branches, an implementation can give more weight to properties tha
...
+
+#### Remote Endpoint Candidates
+
+The Remote Endpoint is typically a name that needs to be resolved into a set of possible addresses that can be used for communication.  Resolving the Remote Endpoint is the process of recursively performing such name lookups, until fully resolved, to return the set of candidates for the remote of this connection.
+
+How this is done will depend on the type of the Remote Endpoint, and can also be specific to each Local Endpoint.  A common case is when the Remote Endpoint is a DNS name, in which case it is resolved to give a set of IPv4 and IPv6 addresses representing that name.  Some types of remote might require more complex resolution.  Resolving the Remote Endpoint for a peer-to-peer connection might involve communication with a rendezvous server, which in turn contacts the peer to gain consent to communicate and retrieve its set of candidate locals, which are returned and form the candidate remote addresses for contacting that peer.
```
gorryfair, line 411: At the start of a para /How this is/How this resolution is/

gorryfair: /types of remote/ is this /types of Remote Endpoint/ or /remote peer/
/candidate locals/candidate Local Endpoints/
/candidate remote addresses/set of remote ???what??/

draft-ietf-taps-impl.md
```diff
@@ -410,6 +388,30 @@ When ordering branches, an implementation can give more weight to properties tha
...
+
+Resolving the Remote Endpoint is not a local operation.  It will involve a directory service, and can require communication with the remote to rendezvous and exchange peer addresses.  This can expose some or all of the candidate locals to the remote.
+
```
gorryfair, line 414: /with the remote/with the remote peer/
/candidate locals to the remote/candidate Local Endpoints to the Remote Endpoint/

gorryfair (changes requested):
Heading in what seems the correct direction, but needs changes to improve reading/consistency.

draft-ietf-taps-impl.md
britram:
```suggestion
Connection establishment is divided into two top-level steps: Candidate Gathering, to identify the paths, protocols, and endpoints to use, and Candidate Racing (see Section 4.2.2 of {{I-D.ietf-taps-arch}}), in which the necessary protocol handshakes are conducted so that the transport system can select which set to use.

This document structures the candidates for racing as a tree as terminological convention, While a tree structure is not the only way in which racing can be implemented, it does ease the illustration of how racing works.
```

draft-ietf-taps-impl.md
```diff
@@ -185,7 +185,9 @@ It is expected that the database of system policies and the method of looking up
...
 
-Connection establishment is divided into two top-level steps: Candidate Gathering, to identify the paths, protocols, and endpoints to use, and Candidate Racing (see Section 4.2.2 of {{I-D.ietf-taps-arch}}), in which the necessary protocol handshakes are conducted so that the transport system can select which set to use. This document structures candidates for racing as a tree.
+Connection establishment is divided into two top-level steps: Candidate Gathering, to identify the paths, protocols, and endpoints to use, and Candidate Racing (see Section 4.2.2 of {{I-D.ietf-taps-arch}}), in which the necessary protocol handshakes are conducted so that the transport system can select which set to use.
+
+This document structures the candidates for racing as a tree as terminological convention; it is not necessary for an implementation to do so, but it does ease the illustration of how racing works.
```
britram, line 190:
```suggestion
This document structures the candidates for racing as a tree as terminological convention. While a 
a tree structure is not the only way in which racing can be implemented, it does ease the illustration of how racing works. 
```

draft-ietf-taps-impl.md
britram: (added suggested change)

draft-ietf-taps-impl.md
britram: Differ is better.

britram, line 192:
```suggestion
The most simple example of this process might involve identifying the single IP address to which the implementation wishes to connect, using the system's current default interface or path, and starting a TCP handshake to establish a stream to the specified IP address. However, each step may also differ depending on the requirements of the connection: if the endpoint is defined as a hostname and port, then there may be multiple resolved addresses that are available; there may also be multiple interfaces or paths available, other than the default system interface; and some protocols may not need any transport handshake to be considered "established" (such as UDP), while other connections may utilize layered protocol handshakes, such as TLS over TCP.
```

draft-ietf-taps-impl.md
britram, line 397:
```suggestion
Both Local and Remote Endpoint Candidates must be discovered during connection establishment.  To support Interactive Connectivity Establishment (ICE) {{?RFC8445}}, or similar protocols that involve out-of-band indirect signalling to exchange candidates with the Remote Endpoint, it is important to query the set of candidate Local Endpoints, and provide the protocol stack with a set of candidate Remote Endpoints, before the Local Endpoint attempts to establish connections.
```

britram: "be able to" is actually doing something here but I won't quibble. made all these in a suggested change.

draft-ietf-taps-impl.md
britram: I'm not sure what transient is doing here (all these are from the original text). Made a suggested change.

britram, line 401:
```suggestion
The set of possible Local Endpoints is gathered.  In the simple case, this merely enumerates the local interfaces and protocols, and allocates ephemeral source ports.  For example, a system that has WiFi and Ethernet and supports IPv4 and IPv6 might gather four candidate Local Endpoints (IPv4 on Ethernet, IPv6 on Ethernet, IPv4 on WiFi, and IPv6 on WiFi) that can form the source for a transient.
```

draft-ietf-taps-impl.md
britram: Same; accepted suggestions, will pull remaining ??? into a separate issue after merge.

britram, line 411:
```suggestion
How this resolution is done will depend on the type of the Remote Endpoint, and can also be specific to each Local Endpoint.  A common case is when the Remote Endpoint is a DNS name, in which case it is resolved to give a set of IPv4 and IPv6 addresses representing that name.  Some types of Remote Endpoint might require more complex resolution.  Resolving the Remote Endpoint for a peer-to-peer connection might involve communication with a rendezvous server, which in turn contacts the peer to gain consent to communicate and retrieve its set of candidate Local Endpoints, which are returned and form the candidate remote addresses for contacting that peer.
```

draft-ietf-taps-impl.md
```diff
@@ -410,6 +388,30 @@ When ordering branches, an implementation can give more weight to properties tha
...
+
+How this is done will depend on the type of the Remote Endpoint, and can also be specific to each Local Endpoint.  A common case is when the Remote Endpoint is a DNS name, in which case it is resolved to give a set of IPv4 and IPv6 addresses representing that name.  Some types of remote might require more complex resolution.  Resolving the Remote Endpoint for a peer-to-peer connection might involve communication with a rendezvous server, which in turn contacts the peer to gain consent to communicate and retrieve its set of candidate locals, which are returned and form the candidate remote addresses for contacting that peer.
+
+Resolving the Remote Endpoint is not a local operation.  It will involve a directory service, and can require communication with the remote to rendezvous and exchange peer addresses.  This can expose some or all of the candidate locals to the remote.
```
britram, line 413:
```suggestion
Resolving the Remote Endpoint is not a local operation.  It will involve a directory service, and can require communication with the Remote Endpoint to rendezvous and exchange peer addresses.  This can expose some or all of the candidate Local Endpoints to the Remote Endpoint.
```

britram:
Going ahead and merging this ahead of other editorial changes (since it's a big move and highly likely to conflict)

PR #771, opened 2021-04-12
Clarify Events upon various close / abort calls.

### LINKED ISSUE #761: Should CloseGroup() and/or AbortGroup() generate events?
these are analogous to connection-level things that generate events.

mwelzl:
(Gracefully,  he he  :)  ) Closes #761.

### DIFF
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -2916,15 +2916,15 @@ Any calls to Receive once the Final Message has been delivered will result in er
 # Connection Termination {#termination}
 
 A Connection can be terminated i) by the Local Endpoint (i.e., the application calls the Close, CloseGroup, Abort or AbortGroup Action), ii) by the Remote Endpoint (i.e., the remote application calls the Close, CloseGroup, Abort or AbortGroup Action), or iii) because of an error (e.g., a timeout). A local call of the Close Action will
-cause the Connection to send the Closed Event or the ConnectionError Event, and a local call of
-the CloseGroup Action will cause all of the Connections in the group to send the Closed Event
-or the ConnectionError Event. A local call of the Abort Action will cause the Connection to send
+cause the Connection to either send a Closed Event or a ConnectionError Event, and a local call of
+the CloseGroup Action will cause all of the Connections in the group to either send a Closed Event
+or a ConnectionError Event. A local call of the Abort Action will cause the Connection to send
 a ConnectionError Event, indicating local Abort as a reason, and a local call of the AbortGroup Action
 will cause all of the Connections in the group to send a ConnectionError Event, indicating local Abort
 as a reason.
 
 Remote Action calls map to Events similar to local calls (e.g., a remote Close causes the
-Connection to send a Closed Event or a ConnectionError Event), but, different from local Action calls,
+Connection to either send a Closed Event or a ConnectionError Event), but, different from local Action calls,
 it is not guaranteed that such Events will indeed be invoked. When an application needs to free resources
 associated with a Connection, it should therefore not rely on the invocation of such Events due to
 termination calls from the Remote Endpoint, but instead use the local termination Actions.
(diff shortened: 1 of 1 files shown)


martinduke (changes requested):
From this text I still don't understand when TAPS sends Closed<>. Is it (1) in response to a Close(), when all data has been delivered, or when the remote endpoint has closed?

Is it the first of these two events to occur?

mwelzl:
Ah, I see, sorry!  I just committed a fix; is it better now?

The new text describing the Closed Event is only this:

> The Closed Event informs the application that a Close Action has successfully completed, or that the Remote Endpoint has closed the Connection.

tfpauly (approved):

martinduke:
Still, it says that this means either everything has been delivered, or the peer has closed the connection. Does this mean I would get two events on a clean TCP close?

mwelzl:
No, one: because the Connection was "closed". Case 1 in the sentence, where it says "Close Action", refers to the call made by the local application - the application made the call, and it terminated correctly. The remote close is the other case leading to a Connection being closed - since it doesn't say "abort" or "forcefully closed", it's the same thing, a graceful close. So... there are two reasons leading to one thing happening, and when that one thing is over, the applications gets an Event.

I'm not sure how you read this differently... it's clear to me. Would this variant help?

> The Closed Event informs the application that the Connection has been closed - either because a Close Action call has successfully completed, or because the Remote Endpoint has closed the Connection.

martinduke:
I've read this again, and I think it's OK.

philsbln (changes requested):
I am still missing a clear statement whether a developer can rely on eventually receiving either a Closed event or a ConnectionError event per Connection.

tfpauly:
Can we explain in the start of the connection termination section a specific list of the different end states/events a connection can have? (Either Closed or ConnectionError, for example).

mwelzl:
Note to self: also add some introductory text for this section as requested by Anna elsewhere

mwelzl:
Done - please take a look at the new intro.

draft-ietf-taps-interface.md
```diff
@@ -2915,27 +2915,45 @@ Any calls to Receive once the Final Message has been delivered will result in er
 
 # Connection Termination {#termination}
 
+A Connection can be terminated i) by the Local Endpoint (i.e., the application calls the Close, CloseGroup, Abort or AbortGroup Action), ii) by the Remote Endpoint (i.e., the remote application calls the Close, CloseGroup, Abort or AbortGroup Action), or iii) because of an error (e.g., a timeout). A local call of the Close Action will
+cause the Connection to send the Closed Event or the ConnectionError Event, and a local call of
+the CloseGroup Action will cause all of the Connections in the group to send the Closed Event
+or the ConnectionError Event. A local call of the Abort Action will cause the Connection to send
```
gorryfair, line 2921: is this /to either send/ ... I'm unsure what the "or" refers to.

draft-ietf-taps-interface.md
```diff
@@ -2915,27 +2915,45 @@ Any calls to Receive once the Final Message has been delivered will result in er
...
 Connection.Close()
 ~~~
 
-The Closed Event informs the application that the Remote Endpoint has closed the
-Connection. There is no guarantee that a remote Close will indeed be
-signaled.
+The Closed Event informs the application that a Close Action has successfully
+completed, or that the Remote Endpoint has closed the Connection.
+There is no guarantee that a remote Close will indeed be signaled.
+
```
gorryfair, line 2948: why /indeed/ in /will indeed be signaled./?

draft-ietf-taps-interface.md
```diff
@@ -2961,7 +2983,7 @@ Connection.AbortGroup()
 A ConnectionError informs the application that: 1) data could not be delivered to the
 peer after a timeout, 
 or 2) the Connection has been aborted (e.g., because the peer has called Abort).
-There is no guarantee that an Abort will indeed be signaled.
+There is no guarantee that an Abort from the peer will indeed be signaled.
```
gorryfair, line 2987: Similarly, should this /will be signalled/?

gorryfair (commented):
A couple of minor queries

Issue #780, opened 2021-04-12
To me, Safely Replayable text seems unclear ( 7.1.3.4.)


gorryfair:
The current text says:

   If true, Safely Replayable specifies that a Message is safe to send
   to the Remote Endpoint more than once for a single Send Action.  It
   marks the data as safe for certain 0-RTT establishment techniques,
   where retransmission of the 0-RTT data may cause the remote
   application to receive the Message multiple times.

   For protocols that do not protect against duplicated messages, e.g.,
   UDP, all messages need to be marked as "Safely Replayable".  To
   enable protocol selection to choose such a protocol, "Safely
   Replayable" needs to be added to the TransportProperties passed to
   the Preconnection.  If such a protocol was chosen, disabling "Safely
   Replayable" on individual messages MUST result in a SendError.
—
I think this could perhaps be misleading, and suggest:
- 1. I would like to change /UDP/directly over UDP/ .... because clearly it is possible that a transport such as QUIC also uses UDP, but provides this sort of protection above the UDP layer.
- 2. This is still written from the perspective of the protocol. I think it is important here to also therefore add a sentence then clarifies that the application might be able to designed to detect and ignore such duplication. ‘Protection from replay of messages could be provided by an application when a transport system replays messages’.

PR #787, opened 2021-04-14
PR for when a transport system replays messages

gorryfair:
Addition to clarify what happens when a transport system replays messages - #780

### DIFF
--- a/draft-ietf-taps-impl.md
+++ b/draft-ietf-taps-impl.md
@@ -533,7 +533,7 @@ The effect of the application sending a Message is determined by the top-level p
 
 - Ordered: when this is false, this disables the requirement of in-order-delivery for protocols that support configurable ordering. When the protocol stack does not support configurable ordering, this property may be ignored.
 
-- Safely Replayable: when this is true, this means that the Message can be used by a transport mechanism that might transfer it multiple times -- e.g., as a result of racing multiple transports or as part of TCP Fast Open. Also, protocols that do not protect against duplicated messages, such as directlty layered over UDP, can only be used with Messages that are Safely Replayable. When a transport system is permitted to replay messages, replay protection could be provided by the application.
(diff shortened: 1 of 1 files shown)


draft-ietf-taps-impl.md
```diff
@@ -533,7 +533,7 @@ The effect of the application sending a Message is determined by the top-level p
 
 - Ordered: when this is false, this disables the requirement of in-order-delivery for protocols that support configurable ordering. When the protocol stack does not support configurable ordering, this property may be ignored.
 
-- Safely Replayable: when this is true, this means that the Message can be used by mechanisms that might transfer it multiple times -- e.g., as a result of racing multiple transports or as part of TCP Fast Open. Also, protocols that do not protect against duplicated messages, such as UDP, can only be used with Messages that are Safely Replayable.
+- Safely Replayable: when this is true, this means that the Message can be used by a transport mechanism that might transfer it multiple times -- e.g., as a result of racing multiple transports or as part of TCP Fast Open. Also, protocols that do not protect against duplicated messages, such as directlty layered over UDP, can only be used with Messages that are Safely Replayable. When a transport system is permitted to replay messages, replay protection could be provided by the application.
```
mwelzl, line 536:
```suggestion
- Safely Replayable: when this is true, this means that the Message can be used by a transport mechanism that might transfer it multiple times -- e.g., as a result of racing multiple transports or as part of TCP Fast Open. Also, protocols that do not protect against duplicated messages, such as UDP (when used directly, without a protocol layered atop), can only be used with Messages that are Safely Replayable. When a transport system is permitted to replay messages, replay protection could be provided by the application.
```

Just a suggestion, fine by me to phrase this differently - but now, you have:
- typo directlty
- "protocols that do not protect against duplicated messages, such as directly layered over UDP" ... is this really okay to write? It reads awkwardly to me ... I mean, such as asks for, what, a protocol? So this would mean "such as protocols that are directly layered over UDP?" but QUIC is such a protocol. This is about the application being layered directly over UDP!  So ...   hence my suggestion above.

mwelzl (changes requested):
A small suggestion

mwelzl (approved):

Issue #782, opened 2021-04-12
No Segmentation  


gorryfair:
Issue: No Segmentation  property - To me, the final text seems unclear about how it operates.:

   When set to true, this property requests the network layer at the
   sending endpoint to not fragment the packets generated by the
   transport layer.  When running over IPv4, setting this property to
   true will also cause the Don't Fragment bit to be set in the IP
   header.  When this property is set, an attempt to send a message size
   greater than the transport's current estimate of its maximum packet
   size (singularTransmissionMsgMaxLen) will result in a SendError.
   This only takes effect when the transport and network layer support
   this functionality.

- Is this really about the network fragmenting packets? - I do not think it is. To me, this is about the transport system segmenting packets.  I think this should read:

  When set to true, this property requests the transport system at the
   sending endpoint to not segment a single message into multiple IP packets.  
   When this property is set, an attempt to send a message
   greater in size than the transport's current estimate of its maximum packet
   size (singularTransmissionMsgMaxLen) will result in a SendError.
   When set to true, this also implies that noFragmentation is also set to true. 
   This only takes effect when the transport layer supports
   this functionality. Requiring this property prevents use of transports such as TCP.

gorryfair:
Linked to #781

mwelzl:
Good: I think you're right that this should not be about network-layer fragmentation.

Bad: The last sentence tells me that you wish for this to be a Selection Property, not (as it now is) a Message Property. From the minset appendix, we can see that the present property maps to "Request not to bundle messages" from SCTP; checking in RFC 8303, I find that this is the "no-bundle flag" of the send call, which we find in RFC 4960.  So that's what it is - a per-message request at the transport layer.

So, I suggest to update the text as follows (this only replaces the first two sentences with your new first sentence):

> When set to true, this property requests the transport system at the sending endpoint to not segment a single message into multiple IP packets. When this property is set, an attempt to send a message size greater than the transport's current estimate of its maximum packet size (singularTransmissionMsgMaxLen) will result in a SendError. This only takes effect when the transport and network layer support this functionality.

... and, if you NEED to be able to choose a transport that supports this functionality, look at the selection properties to see if you need to make a change to them. I'm not sure: we currently have "Preservation of Message Boundaries" there ("This property specifies whether the application needs or prefers to use a transport protocol that preserves message boundaries."), why is this not good enough?

gorryfair:
[...]

PR #793, opened 2021-04-14
Tiny minset fix

mwelzl:
This fixes a wrong reference to the "no segmentation" property: in fact, the "request not to bundle messages" is SCTP's Nagle, and covered by the Capacity profile.
(I mentioned this mistake in #782).


gorryfair (approved):
Looks good, as discussed.

PR #794, opened 2021-04-15
Fixes around Clone and entanglement

### LINKED ISSUE #774: What is "entangled"?
I don’t actually understand the difference between a connection in a Connection Group and an “entangled connection” ? 

Both terms are used in API, but in different ways (possibly by different authors) - do we really really need the word “entangled”??  if we do we need, can we define and explain this.

### LINKED ISSUE #788: Are message framers entangled in a connection group?
This is not fully clear from section 5.4., however, if connections in a group are each QUIC streams, I think it should be possible to use different framers on each stream. So it might be okay that a framers is copied when clone is called but changing a framer on one connection should not change the framer on the other connection in a group...?

mwelzl:
Closes #774. Closes #788.
This changes the use of the term "entanglement" as discussed in #774, and moves some text around in this section for better clarity and less repetition. Also, addressing #788, it introduces a framer as an optional parameter to the Clone call.

### DIFF
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -1616,16 +1616,16 @@ Calling Clone on a Connection yields a Connection Group containing two Connectio
 Connection on which Clone was called, and a resulting cloned Connection.
 The new Connection is actively openend, and it will send a Ready Event or an EstablishmentError Event.
 Calling Clone on any of these Connections adds another Connection to
-the Connection Group, and so on. Connections in a Connection Group share all
+the Connection Group. Connections in a Connection Group share all
 Connection Properties except `Connection Priority` (see {{conn-priority}}),
 and these Connection Properties are entangled: Changing one of the
 Connection Properties on one Connection in the Connection Group
-automatically changes it for all others. For example, changing
+automatically changes the Connection Property for all others. For example, changing
 `Timeout for aborting Connection` (see
 {{conn-timeout}}) on one Connection in a Connection Group will automatically
-change this Connection Property for all other Connections in the Connection Group in the same way.
+make the same change to this Connection Property for all other Connections in the Connection Group.
 Like all other Properties, `Connection Priority` is copied 
-to the new Connection when calling Clone(), but changing 
+to the new Connection when calling Clone(), but in this case, a later change to the 
 `Connection Priority` on one Connection does not change it on the
 other Connections in the same Connection Group.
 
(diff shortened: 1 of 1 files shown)


gorryfair (approved):
This resolves the issues with the entangled wording, many thanks

draft-ietf-taps-interface.md
```diff
@@ -1606,26 +1606,37 @@ Preconnection -> EstablishmentError<reason?>
...
-the cloned Connection when calling Clone. In other words, a cloned Connection 
+Calling Clone on any of these Connections adds another Connection to
+the Connection Group, and so on. Connections in a Connection Group share all
+Connection Properties except `Connection Priority` (see {{conn-priority}}),
+and these Connection Properties are entangled: Changing one of the
+Connection Properties on one Connection in the Connection Group
+automatically changes it for all others. For example, changing
+`Timeout for aborting Connection` (see
+{{conn-timeout}}) on one Connection in a Connection Group will automatically
+change this Connection Property for all other Connections in the Connection Group in the same way.
```
gorryfair, line 1626: Suggest: 
/will automatically make the same change to this Connection Property for all other Connections in the Connection Group./

draft-ietf-taps-interface.md
```diff
@@ -1606,26 +1606,37 @@ Preconnection -> EstablishmentError<reason?>
...
+Like all other Properties, `Connection Priority` is copied 
+to the new Connection when calling Clone(), but changing 
+`Connection Priority` on one Connection does not change it on the
```
gorryfair, line 1629: /changing/in this case, a later change to the/

draft-ietf-taps-interface.md
```diff
@@ -1606,26 +1606,37 @@ Preconnection -> EstablishmentError<reason?>
...
+other Connections in the same Connection Group.
+
+Message Properties are also not entangled.  For example,
+changing `Lifetime` (see {{msg-lifetime}}) of a Message will only affect a
+single Message on a single Connection.
+
+A new Connection created by Clone can get a framer assigned via the optional
```
gorryfair, line 1636: /get/have/
and 
/framer/Message Framer/

draft-ietf-taps-interface.md
```diff
@@ -1642,22 +1653,14 @@ Passive Connections can also be added to the same group -- e.g., when a Listener
...
-functionality to implement Clone. In that case, entangled Connections are
+functionality to implement Clone. In that case, Connections in a Connection Group are
 multiplexed together, giving them similar treatment not only inside endpoints,
 but also across the end-to-end Internet path.
 
 Note that calling Clone() can result in on-the-wire signaling, e.g., to open a new
-connection, depending on the underlying Protocol Stack. When Clone() leads to
-multiple connections being opened instead of multi-streaming,
+transport connection, depending on the underlying Protocol Stack. When Clone() leads to
+multiple such connections being opened instead of multi-streaming,
```
gorryfair, line 1663: /multiple such connections being opened/the opening of multiple such connections,/

gorryfair (changes requested):
I like the way this is going, but I have suggestions for various changes that might improve this.

mwelzl:
@gorryfair I incorporated all of your suggestions, thanks!

gorryfair (approved):
Looks good to me.

PR #795, opened 2021-04-21
Hopefully clearer text for the ReceiveError Event.

### LINKED ISSUE #792: Clarify handling of receive errors
Actually handling of receive error is also not fully clear to me. Section 7.3.2.3 says:
"A ReceiveError occurs when data is received by the underlying Protocol Stack that cannot be fully retrieved or parsed"

However, I assume that this error should rather not be called if a transport is reliable because the assumption is that the packet will be retransmitted, right?

E.g. in case of QUIC if a packet cannot be decrypted would that trigger an error or not? I assume not, because the QUIC stack cannot even know if a packet contained any payload data, right?

mwelzl:
Closes #792.

### DIFF
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -2863,8 +2863,8 @@ to be notified of such errors can be useful. For example, a ReceiveError can
 indicate that a Message (identified via the optionally passed MessageContext)
 that was being partially received previously, but had not
 completed, encountered an error and will not be completed. This can be useful
-for the application, which may want to use this error as a hint to remove the
-previously received Message part(s) from its memory. As another example,
+for an application, which may want to use this error as a hint to remove
+previously received Message parts from memory. As another example,
 if an incoming Message does not fulfill the Required Minimum Corruption
 Protection Coverage for Receiving property (see {{conn-recv-checksum}}),
 an application can use this error as a hint to inform the peer application
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -2859,7 +2859,7 @@ Connection -> ReceiveError<messageContext?, reason?>
 
 A ReceiveError occurs when data is received by the underlying Protocol Stack
 that cannot be fully retrieved or parsed, and when it is useful for the application
-to be notified of such errors can be useful. For example, a ReceiveError can
+to be notified of such errors. For example, a ReceiveError can
 indicate that a Message (identified via the optionally passed MessageContext)
 that was being partially received previously, but had not
 completed, encountered an error and will not be completed. This can be useful
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -2860,7 +2860,7 @@ Connection -> ReceiveError<messageContext, reason?>
 A ReceiveError occurs when data is received by the underlying Protocol Stack
 that cannot be fully retrieved or parsed, and when it is useful for the application
 to be notified of such errors. For example, a ReceiveError can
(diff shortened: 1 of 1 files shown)


draft-ietf-taps-interface.md
```diff
@@ -2854,19 +2854,27 @@ large Message of indeterminate length.
 #### ReceiveError {#receive-error}
 
 ~~~
-Connection -> ReceiveError<messageContext, reason?>
+Connection -> ReceiveError<messageContext?, reason?>
```
mirjak, line 2857: Why did you make the messageContext option? Don't you always need this to assign it to a message somehow?

mirjak: And you also removed the text below that said the messageContext would be optional (e.g. for partially received messages). Was that on purpose? However, I think for the two examples you mention below, it seems to make sense to provide some messageContext.

mwelzl: I made it optional because, according to the text that was there, it was optional - it just wasn't written like that in the signature, which I thought was an oversight.

Then I removed the text that said it would be optional because I didn't see the added value of repeating what the function Event signature above already shows. I checked, the document doesn't consistently always say "and there is this parameter, and that parameter".

Regarding really making the messageContext optional or not... well... I can't see other uses than the two examples, and I agree that the messageContext makes sense to include in these two cases. But that doesn't mean that there couldn't be any other reception errors, where handing over a messageContext might not be possible. I don't know, personally, I can't think of any such case...

draft-ietf-taps-interface.md
```diff
@@ -2854,19 +2854,27 @@ large Message of indeterminate length.
...
+for an application, which may want to use this error as a hint to remove
+previously received Message parts from memory. As another example,
+if an incoming Message does not fulfill the Required Minimum Corruption
+Protection Coverage for Receiving property (see {{conn-recv-checksum}}),
+an application can use this error as a hint to inform the peer application
+to adjust the Sending Corruption Protection Length property (see {{msg-checksum}}).
+
+In contrast, internal protocol reception errors (e.g., loss causing retransmissions
+in TCP) are not signalled by this Event. Conditions that irrevocably lead to
+the termination of the Connection are signaled using ConnectionError
```
philsbln, line 2875: Are these _only_ signalled only by an Connection Error or (if one ore more effected messages are known) _in addition_ to a ReceiveError for these messages?

philsbln (changes requested):
Mich clearer, but need more refinement.

draft-ietf-taps-interface.md
```diff
@@ -2854,19 +2854,27 @@ large Message of indeterminate length.
...
+Connection -> ReceiveError<messageContext?, reason?>
 ~~~
 
 A ReceiveError occurs when data is received by the underlying Protocol Stack
-that cannot be fully retrieved or parsed, or when some other indication is
-received that reception has failed. In contrast, conditions that irrevocably lead to
-the termination of the Connection are instead signaled using ConnectionError
+that cannot be fully retrieved or parsed, and when it is useful for the application
+to be notified of such errors. For example, a ReceiveError can
+indicate that a Message (identified via the optionally passed MessageContext)
```
tfpauly, line 2863:
```suggestion
indicate that a Message (identified via the MessageContext)
```

draft-ietf-taps-interface.md
mirjak:
```suggestion
Connection -> ReceiveError<messageContext, reason?>
```

tfpauly (approved):

PR #796, opened 2021-04-21
Remove one confusing sentence (#790)

### LINKED ISSUE #790: Priority for single stream
Section 7.2.6. says " for example, if a transport stack that only provides a single in-order reliable stream is selected, prioritization information can only be ignored."

I don't think this is entirely true as you can still decide in which order you schedule independent messages on one stream.

mwelzl:
Closes #790.


gorryfair (approved):
I like simple examples, this one wasn't great so this is OK.

PR #824, opened 2021-05-05
minset appendix: remove sentence about framers

### LINKED ISSUE #822: Minset recieve data - mapping seems strange
> Receive data (with no message delimiting): "Received" Event (Section 7.3.2.1). See Section 7.1.2 for handling Message framing
in situations where the Protocol Stack only provides a byte-stream transport.

Why is the part about Message framing here where we talk about receiving a byte stream rather than in the Receive message mapping?

mwelzl:
... in text that refers to receiving a byte stream. Closes #822.


tfpauly (approved):

philsbln (approved):

abrunstrom (approved):
Fine to remove the framer part. Reading the update I am wondering if it should also mention the Receive Action?

PR #826, opened 2021-05-07
Clarify applicability of specific properties; fix #789

### LINKED ISSUE #789: MUST ignore setting properties that does not exists
Section 6 says. "Attempts to set Specific Protocol Properties on a protocol stack not containing that specific protocol are simply ignored, and do not raise an error"

Should this be normative?

"Attempts to set Specific Protocol Properties on a protocol stack not containing that specific protocol MUST be ignored, and MUST NOT raise an error"

I note that this a random sentence where spotted that normative language could be used. I don't look for these kind of occasion systematically...

britram:


draft-ietf-taps-interface.md
```diff
@@ -1692,13 +1692,12 @@ Connection Properties represent the configuration and state of the selected
...
-Properties are defined in {{connection-props}} below. Specific Protocol
-Properties are defined in a transport- and implementation-specific way, and MUST
-NOT be assumed to apply across different protocols. Attempts to set Specific
-Protocol Properties on a protocol stack not containing that specific protocol
-are simply ignored, and do not raise an error; however, too much reliance by an
-application on Specific Protocol Properties can significantly reduce the
-flexibility of a transport services implementation.
+Properties are defined in {{connection-props}} below. 
+
+Specific Protocol Properties are defined in a transport- and
```
philsbln, line 1697:
```suggestion
Protocol Specific Properties are defined in a transport- and 
```

draft-ietf-taps-interface.md
```diff
@@ -1692,13 +1692,12 @@ Connection Properties represent the configuration and state of the selected
...
-NOT be assumed to apply across different protocols. Attempts to set Specific
-Protocol Properties on a protocol stack not containing that specific protocol
-are simply ignored, and do not raise an error; however, too much reliance by an
-application on Specific Protocol Properties can significantly reduce the
-flexibility of a transport services implementation.
+Properties are defined in {{connection-props}} below. 
+
+Specific Protocol Properties are defined in a transport- and 
+implementation-specific way, and MUST NOT apply across different protocols. 
+Too much reliance by an application on Specific Protocol Properties can significantly
```
philsbln, line 1699:
```suggestion
Too much reliance by an application on Protocol Specific Properties can significantly 
```

philsbln (changes requested):
Let's make it consistent to the terms in 3.2

draft-ietf-taps-interface.md
tfpauly: This terminology is what's already used in the doc, so I say to keep it.

tfpauly (approved):

draft-ietf-taps-interface.md
philsbln: Yes, in the current version, we have three times "Protocol Specific Properties" and three times "Specific Protocol Properties" meaning the same thing. We should really unify on one term.

mwelzl: Oh yes, absolutely, it should be Protocol Specific Properties everywhere  (actually I wonder if this wouldn't be even better with a dash?  I.e., "Protocol-specific Properties"?).

To me, "Specific Protocol Properties" doesn't read like it would even mean the same thing, it reads to me as "specific ones as opposed to very vague ones"  (well, as opposed to "generic" ones ... so that kinda works, but it's less clear that these are specific *to* a protocol, IMO).

draft-ietf-taps-interface.md
gorryfair: @philsbln - Looking good, but would "Protocol-Specific Properties" also be OK for all the cases?

Issue #775, opened 2021-04-12
What's a callback?


gorryfair:
We have a “ Terminology and Notation” section, but we don’t define these terms:

Clone: Calling Clone on a Connection yields a group of Connections: the parent Connection on which Clone was called, and a resulting cloned Connection. 

Entangled connection:   if we do we need this term, we should define and explain what is different.

Callback: can we have a one line definition that explains the different with respect to an event that an application can also asynchronously receive.

mwelzl:
I can take care of the first two

mwelzl:
As I looked at this now, I felt that such definitions would seem a bit out of place in the "Terminology and Notation" section; rather, if we need to define it globally, the concepts of Cloning and Entanglement should be defined where we also define what our Connections are, what Transport Properties are, etc.  - and this is in -arch.

Thoughts?

gorryfair:
This isn't need for entangled with the way used in PR #774

mwelzl:
I'm adding the arch label so people notice  :-)

tfpauly:
The one action remaining is defining "callback" within the scope of section 4.3.

gorryfair:
Looks like this will be addressed.

PR #827, opened 2021-05-07
Explicitly distinguish Callbacks from Events

philsbln:
Closes issue #775

### DIFF
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -1364,8 +1364,9 @@ Representation of Security Parameters in implementations should parallel that ch
 
 Security decisions, especially pertaining to trust, are not static. Once configured,
 parameters may also be supplied during connection establishment. These are best
-handled as client-provided callbacks. Security handshake callbacks that may be
-invoked during connection establishment include:
+handled as client-provided callbacks. 
+Callbacks block the progress of a connection, which distinguishes them from other Events in the transport system, while implementations may choose to realize them analogous.
+Security handshake callbacks that may be invoked during connection establishment include:
 
 - Trust verification callback: Invoked when a Remote Endpoint's trust must be validated before the
 handshake protocol can continue.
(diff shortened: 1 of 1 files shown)


draft-ietf-taps-interface.md
```diff
@@ -1364,8 +1364,9 @@ Representation of Security Parameters in implementations should parallel that ch
 
 Security decisions, especially pertaining to trust, are not static. Once configured,
 parameters may also be supplied during connection establishment. These are best
-handled as client-provided callbacks. Security handshake callbacks that may be
-invoked during connection establishment include:
+handled as client-provided callbacks. 
+Callbacks block the progress of a connection, which distinguishes them from other Events in the transport system, while implementations may choose to realize them analogous.
```
britram, line 1368:
```suggestion
Callbacks block the progress of a connection, which distinguishes them from other Events in the transport system. How callbacks and events are implemented is specific to each implementation.
```

tfpauly (approved):

draft-ietf-taps-interface.md
```diff
@@ -1364,8 +1364,9 @@ Representation of Security Parameters in implementations should parallel that ch
 
 Security decisions, especially pertaining to trust, are not static. Once configured,
 parameters may also be supplied during connection establishment. These are best
-handled as client-provided callbacks. Security handshake callbacks that may be
-invoked during connection establishment include:
+handled as client-provided callbacks. 
+Callbacks block the progress of a connection, which distinguishes them from other Events in the transport system. How callbacks and events are implemented is specific to each implementation.
```
mwelzl, line 1368:
```suggestion
Callbacks block the progress of a Connection's establishment, which distinguishes them from other Events in the transport system. How callbacks and events are implemented is specific to each implementation.
```

mwelzl (changes requested):
Minor suggestion

draft-ietf-taps-interface.md
philsbln: I would prefer to leave the connection on lower case (as I am not sure that the connection object we refer to in upper case has/makes progress, but something in the underlaying transport system).

 ```suggestion
Callbacks block the progress of the connection establishment, which distinguishes them from other Events in the transport system. How callbacks and events are implemented is specific to each implementation.
```

mwelzl: ok, this makes sense to me

mwelzl (approved):

britram (approved):

PR #830, opened 2021-05-08
Add appendix note about object freeing

### LINKED ISSUE #762: Can you close a Preconnection?
From reading -api I am not certain of the lifecycle of preconnections?

Is it
-Generate preconnection
- initiate() or listen()
- preconnection is immediately destroyed?

or
-Generate preconnection
- initiate() or listen()
- initiate()/listen()
- ...

Either way, there ought to be a way for taps to throw away this state, either because we never called initiate/listen (1st case) or we're done creating new connections (2nd case).

Or is TAPS not holding any state, with the application fully responsible for holding the object? I'm playing with treating connections and preconnections as socket-style file descriptors, and that model won't work without a preconnection.destroy() or whatever. But maybe that model is just wrong!

tfpauly:
Closes #762

### DIFF
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -3140,6 +3140,16 @@ implementation-specific limitations. For example:
 - A Collection may be represented as a `std::set` in C++ or as a `set` in
   Python. In C, it may be represented as an array or as a higher-level data
   structure with appropriate accessors defined.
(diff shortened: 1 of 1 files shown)


martinduke (commented):
LGTM

mwelzl (approved):

PR #832, opened 2021-05-08
Make disabling security explicit

### LINKED ISSUE #804: S4: Why are only SecurityParameters optional in NewPreconnection?
For other parameters you have to pass in an empty object if you do not want to set anything but not for SecurityParameters. Unclear why the parameters are treated differently?

tfpauly:
Closes #804


abrunstrom (approved):
LGTM, thanks!

mwelzl (approved):

PR #834, opened 2021-05-13
Suggestion for no fragmentation

gorryfair:
This attempts to set "no segmentation" as the usual case, and this as a corner case - such as would be from a transport that segments at layer 4, but does not fragment at layer 3. This latter case is what is desired to send one message using a sequence of seprate IP packets.

### DIFF
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -2512,14 +2512,21 @@ Type:
 Default:
 : false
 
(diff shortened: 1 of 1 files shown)


tfpauly (approved):

gorryfair:
Closing the Frag part of the story

gorryfair:
Closing this and merging the Frag parts

Issue #701, opened 2020-12-03
Network Policies


martinduke:
In Sec 2:

Should Section 3.2 also talk about network policies, e.g. a certain network supports a DSCP that the application desires?

britram:
As discussed in Feb 2021 interim, this seems like a rabbithole. Gorry will figure out whether it is and write language if not, or close if so.

gorryfair:
I'd like to add a couple of lines of text here to bullet 2 that discusses how you might learn that an attached network supports certain features - e.g. through a PvD, routing advertisement, or local discovery method, and suggest that this could enable a system to know which paths have specific characteristics.

gorryfair:
I propose this as a starter:

When additional information (such as Provisioning Domain (PvD) information {{RFC7556}}) is provided about the networks over which an endpoint can operate, this can inform the selection between alternate network paths. Such information includes network segment PMTU, set of supported DSCPs, expected usage, cost, etc. The usage of this information by the Transport Services API is generally independent of the specific mechanism/protocol that was used to receive the information (e.g. zero-conf, DHCP, or IPv6 RA).

Thoughts... ?

gorryfair:
Tried to create a PR to add some policy to the path selection #836

PR #836, opened 2021-05-19
Trying to add some policy to the path selection

gorryfair:
This tries to write something small but specific about Network Policies #701

### DIFF
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -835,7 +835,7 @@ between multiple local interfaces that are connected to different access
 networks.
 
 When additional information (such as Provisioning Domain (PvD) information 
-{{RFC7556}}) is provided about the networks over which an endpoint can operate,
+{{RFC7556}}) is available about the networks over which an endpoint can operate,
 this can inform the selection between alternate network paths.  
 Path information can include network segment PMTU, set of supported DSCPs, 
 expected usage, cost, etc. The usage of this information by the Transport 
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -271,7 +271,7 @@ provides:
   application interactions with the transport layer, in line with developments
   in modern platforms and programming languages;
   
-- Selection between alternate network paths that can be informed by additional information that is available about the networks over which an endpoint can operate (e.g. Provisioning Domain (PvD) information {{RFC7556}});  
+- Selection between alternate network paths that can be informed when additional information is available about the networks over which an endpoint can operate (e.g. Provisioning Domain (PvD) information {{RFC7556}});  
 
 - Explicit support for transport-specific features to be applied should that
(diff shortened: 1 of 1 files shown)


draft-ietf-taps-interface.md
```diff
@@ -273,6 +273,8 @@ provides:
 
 - Explicit support for transport-specific features to be applied should that
   particular transport be part of a chosen Protocol Stack;
+  
+- Support for using additional information about the networks over which an endpoint can operate (e.g. Provisioning Domain (PvD) information {{RFC7556}}), to inform the selection between alternate network paths;
```
abrunstrom, line 277: This reads strangely to me as it is not really the API that provides support for this. Not sure anything needs to be added to this section.

draft-ietf-taps-interface.md
```diff
@@ -833,6 +834,14 @@ selection is necessarily tied to path selection. This may involve choosing
 between multiple local interfaces that are connected to different access
 networks.
 
+When additional information (such as Provisioning Domain (PvD) information 
+{{RFC7556}}) is provided about the networks over which an endpoint can operate,
+this can inform the selection between alternate network paths.  
+Path information can include network segment PMTU, set of supported DSCPs, 
+expected usage, cost, etc. The usage of this information by the Transport 
+Services API is generally independent of the specific mechanism/protocol
```
abrunstrom, line 842:
```suggestion
Services System is generally independent of the specific mechanism/protocol 
```

martinduke:
LGTM

abrunstrom (changes requested):
I think the policy is separate from the API.

draft-ietf-taps-interface.md
```diff
@@ -833,6 +834,14 @@ selection is necessarily tied to path selection. This may involve choosing
 between multiple local interfaces that are connected to different access
 networks.
 
+When additional information (such as Provisioning Domain (PvD) information 
+{{RFC7556}}) is provided about the networks over which an endpoint can operate,
+this can inform the selection between alternate network paths.
```
mwelzl, line 839: This seems like the same issue that @abrunstrom pointed out about a different part of this text: "is provided" gives me the impression that it might be the application's job to provide this information to the transport system via the API. That's not what this is, right? In this case, I'd rather go with something like:

When the transport system has additional information (..). ... available ... about the networks over which ...

mwelzl (changes requested):
Thanks for doing this!  I like it, but I have inserted one comment that needs clarification, I think

draft-ietf-taps-interface.md
gorryfair: available will do. I will rework.

mwelzl: I think you've done this but not resolved this conversation. Clicking the button.

mwelzl (approved):

philsbln (approved):
LGTM

PR #837, opened 2021-05-19
Not unreliable, but no reliability guarentees.

gorryfair:
Not requiring things to be unreliable, but allow no reliability guarentees.
and /throw/generate/

### DIFF
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -3316,7 +3316,7 @@ this is offered by the `Close` Action with slightly changed semantics in line wi
 * "Abort without delivering remaining data, causing an event informing the application on the other side" and "Abort without delivering remaining data, not causing an event informing the application on the other side":
 this is offered by the `Abort` action without promising that this is signaled to the other side. If it is, a `ConnectionError` Event will fire at the peer ({{termination}}).
 
-* "Reliably transfer data, with congestion control", "Reliably transfer a message, with congestion control" and "Transfer a message without reliability":
+* "Reliably transfer data, with congestion control", "Reliably transfer a message, with congestion control" and "Transfer a message without reliability guarantees":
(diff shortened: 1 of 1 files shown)


draft-ietf-taps-interface.md
```diff
@@ -3305,7 +3316,7 @@ this is offered by the `Close` Action with slightly changed semantics in line wi
 * "Abort without delivering remaining data, causing an event informing the application on the other side" and "Abort without delivering remaining data, not causing an event informing the application on the other side":
 this is offered by the `Abort` action without promising that this is signaled to the other side. If it is, a `ConnectionError` Event will fire at the peer ({{termination}}).
 
-* "Reliably transfer data, with congestion control", "Reliably transfer a message, with congestion control" and "Unreliably transfer a message":
+* "Reliably transfer data, with congestion control", "Reliably transfer a message, with congestion control" and "Transfer a message without reliability":
```
abrunstrom, line 3319: Perhaps "without reliability guarantees"?

abrunstrom (approved):
LGTM, only one possible small change suggested.

draft-ietf-taps-interface.md
```diff
@@ -3305,7 +3316,7 @@ this is offered by the `Close` Action with slightly changed semantics in line wi
 * "Abort without delivering remaining data, causing an event informing the application on the other side" and "Abort without delivering remaining data, not causing an event informing the application on the other side":
 this is offered by the `Abort` action without promising that this is signaled to the other side. If it is, a `ConnectionError` Event will fire at the peer ({{termination}}).
 
-* "Reliably transfer data, with congestion control", "Reliably transfer a message, with congestion control" and "Unreliably transfer a message":
+* "Reliably transfer data, with congestion control", "Reliably transfer a message, with congestion control" and "Transfer a message without reliability guarantees":
```
mwelzl, line 3319:
```suggestion
* "Reliably transfer data, with congestion control", "Reliably transfer a message, with congestion control" and "Unreliably transfer a message":
```

Like it or not, but this is a reference to a specific term from RFC 8923, and should therefore be kept as it is. The point of this appendix is to enable back-tracking to the original spec, via 8923 => 8303 or 8304.

mwelzl (changes requested):
LGTM except for one detail (I made a suggestion)

PR #838, opened 2021-05-19
Local/Remote Endpoints

### LINKED ISSUE #755: This use of “remote” on its own offends me...
The use of “remote” on its own offends me:
“Some types of remote might …”
could we make this “Remote Endpoint” to ease my reading? or whatever else is intended?

Similarly can we change “locals” and “remote” here to something that is defined:
“This can expose some or all of the candidate locals to the remote”

gorryfair:
Fix #755

### DIFF
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -292,7 +292,8 @@ Architecture {{I-D.ietf-taps-arch}}.
 An application primarily interacts with this API through two Objects:
 Preconnections and Connections. A Preconnection represents a set of properties
 and constraints on the selection and configuration of paths and protocols to
-establish a Connection with a Remote Endpoint. A Connection represents an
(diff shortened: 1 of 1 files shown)


draft-ietf-taps-interface.md
```diff
@@ -292,7 +292,8 @@ Architecture {{I-D.ietf-taps-arch}}.
 An application primarily interacts with this API through two Objects:
 Preconnections and Connections. A Preconnection represents a set of properties
 and constraints on the selection and configuration of paths and protocols to
-establish a Connection with a Remote Endpoint. A Connection represents an
+establish a Connection with a
```
mwelzl, line 295:
```suggestion
establish a Connection with an 
```

mwelzl (changes requested):
just a nit, else LGTM

PR #845, opened 2021-06-11
Update names full checksum conn. properties

abrunstrom:


gorryfair (commented):
I do not mind what the name is called as long as it is clear. I think the proposal looks good.

Issue #846, opened 2021-06-11
unified interface to datagram and connection-oriented transports


csperkins:
Interface, Section 2, "A unified interface to datagram and connection-oriented transports, allowing use of a common API for connection-establishment and closing". Should this be "…to datagram and stream-oriented transports"?

At what point do we explain that `Connection` and `Preconnection` are also used for connectionless protocols? This change would finesse the question here, leaving it for Section 3, or we can restructure the draft to clarify earlier.

gorryfair:
Yes to (1): Datagram protocols can be connection-oriented or not; however the point here seems to be either:
/interface to datagram and stream-oriented transports/ or
/interface to both connection-less and connection-oriented transports/
... I'd favour the first, it's clear enough and doesn't use terms that we later re-use.

Also to (2): I think we could insert a specific sentence after: " A Connection represents an instance of a transport Protocol Stack on which data can be sent to and/or received from a Remote Endpoint (i.e., depending on the kind of transport, connections can be bi-directional or unidirectional). " 
That explicitly says something like:
"This connections are presently consistently to the application, irrespective of whether the underlying transport is connection-less and connection-oriented."

PR #847, opened 2021-06-11
Clarifications around Connections

csperkins:
Tries to clarify that a Connection is a logical connection, and doesn't necessarily require a connection oriented underlying transport protocol. 

Cross-reference to the descriptions of Preconnection and the connection establishment procedures.


gorryfair (commented):
Looks good.
Can we also change to something like this in section 1: /A unified interface to datagram and connection-oriented transport protocols, allowing use of a common API for connection-establishment and closing;/

csperkins:
That's #846

PR #853, opened 2021-06-11
Clarify Setting Message Properties on Preconnections and Connections

### LINKED ISSUE #808: Be more clear about Setting Message Properties on Preconnections and Connections
In 3.2 it is clear that you can set Message Properties on Connections and Preconnections, but the rest of the document is not very clear about this.

In 4.2 for instance the text only talks about setting Selection Properties and Connection Properties on PreConnections. 

In 7.1.3 the text says "Connection Properties describe the default behavior for all Messages on a Connection". But I think that Message Properties set on a Connection or PreConnection may also describe the default behavior so this relationship should be clarified. Also, the Message Property defaults given for Properties that do not depend on other Properties sets the defaults for all Messages in a Connection I assume? I also assume that a Message Property set on a Message Context overrides a Message Property set on a PreConnection or Connection, but this is also not specified.

In 5.4. Connection Groups, I assume that setting Message Properties on a Connection applies to all Connections in the Connection Group? This is not clear.

philsbln:
closes #808

### DIFF
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -1647,7 +1647,7 @@ to the new Connection when calling Clone(), but in this case, a later change to 
 `Connection Priority` on one Connection does not change it on the
 other Connections in the same Connection Group.
 
-Message Properties set on a Connection are also not entangled. For example,
+Message Properties set on a Connection also apply only to that Connection. For example,
 changing `Lifetime` (see {{msg-lifetime}}) of a Message will only affect a
 single Message on a single Connection.
 
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -1647,9 +1647,7 @@ to the new Connection when calling Clone(), but in this case, a later change to 
 `Connection Priority` on one Connection does not change it on the
 other Connections in the same Connection Group.
 
-Message Properties set on a Connection also apply only to that Connection. For example,
-changing `Lifetime` (see {{msg-lifetime}}) of a Message will only affect a
-single Message on a single Connection.
+Message Properties set on a Connection also apply only to that Connection.
 
 A new Connection created by Clone can have a Message Framer assigned via the optional
 `framer` parameter of the Clone Action. If this parameter is not supplied, the
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -1647,9 +1647,7 @@ to the new Connection when calling Clone(), but in this case, a later change to 
 `Connection Priority` on one Connection does not change it on the
 other Connections in the same Connection Group.
 
-Message Properties set on a Connection are also not entangled. Chaning these on an individual Connection
-which is part of a Connections Group SHOULD have the same effect as setting the Message property on each
-Message sent though that Connection. For example,
+Message Properties set on a Connection are also not entangled. For example,
 changing `Lifetime` (see {{msg-lifetime}}) of a Message will only affect a
 single Message on a single Connection.
 
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -821,8 +821,8 @@ RemoteSpecifier.WithIPv4Address(192.0.2.22)
 A Preconnection Object holds properties reflecting the application's
 requirements and preferences for the transport. These include Selection
 Properties for selecting protocol stacks and paths, as well as Connection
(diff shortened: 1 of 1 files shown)


draft-ietf-taps-interface.md
```diff
@@ -821,8 +821,8 @@ RemoteSpecifier.WithIPv4Address(192.0.2.22)
 A Preconnection Object holds properties reflecting the application's
 requirements and preferences for the transport. These include Selection
 Properties for selecting protocol stacks and paths, as well as Connection
-Properties for configuration of the detailed operation of the selected Protocol
-Stacks.
+Properties and Message Properties for configuration of the detailed operation
+of the selected Protocol Stacks on a per connection and message level.
```
britram, line 825: Might be nitpicking here, but -- the Message Properties at this time are not configuring the Protocol Stack, they're just setting common defaults for Send, correct?

draft-ietf-taps-interface.md
```diff
@@ -1624,7 +1624,7 @@ Preconnection -> EstablishmentError<reason?>
 
 ## Connection Groups {#groups}
 
-Connection Groups can be created using the Clone Action:
+Connection Groups represent sets of entagled connections, e.g., streams in a multi-streaming protocol, and can be created using the Clone Action:
```
britram, line 1627: This change seems unrelated to the main change (and we've gone back and forth here); propose removing it.

britram (changes requested):

draft-ietf-taps-interface.md
philsbln: Correct, but there is no implication when the configuration will actually take effect.

draft-ietf-taps-interface.md
philsbln: Reading it without this hint felt a little confusing to me after not reading it a long time. I am happy to remove it, but would prefer third opinion here.

draft-ietf-taps-interface.md
mwelzl:
```suggestion
of the selected Protocol Stacks on a per-Connection and Message level.
```

I agree that @britram is nitpicking  :-)   ... because I think that pointing out that Message Properties only influence the send call will only make this harder to read, at no real benefit.

draft-ietf-taps-interface.md
mwelzl: I agree to remove it. The text below the Clone call explains it all, and it was revised not too long ago in a PR, so this sits with agreement behind it.

A side note, we no longer talk about Connections being "entangled": the term only applies to their Connection Properties.  Also, using the term here before it's explained isn't good... at best, we could use the "streams in a multi-streaming protocol" example, but if you're missing this, it also fits better as an addition to the text below the call.

draft-ietf-taps-interface.md
```diff
@@ -1647,7 +1647,9 @@ to the new Connection when calling Clone(), but in this case, a later change to
 `Connection Priority` on one Connection does not change it on the
 other Connections in the same Connection Group.
 
-Message Properties are also not entangled.  For example,
+Message Properties set on a Connection are also not entangled. Chaning these on an individual Connection
+which is part of a Connections Group SHOULD have the same effect as setting the Message property on each
+Message sent though that Connection. For example,
```
mwelzl, line 1652: The sentence that you included here is only making things more confusing, IMO. Suggest to remove it.

mwelzl (changes requested):
I also wonder if this really does all that's requested in #808...

draft-ietf-taps-interface.md
abrunstrom: Agree to remove.

draft-ietf-taps-interface.md
abrunstrom: I think a first modified sentence is sufficient "Message Properties set on a Connection also apply only to that Connection". Actually reading it again now I realize that it is the example that confuses things, because that talks about setting a Message Property on a Message, so maybe we can just remove the example as it does not belong in this section.

draft-ietf-taps-interface.md
philsbln: Removed it

draft-ietf-taps-interface.md
philsbln: applied

philsbln:
Applied requested changes

mwelzl (approved):
Thanks!!

britram (approved):

PR #857, opened 2021-06-11
Fixes #812

### LINKED ISSUE #812: S4.2.14. Multipath Transport - why different?
I know we have discussed Multipath Transport many times, but when reading is seems strange when you get to Multipath Transport and here there is no option to "Require" this property, only to "Prefer" it. We should at least say why. It is perhaps intuitive that you do not need to distinguish between Avoid and Prohibit, but not why Require would not be useful.

philsbln:
closes #812


mwelzl (approved):
This reflects what we agreed on, during the last interim.

Issue #859, opened 2021-06-11
When are objects frozen?


martinduke:
Discussion at the 11 Jun 2021 Interim:

The content of any objects passed as part of NewPreconnection(), Initiate(), Listen(), or Rendezvous() is frozen at the moment of Initiate/Listen/Rendezvous.

Thus, objects sent to NewPreconnection (Endpoints and Properties) are mutable, and it might be possible (depending on the implementation) to change them in the preconnection by modifying the object directly. Applications should either
(1) Instantiate the connection before modifying Endpoint or Property objects for another preconnection.
(2) If there are open preconnections that will initiate further connections, create new instances of Properties and Endpoints to modify their attributes in a different preconnection.

There are other ways to do this. For example, NewPreconnection() could do a deep copy of the properties and endpoints, and therefore reusing or freeing the argument objects would always be safe. But this was not the sense of the group at the interim.

While this issue is uncomfortably close to the "implementation dependent", not having fixed assumptions about this increasing the friction of switching platforms using TAPS, and would probably be the source of nasty bugs.

martinduke:
Reference: (https://mailarchive.ietf.org/arch/msg/taps/ZHCc4sLCXb2984hwob_F5aiZXsY/)

theri:
Comment by @MaxF12 in the interim chat:

Also in section 2 in the impl draft it says "Preconnection objects should be implemented as bundles of properties that an application can both read and write. Once a Preconnection has been used to create an outbound Connection or a Listener, the implementation should ensure that the copy of the properties held by the Connection or Listener is immutable." which I assume means that the properties the bundle is made of can also be overwritten

martinduke:
to be clear, IMO the draft is quite clear about freezing at Initiate(), but it is not clear about NOT freezing at NewPreconnection().

gorryfair:
At Interim, we heard that this might be fixed by emphasising in the preconception text  that everything is frozen (aka deep-copied) at Initiate.

martinduke:
I will also point out that this sits somewhat uneasily with Sec 3.1 of taps-impl, where we should fail the preconnection immediately if nothing matches the desired properties. What if I change the desired properties after getting the preconnection?

It's not insurmountable, but it's conceptually not very clean.

mwelzl:
[...]

PR #860, opened 2021-06-14
Freezing pre-connection state.

gorryfair:
To clarify when and how objects are froze.  #859

Arch: No current change needed
API: Clarify 
Impl: Explain how


draft-ietf-taps-impl.md
```diff
@@ -146,10 +146,13 @@ The connection objects that are exposed to applications for Transport Services a
 - the Connection, the basic object that represents a flow of data as Messages in either direction between the Local and Remote Endpoints;
 - and the Listener, a passive waiting object that delivers new Connections.
 
-Preconnection objects should be implemented as bundles of properties that an application can both read and write. Once a Preconnection has been used to create an outbound Connection or a Listener, the implementation should ensure that the copy of the properties held by the Connection or Listener is immutable. This may involve performing a deep-copy, copying the object with all the objects it references, if the application is still able to modify properties on the original Preconnection object.
+Preconnection objects should be implemented as bundles of properties that an application can both read and write. An implementation could allow properties of a preconnection to be updated, but changes to the Preconnection Selection and Connection may be language dependent and is to be avoided in code that could be used across different implementations.
```
csperkins, line 150: The original seems clearer to me: it's more explicit about what happens, and when.

gorryfair: Oh dear we don't seem to agree - I think the original hints at everything being immutable. Instead, I think everything should be copied; but only the selection params are immutable, the other params are mutable on the copy.

csperkins: That would be fine, but I didn't get that meaning from the new text.

gorryfair: So... what's missing? - The text on copying was moved to the section after "once a connection is established.

csperkins: I can't parse "An implementation could allow properties of a preconnection to be updated, but changes to the Preconnection Selection and Connection may be language dependent and is to be avoided in code that could be used across different implementations". I read it as trying to simplify the original wording, but actually making it less clear to me. I don't see how it relates to the selection parameters being immutable and while the other parameters remain mutable.

gorryfair: Improved text

draft-ietf-taps-impl.md
```diff
@@ -146,10 +146,13 @@ The connection objects that are exposed to applications for Transport Services a
 - the Connection, the basic object that represents a flow of data as Messages in either direction between the Local and Remote Endpoints;
 - and the Listener, a passive waiting object that delivers new Connections.
 
-Preconnection objects should be implemented as bundles of properties that an application can both read and write. Once a Preconnection has been used to create an outbound Connection or a Listener, the implementation should ensure that the copy of the properties held by the Connection or Listener is immutable. This may involve performing a deep-copy, copying the object with all the objects it references, if the application is still able to modify properties on the original Preconnection object.
+Preconnection objects should be implemented as bundles of properties that an application can both read and write. An implementation could allow properties of a preconnection to be updated. However, an application should avoid changes to the Preconnection Selection and Connection to improve consistent operation across different implementations.
```
martinduke, line 149:
```suggestion
Preconnection objects should be implemented as bundles of properties that an application can both read and write. An implementation could allow applications to update the properties or endpoints of a preconnection after it is initiated. However, an application should avoid changes to the Preconnection properties and endpoints to improve consistent operation across different implementations.
```

martinduke (approved):
I found the first change a bit unclear. Here's proposed rewording:

martinduke:
LGTM.

draft-ietf-taps-impl.md
csperkins: The last two sentences here still seem vague and contradictory with the below. I think it'd be clearer just to remove them, and leave the detailed following explanations.

gorryfair: @martinduke :  The expected operation is now clearer, csp suggests simply not mentioning the "update during pre-connection" possibility. What do you think?

mwelzl: I have to agree with @csperkins that the original was clearer, mainly for being more explicit - but I think that, after looking at this for some time, I now see where the problem is (which @csperkins is perhaps missing, and which @gorryfair is perhaps trying to fix).

I suspect that @gorryfair wants to avoid that we say, here, that properties on a Connection cannot be changed after creating an outbound Connection or a Listener. This would, of course, be wrong, but the text says that they are all "immutable", which gives a wrong indication.

Then again, the proposed text now doesn't look like a good fix to me. E.g., it says "an implementation could allow applications to...", followed by "however, applications should not do this". This is awkward. Maybe you meant to write "Connection", not "preconnection" in the  second sentence?

Here's a proposal: a change to the original text that avoids falsely indicating that everything becomes immutable:

***
Preconnection objects should be implemented as bundles of properties that an application can both read and write. A Preconnection object influences a Connection only at one point in time: when the Connection is created. Once a Preconnection has been used to create an outbound Connection or a Listener, the implementation should ensure that the copy of the properties held by the Connection or Listener is not affected when the application makes changes to the Preconnection object. This may involve performing a deep-copy, copying the object with all the objects it references.
***

I also removed the trailing "if the application is still able..." because I don't think this added clarity.

draft-ietf-taps-impl.md
```diff
@@ -146,10 +146,12 @@ The connection objects that are exposed to applications for Transport Services a
...
+Connection objects represent the interface between the application and the implementation to manage transport state, and conduct data transfer. During the process of establishment ({{conn-establish}}), the Connection will not be bound to a specific transport flow, since multiple candidate Protocol Stacks might be raced. 
 
+Once the Connection is established, its interface maps actions and events to the details of the chosen Protocol Stack. For example, the same Connection object may ultimately interface to a single instance of one transport protocol (e.g., a TCP connection, a TLS session over TCP, a UDP flow with fully-specified Local and Remote Endpoints, a DTLS session, a SCTP stream, a QUIC stream, or an HTTP/2 stream).
```
mwelzl, line 153: Here, the original text said "may ultimately represent the interface into", and the proposed text now says "may ultimately interface to"...  all this reads a bit clumsy to me.

Could we not just say "may ultimately represent a single instance ......"
Maybe it's a bit vague, that's in the nature of the word "represent", but I think the meaning is clear enough and this would read better.

draft-ietf-taps-impl.md
```diff
@@ -146,10 +146,12 @@ The connection objects that are exposed to applications for Transport Services a
...
+The implementation should ensure that the copy of all Properties held by a Connection or Listener is independent of other connections. This may involve performing a deep-copy, copying the object with all the objects it references. Once Initate has been called, the Selection Properties and Endpoint information are immutable (i.e, an application is not able to later modify Selection Properties on the original Preconnection object).
```
mwelzl, line 154: I see that this adds the explicit parts from the original text. I like it for being clear about what's mutable and what isn't, and I still think that the whole thing would be best with my suggested text change above - this here just adds further clarification.

The first sentence of this paragraph says "... is independent of other connections."  I guess this should really say ""... is independent of any other connections that are not part of the same Connection Group."

draft-ietf-taps-interface.md
```diff
@@ -494,7 +494,8 @@ the lifetime of a connection using Transport Properties, as defined in
 
 Transport Properties are divided into Selection, Connection, and Message
 Properties. Selection Properties (see {{selection-props}}) can only be set during pre-establishment. They are only used to specify which paths and protocol stacks can be used and are preferred by the application. 
-Although Connection Properties (see {{connection-props}}) can be set during pre-establishment, they may be changed later. They are used to inform decisions made during establishment and to fine-tune the established connection.  
+Although Connection Properties (see {{connection-props}}) can be set during pre-establishment, they may be changed later. They are used to inform decisions made during establishment and to fine-tune the established connection. Calling Initiate on a Preconnection creates an outbound Connection or a Listener, the Selection Properties remain readable from the Connection or Listener, but become immutable.
```
mwelzl, line 497:
```suggestion
Although Connection Properties (see {{connection-props}}) can be set during pre-establishment, they may be changed later. They are used to inform decisions made during establishment and to fine-tune the established connection. Calling Initiate on a Preconnection creates an outbound Connection or a Listener, and the Selection Properties remain readable from the Connection or Listener, but become immutable. 
```

mwelzl (changes requested):
This is going in a good direction, I like it!

Can I make a request to change one more thing as a part of this PR, because it really belongs here?

This sentence in section 3.1 of the implementation draft is broken:

"To avoid allocating resources, it is important that such cases fail as early as possible, e.g., prior to endpoint resolution, only to find out later that there is no protocol that satisfies the requirements."

Plese remove the "only to find out later..." part from it.

draft-ietf-taps-impl.md
gorryfair: Could is /use/ better than /represent/?:  
(1) For example, the same Connection object may ultimately use a single instance of one transport protocol

draft-ietf-taps-impl.md
gorryfair: I tried, and merged. Please check the new diff and suggest improvements.

draft-ietf-taps-impl.md
gorryfair: I think this was done.

draft-ietf-taps-impl.md
mwelzl: FWIW, I agree with this update. Thanks!

mwelzl (approved):

Issue #821, opened 2021-05-01
B.2.3. unreliable-datagram profile


abrunstrom:
Why are the settings for reliability, preserveOrder, congestionControl Ignore rather than Avoid? The current profile will match SCTP equally well as UDP.

gorryfair:
To me, there are multiple ways a DG service cabn be used. I think there are at least two distinct cases here:
* A datagram usage that WANTS full control, e.g. because something like RTP is layered above that handles these things, and any lower layer additions might interfere with these mechanisms.
* A datagram usage that PREFS no hinderance, but does not mind how these are handled as long as the packets get through - i.e. connects.

gorryfair:
I'm OK with:
reliability | avoid
preserveOrder | avoid
Can't this be ignore?:
congestionControl |  Ignore

gorryfair:
Merged

PR #861, opened 2021-06-14
Change B.2.3. unreliable-datagram profile

gorryfair:
Update and explain choices
Seeks to address #821


mwelzl (approved):
I like the way this is phrased, it gives a good argument for the default property values.

gorryfair:
Merged copy after Interim

PR #862, opened 2021-06-14
Help UDP fit wrt connection

gorryfair:
Added a couple of sentences to Arch that are hoped to early introduce that connections are things in the API, not aretfacts of the transport protocol instances.

Different to, but related to an issue for the description in API.


csperkins:
"A `Connection` is an abstraction that represents the communication. If the transport services interface selects a protocol such as TCP for the communication, a `Connection` will correspond to an underlying protocol connection. In the other hand, if a protocol such as UDP is selected, the `Connection` remains an abstraction in the end points and doesn't correspond to explicit protocol state" ?

gorryfair:
@csperkins : Could you live with the update in this PR?

csperkins:
yes, I think so (assuming I'm reading the diff right)

PR #866, opened 2021-07-05
Preconnections are just structs

### LINKED ISSUE #859: When are objects frozen?

mwelzl:
Closes #859

This is quite minimal... meant as a start. Or is this already enough?

### DIFF
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -436,7 +436,7 @@ Connection.Close()
 Connection2.Close()
 ~~~
 
-Preconnections are reusable. Hence, for example, after the Connections were closed,
+Preconnections are reusable after being used to initiate a Connection. Hence, for example, after the Connections were closed,
 the following would be correct:
 ~~~
 //.. carry out adjustments to the Preconnection, if desire
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -436,6 +436,14 @@ Connection.Close()
 Connection2.Close()
 ~~~
 
+Preconnections are reusable. Hence, for example, after the Connections were closed,
+the following would be correct:
+~~~
+//.. carry out adjustments to the Preconnection, if desire
+Connection := Preconnection.Initiate()
+~~~
+
+
 ### Peer Example
 
 This is an example of how an application might establish a connection with a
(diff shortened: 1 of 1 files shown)


draft-ietf-taps-interface.md
```diff
@@ -436,6 +436,14 @@ Connection.Close()
 Connection2.Close()
 ~~~
 
+Preconnections are reusable. Hence, for example, after the Connections were closed,
+the following would be correct:
```
gorryfair, line 440: After Initiated wouldn't this also be OK?

mwelzl: Of course - anywhere, really, it's just a struct. But this gets a bit harder to write about ("a similar line would also work below the 'initiate' call higher above"), unless we use more code just to show this; I think, with the explanation below, it should be clear enough as it is.

tfpauly (approved):
I'm good with this!

draft-ietf-taps-interface.md
gorryfair: Could we try: 
"Preconnections are reusable after being used to initiate a Connection. Hence, for example, after the Connections were closed,
 the following would be correct:"

mwelzl: Your wish has been granted.

gorryfair (approved):
Looks good to me...

PR #867, opened 2021-07-08
Many editorial fixes (rollup 1)

### LINKED ISSUE #798: S1.1: Definition of Tuple refers to property before it is defined
S1.1, Definition of Tuple refers to property before it is defined

### LINKED ISSUE #799: S2. Very long bullets are hard to read
The list of what the interface provides is in the form of one sentence per bullet. That makes some sentences very long with many commas. I find this hard to read, particularly the first and fourth bullet. Maybe allow multiple sentences per bullet for some of them?

### LINKED ISSUE #802: S3.2: Reference to single namespace is unclear
Introductory text says "All Transport Properties, regardless of the phase in which they are used, are organized within a single namespace." But later in the text multiple name spaces are mentioned. Even if they are not related to the phase, this does not seem consistent.

### LINKED ISSUE #803: S3.2.2: Enumeration is a basic type so the two bullets do not read correctly

### LINKED ISSUE #805: S4.2 Use of recommended
Text that discusses the recommended default values for Selection Properties uses the word "recommended" in multiple places. It is unclear to me if any of them should be "RECOMMENDED". Otherwise using another word may make it more clear.

### LINKED ISSUE #807: S4.2.3. What makes Configure Per-Message Reliability special?
For this property we state. "This property applies to Connections and Connection Groups." But I think this is true for all Selection Properties, so it read strangely to me to point it out for only one property. Particularly the part about the Connection Groups. It makes you wonder if the other properties do not apply to Connection Groups. I guess it is here because the Property relates to messages, but maybe the text can be reformulated and the part about Connection Groups can be removed.

Maybe something like "This property specifies whether an application considers it useful to indicate its reliability requirements on a per-Message basis for the Messages within the Connection." or "This property specifies whether an application considers it useful to specify different reliability requirements for individual Messages in a Connection.

### LINKED ISSUE #811: S4.2.11. Interface Instance or Type - text clarification needed
I can not parse the "to all remote systems" part of the following sentence.

> An implementation should provide all types that are supported on the local system to all remote systems, to allow applications to be written generically.

### LINKED ISSUE #813: 6. Managing Connections - overlapping bullets
I do not understand the relationship between bullets four and five. They seem to overlap?

### LINKED ISSUE #814: S7.1.3.3. Ordered - Default seems to point to wrong Selection Property

### LINKED ISSUE #816: S7.3.2.2. ReceivedPartial Example
I assume the example can only happen if preserveOrder is not required? As Require is the default for preserveOrder this should be made clear in the example.

### LINKED ISSUE #817: S7.3.3.1. UDP(-Lite)-specific Property: ECN - Motivation
I think this protocol specific Property is only included because it comes from RFC8923? Should be made clear in the text.

britram:
fix #798, fix #799, fix #802, fix #803, fix #805, fix #807, fix #811, fix #813, fix #814, fix #816, fix #817.

More to follow in a subsequent PR (it's lunchtime. :) )

### DIFF
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -274,7 +274,7 @@ provides:
   
 - Selection between alternate network paths, also allowing applications to be
   notified additional information is available about the networks over which 
-  an connetcion can operate (e.g. Provisioning Domain (PvD) information {{RFC7556}}).  
+  a connection can operate (e.g. Provisioning Domain (PvD) information {{RFC7556}}).  
 
 - Explicit support for transport-specific features to be applied, should that
   particular transport be part of a chosen Protocol Stack.
@@ -548,7 +548,7 @@ for Protocol Specific Properties and MUST NOT be used for vendor or implementati
 
 ### Transport Property Types {#property-types}
 
-Each Transport Property has a one ot the basic types described in {{notation}}.
+Each Transport Property has a one of the basic types described in {{notation}}.
 
 Most Selection Properties (see {{selection-props}}) are of the Enumeration type, 
 and use the Preference Enumeration, which takes one of five possible values
@@ -919,7 +919,7 @@ and, as requested, the Connection is not congestion controlled, querying
 the `congestionControl` property also yields the value `false`.
 
 An implementation of this interface must provide sensible defaults for Selection
-Properties. The suggested default values for each property below represent a
+Properties. The default values for each property below represent a
 configuration that can be implemented over TCP. If these default values are used
 and TCP is not supported by a Transport Services implementation, then an application using the
 default set of Properties might not succeed in establishing a connection. Using
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -272,9 +272,9 @@ provides:
   application interactions with the transport layer, in line with developments
   in modern platforms and programming languages;
   
-- Selection between alternate network paths, also allowing applications to be
-  notified additional information is available about the networks over which 
-  a connection can operate (e.g. Provisioning Domain (PvD) information {{RFC7556}}).  
+- Selection between alternate network paths, using additional information about the 
+  networks over which a connection can operate (e.g. Provisioning Domain (PvD) 
+  information {{RFC7556}}) where available.  
 
 - Explicit support for transport-specific features to be applied, should that
   particular transport be part of a chosen Protocol Stack.
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -227,8 +227,7 @@ We also make use of the following basic types:
 - Tuple: An ordered grouping of multiple value types, represented as a
   comma-separated list in parentheses, e.g., ```(Enumeration, Preference)```.
   Instances take a sequence of values each valid for the corresponding value
-  type. The composition of types and their order depends on the property and is
-  fixed for the property.
+  type. 
 - Array: Denoted []Type, an instance takes a value for each of zero or more
   elements in a sequence of the given Type. An array may be of fixed or
   variable length.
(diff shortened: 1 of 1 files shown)


draft-ietf-taps-interface.md
```diff
@@ -252,36 +251,42 @@ principles, themselves an elaboration on the architectural design principles
...
-  concurrent operations during establishment and supporting event-driven
+- Asynchronous Connection establishment, transmission, and reception. 
+  This allows concurrent operations during establishment and event-driven
   application interactions with the transport layer, in line with developments
   in modern platforms and programming languages;
   
-- Selection between alternate network paths that can be informed when additional information is available about the networks over which an endpoint can operate (e.g. Provisioning Domain (PvD) information {{RFC7556}});  
+- Selection between alternate network paths, also allowing applications to be
+  notified additional information is available about the networks over which 
+  an connetcion can operate (e.g. Provisioning Domain (PvD) information {{RFC7556}}).
```
gorryfair, line 277: /a connection/

draft-ietf-taps-interface.md
```diff
@@ -252,36 +251,42 @@ principles, themselves an elaboration on the architectural design principles
...
-- Asynchronous Connection establishment, transmission, and reception, allowing
-  concurrent operations during establishment and supporting event-driven
+- Asynchronous Connection establishment, transmission, and reception. 
+  This allows concurrent operations during establishment and event-driven
   application interactions with the transport layer, in line with developments
   in modern platforms and programming languages;
   
-- Selection between alternate network paths that can be informed when additional information is available about the networks over which an endpoint can operate (e.g. Provisioning Domain (PvD) information {{RFC7556}});  
+- Selection between alternate network paths, also allowing applications to be
+  notified additional information is available about the networks over which
```
gorryfair, line 276: English messy: However, I intended ... informing the selection ... rather than the other possibility of informing the app.

draft-ietf-taps-interface.md
```diff
@@ -543,14 +548,12 @@ for Protocol Specific Properties and MUST NOT be used for vendor or implementati
 
 ### Transport Property Types {#property-types}
 
-Transport Properties each have a type, which can be:
+Each Transport Property has a one ot the basic types described in {{notation}}.
```
gorryfair, line 551: /ot/to/

draft-ietf-taps-interface.md
```diff
@@ -916,14 +919,15 @@ and, as requested, the Connection is not congestion controlled, querying
 the `congestionControl` property also yields the value `false`.
 
 An implementation of this interface must provide sensible defaults for Selection
-Properties. The recommended default values for each property below represent a
+Properties. The suggested default values for each property below represent a
```
gorryfair, line 922: could simply remove /suggested/  entirely?

gorryfair (commented):
Thanks for getting to this: My lunch-time was offset from yours, so here are a few initial NiTs to help finish this.

draft-ietf-taps-interface.md
britram:
```suggestion
  a connection can operate (e.g. Provisioning Domain (PvD) information {{RFC7556}}).  
```

draft-ietf-taps-interface.md
britram:
```suggestion
Each Transport Property has a one of the basic types described in {{notation}}.
```

draft-ietf-taps-interface.md
britram:
```suggestion
Properties. The default values for each property below represent a
```

draft-ietf-taps-interface.md
britram: (will resolve outside this thread, for some reason having a comment here breaks multiline suggestions)

draft-ietf-taps-interface.md
```diff
@@ -252,36 +251,42 @@ principles, themselves an elaboration on the architectural design principles
...
-  concurrent operations during establishment and supporting event-driven
+- Asynchronous Connection establishment, transmission, and reception. 
+  This allows concurrent operations during establishment and event-driven
   application interactions with the transport layer, in line with developments
   in modern platforms and programming languages;
   
-- Selection between alternate network paths that can be informed when additional information is available about the networks over which an endpoint can operate (e.g. Provisioning Domain (PvD) information {{RFC7556}});  
+- Selection between alternate network paths, also allowing applications to be
+  notified additional information is available about the networks over which 
+  a connection can operate (e.g. Provisioning Domain (PvD) information {{RFC7556}}).
```
britram, line 277:
```suggestion
- Selection between alternate network paths, using additional information about the 
  networks over which a connection can operate (e.g. Provisioning Domain (PvD) 
  information {{RFC7556}}) where available.  
```

PR #871, opened 2021-07-08
Resolve #852: text disclaiming multicast rendezvous

### LINKED ISSUE #852: Using Multicast Endpoints
The interface draft, section 4.1.1, talks about multicast endpoints, but doesn't mention whether such endpoints can be used with Rendezvous().

### LINKED ISSUE #872: Add CloseGroup to Mapping template
The implementation draft defines a template for what mappings should define. This needs to include newer group operations like CloseGroup/AbortGroup

GrumpyOldTroll:


mwelzl (approved):
This works for me; thanks a lot for doing it!

philsbln (approved):
LGMT

draft-ietf-taps-interface.md
```diff
@@ -738,7 +738,12 @@ a Preconnection will only support calling Listen(), not Initiate(). The accepted
 are receive-only.
 
 Similarly, specifying a multicast group address on the Remote Endpoint will indicate that the
-resulting connection will be used to send multicast messages.
+resulting connection will be used to send multicast messages, and that the Preconnection will
+support Initiate() but not Listen(). The Connection created this way is send-only.
+
```
csperkins, line 743: This part is good.

draft-ietf-taps-interface.md
```diff
@@ -738,7 +738,12 @@ a Preconnection will only support calling Listen(), not Initiate(). The accepted
...
+No semantics for creating bidirectional connections for multicast group members or the use of
+Rendezvous() with Preconnections that have endpoints containing group addresses are defined in
+this document.
```
csperkins, line 747: Not sure I agree with this. We need to either prohibit the use of `Rendezvous()` with multicast in this version of the API, or define some semantics for what its use means.

gorryfair: I'm OK just not to do this. I'd be against delaying this until we've thought it through - maybe there'll be more to add in future? so much of TAPS has actually been implemented, I'd not be in favour of adding last-minute things as we finish.

GrumpyOldTroll: Hmm, fair point.  Best if different implementations don't use it in different incompatible ways.

I guess one option is to replace this with something like "A Rendezvous() call on Preconnections containing group addresses results in an EstablishmentError as described in {{rendezvous}}.", and another option is to leave it implicit and remove this sentence.

GrumpyOldTroll: Hmm, fair point.  Best if different implementations don't use it in different incompatible ways.

I guess one option is to replace this with something like "A Rendezvous() call on Preconnections containing group addresses results in an EstablishmentError as described in {{rendezvous}}.", and another option is to leave it implicit and remove this sentence.

mwelzl: It seems to me that "A Rendezvous() call on Preconnections containing group addresses results in an EstablishmentError as described in {{rendezvous}}." addresses what @csperkins says but removing the sentence doesn't. So, this suggested sentence seems to be the better option?

PR #875, opened 2021-08-23
Add more multicast example code

### LINKED ISSUE #874: Add examples for ASM and SSM
In the multicast endpoint section, we should have explicit examples for patterns of ASM and SSM.

Listen(multicast address, port, interface)
Get a connection from a specific remote endpoint
Initiate a connection back to talk to that endpoint,
etc

tfpauly:
Closes #874

### DIFF
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -736,7 +736,7 @@ system that the resulting connection will be used to receive multicast messages.
 Remote Endpoint can be used to filter incoming multicast from specific senders. Such
 a Preconnection will only support calling Listen(), not Initiate(). Calling Listen()
 will cause the Transport Services system to register for receiving multicast, such
-as issuing an IGMP join {{?RFC3376}}. Any Connections that are accepted from this
+as issuing an IGMP join {{?RFC3376}} or using MLDv2 for IPV6 {{?RFC4604}}. Any Connections that are accepted from this
 Listener are receive-only.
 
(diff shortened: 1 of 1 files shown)


draft-ietf-taps-interface.md
```diff
@@ -734,14 +734,19 @@ early binding when required, for example with some Network Address Translator
 Specifying a multicast group address on a Local Endpoint will indicate to the transport
 system that the resulting connection will be used to receive multicast messages. The
 Remote Endpoint can be used to filter incoming multicast from specific senders. Such
-a Preconnection will only support calling Listen(), not Initiate(). The accepted Connections
-are receive-only.
+a Preconnection will only support calling Listen(), not Initiate(). Calling Listen()
+will cause the Transport Services system to register for receiving multicast, such
+as issuing an IGMP join {{?RFC3376}}. Any Connections that are accepted from this
```
gorryfair, line 739: or MLDv2 for IPv6.

tfpauly:
```suggestion
as issuing an IGMP join {{?RFC3376}} or using MLDv2 for IPV6 {{?RFC4604}}. Any Connections that are accepted from this
```

csperkins:
I realise I'm late reviewing this, but this approach seems pretty cumbersome. Why do we need two connections, once send only and one receive only, rather than a single bidirectional connection? 

It also seems cumbersome to implement: both `Connection` objects will be bound to the same port, and so will receive the multicast packets once the group has been joined. The send-only `Connection` must then filter out those received packets?

PR #877, opened 2021-09-17
Mullticast reuse draft-ietf-taps-impl.md

### LINKED ISSUE #876: Add implementation text for multicast
The implementation document should explain IGMP joining and SO_REUSEADDR for multicast

gorryfair:
This adds SO_reuse commentry on the multicast UDP service. It seeks to fix #876

### DIFF
--- a/draft-ietf-taps-impl.md
+++ b/draft-ietf-taps-impl.md
@@ -1024,6 +1024,13 @@ ConnectionError:
 Listen:
 : LISTEN.UDP. Calling `Listen` for UDP Multicast Receive binds a local port, prepares it to receive inbound UDP datagrams from peers, and issues a multicast host join.  If a Remote Endpoint with an address is supplied, the join is Source-specific Multicast, and the path selection is based on the route to the Remote Endpoint.  If a Remote Endpoint is not supplied, the join is Any-source Multicast, and the path selection is based on the outbound route to the group supplied in the Local Endpoint.
 
+There are cases where it is required to open multiple connections for the same address(es). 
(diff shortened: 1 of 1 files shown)


draft-ietf-taps-impl.md
```diff
@@ -1024,6 +1024,13 @@ ConnectionError:
 Listen:
 : LISTEN.UDP. Calling `Listen` for UDP Multicast Receive binds a local port, prepares it to receive inbound UDP datagrams from peers, and issues a multicast host join.  If a Remote Endpoint with an address is supplied, the join is Source-specific Multicast, and the path selection is based on the route to the Remote Endpoint.  If a Remote Endpoint is not supplied, the join is Any-source Multicast, and the path selection is based on the outbound route to the group supplied in the Local Endpoint.
 
+There are cases where it is required to open multiple connections for the same address(es). 
+For example, one Connection might be opened to listen to a multicast group, and later a separate one to send to the same group; 
+two applications might independently listen to the same address(es) to send signals to and/or receive signals from a 
+common multicast control bus. 
+In such cases, the interface needs to explicitly enable re-use of the same set of addresses (equivalent to setting SO_REUSEADDR 
+in the socket API).
+
```
csperkins, line 1033: This is conflating two issues. 

If two applications want to listen to the same multicast group, then they both need sockets that are bound to that group address and having called `setsockopt(… IP_ADD_MEMBERSHIP…)`. They need to set `SO_REUSEADDR` and `SO_REUSEPORT` to allow them to both bind to the same address. 

For an application to send to and receive from a single group, you only need a single socket and so don't need `SO_REUSEADDR`

PR #878, opened 2021-09-17
"connected" doesn't mean send a packet to draft-ietf-taps-impl.md

gorryfair:
Added something directly from RFC8085 that UDP can be connected in a sense.

### DIFF
--- a/draft-ietf-taps-impl.md
+++ b/draft-ietf-taps-impl.md
@@ -136,7 +136,10 @@ event-driven interaction pattern. This document serves as a guide to implementat
 
 The Transport Services architecture {{I-D.ietf-taps-arch}} defines a system that allows applications to use transport networking protocols flexibly. The interface such a system exposes to applications is defined as the Transport Services API {{I-D.ietf-taps-interface}}. This API is designed to be generic across multiple transport protocols and sets of protocols features.
 
-This document serves as a guide to implementation on how to build a system that provides a Transport Services API. It is the job of an implementation of a Transport Services system to turn the requests of an application into decisions on how to establish connections, and how to transfer data over those connections once established. The terminology used in this document is based on the Architecture {{I-D.ietf-taps-arch}}.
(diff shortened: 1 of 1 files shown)


draft-ietf-taps-impl.md
```diff
@@ -136,7 +136,10 @@ event-driven interaction pattern. This document serves as a guide to implementat
...
 
-This document serves as a guide to implementation on how to build a system that provides a Transport Services API. It is the job of an implementation of a Transport Services system to turn the requests of an application into decisions on how to establish connections, and how to transfer data over those connections once established. The terminology used in this document is based on the Architecture {{I-D.ietf-taps-arch}}.
+This document serves as a guide to implementation on how to build a system that provides a Transport Services API. It is the job of an implementation of a Transport Services system to turn the requests of an application into decisions on how to establish connections, and how to transfer data over those connections once established. 
+Note that many operating systems also allow a UDP(-Lite) socket to be connected, i.e., to bind a UDP socket to a specific pair of addresses and ports. This is similar to the corresponding TCP sockets API functionality. However, for UDP(-Lite), this is only a local operation that serves to simplify the local send/receive functions and to filter the traffic for the specified addresses and ports {{?RFC8085}}.
```
tfpauly, line 140: I don't see why this paragraph belongs here in the introduction.

Did you mean to put this in Section 2, about connection objects?

It seems more related to this:
```
For example, the same Connection object may ultimately represent a single instance 
of one transport protocol (e.g., a TCP connection, a TLS session over TCP, a UDP flow
with fully-specified Local and Remote Endpoints, a DTLS session, a SCTP stream, a
QUIC stream, or an HTTP/2 stream).
```

tfpauly (changes requested):

draft-ietf-taps-impl.md
gorryfair: OK, so you are so correct - this is just a clarification of section 2, where I'll propose a simple change instead.

Issue #891, opened 2021-09-19
ART-ART: Tidy section 4.1.2 arch


gorryfair:
The big paragraph that is the first bullet of section 4.1.2 doesn't read well.
Pleas consider restructuring it, and simplifying the sentences as much as
possible.

PR #893, opened 2021-09-19
attempt to improve preconnection draft-ietf-taps-arch.md

gorryfair:
Gorry's attempt to improve preconnection (#891)


mwelzl (approved):
I think this reads better

tfpauly (approved):

PR #921, opened 2021-09-21
Add a ReceivedPartial example

### LINKED ISSUE #791: be more concrete about receiving
Reading 7.3.2 I found the concepts there quite abstract. I think it would really help to be a bit more specific and e.g. take UDP and TCP as examples basically saying some like: as UDP is consider a message based protocol, each received datagram will trigger receive() of a complete message. As TCP is stream-based all TCP segments will be delivered by ReceivedPartial().

tfpauly:
Closes  #791


philsbln (approved):
LGMT

PR #931, opened 2021-10-15
Add a Connection to a group at any time, also on the passive side

mwelzl:
This also updates a reference in the API and implementation drafts.


abrunstrom (commented):
Text looks god to me. We may want to discuss at the interim to agree on the added functionality. From a research perspective it is nice to have.

PR #934, opened 2021-10-15
Explicit context in example

### LINKED ISSUE #901: ART-ART: Explicit context in the Send in example in 3.1.1
Consider showing an explicit context in the Send in the example in 3.1.1 - it
would make the reader guess less to wait to introduce implicit contexts later.

mwelzl:
Resolves #901.

Is this really what was being asked for here?  Just to create a MessageContext object? Or am I missing something?

### DIFF
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -389,6 +389,7 @@ Listener -> ConnectionReceived<Connection>
 // Only receive complete messages in a Conn.Received handler
 Connection.Receive()
 
+messageContext := NewMessageContext()
 Connection -> Received<messageDataRequest, messageContext>
 
 //---- Receive event handler begin ----
@@ -440,6 +441,7 @@ C.Send(messageDataRequest)
 // Only receive complete messages
 C.Receive()
 //---- Ready event handler for any Connection C end ----
+messageContext := NewMessageContext()
 
 Connection -> Received<messageDataResponse, messageContext>
 Connection2 -> Received<messageDataResponse, messageContext>
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -393,7 +393,9 @@ messageContext := NewMessageContext()
 Connection -> Received<messageDataRequest, messageContext>
 
 //---- Receive event handler begin ----
-Connection.Send(messageDataResponse)
+messageContext.add(msgPrio, 200)
+messageContext.add(final, true)
+Connection.Send(messageDataResponse, messageContext)
 Connection.Close()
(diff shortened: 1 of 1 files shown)


draft-ietf-taps-interface.md
```diff
@@ -440,6 +443,7 @@ C.Send(messageDataRequest)
 // Only receive complete messages
 C.Receive()
 //---- Ready event handler for any Connection C end ----
+messageContext := NewMessageContext()
```
abrunstrom, line 446: This part did not change so why is this now needed to create the messageContext?

draft-ietf-taps-interface.md
```diff
@@ -389,10 +389,13 @@ Listener -> ConnectionReceived<Connection>
 // Only receive complete messages in a Conn.Received handler
 Connection.Receive()
 
+messageContext := NewMessageContext()
```
abrunstrom, line 392: Is this in the right place? Should it not be in the event handler where you send the message?

draft-ietf-taps-interface.md
```diff
@@ -389,10 +389,13 @@ Listener -> ConnectionReceived<Connection>
...
 Connection -> Received<messageDataRequest, messageContext>
 
 //---- Receive event handler begin ----
-Connection.Send(messageDataResponse)
+messageContext.add(msgPrio, 200)
+messageContext.add(final, true)
+Connection.Send(messageDataResponse, messageContext)
```
abrunstrom, line 398: Where does the messageDataResponse come from. I guess the example does not show everything? But it gets a bit confusing when the messageContext is filled explicitly.

draft-ietf-taps-interface.md
```diff
@@ -389,10 +389,13 @@ Listener -> ConnectionReceived<Connection>
 // Only receive complete messages in a Conn.Received handler
 Connection.Receive()
 
+messageContext := NewMessageContext()
 Connection -> Received<messageDataRequest, messageContext>
```
tfpauly, line 393:
```suggestion
Connection -> Received<messageDataRequest, receiveMessageContext>
```

draft-ietf-taps-interface.md
tfpauly: sendMessageContext := NewMessageContext()
sendMessageContext.add(msgPrio, 200)
sendMessageContext.add(final, true)
Connection.Send("response", sendMessageContext)

draft-ietf-taps-interface.md
tfpauly:
```suggestion
```

PR #938, opened 2021-10-15
Make architecture nicer (Many edits)

gorryfair:
Nicer text throughout to introduce key things consistently; firm-up architecture v interface; introduce properties; resolve some NiTs.


draft-ietf-taps-arch.md
```diff
@@ -220,33 +220,36 @@ Flexibility after connection establishment is also important. Transport protocol
...
+There are several degrees in which a Transport Services system is intended to offer flexibility to an application: it can provide access to multiple sets of protocols and protocol features; it can use these protocols across multiple paths that could have different performance and functional characteristics; and it can communicate with different remote systems to optimize performance, robustness to failure, or some other metric. Beyond these, if the API for the system remains the same over time, new protocols and features can be added to the system's implementation without requiring changes in applications for adoption. Similarly this can provide a common basis for utilising information about a a network path or interface, enabling evolution below the transport layer.
```
tfpauly, line 225:
```suggestion
There are several degrees in which a Transport Services system is intended to offer flexibility to an application: it can provide access to multiple sets of protocols and protocol features; it can use these protocols across multiple paths that could have different performance and functional characteristics; and it can communicate with different remote systems to optimize performance, robustness to failure, or some other metric. Beyond these, if the API for the system remains the same over time, new protocols and features can be added to the system's implementation without requiring changes in applications for adoption. Similarly, this can provide a common basis for utilizing information about a network path or interface, enabling evolution below the transport layer.
```

draft-ietf-taps-arch.md
```diff
@@ -220,33 +220,36 @@ Flexibility after connection establishment is also important. Transport protocol
...
+An application can specify constraints and preferences for the protocols, features, and network interfaces it will use via Properties. Properties are used by an applicatioon to  declares its preferences for how the transport service should operate at each stage of the lifetime of a connection. Transport Properties are subdivided into Selection Properties, which specify which paths and protocol stacks can be used and are preferred by the application; Connection Properties, which inform decisions made during connection establishment and to fine-tune the established connection and Message Properties, set on indvidual Messages.
```
tfpauly, line 233:
```suggestion
An application can specify constraints and preferences for the protocols, features, and network interfaces it will use via Properties. Properties are used by an application to declare its preferences for how the transport service should operate at each stage in the lifetime of a connection. Transport Properties are subdivided into Selection Properties, which specify which paths and protocol stacks can be used and are preferred by the application; Connection Properties, which inform decisions made during connection establishment and fine-tune the established connection; and Message Properties, set on individual Messages.
```

draft-ietf-taps-arch.md
```diff
@@ -220,33 +220,36 @@ Flexibility after connection establishment is also important. Transport protocol
...
 
 ## Select Equivalent Protocol Stacks {#equivalence}
 
-A Transport Services implementation can select Protocol Stacks based on the Properties communicated by the application. If two different Protocol Stacks can be safely swapped, or raced in parallel (see {{racing}}), then they are considered to be "equivalent". Equivalent Protocol Stacks are defined as stacks that can provide the same transport properties and interface expectations as requested by the application.
+A Transport Services implementation can select Protocol Stacks based on the Selection and Connection Properties communicated by the application and any security parameters. If two different Protocol Stacks can be safely swapped, or raced in parallel (see {{racing}}), then they are considered to be "equivalent". Equivalent Protocol Stacks are defined as stacks that can provide the same Transport Properties and interface expectations as requested by the application.
```
tfpauly, line 252:
```suggestion
A Transport Services implementation can select Protocol Stacks based on the Selection and Connection Properties communicated by the application, along with any security parameters. If two different Protocol Stacks can be safely swapped, or raced in parallel (see {{racing}}), then they are considered to be "equivalent". Equivalent Protocol Stacks are defined as stacks that can provide the same Transport Properties and interface expectations as requested by the application.
```

draft-ietf-taps-arch.md
```diff
@@ -264,11 +267,11 @@ To ensure that security protocols are not incorrectly swapped, Transport Service
...
+The Transport Services system makes decisions that select protocols and interfaces. In normal use, a Transport Services system SHOULD result in consistent protocol and interface selection decisions for the same network conditions given the same set of Properties. This is intended to provide predictable outcomes to the application using the API.
 
 # Transport Services Architecture and Concepts {#concepts}
 
-The concepts defined in this document are intended primarily for use in the documents and specifications that describe the Transport Services architecture and API. While the specific terminology can be used in some implementations, it is expected that there will remain a variety of terms used by running code.
+This Section and the remainder of this document describe the architecture non-normatively. The concepts defined in this document are intended primarily for use in the documents and specifications that describe the Transport Services architecture and API. While the specific terminology can be used in some implementations, it is expected that there will remain a variety of terms used by running code.
```
tfpauly, line 274:
```suggestion
This section and the remainder of this document describe the architecture non-normatively. The concepts defined in this document are intended primarily for use in the documents and specifications that describe the Transport Services architecture and API. While the specific terminology can be used in some implementations, it is expected that there will remain a variety of terms used by running code.
```

draft-ietf-taps-arch.md
```diff
@@ -385,7 +388,7 @@ The diagram below provides a high-level view of the actions and events during th
...
+* Selection: The Selection Properties consist of the properties that an application can set to influence the selection of paths between the local and remote systems, to influence the selection of transport protocols, or to configure the behavior of generic transport protocol features. These properties can take the form of requirements, prohibitions, or preferences. Examples of properties that influence path selection include the interface type (such as a Wi-Fi connection, or a Cellular LTE connection), requirements around the largest Message that can be sent, or preferences for throughput and latency. Examples of properties that influence protocol selection and configuration of transport protocol features include reliability, multipath support, and fast open support.
```
tfpauly, line 391: Should we change more? It seems odd to say "Selection", and then describe "Selection Properties"

tfpauly (approved):

draft-ietf-taps-arch.md
gorryfair: Looks good.

PR #942, opened 2021-10-20
Caching and racing in implementation intro

### LINKED ISSUE #883: ART-ART: Introduce Caching
Describe what you mean by caching and why you might want to isolate sessions
earlier.

### LINKED ISSUE #890: ART-ART: Privacy in Racing
In (Arch) section 3.3, you are talking about racing before you've described what it is. At the
point where you note that racing should only happen over stacks with identical
security properties, note when doing so might introduce privacy issues and what
can be done about them.

britram:
Fixes #883. Fixes #890.


draft-ietf-taps-arch.md
```diff
@@ -212,9 +212,16 @@ Allowing applications to interact with messages is backwards-compatible with exi
...
+- to handle multiple candidate endpoints, protocols, and paths;
+- to support candidate protocol racing to select the most optimal stack in each situation;
+- to support multipath and multistreaming protocols;  
+- to provide state caching and application control over it.
+
+Transport Services implementations are meant to be flexible at connection establishment time, considering many different options and trying to select the most optimal combinations by racing them and measuring the results (see {{gathering}} and {{racing}}). This requires applications to provide higher-level endpoints than IP addresses, such as hostnames and URLs, which are used by a Transport Services implementation for resolution, path selection, and racing. Transport services implementations can further implement fallback mechanisms if connection establishment of one protocol fails or performance is detected to be unsatisfactory.
```
mwelzl, line 222:
```suggestion
Transport Services implementations are meant to be flexible at connection establishment time, considering many different options and trying to select the most optimal combinations by racing them and measuring the results (see {{gathering}} and {{racing}}). This requires applications to provide higher-level endpoints than IP addresses, such as hostnames and URLs, which are used by a Transport Services implementation for resolution, path selection, and racing. Transport Services implementations can further implement fallback mechanisms if connection establishment of one protocol fails or performance is detected to be unsatisfactory.
```

PR #950, opened 2021-10-21
API: PVD  and other queriable info

gorryfair:
Rewording the path information with respect to PVD to reflect Info concerning the 
path(s) used by the Protocol Stack.

### DIFF
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -1855,11 +1855,11 @@ Properties will include different information:
   Protocol Selection Properties that the application specified on the
   Preconnection. 
 
-
-* For Connections that are Established, additional properties of the path(s) in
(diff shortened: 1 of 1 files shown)


draft-ietf-taps-interface.md
```diff
@@ -1855,11 +1855,11 @@ Properties will include different information:
   Protocol Selection Properties that the application specified on the
   Preconnection. 
 
-
-* For Connections that are Established, additional properties of the path(s) in
-  use. These properties can be derived from the local provisioning domain
-  {{RFC7556}}, measurements by the Protocol Stack, or other sources.
-
+* For Connections that are Established: Information concerning the
```
mwelzl, line 1858:
```suggestion
* For Connections that are Established: information concerning the 
```

PR #951, opened 2021-10-21
ARCH components.

gorryfair:
This is the second sldie trying to unravel what things are called.

### DIFF
--- a/draft-ietf-taps-arch.md
+++ b/draft-ietf-taps-arch.md
@@ -237,7 +237,7 @@ The normative requirements described in this section allow Transport Services AP
 
 ## Provide Common APIs for Common Features
 
-Any functionality that is common across multiple transport protocols SHOULD be made accessible through a unified set of calls using the Transport Services API. As a baseline, any Transport Services API MUST allow access to the minimal set of features offered by transport protocols {{?RFC8923}}.
+Any functionality that is common across multiple transport protocols SHOULD be made accessible through a unified set of calls using the Transport Services API. As a baseline, any Transport Services API SHOULD allow access to the minimal set of features offered by transport protocols {{?RFC8923}}.
 
 An application can specify constraints and preferences for the protocols, features, and network interfaces it will use via Properties. Properties are used by an application to declare its preferences for how the transport service should operate at each stage in the lifetime of a connection. Transport Properties are subdivided into Selection Properties, which specify which paths and protocol stacks can be used and are preferred by the application; Connection Properties, which inform decisions made during connection establishment and fine-tune the established connection; and Message Properties, set on individual Messages.
 
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -1801,11 +1801,15 @@ Protocol Stack(s) backing a Connection. These Connection Properties may be
 Generic, applying regardless of transport protocol, or Specific, applicable to a
 single implementation of a single transport protocol stack. Generic Connection
 Properties are defined in {{connection-props}} below. 
-
+ 
 Protocol Specific Properties are defined in a transport- and 
-implementation-specific way, and MUST NOT apply across different protocols. 
-Too much reliance by an application on Protocol Specific Properties can significantly 
-reduce the flexibility of a transport services implementation.
+implementation-specific way to
+permit more specialized protocol features to be used. 
+Too much reliance by an application on Protocol Specific Properties can significantly reduce the flexibility 
+of a transport services implementation to make appropriate
+selection and configuration choices. Therefore, it is RECOMMENDED that
+Protocol Properties are used for properties common across different protocols and that
+Protocol Specific Properties are only used where specific protocols or properties are necessary.
 
 The application can set and query Connection Properties on a per-Connection
 basis. Connection Properties that are not read-only can be set during


gorryfair:
This is a PR that came out of  my unvoiced frustration with the terms. In particular when we talk about a "system" or and "endpoint"; what do we call the thing below the API: the "transport system", and what is the architecture (the whole thing). I don't much mind if we use "Transport System" or "Transport Services System" ... hopefully that change would be small. Is this PR sane?

draft-ietf-taps-arch.md
```diff
@@ -239,42 +239,42 @@ Any functionality that is common across multiple transport protocols SHOULD be m
...
+It is important to note that neither the Transport Services API {{!I-D.ietf-taps-interface}} nor the guidelines for implementation of the Transport System {{?I-D.ietf-taps-impl}} define new protocols or protocol capabilities that affect what is communicated across the network. Use of the Transport Service Architecture MUST NOT require that a peer on the other side of a connection uses the same API or implementation. A Transport System acting as a connection initiator is able to communicate with any existing ednpoint that implements the transport protocol(s) and all the required properties selected by the Transport System. Similarly, a Transport System acting as a listener can receive connections for any protocol that is supported by the Transport System from an existing initiator that implements the protocol, independent of whether the initiator uses the Transport Services Archhitecture or not.
```
mwelzl, line 275:
```suggestion
It is important to note that neither the Transport Services API {{!I-D.ietf-taps-interface}} nor the guidelines for implementation of the Transport System {{?I-D.ietf-taps-impl}} define new protocols or protocol capabilities that affect what is communicated across the network. Use of the Transport Services Architecture MUST NOT require that a peer on the other side of a connection uses the same API or implementation. A Transport System acting as a connection initiator is able to communicate with any existing endpoint that implements the transport protocol(s) and all the required properties selected by the Transport System. Similarly, a Transport System acting as a listener can receive connections for any protocol that is supported by the Transport System from an existing initiator that implements the protocol, independent of whether the initiator uses the Transport Services Architecture or not.
```

draft-ietf-taps-arch.md
```diff
@@ -239,42 +239,42 @@ Any functionality that is common across multiple transport protocols SHOULD be m
...
 
-The Transport Services system makes decisions that select protocols and interfaces. In normal use, a Transport Services system SHOULD result in consistent protocol and interface selection decisions for the same network conditions given the same set of Properties. This is intended to provide predictable outcomes to the application using the API.
+The Transport System makes decisions that select protocols and interfaces. In normal use, a Transport Ssystem SHOULD result in consistent protocol and interface selection decisions for the same network conditions given the same set of Properties. This is intended to provide predictable outcomes to the application using the API.
```
mwelzl, line 277:
```suggestion
The Transport System makes decisions that select protocols and interfaces. In normal use, a Transport System SHOULD result in consistent protocol and interface selection decisions for the same network conditions given the same set of Properties. This is intended to provide predictable outcomes to the application using the API.
```

draft-ietf-taps-arch.md
```diff
@@ -283,7 +283,7 @@ This section and the remainder of this document describe the architecture non-no
 The architecture divides the concepts for Transport Services into two categories:
 
 1. API concepts, which are intended to be exposed to applications; and
-2. System-implementation concepts, which are intended to be internally used when building systems that implement Transport Services.
+2. Implementation concepts that are intended to be internally used when building Transport System that implements the Transport Services Architecture.
```
mwelzl, line 286:
```suggestion
2. Implementation concepts that are intended to be internally used when building a Transport System that implements the Transport Services Architecture.
```

draft-ietf-taps-arch.md
```diff
@@ -381,31 +381,31 @@ The diagram below provides a high-level view of the actions and events during th
...
+* Connection: A Connection object represents one or more active transport protocol instances that can send and/or receive Messages between local and remote ednpoints. It is an abstraction that represents the communication. The Connection object holds state pertaining to the underlying transport protocol instances and any ongoing data transfers. For example, an active Connection can represent a connection-oriented protocol such as TCP, or can represent a fully-specified 5-tuple for a connectionless protocol such as UDP, where the Connection remains an abstraction at the end points. It can also represent a pool of transport protocol instances, e.g., a set of TCP and QUIC connections to equivalent Endpoints, or a stream of a multi-streaming transport protocol instance. Connections can be created from a Preconnection or by a Listener.
```
mwelzl, line 392:
```suggestion
* Connection: A Connection object represents one or more active transport protocol instances that can send and/or receive Messages between local and remote endpoints. It is an abstraction that represents the communication. The Connection object holds state pertaining to the underlying transport protocol instances and any ongoing data transfers. For example, an active Connection can represent a connection-oriented protocol such as TCP, or can represent a fully-specified 5-tuple for a connectionless protocol such as UDP, where the Connection remains an abstraction at the end points. It can also represent a pool of transport protocol instances, e.g., a set of TCP and QUIC connections to equivalent endpoints, or a stream of a multi-streaming transport protocol instance. Connections can be created from a Preconnection or by a Listener. 
```

You haven't capitalized "endpoints" elsewhere, and I think it's better without capitals (unless it's a defined term in the arch document? I guess it is... hmmm ... then you have to use it with capitals but only after its definition?  oh my oh my...)

draft-ietf-taps-arch.md
```diff
@@ -381,31 +381,31 @@ The diagram below provides a high-level view of the actions and events during th
...
 
-* Listen: Enables a listener to accept incoming Connections. The Listener will then create Connection objects as incoming connections are accepted ({{events}}). Listeners by default register with multiple paths, protocols, and Local Endpoints, unless constrained by Selection Properties and/or the specified Local Endpoint(s). Connections can be accepted on any of the available paths or endpoints.
+* Listen: Enables a listener to accept incoming Connections. The Listener will then create Connection objects as incoming connections are accepted ({{events}}). Listeners by default register with multiple paths, protocols, and Local Endpoints, unless constrained by Selection Properties and/or the specified Local Endpoint(s). Connections can be accepted on any of the available paths or Endpoints.
```
mwelzl, line 408:
```suggestion
* Listen: Enables a listener to accept incoming Connections. The Listener will then create Connection objects as incoming connections are accepted ({{events}}). Listeners by default register with multiple paths, protocols, and Local Endpoints, unless constrained by Selection Properties and/or the specified Local Endpoint(s). Connections can be accepted on any of the available paths or endpoints.
```

See above for a discussion of "endpoints" vs "Endpoints".

draft-ietf-taps-arch.md
```diff
@@ -426,15 +426,15 @@ The diagram below provides a high-level view of the actions and events during th
...
+* Send: The action to transmit a Message over a Connection to the Remote Endpoint. The interface to Send can accept Message Properties specific to how the Message content is to be sent. The status of the Send operation is delivered back to the sending application in an Event ({{events}}).
 
-* Receive: An action that indicates that the application is ready to asynchronously accept a Message over a Connection from a remote system, while the Message content itself will be delivered in an event ({{events}}). The interface to Receive can include Message Properties specific to the Message that is to be delivered to the application.
+* Receive: An action that indicates that the application is ready to asynchronously accept a Message over a Connection from a remote endpoint, while the Message content itself will be delivered in an Event ({{events}}). The interface to Receive can include Message Properties specific to the Message that is to be delivered to the application.
```
mwelzl, line 435:
```suggestion
* Receive: An action that indicates that the application is ready to asynchronously accept a Message over a Connection from a Remote Endpoint, while the Message content itself will be delivered in an Event ({{events}}). The interface to Receive can include Message Properties specific to the Message that is to be delivered to the application.
```

So "Remote Endpoint" is capitalized but "endpoint" is not? At least for now, that's how it seems to be...

draft-ietf-taps-arch.md
```diff
@@ -448,33 +448,34 @@ The following categories of events can be delivered to an application:
...
 instance.
 
-While Connection Groups are managed by the transport system, an application can define Connection Contexts to control caching boundaries, as discussed in {{conn-context}}.
+While Connection Groups are managed by the Transport System, an application can define Connection Contexts to control caching boundaries, as discussed in {{conn-context}}.
 
-## Transport Services Implementation Concepts
+## Transport System Implementation
 
-This section defines the set of objects used internally to a system or library to implement the functionality needed to provide a transport service across a network, as required by the abstract interface.
+This section defines the key concepts of the Transport System component of the Transport Services Architecture.
```
mwelzl, line 474: This gets odd. If the Transport System is not a complete instantiation of the architecture (as I thought up to here), but just a component, then how can it be one of the "key concepts" of itself?  because it's also the next item in the list.

draft-ietf-taps-arch.md
```diff
@@ -553,14 +556,14 @@ the Encapsulating Security Protocol (ESP, part of IPsec) {{?RFC4303}} with QUIC,
...
 secure protocols to insecure protocols, or to weaker versions of secure
 protocols (see {{equivalence}}). For example, if an application requests a specific version of TLS,
 but the desired version of TLS is not available, its connection will fail.
 Applications are thus responsible for implementing security protocol fallback
-or version fallback by creating multiple Transport Services Connections, if so
-desired. Alternatively, a Transport Services system MAY allow applications to
-specify that fallback to a specific other version of a protocol is allowed.
+or version fallback by creating multiple Connections, if so
+desired. Alternatively, the Transport Services API MAY allow applications to
+specify that the Transport System is allowed to fallback to a specific other version of a protocol.
```
mwelzl, line 566: in the replaced text, "fallback" is used as a noun, but here it is a verb - I'm not sure it's a verb that exists? Maybe "falling back" ?

draft-ietf-taps-arch.md
gorryfair: Made the Transport System more clearly sit below the API when first introduced earlier (subtle because the "API" is only realised when you have a system designed to implement it, but at least tries to make reading less painful. We didn't talk about a "component" here, so suggest we don't. Text updated.

draft-ietf-taps-arch.md
```diff
@@ -448,33 +450,34 @@ The following categories of events can be delivered to an application:
...
 instance.
 
-While Connection Groups are managed by the transport system, an application can define Connection Contexts to control caching boundaries, as discussed in {{conn-context}}.
+While Connection Groups are managed by the Transport System, an application can define Connection Contexts to control caching boundaries, as discussed in {{conn-context}}.
 
-## Transport Services Implementation Concepts
+## Transport System Implementation
 
-This section defines the set of objects used internally to a system or library to implement the functionality needed to provide a transport service across a network, as required by the abstract interface.
+This section defines the key concepts of the Transport System of the Transport Services Architecture.
```
mwelzl, line 476: I still don't like the recursion, where the Transport System is a "key concept" of the Transport System. Are these maybe just key concepts of the Transport Services Architecture?

draft-ietf-taps-arch.md
```diff
@@ -381,23 +383,23 @@ The diagram below provides a high-level view of the actions and events during th
...
 
 * Security Parameters: Security Parameters define an application's requirements for authentication and encryption on a Connection. They are used by Transport Security protocols (such as those described in {{?RFC8922}}) to establish secure Connections. Examples of parameters that can be set include local identities, private keys, supported cryptographic algorithms, and requirements for validating trust of remote identities. Security Parameters are primarily associated with a Preconnection object, but properties related to identities can be associated directly with Endpoints.
```
mwelzl, line 404: Sorry, not this PR's problem, but I just noticed a capitalized Endpoint here:

```suggestion
* Security Parameters: Security Parameters define an application's requirements for authentication and encryption on a Connection. They are used by Transport Security protocols (such as those described in {{?RFC8922}}) to establish secure Connections. Examples of parameters that can be set include local identities, private keys, supported cryptographic algorithms, and requirements for validating trust of remote identities. Security Parameters are primarily associated with a Preconnection object, but properties related to identities can be associated directly with endpoints.
```

draft-ietf-taps-arch.md
```diff
@@ -212,78 +214,78 @@ Allowing applications to interact with messages is backwards-compatible with exi
...
 
 The following two examples show non-equivalent Protocol Stacks:
 
- - If the application requires preservation of message boundaries, a Protocol Stack that runs UDP as the top-level interface to the application is not equivalent to a Protocol Stack that runs TCP as the top-level interface. A UDP stack would allow an application to read out message boundaries based on datagrams sent from the remote system, whereas TCP does not preserve message boundaries on its own, but needs a framing protocol on top to determine message boundaries.
+ - If the application requires preservation of message boundaries, a Protocol Stack that runs UDP as the top-level interface to the application is not equivalent to a Protocol Stack that runs TCP as the top-level interface. A UDP stack would allow an application to read out message boundaries based on datagrams sent from the Remote Endpoint, whereas TCP does not preserve message boundaries on its own, but needs a framing protocol on top to determine message boundaries.
```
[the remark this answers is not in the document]
gorryfair: sure, done.

draft-ietf-taps-arch.md
```diff
@@ -171,7 +171,7 @@ The Transport Services architecture evolves this general model of interaction, a
 +-----------------------------------------------------+
                           |
 +-----------------------------------------------------+
-|          Transport Services Implementation          |
+|         Transport  Services Implementation          |
```
tfpauly, line 174:
```suggestion
|          Transport Services Implementation          |
```

tfpauly (approved):

PR #962, opened 2021-10-24
Handle changes to connection endpoints

### LINKED ISSUE #865: Should Endpoint Object identifiers be discoverable?
It feels to me like there's a bit of a gap between how software ought to discover Endpoint Object identifiers for Remote Endpoints versus Local Endpoints. Remote Endpoint identifiers are likely to be defined by the kind of software being used acting on some user input (web browser requesting a page, DNS server recursing, etc.), and these identifiers are resolved through existing internet architecture. Importantly, there's not really any way to check a valid input would work until `Initiate` for a Remote Endpoint.

This isn't really true for identifiers on Local Endpoints. While it's true that discovering these things is system-specific, a TAPS API implementation ought to provide a vehicle for discovering what inputs are valid. Two reasons I think it's reasonable to support discovery of local addresses and interfaces include:

 * It'd be unfortunate for TAPS-consuming software to rely on system-specific APIs to implement a UX for configuring that software.
 * Given diversity of public cloud offerings and wide use of multi-cloud deployments, it's reasonable to expect folks to want to dynamically discover what local network addresses and interfaces exist in their environment.

If the intent is to not define an API to discover these identifiers, the document should be clarified to discuss what Errors can be returned by `WithInterface` and `WithIPv[46]Address` Actions on Local Endpoints. (This is coming from a perspective that "error early, error often" is a best practice, which is admittedly subjective.)

If there is agreement to add (for example) `GetInterfaces` and `GetIPv[46]Addresses` Actions on a Local Endpoint, these should be documented. Further, the `WithInterface` and `WithIPv[46]Address` Actions on a Local Endpoint should be described to return an Error when the supplied input would not exist in the results returned by the corresponding `Get<identifier>` call at the time the `With<identifier>` Action was called.

In either case, it would be useful to mention that a successful invocation of a `With<identifier>` Action is not a guarantee that initiating the precondition will succeed.

### LINKED ISSUE #911: ART-ART: how does an application doing ICE learn about candidates?
I would really like to see an example of how an application that relies on ICE
learns about new candidates as they are discovered - the callback pattern is
not clear to me yet.

### LINKED ISSUE #961: How to add/remove endpoints from an existing connection?
For multipath transport protocols, the transport can automatically manage the set of local and remote endpoints its using, and the application doesn't need to be aware of this.

For non-multipath transport protocols used peer-to-peer, the peers need to be aware of changes to the set of local and remote endpoints, so they can trigger updates to the NAT binding discovery. This can't be done automatically, since to implement Trickle ICE, it's necessary to pass the updated set of candidates to the peer via an out-of-band signalling channel, before passing them in to the connection to probe connectivity. Accordingly, we need:

* A way to add/remove remote endpoints on an existing connection
* A way to add/remove local endpoints on an existing connection
* A way to detect when the set of local endpoints has changed

To implement Trickle ICE, adding and removing remote endpoints has to be an explicit action, since changes to the set of remote endpoints are communicated to the peer via out-of-band signalling.

Similarly, when implementing Trickle ICE, changes to the set of local endpoints have to be communicated to the peer via out-of-band signalling, so the application has to be notified about them. 

It could be possible to automatically add/remove local endpoints to an existing connection, if the Preconnection left the local endpoints unspecified. But if the set of local endpoints was specified, the application will need to make a choice which ones it adds/removes when the set changes.

csperkins:
This makes the following changes:
* Add `AddRemote()`, `RemoveRemote()`, `AddLocal()`, and `RemoveLocal()`
  actions to existing connections
* Update the description of the `PathChanged()` event to note that it's
  generated if the set of local endpoints available changes
* Update the peer-to-peer example to show how these can be used to support
  Trickle ICE

Closes #961
Closes #911
Closes #865

### DIFF
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -1841,7 +1841,7 @@ stop using one or more remote endpoints:
 Connection.RemoveRemote([]RemoteEndpoint)
 ~~~
 
-Removing all known remote endpoints will have the effect of aborting the
+Removing all known remote endpoints can have the effect of aborting the
 connection. The effect of removing the active remote endpoint(s) depends
 on the underlying transport: multipath aware transports might be able to
 switch to a new path if other reachable remote endpoints exist, or the
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -495,6 +495,25 @@ Connection.Receive()
 
 Connection -> Received<messageDataResponse, messageContext>
 
+// If new remote endpoint candidates are received from the peer over
+// the signalling channel, for example if using Trickle ICE, then add
+// them to the Connection:
+Connection.AddRemote(NewRemoteCandidates)
+
+// On a PathChange<> events, resolve the local endpoints to see if a
+// new local endpoint has become available and, if so, send to the peer
+// as a new candidate and add to the connection:
+Connection -> PathChange<>
+
+//---- PathChange event handler begin ----
+ResolvedLocal, ResolvedRemote = Preconnection.Resolve()
+if ResolvedLocal has changed:
+  // ...Send the ResolvedLocal list to peer via signalling channel
+  // Add the new local endpoints to the connection:
+  Connection.AddLocal(ResolvedLocal)
+//---- PathChange event handler end ----
+
+
 // Close the Connection in a Receive event handler
 Connection.Close()
 ~~~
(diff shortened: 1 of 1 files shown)


draft-ietf-taps-interface.md
```diff
@@ -1790,6 +1809,48 @@ lower Priority values. Capacity will be shared among these Connections according
...
+
+~~~
+Connection.RemoveRemote([]RemoteEndpoint)
+~~~
+
+Removing all known remote endpoints will have the effect of aborting the
+connection. The effect of removing the active remote endpoint(s) depends
+on the underlying transport: multipath aware transports might be able to
+switch to a new path if other reachable remote endpoints exist, or the
+connection might abort.
```
tfpauly, line 1848:
```suggestion
Removing all known remote endpoints can have the effect of aborting the
connection. The effect of removing the active remote endpoint(s) depends
on the underlying transport: multipath aware transports might be able to
switch to a new path if other reachable remote endpoints exist, or the
connection might abort.
```

PR #965, opened 2021-10-28
Clarify Pooled Connections vs. Connection Groups. Resolves #956.

### LINKED ISSUE #956: Impl: Pooled Connection
## Pooled Connection {#pooled-connections}

- So implementation talks about pooled connections - I must have blinked and missed that, obviously it is different to a connection group - or is it? If I am asking a silly question then I suspect we need  just a few lines of text to demuddle me.

mwelzl:
Resolves #956.

### DIFF
--- a/draft-ietf-taps-impl.md
+++ b/draft-ietf-taps-impl.md
@@ -145,7 +145,7 @@ The Transport Services system should have a list of supported protocols availabl
 
 In the following cases, failure should be detected during pre-establishment:
 
-- A request by an application for Protocol Properties that include requirements (properties required or prohibited) that cannot be satisfied by any of the available protocols. For example, if an application requires "Configure Reliability per Message", but no such feature is available in any protocol the host running the transport system on the host running the transport system this should result in an error, e.g., when SCTP is not supported by the operating system.
+- A request by an application for Protocol Properties that cannot be satisfied by any of the available protocols. For example, if an application requires "Configure Reliability per Message", but no such feature is available in any protocol the host running the transport system on the host running the transport system this should result in an error, e.g., when SCTP is not supported by the operating system.
 - A request by an application for Protocol Properties that are in conflict with each other, i.e., the required and prohibited properties cannot be satisfied by the same protocol. For example, if an application prohibits "Reliable Data Transfer" but then requires "Configure Reliability per Message", this mismatch should result in an error.
 
 To avoid allocating resources that are not finally needed, it is important that configuration-time errors fail as early as possible.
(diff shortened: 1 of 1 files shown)


gorryfair (approved):
Seems to answer my issue.

draft-ietf-taps-impl.md
```diff
@@ -738,10 +738,11 @@ errors that can be delivered to the application as Soft Errors. These allow the
...
-For these kinds of protocols, implementations may hide the connection management and only expose a single Connection object and the individual requests/responses as messages.
-These Pooled Connections can use multiple connections or multiple streams of multi-streaming connections between endpoints, as long as all of these satisfy the requirements, and prohibitions specified in the Selection Properties of the Pooled Connection. 
-This enables implementations to realize transparent connection coalescing, connection migration, and to perform per-message endpoint and path selection by choosing among these underlying connections.
+For applications that do not need in-order delivery of Messages, the Transport Services implementation may distribute Messages of a single Connection across several underlying transport connections or multiple streams of multi-streaming connections between endpoints, as long as all of these satisfy the requirements and prohibitions specified via Selection Properties.
```
philsbln, line 741: For me, prohibitions is redundant and just makes reading harder 
```suggestion
For applications that do not need in-order delivery of Messages, the Transport Services implementation may distribute Messages of a single Connection across several underlying transport connections or multiple streams of multi-streaming connections between endpoints, as long as all of these satisfy the requirements specified via Selection Properties.
```

philsbln (approved):
Works for me and makes the reasoning clearer and less theoretical.
The small change suggestion is optional, thus, improving readability IMHO.

draft-ietf-taps-impl.md
gorryfair: I think you do need to check the prohibitions also?

abrunstrom: I would think so as well.

abrunstrom (approved):
LGTM, thanks!

draft-ietf-taps-impl.md
philsbln: Oh … for me "requirements" actually means everything you express by using "require" or "prohibit" … would probably benefit explicitly stating that and making the remainder easier to read.

gorryfair: I tried a fix in the text - if it doesn't please suggest something else.

mwelzl: I don't see the fix. But a solution might be: ".... as long as all of these satisfy the specified Selection Properties."

tfpauly:
```suggestion
For applications that do not need in-order delivery of Messages, the Transport Services implementation may distribute Messages of a single Connection across several underlying transport connections or multiple streams of multi-streaming connections between endpoints, as long as all of these satisfy the Selection Properties.
```

tfpauly (approved):

PR #966, opened 2021-10-28
Be explicit about objects being treated as call-by-value

### LINKED ISSUE #912: ART-ART: improve text about local copies of objects
There are several places where the document says "Changes to a <foo> after
func() has been called do not have any connection on existing <bar>s". This
feels like an opportunity for more clarity about what a <foo> really is. The
document uses the word Object, and maybe there's an assumption that there's a
shared understanding of what the properties of an Object are. Being explicit
with something like "An Object is a collection of properties that
implementations of the API read and use..." to avoid the "here's a chunk of
memory that can get passed into the API that it can operate on, and expect the
caller to see the effects of the operation" interpretation. I _think_ what the
current text is trying to say is that the user of the API should assume the
implementation of the API made its own private copy of the Object at the point
of the given func() call and that it will not see any changes to the copy the
user has at any later point. But then there's some text that looks like "Once a
<foo> has been used with a func(), it is invalid to modify any of its
properties." So...

mwelzl:
Resolves #912.

### DIFF
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -214,6 +214,10 @@ The following notations, which can be combined, are used in this document:
       Action(param0, param1?, ...) / Event<param0, param1, ...>
 ~~~
 
+When an Object is used as a Parameter of an Action, the Action will not change this Object, and later changes
+to the Object will not have any automatic effect on the Transport Services system. Irrespective of the actual implementation,
+this behavior is as if the Transport Services system makes a complete copy
+of the Object and any other the Objects that it may reference when an Action is called.
 Actions associated with no Object are Actions on the API; they are equivalent to Actions on a per-application global context.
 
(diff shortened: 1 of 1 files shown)


britram (approved):

draft-ietf-taps-interface.md
```diff
@@ -214,6 +214,10 @@ The following notations, which can be combined, are used in this document:
       Action(param0, param1?, ...) / Event<param0, param1, ...>
 ~~~
 
+When an Object is used as a Parameter of an Action, the Action will not change this Object, and later changes
+to the Object will not have any automatic effect on the Transport Services system. Irrespective of the actual implementation,
+this behavior is as if the Transport Services system makes a complete copy
+of the Object and any other Objects that it may reference when an Action is called.
```
tfpauly, line 220: Objects that are passed as parameters to Actions use call-by-value behavior.

tfpauly (approved):

PR #972, opened 2021-12-08
Clarify framer-specific message properties

### LINKED ISSUE #918: ART-ART: scope vs. framer
Is the "scope" used for message contexts (9.1.1) ever anything but a framer? If
not, could this section just say framer?

tfpauly:
Closes #918

### DIFF
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -2458,12 +2458,10 @@ Framers must happen before any operation that may result in the creation of a Co
 When sending Messages, applications can add Framer-specific
 properties to a MessageContext ({{msg-ctx}}).
 In order to set these properties, the `add` and `get` actions
-on the MessageContext take an additional framer parameter.
-
-~~~
-MessageContext.add(framer, property, value)
-PropertyValue := MessageContext.get(framer, property)
-~~~
+on the MessageContext. To avoid naming conflicts, the property
+names SHOULD be prefixed with a namespace referencing the
+framer implementation or the protocol it implements as described
+in {{property-names}}. 
 
 This mechanism can be used, for example, to set the type of a Message for a TLV format.
 The namespace of values is custom for each unique Message Framer.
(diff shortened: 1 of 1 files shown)


tfpauly:
The other option is to tell people to prepend the name of the framer to the property name.

britram (approved):

mwelzl (commented):
I find this a little strange: given that we do have optional parameters, what's the benefit of first introducing the "add" and "get" calls as if the "framer" parameter didn't exist, and then later changing that?   It would seem more consistent to me to introduce it where you had the scope parameter before, called "framer", and use a forward reference to say that framers will be explained later.

draft-ietf-taps-interface.md
```diff
@@ -2456,7 +2456,15 @@ Framers must happen before any operation that may result in the creation of a Co
...
-key/value pairs to a MessageContext ({{msg-ctx}}).
+properties to a MessageContext ({{msg-ctx}}).
+In order to set these properties, the `add` and `get` actions
+on the MessageContext take an additional framer parameter.
+
+~~~
+MessageContext.add(framer, property, value)
+PropertyValue := MessageContext.get(framer, property)
+~~~
+
```
tfpauly, line 2467: @philsbln can you propose text for how we can prepend the name of the framer as a namespace for the properties we add/get.

philsbln, line 2467:
```suggestion
properties to a MessageContext ({{msg-ctx}}).
In order to set these properties, use the `add` and `get` actions
on the MessageContext. To avoid naming conflicts, the property names SHOULD be prefixed with a namespace referencing the framer implementation or the protocol it implements as described in {{property-names}}. 

~~~
MessageContext.add(property, value)
PropertyValue := MessageContext.get(property)
~~~

```

tfpauly:
```suggestion
properties to a MessageContext ({{msg-ctx}}).
In order to set these properties, the `add` and `get` actions
on the MessageContext. To avoid naming conflicts, the property
names SHOULD be prefixed with a namespace referencing the
framer implementation or the protocol it implements as described
in {{property-names}}. 

```

PR #974, opened 2021-12-09
Endpoints SHOULD/MAY for Preconnection

### LINKED ISSUE #973: secdir: s6: MUST/MAY for Local and Remote Endpoint requirements
0 from this [secdir review](https://mailarchive.ietf.org/arch/msg/secdir/yUaaas7vohCXnERvfU7DvlILh4U/). Note that in that review I really was not very clear about the issue, I tried to do better here.

The MUST/MAY combination of the following Local Endpoint requirement:

    At least one Local Endpoint MUST be specified if the Preconnection is
    used to Listen() for incoming Connections, but the list of Local Endpoints MAY be
    empty if the Preconnection is used to Initiate() connections.

and Remote Endpoint requirement:

    At least one Remote Endpoint MUST be specified if the Preconnection is
    used to Initiate() Connections, but the list of Remote Endpoints MAY be
    empty if the Preconnection is used to Listen() for incoming Connections.

It says you MUST have a value but then it says maybe not under this one condition. Isn't the MUST then a SHOULD? I.e., you really, really need to do this unless there's this one condition and then it's okay to not set it? I realize I may be dancing on a pinhead here so I will bow to collective wisdom of the group.

seanturner:
Closes #973.

### DIFF
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -642,12 +642,12 @@ of the potential Connection (see {{endpointspec}}), the Selection Properties
                                      SecurityParameters)
 ~~~
 
-At least one Local Endpoint MUST be specified if the Preconnection is used to Listen()
+At least one Local Endpoint SHUOLD be specified if the Preconnection is used to Listen()
 for incoming Connections, but the list of Local Endpoints MAY be empty if
 the Preconnection is used to Initiate()
 connections. If no Local Endpoint is specified, the Transport Services system will
 assign an ephemeral local port to the Connection on the appropriate interface(s).
-At least one Remote Endpoint MUST be specified if the Preconnection is used
+At least one Remote Endpoint SHOULD be specified if the Preconnection is used
 to Initiate() Connections, but the list of Remote Endpoints MAY be empty if 
 the Preconnection is used to Listen() for incoming Connections.
 At least one Local Endpoint and one Remote Endpoint MUST be specified if a


gorryfair (commented):
I thought this was needed for the preconnection to be acceptable, so I'm not sure about SHOULD. To me this is either required for the preconnection, or this is a MAY.

draft-ietf-taps-interface.md
```diff
@@ -642,12 +642,12 @@ of the potential Connection (see {{endpointspec}}), the Selection Properties
                                      SecurityParameters)
 ~~~
 
-At least one Local Endpoint MUST be specified if the Preconnection is used to Listen()
+At least one Local Endpoint SHUOLD be specified if the Preconnection is used to Listen()
```
tfpauly, line 645:
```suggestion
At least one Local Endpoint SHOULD be specified if the Preconnection is used to Listen()
```

tfpauly (changes requested):
I don't think this change is necessarily correct. The logic is:

- If Listener: MUST have local endpoint, MAY have remote
- If initiator: MUST have remote endpoint, MAY have local

draft-ietf-taps-interface.md
```diff
@@ -642,12 +642,12 @@ of the potential Connection (see {{endpointspec}}), the Selection Properties
...
 for incoming Connections, but the list of Local Endpoints MAY be empty if
 the Preconnection is used to Initiate()
 connections. If no Local Endpoint is specified, the Transport Services system will
 assign an ephemeral local port to the Connection on the appropriate interface(s).
-At least one Remote Endpoint MUST be specified if the Preconnection is used
+At least one Remote Endpoint SHOULD be specified if the Preconnection is used
```
tfpauly, line 650:
```suggestion
At least one Remote Endpoint MUST be specified if the Preconnection is used
```
```suggestion
At least one Remote Endpoint SHOULD be specified if the Preconnection is used
```

tfpauly:
Interim discussion: leave as-is.

PR #985, opened 2021-12-09
connCapacityProfile Type and Default

### LINKED ISSUE #984: secdir: s8.1.6: connCapacityProfile Type and Default
7 from this [secdir review](https://mailarchive.ietf.org/arch/msg/secdir/yUaaas7vohCXnERvfU7DvlILh4U/). And slightly ammended.

Don’t you need to specify the Type? I guess Enumeration is what you’re after?

Also, is the Enumeration value supposed to be empty or is it `Default`? Assumed the later, because `Default` is listed as a value.

seanturner:
Closes #984.

### DIFF
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -2056,7 +2056,7 @@ Type:
 : Enumeration
 
 Default:
-: Default
+: Default Profile (Best Effort)
 
 This property specifies the desired network treatment for traffic sent by the
 application and the tradeoffs the application is prepared to make in path and
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -2052,6 +2052,12 @@ be taken from {{?RFC8260}}.
 Name:
 : connCapacityProfile
 
+Type:
+: Enumeration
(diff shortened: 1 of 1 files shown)


gorryfair (approved):
Looks good to me. Default: Default looks odd, but seems correct.

mwelzl (commented):
I think this PR is doing the right thing, but it also makes clear that "Default" is a strange name for a value that can be chosen here. Can we rename it to "Best effort" ?

gorryfair:
The class is actually called "Default" ... but I wondered if we could call it something like in my order of decreasing preference:

Default Profile
Best Effort (Default Profile)
Default Profile (Best Effort)

... do any of these resonate?

mwelzl:
I like "Default Profile (Best Effort)" and "Best Effort (Default Profile)".  I just don't like only having "default" or "default profile", as it leads to: the default choice is "default".

gorryfair:
let's try: Default Profile (Best Effort)

mwelzl (approved):

gorryfair:
Fixes default capacity profile

PR #1002, opened 2022-01-10
Update author list as previously agreed

mwelzl:
I think this reflects prior agreement, which for some reason just wasn't applied to the API draft yet.


tfpauly (approved):

gorryfair (approved):
Looks like this reflect my notes.

PR #1003, opened 2022-02-01
Mention framer synchronization

### LINKED ISSUE #750: Framer changes and send/receive buffers (Sec 6.1)
Dynamic framer add and passthrough (last two paragraphs of section 6.1) assume synchronous operation; what happens to bytes in the send/receive buffers when the framer stack changes? 

There either needs to be an explicit mutex here (allowing the framer/implementation to lock the buffers at the exact point the framer change happens, not allowing any bytes to be processed while the dynamic framer decision is made), or some way to annotate in a buffer the replay point after which a dynamic reframing is necessary.

Are we going to leave this up to implementors to figure out, or should we point out that here be dragons?

tfpauly:
Closes #750

### DIFF
--- a/draft-ietf-taps-impl.md
+++ b/draft-ietf-taps-impl.md
@@ -602,6 +602,13 @@ implementation or the application itself. The Message Framer refers to the objec
 or function within the main Connection implementation that delivers events
 to the custom framer implementation whenever data is ready to be parsed or framed.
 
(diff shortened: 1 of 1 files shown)


mwelzl:
It seems blindingly obvious to me that an implementation would need to take care of this... and this note of caution seems quite appropriate (if unnecessary... but surely it doesn't hurt).

PR #1004, opened 2022-02-01
Preconnection immutability

### LINKED ISSUE #744: mutability of Preconnection objects is unclear (Sec 2)
Preconnections are mutable, until they are not. The second paragraph of section 2 states "bundles of properties an application can both read and write", which seems to conflict with this. Para should lead with the fact of mutable mutability over the preconnection lifecycle, not "is mutable".

tfpauly:
Closes #744


draft-ietf-taps-impl.md
```diff
@@ -123,7 +123,7 @@ The connection objects that are exposed to applications for Transport Services a
...
 
-Once a Preconnection has been used to create an outbound Connection or a Listener, the implementation should ensure that the copy of the properties held by the Connection or Listener is not affected when the application makes changes to a Preconnection object. This may involve the implementation performing a deep-copy, copying the object with all the objects that it references.
+Once a Preconnection has been used to create an outbound Connection or a Listener, the implementation should ensure that the copy of the properties held by the Connection or Listener cannot be mutated by the application makes changes to the original Preconnection object. This may involve the implementation performing a deep-copy, copying the object with all the objects that it references, or making the Preconnection object immutable.
```
abrunstrom, line 126: I do not see how the last addition solves the contradiction complained about by @britram. It seems to make it worse? But I also do not see the problem with the original text. maybe the addition of "the original" is enough?

Nit: makes --> making.

tfpauly:
```suggestion
Once a Preconnection has been used to create an outbound Connection or a Listener, the implementation should ensure that the copy of the properties held by the Connection or Listener cannot be mutated by the application makes changes to the original Preconnection object. This may involve the implementation performing a deep-copy, copying the object with all the objects that it references.
```
```suggestion
Once a Preconnection has been used to create an outbound Connection or a Listener, the implementation should ensure that the copy of the properties held by the Connection or Listener cannot be mutated by the application makes changes to the original Preconnection object. This may involve the implementation performing a deep-copy, copying the object with all the objects that it references, or making the Preconnection object immutable.
```

draft-ietf-taps-impl.md
```diff
@@ -123,7 +123,7 @@ The connection objects that are exposed to applications for Transport Services a
...
 
-Once a Preconnection has been used to create an outbound Connection or a Listener, the implementation should ensure that the copy of the properties held by the Connection or Listener is not affected when the application makes changes to a Preconnection object. This may involve the implementation performing a deep-copy, copying the object with all the objects that it references.
+Once a Preconnection has been used to create an outbound Connection or a Listener, the implementation should ensure that the copy of the properties held by the Connection or Listener cannot be mutated by the application makes changes to the original Preconnection object. This may involve the implementation performing a deep-copy, copying the object with all the objects that it references.
```
tfpauly, line 126:
```suggestion
Once a Preconnection has been used to create an outbound Connection or a Listener, the implementation should ensure that the copy of the properties held by the Connection or Listener cannot be mutated by the application making changes to the original Preconnection object. This may involve the implementation performing a deep-copy, copying the object with all the objects that it references.
```

PR #1005, opened 2022-02-01
Explain application preferences for addresses

### LINKED ISSUE #842: How does the application express preference for derived endpoints?
In the impl draft, I'm supposed to build a tree based on, in order

(1) "path": AFAICT this means "interface". The -api draft provides in interface preference framework in 4.2.11. (BTW if it really means interface maybe we should say 'interface').

(2) Protocol: almost all of 4.2 in -api explains how to express this. (with an assist in -impl 4.1.3 explaining how to handle PREFER vs. AVOID)

(3) "derived endpoint" I gather this is basically multiple addresses coming back from remote endpoint resolution. But where is the construct in -api that allows applications to express a preference here?

tfpauly:
Closes #842


abrunstrom (approved):
Thanks, LGTM!

PR #1008, opened 2022-03-19
Add CloseGroup/AbortGroup to mappings

tfpauly:


mwelzl:
Why this "calling CloseGroup is identical to calling Close" everywhere?  This seems counter-intuitive... am I missing something?   If I create a Connection Group by cloning a Connection a couple of times, no matter what the underlying protocol is, I would expect that calling CloseGroup on one of them closes them all.  This would match the semantics described in the interface draft, section 10.

mwelzl (changes requested):
Sorry, I guess I should have made this a review, not a comment:

Why this "calling CloseGroup is identical to calling Close" everywhere? This seems counter-intuitive... am I missing something? If I create a Connection Group by cloning a Connection a couple of times, no matter what the underlying protocol is, I would expect that calling CloseGroup on one of them closes them all. This would match the semantics described in the interface draft, section 10.

tfpauly:
So, for a multiplexing protocol (SCTP, QUIC), CloseGroup and AbortGroup has a specific meaning to treat the group vs a single connection/stream.

For TCP/UDP I wrote it to be a no-op, but I can say that it affects any other connections that were cloned out? But I wonder if that might be a little weird too.

mwelzl:
I think we need the system to treat it that way for consistency. Clearly, for a mux'ing protocol this is the obvious choice - but the application is not aware of the protocol below... so if it creates a group of TCP connections, with or without congestion control coupling, it will expect:
- that changing a connection property on one of the connections will affect the others (so that also must be implemented - because it WILL happen with a mux'ing protocol),
- that closing the group will close them all.

Imagine an application getting QUIC once, and not getting it another time...  surely we need a consistent behavior towards the app for both of these cases?

draft-ietf-taps-impl.md
```diff
@@ -920,6 +920,12 @@ Close:
 Abort:
 : Calling `Abort` on a TCP Connection indicates that the Connection should be immediately closed by sending a RST to the peer (ABORT.TCP).
 
+CloseGroup:
+: Calling `CloseGroup` on a TCP Connection (CLOSE.TCP) is identical to calling `Close`.
+
+AbortGroup:
+: Calling `AbortGroup` on a TCP Connection (ABORT.TCP) is identical to calling `Abort`.
+
```
tfpauly, line 928:
```suggestion
CloseGroup:
: Calling `CloseGroup` on a TCP Connection (CLOSE.TCP) is identical to calling `Close` on all TCP Connections
within a group.

AbortGroup:
: Calling `AbortGroup` on a TCP Connection (ABORT.TCP) is identical to calling `Abort` on all TCP Connections
within a group.

```

draft-ietf-taps-impl.md
```diff
@@ -976,6 +982,12 @@ Close:
 Abort:
 : Calling `Abort` on a UDP Connection (ABORT.UDP(-Lite)) is identical to calling `Close`.
 
+CloseGroup:
+: Calling `CloseGroup` on a UDP Connection (ABORT.UDP(-Lite)) is identical to calling `Close`.
+
+AbortGroup:
+: Calling `AbortGroup` on a UDP Connection (ABORT.UDP(-Lite)) is identical to calling `Close`.
+
```
tfpauly, line 990:
```suggestion
CloseGroup:
: Calling `CloseGroup` on a UDP Connection (ABORT.UDP(-Lite)) is identical to calling `Close` on all
UDP Connections within a group.

AbortGroup:
: Calling `AbortGroup` on a UDP Connection (ABORT.UDP(-Lite)) is identical to calling `Close` on all
UDP Connections within a group.

```

draft-ietf-taps-impl.md
```diff
@@ -1039,6 +1051,12 @@ Close:
 Abort:
 : Calling `Abort` on a UDP Multicast Receive Connection (ABORT.UDP(-Lite)) is identical to calling `Close`.
 
+CloseGroup:
+: Calling `CloseGroup` on a UDP Multicast Receive Connection (ABORT.UDP(-Lite)) is identical to calling `Close`.
```
tfpauly, line 1055:
```suggestion
: Calling `CloseGroup` on a UDP Multicast Receive Connection (ABORT.UDP(-Lite)) is identical to calling `Close` on all UDP Multicast Receive  Connections within a group.
```

draft-ietf-taps-impl.md
```diff
@@ -1039,6 +1051,12 @@ Close:
...
+
+AbortGroup:
+: Calling `AbortGroup` on a UDP Multicast Receive Connection (ABORT.UDP(-Lite)) is identical to calling `Close`.
```
tfpauly, line 1058:
```suggestion
: Calling `AbortGroup` on a UDP Multicast Receive Connection (ABORT.UDP(-Lite)) is identical to calling `Close`
on all UDP Multicast Receive  Connections within a group.
```

tfpauly:
@mwelzl sounds good, please check the updated text (or suggest text if you have further ideas!)

draft-ietf-taps-impl.md
```diff
@@ -920,6 +920,14 @@ Close:
 Abort:
 : Calling `Abort` on a TCP Connection indicates that the Connection should be immediately closed by sending a RST to the peer (ABORT.TCP).
 
+CloseGroup:
+: Calling `CloseGroup` on a TCP Connection (CLOSE.TCP) is identical to calling `Close` on all TCP Connections
+within a group.
```
theri, line 925: How about:

```suggestion
: Calling `CloseGroup` on a TCP Connection (CLOSE.TCP) is identical to calling `Close` on this Connection and on all Connections in the same ConnectionGroup.
```
This is essentially what it says in Section 10 in the Interface draft, and perhaps the clearest way to phrase this.

theri (approved):
This is probably fine as is, but I suggested alternative phrasing which might be even more clear.

draft-ietf-taps-impl.md
mwelzl: I like this - I think it should be phrased using that style for all the calls below. "within a group" is just a touch too vague IMO... it needs to be the group this connection is in, even though this is super obvious.

mwelzl (changes requested):
LGTM if the suggestion by @theri is applied everywhere.  Many thanks for doing this BTW!

draft-ietf-taps-impl.md
```diff
@@ -920,6 +920,14 @@ Close:
...
+
+AbortGroup:
+: Calling `AbortGroup` on a TCP Connection (ABORT.TCP) is identical to calling `Abort` on all TCP Connections
+within a group.
```
tfpauly, line 929:
```suggestion
AbortGroup:
: Calling `AbortGroup` on a TCP Connection (ABORT.TCP) is identical to calling `Abort` on this Connection and on all Connections in the same ConnectionGroup.
```

draft-ietf-taps-impl.md
```diff
@@ -976,6 +984,14 @@ Close:
...
 : Calling `Abort` on a UDP Connection (ABORT.UDP(-Lite)) is identical to calling `Close`.
 
+CloseGroup:
+: Calling `CloseGroup` on a UDP Connection (ABORT.UDP(-Lite)) is identical to calling `Close` on all
+UDP Connections within a group.
+
+AbortGroup:
+: Calling `AbortGroup` on a UDP Connection (ABORT.UDP(-Lite)) is identical to calling `Close` on all
+UDP Connections within a group.
+
```
tfpauly, line 994:
```suggestion
CloseGroup:
: Calling `CloseGroup` on a UDP Connection (ABORT.UDP(-Lite)) is identical to calling `Close` on this Connection and on all Connections in the same ConnectionGroup.

AbortGroup:
: Calling `AbortGroup` on a UDP Connection (ABORT.UDP(-Lite)) is identical to calling `Close` on this Connection and on all Connections in the same ConnectionGroup.
```

draft-ietf-taps-impl.md
```diff
@@ -1039,6 +1055,13 @@ Close:
 Abort:
 : Calling `Abort` on a UDP Multicast Receive Connection (ABORT.UDP(-Lite)) is identical to calling `Close`.
 
+CloseGroup:
+: Calling `CloseGroup` on a UDP Multicast Receive Connection (ABORT.UDP(-Lite)) is identical to calling `Close` on all UDP Multicast Receive  Connections within a group.
+
+AbortGroup:
+: Calling `AbortGroup` on a UDP Multicast Receive Connection (ABORT.UDP(-Lite)) is identical to calling `Close`
+on all UDP Multicast Receive  Connections within a group.
+
```
tfpauly, line 1064:
```suggestion
CloseGroup:
: Calling `CloseGroup` on a UDP Multicast Receive Connection (ABORT.UDP(-Lite)) is identical to calling `Close` on this Connection and on all Connections in the same ConnectionGroup.

AbortGroup:
: Calling `AbortGroup` on a UDP Multicast Receive Connection (ABORT.UDP(-Lite)) is identical to calling `Close`
on this Connection and on all Connections in the same ConnectionGroup.

```

draft-ietf-taps-impl.md
tfpauly: Applied, thanks!

mwelzl (approved):
Great, thanks!!

abrunstrom (approved):
LGTM with the updates, thanks!

PR #1014, opened 2022-05-10
readability edit of section 4.0, complementing my ongoing draft review

benthor:
While reviewing the draft, I found it easy to address some of the issues I saw in section 4.0 in a light edit. I also opened an accompanying issue #1015 containing the relevant review feedback.

I understand this PR as a low-key "take it or leave it" contribution, it wouldn't bother me if it gets summarily rejected

*Feedback welcome: is the level of change in this PR too heavy handed or is there interest in similar feedback about other sections of the draft?*


draft-ietf-taps-impl.md
```diff
@@ -166,14 +166,11 @@ It is expected that the database of system policies and the method of looking up
 
 The process of establishing a network connection begins when an application expresses intent to communicate with a Remote Endpoint by calling Initiate. (At this point, any constraints or requirements the application may have on the connection are available from pre-establishment.) The process can be considered complete once there is at least one Protocol Stack that has completed any required setup to the point that it can transmit and receive the application's data.
 
-Connection establishment is divided into two top-level steps: Candidate Gathering, to identify the paths, protocols, and endpoints to use, and Candidate Racing (see Section 4.2.2 of {{I-D.ietf-taps-arch}}), in which the necessary protocol handshakes are conducted so that the transport system can select which set to use.
+Connection establishment is divided into two top-level steps: Candidate Gathering, to identify the potential paths, protocols, and endpoints to use, and Candidate Racing (see Section 4.2.2 of {{I-D.ietf-taps-arch}}), in which the necessary protocol handshakes are conducted so that the transport system can select which set to use.
```
tfpauly, line 169: I think this may be more clear as "possible" or "available" than "potential"

draft-ietf-taps-impl.md
```diff
@@ -182,16 +179,14 @@ Aggregate [Endpoint: www.example.com:80] [Interface: Any]   [Protocol: TCP]
 |-> [Endpoint: 2001:DB8::1.80]     [Interface: LTE]   [Protocol: TCP]
 ~~~~~~~~~~
 
-Any one of these sub-entries on the aggregate connection attempt would satisfy the original application intent. The concern of this section is the algorithm defining which of these options to try, when, and in what order.
+Any one of these sub-entries on the aggregate connection attempt would satisfy the original application intent. The concern of this section is the algorithm defining which of these "Candidate" options to try, when, and in what order.
```
tfpauly, line 182:
```suggestion
Any one of these sub-entries on the aggregate connection attempt would satisfy the original application intent. The concern of this section is the algorithm defining which of these options (referred to for the rest of this section as "candidates") to try, when, and in what order.
```

draft-ietf-taps-impl.md
```diff
@@ -182,16 +179,14 @@ Aggregate [Endpoint: www.example.com:80] [Interface: Any]   [Protocol: TCP]
...
 
-During Candidate Gathering, an implementation first excludes all protocols and
-paths that match a Prohibit or do not match all Require properties.
-Then, the implementation will sort branches according to Preferred
-properties, Avoided properties, and possibly other criteria.
+During Candidate Gathering, an implementation first excludes all protocols and paths with Prohibited features as well as those which do not satisfy a Required property.
```
tfpauly, line 184:
```suggestion
During Candidate Gathering, an implementation first excludes all protocols and paths that match Prohibited properties or do not satisfy the Required properties.
```

draft-ietf-taps-impl.md
```diff
@@ -182,16 +179,14 @@ Aggregate [Endpoint: www.example.com:80] [Interface: Any]   [Protocol: TCP]
...
+Then, the remaining candidates will be ranked according to Preferred properties, Avoided properties, and possibly other criteria.
 
 ## Structuring Candidates as a Tree
 
-As noted above, the considereration of multiple candidates in a gathering and racing process can be conceptually structured as a tree; this terminological convention is used throughout this document.
+The considereration of multiple candidates in a gathering and racing process can be conceptually structured as a tree; this terminological convention is used throughout this document. While a tree structure is not the only way in which racing can be implemented, it does ease the illustration of how racing works.
```
tfpauly, line 189:
```suggestion
The considereration of multiple candidates in a gathering and racing process can be conceptually structured as a tree; this terminological convention is used throughout this document. While a tree structure is not the only way in which racing can be implemented, it is a useful model to use to understand and illustrate connection racing.
```

draft-ietf-taps-impl.md
```diff
@@ -166,14 +166,11 @@ It is expected that the database of system policies and the method of looking up
...
 
-This document structures the candidates for racing as a tree as terminological convention. While a 
-a tree structure is not the only way in which racing can be implemented, it does ease the illustration of how racing works. 
+In its most basic form, the connection-establishment process might just involve identifying a single IP address (and port) to which the application wishes to connect, and starting a TCP handshake to establish a stream to that IP address, using the system's current default path (i.e., by using the current default network interface). However, each step may also differ depending on the requirements for the connection: if, instead of a known IP address, the endpoint is identified by a hostname, then there may be a choice between multiple resolved addresses; some protocols may not need any transport handshake to establish communication (such as UDP), while others may require several layers of handshakes, such as TLS over TCP; finally, there may also be multiple paths available (i.e., via additional network interfaces besides the system default);
```
tfpauly, line 171: Nit: I think "connection-establishment" should just be "connection establishment".

More broadly, I don't love how this makes the default seem like connecting to an IP address — "if, instead of a known IP address...". The default is going to be higher-level endpoints, and that's a core tenet of the API.

tfpauly (commented):
Thanks very much for the suggested edits! I think we can have a bit of refinement, but these are much appreciated.

draft-ietf-taps-impl.md
abrunstrom: Agreed. Perhaps replacing the start of that sentence with something like "more commonly, when the endpoint..." will solve this.

abrunstrom:
Thanks for the suggestions @benthor! I agree on the refinements from @tfpauly, but this will improve the readability so much appreciated.

draft-ietf-taps-impl.md
benthor: Doesn't
> a useful model to use to

sound slightly odd and redundant?

How about simply saying
> it can serve as a useful model to illustrate connection racing

draft-ietf-taps-impl.md
benthor: We could change

> if, instead of a known IP address, the endpoint is identified by a hostname, then there may be a choice between multiple resolved addresses;

to simply state that

> an endpoint is more commonly identified by a hostname which may resolve into more than one IP address;

benthor: I think Tommy and Anna wanted to emphasize that having a name resolution step is the more common case.

gorryfair: I see, I'm a little unsure this specific change is really needed... and it doesn't seem to read well, but if it is needed then please consider  /which may/that can/ ... and /resolve into/resolve to/

draft-ietf-taps-impl.md
mwelzl: I don't understand what's slightly odd and redundant about the first version, but I guess both will work. This is for @tfpauly to answer though...

tfpauly:
Many of these changes have been overtaken by other PRs, and I don't think we have consensus on the other changes. As such, I'm closing this PR.

PR #1017, opened 2022-05-15
Clarify leaf node establishment

### LINKED ISSUE #1016: Implementation Draft Review: Section 4.1 - Choice of terms leads to confusing language
Raising this issue as part of my review of the the Implementation Draft.

Quoting from Section 4.1 "Structuring Candidates as a Tree"

~~~~~~~~~~
1 [www.example.com:80, Any, TCP]
  1.1 [www.example.com:80, Wi-Fi, TCP]
    1.1.1 [192.0.2.1:80, Wi-Fi, TCP]
  1.2 [www.example.com:80, LTE, TCP]
    1.2.1 [192.0.2.1:80, LTE, TCP]
    1.2.2 [2001:DB8::1.80, LTE, TCP]
~~~~~~~~~~

> When an implementation views this aggregate set of connection attempts as a single connection establishment, it only will use one of the leaf nodes to transfer data. Thus, when a single leaf node becomes ready to use, then the entire connection attempt is ready to use by the application. Another way to represent this is that every leaf node updates the state of its parent node when it becomes ready, until the trunk node of the tree is ready, which then notifies the application that the connection as a whole is ready to use.

Feedback: 

- While it is understood that "connection attempt" is here used to refer to a given node in the tree, rather than the actual (stateful) process of attempting to connect to the given endpoint (using the given parameters), it does not make for easy reading.
- Similarly, it is _possible_ to grasp how an "aggregate set" can be "viewed" as "a single connection establishment", but the language could be clearer here as well.
- Couldn't an "attempt" that has succeeded and is "ready to use" be referred to differently?
- Couldn't "candidate" or "option" be substituted for "attempt" to improve readability? 

 Proposal:

> When an implementation is asked to establish a single connection, only one of the leaf nodes in the candidate set is needed to transfer data. Thus, when a single leaf node becomes ready to use, then the entire connection attempt has been successful. One way to implement this is by having every leaf node update the state of its parent node when it becomes ready, until the trunk node of the tree is ready, which then notifies the application that the connection as a whole is ready to use.

tfpauly:
Closes #1016

### DIFF
--- a/draft-ietf-taps-impl.md
+++ b/draft-ietf-taps-impl.md
@@ -223,7 +223,7 @@ The rest of this section will use a notation scheme to represent this tree. The 
     1.2.2 [2001:DB8::1.80, LTE, TCP]
 ~~~~~~~~~~
 
-When an implementation views this aggregate set of connection attempts as a single connection establishment, it only will use one of the leaf nodes to transfer data. Thus, when a single leaf node becomes ready to use, then the entire connection attempt is ready to use by the application. Another way to represent this is that every leaf node updates the state of its parent node when it becomes ready, until the trunk node of the tree is ready, which then notifies the application that the connection as a whole is ready to use.
+When an implementation is asked to establish a single connection, only one of the leaf nodes in the candidate set is needed to transfer data. Thus, when a single leaf node becomes ready to use, then the entire connection attempt has been successful. One way to implement this is by having every leaf node update the state of its parent node when it becomes ready, until the trunk node of the tree is ready, which then notifies the application that the connection as a whole is ready to use.
 
 A connection establishment tree may be degenerate, and only have a single leaf node, such as a connection attempt to an IP address over a single interface with a single protocol.
 
--- a/draft-ietf-taps-impl.md
+++ b/draft-ietf-taps-impl.md
@@ -223,7 +223,7 @@ The rest of this section will use a notation scheme to represent this tree. The 
     1.2.2 [2001:DB8::1.80, LTE, TCP]
 ~~~~~~~~~~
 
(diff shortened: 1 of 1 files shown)


draft-ietf-taps-impl.md
```diff
@@ -223,7 +223,7 @@ The rest of this section will use a notation scheme to represent this tree. The
     1.2.2 [2001:DB8::1.80, LTE, TCP]
 ~~~~~~~~~~
 
-When an implementation views this aggregate set of connection attempts as a single connection establishment, it only will use one of the leaf nodes to transfer data. Thus, when a single leaf node becomes ready to use, then the entire connection attempt is ready to use by the application. Another way to represent this is that every leaf node updates the state of its parent node when it becomes ready, until the trunk node of the tree is ready, which then notifies the application that the connection as a whole is ready to use.
+When an implementation is asked to establish a single connection, only one of the leaf nodes in the candidate set is needed to transfer data. Thus, when a single leaf node becomes ready to use, then the entire connection attempt has been successful. One way to implement this is by having every leaf node update the state of its parent node when it becomes ready, until the trunk node of the tree is ready, which then notifies the application that the connection as a whole is ready to use.
```
abrunstrom, line 226: The "entire connection attempt" may still be confusing here. Perhaps better "entire connection establishment" or just "connection establishment"? Or we should replace "connection attempt" in the paragraph that follows, where it refers to an individual node.

philsbln: Perhaps "the connection object becomes ready" would be a clearer alternative here

tfpauly:
```suggestion
When an implementation is asked to establish a single connection, only one of the leaf nodes in the candidate set is needed to transfer data. Thus, once a single leaf node becomes ready to use, then the connection establishment tree is considered ready. One way to implement this is by having every leaf node update the state of its parent node when it becomes ready, until the trunk node of the tree is ready, which then notifies the application that the connection as a whole is ready to use.
```

PR #1018, opened 2022-05-16
add a stream-id: assignable and readable. Closes #1010

### LINKED ISSUE #1010: Add stream id's to Connection metadata
... combined with "if the transport is mux'ing, else ..."

mwelzl:
Closes #1010

### DIFF
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -1747,7 +1747,7 @@ Preconnection -> EstablishmentError<reason?>
 Connection Groups can be created using the Clone Action:
 
 ~~~
-Connection := Connection.Clone(framer?)
+Connection := Connection.Clone(framer?, streamid?)
 ~~~
 
 Calling Clone on a Connection yields a Connection Group containing two Connections: the parent
@@ -1792,7 +1792,9 @@ protocol instance.
 If the underlying protocol supports multi-streaming, it is natural to use this
 functionality to implement Clone. In that case, Connections in a Connection Group are
 multiplexed together, giving them similar treatment not only inside endpoints,
-but also across the end-to-end Internet path.
+but also across the end-to-end Internet path. Then, if possible, and if this optional
+parameter was provided, the parameter `streamid` can be used to assign a stream number
+to the stream in the underlying protocol.
 
 Note that calling Clone() can result in on-the-wire signaling, e.g., to open a new
 transport connection, depending on the underlying Protocol Stack. When Clone() leads to
@@ -2247,6 +2249,16 @@ Type:
 
 This numeric property represents the maximum Message size that an application can receive.
 It specified as the number of bytes.
+
+#### Stream ID {#conn-streamid}
+
+Name:
+: streamID
+
+Type:
+: Integer (non-negative) or `Not applicable`
+
+If the Connection is a stream of a multistreaming transport protocol, then this property represents the associated stream number. Otherwise, the value will be `Not applicable`.
 
 ## TCP-specific Properties: User Timeout Option (UTO) {#tcp-uto}
 


philsbln:
Does this make sense outside of a protocol-specific context?

I am aware that this is a different can of worms adding protocol specific parameters to clone, but this looks like a strange concept to add an Integer here that may have different side effects in QUIC and SCTP…

mwelzl:
That's a very good question. This feature request came up in the QuIC mapping part of the discussion in Vienna, see: https://datatracker.ietf.org/meeting/113/materials/minutes-113-taps-01

Not sure what to do: keep it as it is, remove it, or add the protocol:
- assign: "if protocol = x, then stream ID = y"
- read: "stream ID is y, and your protocol is x"
I have to say, I don't like this particular way, as it would complicate matters quite a bit. But how else to do it?

britram:
All of the multistreaming protocols we know about have an integer stream identifier.

I suppose a future protocol could use some sort of binary blob for identifying streams (which are... pedantically... representable as arbitrary-length integers)

tfpauly:
Interim discussion: let's make Clone() and Initiate() take a group of protocol-specific parameters (optionally) which allow protocols like QUIC and SCTP to extend this.

mwelzl:
I just made a very minimalist update, where "protocolSpecificProperties" is only a parameter of Clone(). Why not list it as readable properties, and why not as a parameter of Initiate, as we discussed?

- What's the point of saying "there can be protocol-specific properties which are currently not defined, and they can be readable"? This is for a future spec to say, when it proposes a new transport property.
- Regarding Initiate(), I think this should in fact not be a parameter to the Initiate Action itself, but a transport property to be supplied when creating a Preconnection. Again, this made me find that I would only write something like "There can also be transport properties that are not yet defined", which doesn't seem to be useful text.

In conclusion, I found that only the Clone() Action needs this as a new parameter. Everything else is for future mapping documents.

Thoughts?

draft-ietf-taps-interface.md
```diff
@@ -1747,7 +1747,7 @@ Preconnection -> EstablishmentError<reason?>
 Connection Groups can be created using the Clone Action:
 
 ~~~
-Connection := Connection.Clone(framer?)
+Connection := Connection.Clone(framer?, protocolSpecificProperties?)
```
philsbln, line 1750:
```suggestion
Connection := Connection.Clone(framer?, connectionProperties?)
```
I guess these could not only protocol specific properties, but any property that makes sense on a per-stream basis

draft-ietf-taps-interface.md
```diff
@@ -1767,6 +1767,10 @@ to the new Connection when calling Clone(), but in this case, a later change to
 `Connection Priority` on one Connection does not change it on the
 other Connections in the same Connection Group.
 
+The optional `protocolSpecificProperties` parameter represents protocol-specific
+Transport Properties that are not defined in the present document, but left open for definition
+in future documents.
+
```
philsbln, line 1773:
```suggestion
The optional `connectionProperties` parameter allows passing
Transport Properties that control the behaviour of the stream/underlaying connection to be created, e.g., protocol specific properties to request specific stream IDs for SCTP or QUIC.

```

philsbln (changes requested):
Thank you for your analysis – I really like the minimalistic approach.

I would prefer a less editorial framing as the one proposed.

draft-ietf-taps-interface.md
```diff
@@ -1767,6 +1767,9 @@ to the new Connection when calling Clone(), but in this case, a later change to
 `Connection Priority` on one Connection does not change it on the
 other Connections in the same Connection Group.
 
+The optional `connectionProperties` parameter allows passing
+Transport Properties that control the behaviour of the stream/underlaying connection to be created, e.g., protocol specific properties to request specific stream IDs for SCTP or QUIC.
```
mwelzl, line 1771:
```suggestion
Transport Properties that control the behavior of the underlying stream or connection to be created, e.g., protocol specific properties to request specific stream IDs for SCTP or QUIC.
```

mwelzl:
Many thanks!  I incorporated your suggestion, and made a tiny edit to it (behaviour => behavior, underlaying => underlying, and "stream or connection" instead of "stream/connection").

philsbln (approved):
Thanks – I guess we have a fairly good solution.

tfpauly (approved):

PR #1019, opened 2022-05-16
A few nits; primarily adding "non-negative" in some places

mwelzl:
An alternative would be to remove negative values from the definition of Integer and Numeric, because I think we have no single negative usage anywhere. Still ... to me, this PR seems cleaner (more forward-looking, as we might want to support negative values in some future update?)


philsbln:
LGMT

tfpauly (approved):

PR #1021, opened 2022-05-17
Sketch alternative multicast model

csperkins:
Sketch of alternative multicast model. incomplete.

#874

### DIFF
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -867,70 +867,69 @@ LocalSpecifier.WithStunServer(address, port, credentials)
 
 ### Multicast Examples {#multicast-examples}
 
-Specify a Local Endpoint using an Any-Source Multicast group to join on a named local interface:
+The following examples show how multicast groups can be used.
+
+Join an Any-Source Multicast group in receive-only mode, bound to a known
+port on a named local interface:
 
 ~~~
-LocalSpecifier := NewLocalEndpoint()
-LocalSpecifier.WithIPv4Address(233.252.0.0)
-LocalSpecifier.WithInterface("en0")
+   RemoteSpecifier := NewRemoteEndpoint()
+   RemoteSpecifier.WithIPv4Address(233.252.0.0)
+
+   LocalSpecifier := NewLocalEndpoint()
+   LocalSpecifier.WithPort(5353)
+   LocalSpecifier.WithInterface("en0")
+
+   TransportProperties := ...
+   SecurityParameters  := ...
+
+   Preconnection := newPreconnection(LocalSpecifier,
+                                     RemoteSpecifier,
+                                     TransportProperties,
+                                     SecurityProperties)
+   Listener := Preconnection.Listen()
 ~~~
 
-Source-Specific Multicast requires setting both a Local and Remote Endpoint:
+Create a Source-Specific Multicast group as a sender:
 
 ~~~
-LocalSpecifier := NewLocalEndpoint()
-LocalSpecifier.WithIPv4Address(232.1.1.1)
-LocalSpecifier.WithInterface("en0")
+   RemoteSpecifier := NewRemoteEndpoint()
+   RemoteSpecifier.WithIPv4Address(232.1.1.1)
 
-RemoteSpecifier := NewRemoteEndpoint()
-RemoteSpecifier.WithIPv4Address(192.0.2.22)
+   LocalSpecifier := NewLocalEndpoint()
+   LocalSpecifier.WithIPv4Address(192.0.2.22)
+   RemoteSpecifier.WithInterface("en0")
+
+   TransportProperties := ...
+   SecurityParameters  := ...
+
+   Preconnection := newPreconnection(LocalSpecifier,
+                                     RemoteSpecifier,
+                                     TransportProperties,
+                                     SecurityProperties)
+   Connection := Preconnection.Initiate()
 ~~~
 
-One common pattern for multicast is to both send and receive multicast. For such
-cases, an application can set up both a Listener and a Connection. The Listener
-is only used to accept Connections that receive inbound multicast. The initiated
-Connection is only used to send multicast.
+Join an any-source multicast group as both a sender and a receiver:
 
 ~~~
-// Prepare multicast Listener
-LocalMulticastSpecifier := NewLocalEndpoint()
-LocalMulticastSpecifier.WithIPv4Address(233.252.0.0)
-LocalMulticastSpecifier.WithPort(5353)
-LocalMulticastSpecifier.WithInterface("en0")
+   RemoteSpecifier := NewRemoteEndpoint()
+   RemoteSpecifier.WithIPv4Address(233.252.0.0)
+   RemoteSpecifier.WithPort(5353)
+   RemoteSpecifier.WithInterface("en0")
 
-TransportProperties := NewTransportProperties()
-TransportProperties.Require(preserve-msg-boundaries)
-// Reliable Data Transfer and Preserve Order are Required by default
-                                  
-// Specifying a Remote Endpoint is optional when using Listen()
-Preconnection := NewPreconnection(LocalMulticastSpecifier,
-                                  TransportProperties,
-                                  SecurityParameters)
+   LocalSpecifier := NewLocalEndpoint()
+   LocalSpecifier.WithIPv4Address(192.0.2.22)
+   LocalSpecifier.WithPort(5353)
 
-MulticastListener := Preconnection.Listen()
+   TransportProperties := ...
+   SecurityParameters  := ...
 
-// Handle inbound messages sent to the multicast group
-MulticastListener -> ConnectionReceived<MulticastReceiverConnection>
-MulticastReceiverConnection.Receive()
-MulticastReceiverConnection -> Received<messageDataRequest, messageContext>
-
-// Prepare Connection to send multicast
-LocalSpecifier := NewLocalEndpoint()
-LocalSpecifier.WithPort(5353)
-LocalSpecifier.WithInterface("en0")
-RemoteMulticastSpecifier := NewRemoteEndpoint()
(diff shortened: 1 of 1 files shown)


draft-ietf-taps-interface.md
```diff
@@ -737,23 +737,45 @@ early binding when required, for example with some Network Address Translator
 
 ### Using Multicast Endpoints
 
-Specifying a multicast group address on a Local Endpoint will indicate to the Transport
-Services system that the resulting connection will be used to receive multicast messages. The
-Remote Endpoint can be used to filter incoming multicast from specific senders. Such
-a Preconnection will only support calling Listen(), not Initiate(). Calling Listen()
-will cause the Transport Services system to register for receiving multicast, such
-as issuing an IGMP join {{?RFC3376}} or using MLD for IPV6 {{?RFC4604}}. Any Connections
+When using multicast the Remote Endpoint will indicate the multicast group
```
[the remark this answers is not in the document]
csperkins: Yes, will fix.

csperkins:
I think this is ready to merge.

draft-ietf-taps-interface.md
```diff
@@ -737,23 +737,39 @@ early binding when required, for example with some Network Address Translator
...
+
+Calling Initiate() on that Preconnection creates a Connection that can be
+used to send messages to the multicast group. The Connection object that is
+created will support Send() but not Receive(). Any Connections created this
+way are send-only, and do not join the multicast group. The resulting
+Connection will have a Local Endpoint indicating the local interface to
+which the connection is bound and a Remote Endpoint indicating the
+multicast group.
+
+Calling Listen() on a Preconnection with a multicast group as the Remote
```
MaxF12, line 751: I am unsure on how you would join SSM with these semantics. Maybe add an example for that as well? Where do you specify the source address?

MaxF12 (changes requested):

draft-ietf-taps-interface.md
```diff
@@ -851,70 +867,69 @@ LocalSpecifier.WithStunServer(address, port, credentials)
...
+Join an Any-Source Multicast group in receive-only mode, bound to a known
+port on a named local interface:
 
 ~~~
-LocalSpecifier := NewLocalEndpoint()
-LocalSpecifier.WithIPv4Address(233.252.0.0)
-LocalSpecifier.WithInterface("en0")
-~~~
+   RemoteSpecifier := NewRemoteEndpoint()
+   RemoteSpecifier.WithIPv4Address(233.252.0.0)
```
philsbln, line 877:
```suggestion
   RemoteSpecifier := NewRemoteEndpoint()
   RemoteSpecifier.WithIPv4MulticastGroup(233.252.0.0)
```

draft-ietf-taps-interface.md
```diff
@@ -851,70 +867,69 @@ LocalSpecifier.WithStunServer(address, port, credentials)
...
-LocalSpecifier.WithIPv4Address(232.1.1.1)
-LocalSpecifier.WithInterface("en0")
 
-RemoteSpecifier := NewRemoteEndpoint()
-RemoteSpecifier.WithIPv4Address(192.0.2.22)
+Create a Source-Specific Multicast group as a sender:
+
 ~~~
+   RemoteSpecifier := NewRemoteEndpoint()
+   RemoteSpecifier.WithIPv4Address(232.1.1.1)
```
philsbln, line 897:
```suggestion
   RemoteSpecifier.WithIPv4MulticastGroup(232.1.1.1)
```

draft-ietf-taps-interface.md
```diff
@@ -851,70 +867,69 @@ LocalSpecifier.WithStunServer(address, port, credentials)
...
-// Reliable Data Transfer and Preserve Order are Required by default
+Join an any-source multicast group as both a sender and a receiver:
 
-// Specifying a Remote Endpoint is optional when using Listen()
-Preconnection := NewPreconnection(LocalMulticastSpecifier,
-                                  TransportProperties,
-                                  SecurityParameters)
+~~~
+   RemoteSpecifier := NewRemoteEndpoint()
+   RemoteSpecifier.WithIPv4Address(233.252.0.0)
```
philsbln, line 917:
```suggestion
   RemoteSpecifier.WithIPv4MulticastGroup(233.252.0.0)
```

draft-ietf-taps-interface.md
philsbln: As discussed in the interim, we will habe an extra attribute for the remote endpoint to specify the multicast group.
```suggestion
Calling Listen() on a Preconnection with a multicast group specified on the Remote
```

philsbln (changes requested):
Lets' apply the solution discussed in the Interim first and then have another pass adding the Multicast group to the remote endpoint and adding text for SSM

philsbln (approved):
These changes should implement the consent from the interim, but I really would like to see a review from @GrumpyOldTroll

draft-ietf-taps-interface.md
```diff
@@ -737,23 +737,45 @@ early binding when required, for example with some Network Address Translator
...
+LocalSpecifier.WithPort(5353)
+```
+
+Calling Initiate() on that Preconnection creates a Connection that can be
+used to send messages to the multicast group. The Connection object that is
+created will support Send() but not Receive(). Any Connections created this
+way are send-only, and do not join the multicast group. The resulting
+Connection will have a Local Endpoint indicating the local interface to
+which the connection is bound and a Remote Endpoint indicating the
+multicast group.
```
GrumpyOldTroll, line 755: In sockets the semantics for sending to a destination group are identical to sending to a destination unicast address, except for 2 things I could think of:
 - TTL/Hop Limit defaults to 1 when sending to group addresses with sockets, and must be explicitly overridden with a setsockopt to get a longer limit.  (The value 1 is the most appropriate default for L2 discovery use cases of multicast, but typically causes a few hours of debugging for bulk data delivery use cases that want the traffic to get routed for multiple hops.)
 - The destination MAC address for ethernet has a static mapping that's implied for multicast destination addresses.

I guess I'm OK with using WithIPxMulticastGroup for sending, though I think for sending it's much less helpful than it is for receiving (as compared to just setting the RemoteSpecifier WithIPxAddress for the destination group address when sending).  Also, I think this might be more confusing than helpful if we don't also use another different thing for broadcast addresses, which are different from both multicast and unicast, but like multicast, also not different in ways that are usually very important ways for senders to distinguish.

So I tend to think it might make more sense to just use WithIPxAddress to set the group destination address on the RemoteSpecifier when sending, and leave WithIPxMulticastGroup (including an ASM/SSM distinguisher) to ask for the appropriate join operation for receiving, but my opinion on this point is not strong.

draft-ietf-taps-interface.md
```diff
@@ -737,23 +737,45 @@ early binding when required, for example with some Network Address Translator
...
+
+Calling Listen() on a Preconnection with a multicast group specified on the Remote
```
GrumpyOldTroll, line 757: (As above, I would recommend s/Remote Endpoint/Local Endpoint/ for where the multicast group should be set.)

draft-ietf-taps-interface.md
```diff
@@ -737,23 +737,45 @@ early binding when required, for example with some Network Address Translator
...
+Endpoint will join the multicast group to receive messages. This Listener
+will create one Connection for each Remote Endpoint sending to the group,
+with the Local Endpoint set to the group address. The set of Connection
+objects created forms a Connection Group.
+
+Calling Rendezvous() on a Preconnection with an any-source multicast group
```
GrumpyOldTroll, line 763: For bidirectional ASM, to me it makes better sense to both set both RemoteSpecifier address via WithIPxAddress to the group address and also set the LocalSpecifier group via WithIPxMulticastGroup.  I have no objection to using Rendezvous for this use case, that makes decent sense to me.

draft-ietf-taps-interface.md
```diff
@@ -737,23 +737,45 @@ early binding when required, for example with some Network Address Translator
...
+address as the Remote Endpoint will join the multicast group, and also
+indicates that the resulting connection can be used to send messages to the
+multicast group. The Rendezvous() call will return both a Connection that
+can be used to send to the group, that acts the same as a connection
+returned by calling Initiate() with a multicast Remote Endpoint, and a
+Listener that acts as if Listen() had been called with a multicast Remote
+Endpoint.
+
+Calling Rendezvous() on a Preconnection with a source-specific multicast
+group address as the Remote Endpoint results in an EstablishmentError.
```
GrumpyOldTroll, line 773: SSM group address is not entirely knowable just from the address value, so I'm not sure this is well-founded.

For instance the administratively scoped space (239/8 in IP4 IIRC) I think can by design be either according to how the site wants to treat it, and in practice other addresses outside of 232/8 or ff3x::/96 can likewise generally be configured as ssm optionally in most routers.  And also ASM joins can be requested on the reserved ssm addresses.

So this condition also would be much better handled by prohibiting Rendevous() this way only according to an explicit SSM indicator as part of WithIPxMulticastGroup().  (Or similar, such as a WithSSM(true) or something, though I think it's probably better to make it 'WithIPxMulticastGroup(addr, SSM)', since I don't think an ASM/SSM indicator makes sense unless there is a MulticastGroup.

GrumpyOldTroll (commented):
Main recommendations:
- (strong) use LocalSpecifier.WithIPxMulticastGroup() rather than RemoteSpecifier.WithIPxMulticastGroup() for receive
- (strong) pass an ASM/SSM specifier to WithIPxMulticastGroup() instead of inferring it from group address values
- (medium) use RemoteSpecifier.WithIPxAddress(group_address) for sending, and leave WithIPxMulticastGroup() as a LocalSpecifier-only call indicating a join + receive is desired.

HTH. -Jake

philsbln:
I suggest closing this one and merging #1050 instead

tfpauly:
I agree, since this seems to have stalled out

PR #1022, opened 2022-05-17
Soften order of operations text

### LINKED ISSUE #536: Branching Order-of-Operations
In section 4.1.4 in the impl draft, we have:

"The order of operations for branching, where lower numbers are acted upon first, should be:

1. Alternate Paths
2. Protocol Options
3. Derived Endpoints"

However, I would say there is not really a strict ordering required. The text actually indicates that selection of the protocol is rather independent while there is usually a dependency between the interface and derived endpoint. I think this is related to my other issue about the use of the term path, because a path consists of both the local and remote endpoint and therefore there is of course a dependency. 

So in summary I'm not sure how useful I find this whole section as currently written...

tfpauly:
Closes #536

### DIFF
--- a/draft-ietf-taps-impl.md
+++ b/draft-ietf-taps-impl.md
@@ -266,7 +266,7 @@ DNS-Based Service Discovery {{?RFC6763}} can also provide an endpoint derivation
 
 Applications can influence which derived endpoints are allowed and preferred via Selection Properties set on the Preconnection. For example, setting a preference for `useTemporaryLocalAddress` would prefer the use of IPv6 over IPv4, and requiring `useTemporaryLocalAddress` would eliminate IPv4 options, since IPv4 does not support temporary addresses.
 
-#### Alternate Paths
+#### Network Paths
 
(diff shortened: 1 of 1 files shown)


gorryfair:
SORRY I made a PR on your PR, when I meant to leave a review comment, see 1024.

draft-ietf-taps-impl.md
```diff
@@ -319,11 +319,11 @@ Implementations that support racing protocols and protocol options should mainta
...
+Branch types should occur in a specific order relative to one another to avoid creating leaf nodes with invalid or incompatible settings. In the example above, it would be invalid to branch for derived endpoints (the DNS results for www.example.com) before branching between interface paths, since there are situations when the results will be different across networks due to private names or different supported IP versions. Implementations need to be careful to branch in a consistentorder that results in usable leaf nodes whenever there are multiple branch types that could be used from a single node.
```
gorryfair, line 322: /should occur/ought to occur/ ?
/consistentorder/consistent order/ - with space

tfpauly:
```suggestion
Branch types ought to occur in a specific order relative to one another to avoid creating leaf nodes with invalid or incompatible settings. In the example above, it would be invalid to branch for derived endpoints (the DNS results for www.example.com) before branching between interface paths, since there are situations when the results will be different across networks due to private names or different supported IP versions. Implementations need to be careful to branch in a consistent order that results in usable leaf nodes whenever there are multiple branch types that could be used from a single node.
```

PR #1023, opened 2022-05-17
Explain use of framers for connectionless protocols

### LINKED ISSUE #538: Interface for app-driven racing of "unconnected" protocols?
Impl draft Section 4.5: Handling racing with "unconnected" protocols
"In the case that the application signals that the initial Protocol Stack is failing for some reason and that another option should be attempted, the Connection can be updated to point to the next candidate Protocol Stack. This can be viewed as an application-driven form of Protocol Stack racing."
How does the application indicate that? I don't think this is covered by the API draft?

Also I would recommend to move this section as a subsection within 4.5. (Completing Establishment) or at least switch the order with the previous section (Establishing multiplexed connections) to get a better reading flow.

tfpauly:
Closes #538

### DIFF
--- a/draft-ietf-taps-impl.md
+++ b/draft-ietf-taps-impl.md
@@ -484,7 +484,7 @@ Instead, calling the ConnectionReceived event could be delayed until the first M
 
 While protocols that use an explicit handshake to validate a Connection to a peer can be used for racing multiple establishment attempts in parallel, connectionless protocols such as raw UDP do not offer a way to validate the presence of a peer or the usability of a Connection without application feedback. An implementation should consider such a Protocol Stack to be established as soon as the Transport Services system has selected a path on which to send data.
 
-However, this can cause a problem if a specific peer is not reachable over the network using the connectionless protocol, or data cannot be exchanged with the peer for any other reason. To mitigate this, an application can use a Message Framer {{message-framers}} on top of a connectionless protocol to only mark a specific connection attempt as ready when some data has been received, or after some application-level handshake has been performed. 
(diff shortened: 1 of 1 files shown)


chrysn:
That change would have helped me understand what framers can do. It seems a bit out of sync with " A Framer can encapsulate or encode outbound Messages, and decapsulate or decode inbound data into Messages." which sounds like it'd preclude operations like processing received data (or whole messages) into connection object state without emitting a message again -- maybe that sentence could be extended along with it (if even just to include "for example" or "or modify the connection object").

mwelzl (approved):

tfpauly:
@chrysn we've added some clarifying text in the message framer definition

draft-ietf-taps-impl.md
```diff
@@ -484,7 +484,7 @@ Instead, calling the ConnectionReceived event could be delayed until the first M
...
 
-However, if a peer is not reachable over the network using the connectionless protocol, or data cannot be exchanged for any other reason, the application may want to attempt using another candidate Protocol Stack. The implementation should maintain the list of other candidate Protocol Stacks that were eligible to use.
+However, this can cause a problem if a specific peer is not reachable over the network using the connectionless protocol, or data cannot be exchanged with the peer for any other reason. To mitigate this, an application can use a Message Framer {{message-framers}} on top of a connectionless protocol to only mark a specific connection attempt as ready when some data has been received, or after some application-level handshake has been performed.
```
tfpauly, line 487:
```suggestion
However, this can cause a problem if a specific peer is not reachable over the network using the connectionless protocol, or data cannot be exchanged with the peer for any other reason. To mitigate this, an application can use a Message Framer {{message-framers}} on top of a connectionless protocol to only mark a specific connection attempt as ready when some data has been received, or after some application-level handshake has been performed.
```

PR #1033, opened 2022-05-20
Reading Selection Properties, etc.  Resolves #1012

### LINKED ISSUE #1012: Interface draft: Different Property Types and States still confusing
As [promised](https://datatracker.ietf.org/meeting/113/materials/minutes-113-taps-01), here follows a bit of feedback on the different Property types in the Interface draft, from the perspective of an academic developer who is currently attempting to follow the draft spec as closely as possible while trying to implement a path-aware transport system in a [statically typed programming language](https://golang.org)

In [Chapter 4 of the Interface draft](https://ietf-tapswg.github.io/api-drafts/draft-ietf-taps-interface.html#name-transport-properties), it says:

> Transport Properties are divided into Selection, Connection, and Message Properties. Selection Properties [...] can only be set during pre-establishment. They are only used to specify which paths and protocol stacks can be used and are preferred by the application. Although Connection Properties [...] can be set during pre-establishment, they may be changed later. They are used to inform decisions made during establishment and to fine-tune the established connection. 

At this point, the reader of the document could be expected to conclude the following:

- **Selection Properties** are a set of preferences for protocol/path selection that (obviously) can no longer be meaningfully changed once a connection has been established.
- **Connection Properties** are a set of "run-time preferences" for connections which could still be adjusted when a connection is already established, but should ideally already be known during protocol/path selection

The text goes on to state:

> Calling Initiate on a Preconnection creates an outbound Connection or a Listener, and the Selection Properties remain readable from the Connection or Listener, but become immutable.

With this wording, the reader can now reasonably conclude that the original Selection Properties that have resulted in a given Connection should remain available during the lifetime of that Connection, perhaps in form of a copy of the original data structure. So far so straight forward.

In [Section 6.2](https://ietf-tapswg.github.io/api-drafts/draft-ietf-taps-interface.html#section-6.2) however, this conclusion is called into question. After the text has gone on to introduce the 5 different Preference Levels, it states

> Many properties are only considered during establishment, and can not be changed after a Connection is established; however, they can still be queried. The return type of a queried Selection Property is Boolean, where `true` means that the Selection Property has been applied and `false` means that the Selection Property has not been applied.

At this point, the reader becomes confused.

- Why is the text talking about "many properties" that can no longer be changed, instead of relying on the earlier, nice and clear distinction between Selection and Connection Properties?
- Where earlier, Selection Properties were said to remain _readable_, now they are only _queryable_?
- The type of a Selection Property that has been "applied" mutates from a Preference Level to become a Boolean?
- Are "applied" Selection Properties still Selection Properties? Shouldn't they better be referred to using a different term?
- So the original Selection _Preferences_ that resulted in the current Connection are not available after all? Only the (Boolean) _Properties_ of the chosen Protocol can be "queried"?

But wait, the text goes on:

> Note that `true` does not mean that a request has been honored. For example, if `Congestion control` was requested with preference level `Prefer`, but congestion control could not be supported, querying the `congestionControl` property yields the value `false`. If the preference level `Avoid` was used for `Congestion control`, and, as requested, the Connection is not congestion controlled, querying the `congestionControl` property also yields the value `false`.

This is a complicated way of saying that the values _only_ reflect the _actual_ Properties of the chosen Connection and (in direct contradiction to the previous paragraph) do _not_ actually indicate whether a Selection Property has also been "applied" or not. (A Boolean only carries exactly one bit of information after all.)

In other words, the _Protocol Properties_ can be queried. The usefulness of this is probably limited however. If we already know that our Connection is using, e.g., TCP, we don't have to issue any "queries" to know that the Connection offers reliable transport, is subject to congestion control and preserves packet ordering. (This is indeed how I have chosen to implement this. The chosen _Protocol_ can be queried for the Properties it inherently has. In the Preconnection phase, the Selection _Preferences_ are matched against these and any `Require`/`Prohibit` conflict results in that Protocol being dropped from consideration.)

---

Hoping not to be unreasonable (or somehow block-headed), I have the following suggestions

- Either introduce a new term for Properties that have been "applied", or
- Use the term "Preferences" (e.g., "Selection Preferences", "Connection Preferences") to refer to a collection of adjustable "Settings" and the term "Properties" to refer to measured or queried values that _are_ actual Properties of underlying Connections and Protocols
- Go into a bit more detail about what the draft really means when talking about "querying" Properties (versus setting them), maybe even in terms of exemplary interactions with Property data structures. (After all, the draft elsewhere goes into considerable detail with it's exemplary Event-based system)
- Think about whether it in general would clear things up when the explicit notion of Protocol Properties as a (static) counterpoint to (adjustable) Selection Properties is introduced

mwelzl:
Resolves #1012.

- @benthor wrote: "Why is the text talking about "many properties" that can no longer be changed, instead of relying on the earlier, nice and clear distinction between Selection and Connection Properties?". => In line with the suggestion from @theri, this is now changed, it says Selection Properties.
- @benthor wrote: "Where earlier, Selection Properties were said to remain readable, now they are only queryable?". =>  In line with @theri's answer, this PR replaces "query" with "read" in this text.
- @benthor wrote: "Are "applied" Selection Properties still Selection Properties? Shouldn't they better be referred to using a different term?"  and here, @theri suggested: "In the [Architecture doc](https://www.ietf.org/archive/id/draft-ietf-taps-arch-12.html#name-transport-services-implemen), we define Protocol Stack as a Transport Services implementation concept, and we talk about a certain protocol stacks providing certain features (see [Transport Features defined in the minset RFC 8923](https://www.rfc-editor.org/rfc/rfc8923). Perhaps we could rephrase this part of the API draft to use that terminology, e.g., the application can query the actually used protocol stack and then know which Transport Features it supports."   => I don't know how to apply this change, and it generally sounds a bit complicated to me. I think that Selection Properties should keep their name.
- Regarding the confusing text with the `congestion control` example, I removed this and I think it makes things clearer.

Regarding the return type, I think I now finally understood the bigger problem. Returning a Boolean instead of a Preference should make sense and should be easy to implement (as also described in this PR). The problem is that not *all* Selection Properties are of type Preference. E.g., `multipath` is an enumeration, representing `disabled`, `active` or `passive`. What sense does it make if this becomes only `true` or `false`?  Similarly, `interface` and `pvd` are of type "Collection of (Preference, Enumeration)", and indeed it would make sense to expect a Collection of booleans, not a singular boolean.

This PR now explicitly says that, upon reading, the *Preference type* of a Selection Property (not the whole Selection Property!) changes into Boolean.

### DIFF
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -1001,17 +1001,17 @@ via a Preconnection, by configuration after establishment, or by inheriting them
 from an antecedent via cloning; see {{groups}} for more.
 
 {{connection-props}} provides a list of Connection Properties, while Selection
-Properties are listed in the subsections below. Many properties are
+Properties are listed in the subsections below. Sepection Properties are
 only considered during establishment, and can not be changed after a Connection
-is established; however, they can still be queried. The return type of a queried
+is established; however, they can still be read. The return type of a read
 Selection Property is Boolean, where `true` means that the Selection Property
 has been applied and `false` means that the Selection Property has not
 been applied. Note that `true` does not mean that a request has been honored.
 For example, if `Congestion control` was
 requested with preference level `Prefer`, but congestion control could not
-be supported, querying the `congestionControl` property yields the
+be supported, reading the `congestionControl` property yields the
 value `false`. If the preference level `Avoid` was used for `Congestion control`,
-and, as requested, the Connection is not congestion controlled, querying
+and, as requested, the Connection is not congestion controlled, reading
 the `congestionControl` property also yields the value `false`.
 
 An implementation of the Transport Services API must provide sensible defaults for Selection
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -1004,7 +1004,7 @@ from an antecedent via cloning; see {{groups}} for more.
 Properties are listed in the subsections below. Selection Properties are
 only considered during establishment, and can not be changed after a Connection
 is established; however, they can still be read. Upon reading, the Preference type
-of a Selection Property changes into Boolean, where where `true` means that the
+of a Selection Property changes into Boolean, where `true` means that the
 Preference was honored and `false` means that it has not been honored (implementations
 of Transport Services systems may alternatively use the two Preference values `Require`
 and `Prohibit` to represent `true` and `false`, respectively).
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -1001,7 +1001,7 @@ via a Preconnection, by configuration after establishment, or by inheriting them
 from an antecedent via cloning; see {{groups}} for more.
 
 {{connection-props}} provides a list of Connection Properties, while Selection
-Properties are listed in the subsections below. Sepection Properties are
+Properties are listed in the subsections below. Selection Properties are
 only considered during establishment, and can not be changed after a Connection
 is established; however, they can still be read. The return type of a read
 Selection Property is Boolean, where `true` means that the Selection Property
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -1003,16 +1003,11 @@ from an antecedent via cloning; see {{groups}} for more.
 {{connection-props}} provides a list of Connection Properties, while Selection
 Properties are listed in the subsections below. Selection Properties are
 only considered during establishment, and can not be changed after a Connection
-is established; however, they can still be read. The return type of a read
-Selection Property is Boolean, where `true` means that the Selection Property
-has been applied and `false` means that the Selection Property has not
-been applied. Note that `true` does not mean that a request has been honored.
-For example, if `Congestion control` was
-requested with preference level `Prefer`, but congestion control could not
-be supported, reading the `congestionControl` property yields the
-value `false`. If the preference level `Avoid` was used for `Congestion control`,
-and, as requested, the Connection is not congestion controlled, reading
-the `congestionControl` property also yields the value `false`.
+is established; however, they can still be read. Upon reading, the Preference type
+of a Selection Property changes into Boolean, where where `true` means that the
+Preference was honored and `false` means that it has not been honored (implementations
+of Transport Services systems may alternatively use the two Preference values `Require`
+and `Prohibit` to represent `true` and `false`, respectively).
 
 An implementation of the Transport Services API must provide sensible defaults for Selection
 Properties. The default values for each property below represent a


theri (changes requested):
Thanks for taking a stab at this! I agree with most changes, but I still think the "request has been honored" language is confusing.

draft-ietf-taps-interface.md
```diff
@@ -1001,18 +1001,13 @@ via a Preconnection, by configuration after establishment, or by inheriting them
...
-requested with preference level `Prefer`, but congestion control could not
-be supported, querying the `congestionControl` property yields the
-value `false`. If the preference level `Avoid` was used for `Congestion control`,
-and, as requested, the Connection is not congestion controlled, querying
-the `congestionControl` property also yields the value `false`.
+is established; however, they can still be read. Upon reading, the Preference type
+of a Selection Property changes into Boolean, where `true` means that the
+Preference was honored and `false` means that it has not been honored (implementations
+of Transport Services systems may alternatively use the two Preference values `Require`
+and `Prohibit` to represent `true` and `false`, respectively).
```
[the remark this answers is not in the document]
mwelzl: I see now, and I agree so very much!  Many thanks, this is a great change!

mwelzl:
MANY thanks for the proposed fix (which I committed)!  I hadn't really understood the problem just from reading the issue discussion.

theri (approved):
Great, this looks good to me now. I think now it is simple enough that we don't need an example anymore.

tfpauly (approved):

draft-ietf-taps-interface.md
```diff
@@ -1001,18 +1001,13 @@ via a Preconnection, by configuration after establishment, or by inheriting them
...
-requested with preference level `Prefer`, but congestion control could not
-be supported, querying the `congestionControl` property yields the
-value `false`. If the preference level `Avoid` was used for `Congestion control`,
-and, as requested, the Connection is not congestion controlled, querying
-the `congestionControl` property also yields the value `false`.
+is established; however, they can still be read. Upon reading, the Preference type
+of a Selection Property changes into Boolean, where `true` means that the
+selected Protocol Stack supports the feature or uses the path associated with the Selection Property, and `false` means that the Protocol Stack does not support the feature or use the path. Implementations
+of Transport Services systems may alternatively use the two Preference values `Require`
+and `Prohibit` to represent `true` and `false`, respectively.
```
tfpauly, line 1010:
```suggestion
is established. After a Connection is established, Selection Properties can only
be read to check the properties used by the Connection. Upon reading, the
Preference type of a Selection Property changes into Boolean, where `true` means
that the selected Protocol Stack supports the feature or uses the path associated with the Selection Property, and `false` means that the Protocol Stack does not support the feature or use the path. Implementations
of Transport Services systems may alternatively use the two Preference values `Require`
and `Prohibit` to represent `true` and `false`, respectively.
```

tfpauly (approved):
Looks good, one clarification suggested

draft-ietf-taps-interface.md
philsbln:
```suggestion
is established. After a Connection is established, Selection Properties can only
be read to check the properties used by the Connection. Upon reading, the
Preference type of a Selection Property changes into Boolean, where `true` means
that the selected Protocol Stack supports the feature or uses the path conforms with the Selection Property, and `false` means that the Protocol Stack does not support the feature or use a conforming path. Implementations
of Transport Services systems may alternatively use the two Preference values `Require`
and `Prohibit` to represent `true` and `false`, respectively or expose the supported features of the selected Protocol Stack and path in a different way that is consistent for all Preference typed Selection Properties.
```

philsbln (approved):
Works for me (with @tfpauly's changes).
I would propose a slight rewording `path associated with` -> `path conforming with` and add a hint that this exposing can also be implemented differently.

Issue #1026, opened 2022-05-19
Review by CA - Datagram v Connectionless


gorryfair:
* "Connection establishment is only a local operation for a Datagram
  transport (e.g., UDP(-Lite))": I think the criterion is not Datagram
  transports but "transports without a handshake" or "connectionless
  protocols" as it is used later in the document.

gorryfair:
I wrote, and I suggest we use "connectionless protocols" explicitly mentioning examples of UDP and UDP-Lite when used as a transport without a handshake.

mwelzl:
To whoever plans to address this: I checked occurrences of the word "datagram" in the architecture and API drafts, and I found (only) one sentence that needs fixing on similar grounds, I think, in the API draft:

> Guidance on setting this value for datagram transports is provided in [RFC8085].

in Section 8.1.4 "Timeout for keep alive packets".   Surely the guidance regarding this matter in the UDP usage RFC is related to the fact that UDP(-Lite) is connectionless, not that it's datagram based.

gorryfair:
Starting #1034 to fix this.

PR #1035, opened 2022-05-26
 PR to resolve Datagram v Connectionless #1026

gorryfair:
replace DG by CL.

Changed one oddity: /as to/ to /about/


eyeballs
```diff
@@ -168,8 +168,8 @@ The process of establishing a network connection begins when an application expr
 
 Connection establishment is divided into two top-level steps: Candidate Gathering, to identify the paths, protocols, and endpoints to use, and Candidate Racing (see Section 4.2.2 of {{I-D.ietf-taps-arch}}), in which the necessary protocol handshakes are conducted so that the transport system can select which set to use.
 
-This document structures the candidates for racing as a tree as terminological convention. While a 
-a tree structure is not the only way in which racing can be implemented, it does ease the illustration of how racing works. 
+As a terminological convention, this document structures the candidates for racing as a tree. While a
+a tree structure is not the only way in which racing can be implemented, it does ease the illustration of how racing works.
```
mwelzl, line 172:
```suggestion
tree structure is not the only way in which racing can be implemented, it does ease the illustration of how racing works.
```

eyeballs
```diff
@@ -552,7 +556,7 @@ Since sending a Message may involve a context switch between the application and
...
+Similar to sending, Receiving a Message is determined by the top-level protocol in the established Protocol Stack. The main difference with Receiving is that the size and boundaries of the Message are not known beforehand. The application can communicate in its Receive action the parameters for the Message, which can help the Transport Services implementation know how much data to deliver and when. For example, if the application only wants to receive a complete Message, the implementation should wait until an entire Message (datagram, stream, or frame) is read before delivering any Message content to the application. This requires the implementation to understand where messages end, either via a supplied  or because the top-level protocol in the established Protocol Stack preserves message boundaries. If the top-level protocol only supports a byte-stream and no framers were supported, the application can control the flow of received data by specifying the minimum number of bytes of Message content it wants to receive at one time.
```
mwelzl, line 559:
```suggestion
Similar to sending, receiving a Message is determined by the top-level protocol in the established Protocol Stack. The main difference with receiving is that the size and boundaries of the Message are not known beforehand. The application can communicate in its Receive action the parameters for the Message, which can help the Transport Services implementation know how much data to deliver and when. For example, if the application only wants to receive a complete Message, the implementation should wait until an entire Message (datagram, stream, or frame) is read before delivering any Message content to the application. This requires the implementation to understand where messages end, either via a supplied  or because the top-level protocol in the established Protocol Stack preserves message boundaries. If the top-level protocol only supports a byte-stream and no framers were supported, the application can control the flow of received data by specifying the minimum number of bytes of Message content it wants to receive at one time.
```

mwelzl: I think the action is called "Receive", but these other two uses of the word "receive" should not be capitalised here.

mwelzl (approved):
This LGTM

draft-ietf-taps-impl.md
```diff
@@ -1243,22 +1254,22 @@ The remote or local specifier provided by the application can not be resolved.
...
 * MessageTooLarge:
 The message size is too big for the transport system to handle.
 
 * ProtocolFailed:
-The underlying Protocol Stack failed. 
+The underlying Protocol Stack failed.
 
 * InvalidMessageProperties:
-The message properties are either contradictory to the transport properties or they can not be satisfied by the transport system. 
+The message properties either contradict with the transport properties or they can not be satisfied by the transport system.
```
tfpauly, line 1269:
```suggestion
The message properties either contradict the transport properties or they can not be satisfied by the transport system.
```

draft-ietf-taps-impl.md
```diff
@@ -1243,22 +1254,22 @@ The remote or local specifier provided by the application can not be resolved.
...
 
 * DeframingFailed:
-The data that was received by the underlying Protocol Stack could not be deframed. 
+The data that was received by the underlying Protocol Stack could not be processed by the deframer.
```
tfpauly, line 1272:
```suggestion
The data that was received by the underlying Protocol Stack could not be processed by the Message Framer.
```

tfpauly (approved):

PR #1036, opened 2022-06-02
Addressing Bron Gondwana's art-art review.

mwelzl:
These are the ones that probably don't need much discussion. See the email to the list.

### DIFF
--- a/draft-ietf-taps-impl.md
+++ b/draft-ietf-taps-impl.md
@@ -125,10 +125,12 @@ Preconnection objects should be implemented as bundles of properties that an app
 
 Once a Preconnection has been used to create an outbound Connection or a Listener, the implementation should ensure that the copy of the properties held by the Connection or Listener cannot be mutated by the application making changes to the original Preconnection object. This may involve the implementation performing a deep-copy, copying the object with all the objects that it references.
 
-Once the Connection is established, Transport Services implementation maps actions and events to the details of the chosen Protocol Stack. For example, the same Connection object may ultimately represent a single transport protocol instance (e.g., a TCP connection, a TLS session over TCP, a UDP flow with fully-specified Local and Remote Endpoints, a DTLS session, a SCTP stream, a QUIC stream, or an HTTP/2 stream).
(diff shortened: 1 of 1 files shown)


draft-ietf-taps-impl.md
```diff
@@ -355,7 +355,7 @@ Two examples of how Selection and Connection Properties may be used to sort bran
 
 * "Interface Instance or Type":
 If the application specifies an interface type to be preferred or avoided, implementations should accordingly rank the paths.
-If the application specifies an interface type to be required or prohibited, an implementation is expeceted to not include the non-conforming paths.
+If the application specifies an interface type to be required or prohibited, an implementation is expected to not include the non-conforming paths.
```
philsbln, line 358:
```suggestion
If the application specifies an interface type to be required or prohibited, an implementation is expected to exclude the non-conforming paths.
```

draft-ietf-taps-impl.md
```diff
@@ -942,7 +944,7 @@ Connectedness: Connected
 Data Unit: Byte-stream
 
 the Transport Services API mappings for MPTCP are identical to TCP. MPTCP adds support for multipath properties,
-such as "Multipath Transport" and "Policy for using Multipath Transports".
+such as `multipath` and `multipath-policy`.
```
philsbln, line 947: For sake of consistency: shouldn't we better use `multipathPolicy` here?

philsbln (approved):
All changes make sense and are legitimate, I just annotated some findings while reviewing them that could improve consistency and readability.

draft-ietf-taps-impl.md
tfpauly: I think the API is multipath-policy

tfpauly (approved):

draft-ietf-taps-impl.md
theri: We'll have a separate issue to turn all of them into camel case in the interface draft (another example: `advertises-altaddr`).

PR #1045, opened 2022-06-14
Incorporate Kyle's suggestions

### LINKED ISSUE #718: Some random comments/suggestions/questions
# Implementing Connection Establishment

I'm having difficult parsing "At this point, any constraints or requirements the application may have on the connection are available from pre-establishment." I'm not even sure what meaning is intended. Available? To what and for what?

# Structuring Options as a Tree

Is this really a forest? I.e., is a `2` possible when, for instance, you have a choice of two hostnames? If so, there should be an example illustrating that.

# Derived Endpoints

The last example in the section should be augmented to show multiple hostnames as described by the preceding paragraph.

Additionally, I think stating that there is a clear/natural dependency relationship between branching operations such that later operations depend on earlier decisions could simplify and clarify the justification for the suggested order of operations.

# Sorting Branches

"Scavenger" implies LEDBAT to me: it will take whatever is currently not being used but will defer to everything else. I'm not sure that notion aligns with the preference described.

# Handling of data for fast-open protocols

Is the intent of:

> It is also possible that protocol stacks within a particular leaf node use 0-RTT handshakes without any safely replayable application data. For example, TCP Fast Open could use a Client Hello from TLS as its 0-RTT data, shortening the cumulative handshake time.

that a protocol higher in the stack (or a framer?) could use 0-RTT even if the application doesn't want 0-RTT? Are there any security or privacy implications to this? Should there be a mechanism to disable this entirely?

# Defining Message Framers

I do not understand the syntax employed here. Is `MessageFramer` in:

```
MessageFramer -> Start(Connection)
MessageFramer -> Stop(Connection)
```

a global object or something? It appears to be generating a Start event with the Connection object to the system, but isn't the framer the receiver of such an event? (Nit: isn't Event<> with angle brackets the syntax here?)

# Specific Transport Protocol Considerations

> Protocols also define a notion of Data Unit

doesn't seem quite right. Protocols possess a property of application-layer data unit, usually as an intrinsic property that is a direct consequence of the protocol design. They don't really define it so much as end up with it. Maybe this is too much of a nit, but it reads weirdly.

# UDP

The use of "unconnected" (or "connectionless") directly adjacent to the phrase "UDP connections" is driving me batty. I understand that the TAPS interface provides only a Connection object, but we should not ever use lowercase-c "connection" to refer to UDP, and be clear when we mean the uppercase-C "Connection" object in relation to an underlying UDP transport.

tfpauly:
Closes #718

### DIFF
--- a/draft-ietf-taps-impl.md
+++ b/draft-ietf-taps-impl.md
@@ -358,7 +358,7 @@ If the application specifies an interface type to be preferred or avoided, imple
 If the application specifies an interface type to be required or prohibited, an implementation is expeceted to not include the non-conforming paths.
 
 * "Capacity Profile":
-An implementation can use the Capacity Profile to prefer paths that match an application's expected traffic pattern. This match will use cached performance estimates, see {{performance-caches}}:
+An implementation can use the Capacity Profile to prefer paths that match an application's expected traffic pattern. This match will use cached performance estimates, see {{performance-caches}}. Some examples of path preferences based on Capacity Profiles include:
    * Low Latency/Interactive:
      Prefer paths with the lowest expected Round Trip Time, based on observed round trip time estimates;
    * Low Latency/Non-Interactive:
--- a/draft-ietf-taps-impl.md
+++ b/draft-ietf-taps-impl.md
@@ -364,7 +364,7 @@ An implementation can use the Capacity Profile to prefer paths that match an app
    * Low Latency/Non-Interactive:
      Prefer paths with a low expected Round Trip Time, but can tolerate delay variation;
    * Constant-Rate Streaming:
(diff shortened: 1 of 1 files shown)


gorryfair (commented):
WFM

draft-ietf-taps-impl.md
```diff
@@ -359,14 +359,12 @@ If the application specifies an interface type to be required or prohibited, an
 
 * "Capacity Profile":
 An implementation can use the Capacity Profile to prefer paths that match an application's expected traffic pattern. This match will use cached performance estimates, see {{performance-caches}}:
```
tfpauly, line 361:
```suggestion
An implementation can use the Capacity Profile to prefer paths that match an application's expected traffic pattern. This match will use cached performance estimates, see {{performance-caches}}. Some examples of path preferences based on Capacity Profiles include:
```

PR #1050, opened 2022-07-23
Sketch alternative multicast model – #1021 reloaded

philsbln:
This is a remake of #1021 as @GrumpyOldTroll suggested.
It moves the WithIPvXMulticastGroup for receiving to the LocalEndpoint and adds the ASM/SSM distinguisher.
It does not add TTL support.

### DIFF
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -867,70 +867,69 @@ LocalSpecifier.WithStunServer(address, port, credentials)
 
 ### Multicast Examples {#multicast-examples}
 
-Specify a Local Endpoint using an Any-Source Multicast group to join on a named local interface:
+The following examples show how multicast groups can be used.
+
+Join an Any-Source Multicast group in receive-only mode, bound to a known
+port on a named local interface:
 
 ~~~
-LocalSpecifier := NewLocalEndpoint()
-LocalSpecifier.WithIPv4Address(233.252.0.0)
-LocalSpecifier.WithInterface("en0")
+   RemoteSpecifier := NewRemoteEndpoint()
+   RemoteSpecifier.WithIPv4Address(233.252.0.0)
+
+   LocalSpecifier := NewLocalEndpoint()
+   LocalSpecifier.WithPort(5353)
+   LocalSpecifier.WithInterface("en0")
+
+   TransportProperties := ...
+   SecurityParameters  := ...
+
+   Preconnection := newPreconnection(LocalSpecifier,
+                                     RemoteSpecifier,
+                                     TransportProperties,
+                                     SecurityProperties)
+   Listener := Preconnection.Listen()
 ~~~
 
-Source-Specific Multicast requires setting both a Local and Remote Endpoint:
+Create a Source-Specific Multicast group as a sender:
 
 ~~~
-LocalSpecifier := NewLocalEndpoint()
-LocalSpecifier.WithIPv4Address(232.1.1.1)
-LocalSpecifier.WithInterface("en0")
+   RemoteSpecifier := NewRemoteEndpoint()
+   RemoteSpecifier.WithIPv4Address(232.1.1.1)
 
-RemoteSpecifier := NewRemoteEndpoint()
-RemoteSpecifier.WithIPv4Address(192.0.2.22)
+   LocalSpecifier := NewLocalEndpoint()
+   LocalSpecifier.WithIPv4Address(192.0.2.22)
+   RemoteSpecifier.WithInterface("en0")
+
+   TransportProperties := ...
+   SecurityParameters  := ...
+
+   Preconnection := newPreconnection(LocalSpecifier,
+                                     RemoteSpecifier,
+                                     TransportProperties,
+                                     SecurityProperties)
+   Connection := Preconnection.Initiate()
 ~~~
 
-One common pattern for multicast is to both send and receive multicast. For such
-cases, an application can set up both a Listener and a Connection. The Listener
-is only used to accept Connections that receive inbound multicast. The initiated
-Connection is only used to send multicast.
+Join an any-source multicast group as both a sender and a receiver:
 
 ~~~
-// Prepare multicast Listener
-LocalMulticastSpecifier := NewLocalEndpoint()
-LocalMulticastSpecifier.WithIPv4Address(233.252.0.0)
-LocalMulticastSpecifier.WithPort(5353)
-LocalMulticastSpecifier.WithInterface("en0")
+   RemoteSpecifier := NewRemoteEndpoint()
+   RemoteSpecifier.WithIPv4Address(233.252.0.0)
+   RemoteSpecifier.WithPort(5353)
+   RemoteSpecifier.WithInterface("en0")
 
-TransportProperties := NewTransportProperties()
-TransportProperties.Require(preserve-msg-boundaries)
-// Reliable Data Transfer and Preserve Order are Required by default
-                                  
-// Specifying a Remote Endpoint is optional when using Listen()
-Preconnection := NewPreconnection(LocalMulticastSpecifier,
-                                  TransportProperties,
-                                  SecurityParameters)
+   LocalSpecifier := NewLocalEndpoint()
+   LocalSpecifier.WithIPv4Address(192.0.2.22)
+   LocalSpecifier.WithPort(5353)
 
-MulticastListener := Preconnection.Listen()
+   TransportProperties := ...
+   SecurityParameters  := ...
 
-// Handle inbound messages sent to the multicast group
-MulticastListener -> ConnectionReceived<MulticastReceiverConnection>
-MulticastReceiverConnection.Receive()
-MulticastReceiverConnection -> Received<messageDataRequest, messageContext>
-
-// Prepare Connection to send multicast
-LocalSpecifier := NewLocalEndpoint()
-LocalSpecifier.WithPort(5353)
-LocalSpecifier.WithInterface("en0")
-RemoteMulticastSpecifier := NewRemoteEndpoint()
-RemoteMulticastSpecifier.WithIPv4Address(233.252.0.0)
-RemoteMulticastSpecifier.WithPort(5353)
-RemoteMulticastSpecifier.WithInterface("en0")
-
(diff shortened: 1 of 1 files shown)


philsbln:
I got more in-person input from @GrumpyOldTroll in person in Philly.
We might want to include the src addr for SSN into a "WithSSNMulticastGroup" endpoint attribute and make Multicast addresses "special" that way. 

I guess we don't want to support racing multicast against non-multicast for now.

MaxF12 (approved):
LGTM, one minor nit

britram (approved):
LGTM but I'd like @csperkins eyes on this as well.

GrumpyOldTroll (commented):
Overall I think this looks like it should be fine.

The things that are not currently clear to me (might be covered elsewhere already? I haven't re-reviewed the whole doc):
- What if any difference is there between a sender using RemoteSpecifier.WithMulticastGroupIPvX and RemoteSpecifier.WithIPvXAddress?  Would WithIPvXAddress raise an error if given a group address, or would it use different defaults for the packets, or what exactly? (I'm thinking here of TTL/hoplimit and destination mac address as the 2 examples that come to mind that could have different defaults for these cases.)
- how is the default interface determined if not provided explicitly?  (sockets with apis that allow it I think will use the route to the group address, but I think ssm receive probably should use the route to the source address)
- will the listener merge receive data from different source addresses and source ports as packets are received into the same connection object, or will each new source (addr,port) pair result in a new connection, or is there some kind of controller that can pick between these?

Outside of these, I haven't noticed any questions without answers that seem obvious to me.

draft-ietf-taps-interface.md
```diff
@@ -737,23 +737,60 @@ early binding when required, for example with some Network Address Translator
...
+RemoteSpecifier.WithMulticastGroupIPv6(GroupAddress)
+RemoteSpecifier.WithPort(PortNumber)
+```
+
+Calling Listen() on a Preconnection with a multicast group specified on the Remote
+Endpoint will join the multicast group to receive messages. This Listener
+will create one Connection for each Remote Endpoint sending to the group,
+with the Local Endpoint set to the group address. The set of Connection
+objects created forms a Connection Group.
+
```
csperkins, line 762: I think this makes sense, but I wonder if the Local Endpoint must somehow specify both the group address and the local address/port to which the connection is bound? 

Or, to put this another way, when calling `Listen()` is it a requirement to bind to a specific local interface? If not, does the system pick a default interface from which to join the group, or does it join the group from all available interfaces? If the former, how can the application know what interface was used? If the latter, how can it tell on what interface each packet was received?

draft-ietf-taps-interface.md
```diff
@@ -737,23 +737,60 @@ early binding when required, for example with some Network Address Translator
...
+way are send-only, and do not join the multicast group. The resulting
+Connection will have a Local Endpoint indicating the local interface to
+which the connection is bound and a Remote Endpoint indicating the
+multicast group.
+
+```
+RemoteSpecifier.WithMulticastGroupIPv4(GroupAddress)
+RemoteSpecifier.WithMulticastGroupIPv6(GroupAddress)
+RemoteSpecifier.WithPort(PortNumber)
+```
```
csperkins, line 755: RemoteSpecifier.WithTTL(TTL)

draft-ietf-taps-interface.md
```diff
@@ -737,23 +737,60 @@ early binding when required, for example with some Network Address Translator
...
+group address as the Local Endpoint results in an EstablishmentError.
+
+```
+RemoteSpecifier.WithMulticastGroupIPv4(GroupAddress)
+RemoteSpecifier.WithMulticastGroupIPv6(GroupAddress)
+RemoteSpecifier.WithPort(PortNumber)
+LocalSpecifier.WithAnySourceMulticastGroupIPv4(GroupAddress)
+LocalSpecifier.WithAnySourceMulticastGroupIPv6(GroupAddress)
+LocalSpecifier.WithPort(PortNumber)
+```
```
csperkins, line 789: RemoteSpecifier.WithTTL(TTL)

csperkins:
This will close #874 if merged.

draft-ietf-taps-interface.md
philsbln: I discussed this with @GrumpyOldTroll in Philly … and we decided to leave that unspecified. We assume the implementation will throw an error if it has insufficient information. In the examples, we alway give an Interface, but a source address might also be sufficient.

draft-ietf-taps-interface.md
```diff
@@ -737,23 +737,64 @@ early binding when required, for example with some Network Address Translator
...
-that are accepted from this Listener are receive-only.
-
-Similarly, specifying a multicast group address on the Remote Endpoint will indicate that the
-resulting connection will be used to send multicast messages, and that the Preconnection will
-support Initiate() but not Listen(). Any Connections created this way are send-only.
-
-A Rendezvous() call on Preconnections containing group addresses results in an
-EstablishmentError as described in {{rendezvous}}.
+To use multicast, a Preconnection is first created with the Local/Remote Endpoint
+specifying the ASM/SSM multicast group and destination port number.
```
tfpauly, line 741: We need to expand ASM/SSM the first time I think

draft-ietf-taps-interface.md
```diff
@@ -737,23 +737,64 @@ early binding when required, for example with some Network Address Translator
...
+way are send-only, and do not join the multicast group. The resulting
+Connection will have a Local Endpoint indicating the local interface to
+which the connection is bound and a Remote Endpoint indicating the
+multicast group.
+
+```
+RemoteSpecifier.WithMulticastGroupIPv4(GroupAddress)
+RemoteSpecifier.WithMulticastGroupIPv6(GroupAddress)
+RemoteSpecifier.WithPort(PortNumber)
+RemoteSpecifier.WithTTL(TTL)
```
tfpauly, line 755: Move this up before the prior paragraph?

Can we also describe the TTL parameter?

draft-ietf-taps-interface.md
```diff
@@ -737,23 +737,64 @@ early binding when required, for example with some Network Address Translator
...
+LocalSpecifier.WithAnySourceMulticastGroupIPv4(GroupAddress)
+LocalSpecifier.WithAnySourceMulticastGroupIPv6(GroupAddress)
+LocalSpecifier.WithPort(PortNumber)
+LocalSpecifier.WithTTL(TTL)
+```
 
 See {{multicast-examples}} for more examples.
 
+
```
tfpauly, line 797:
```suggestion
```

draft-ietf-taps-interface.md
philsbln:
```suggestion
The receiving interface can be restricted by passing it as part of the LocalSpecifier or queried through the MessagContext on the messages received (see {{msg-ctx}} for further details).

```

draft-ietf-taps-interface.md
```diff
@@ -737,20 +737,60 @@ early binding when required, for example with some Network Address Translator
...
-that are accepted from this Listener are receive-only.
-
-Similarly, specifying a multicast group address on the Remote Endpoint will indicate that the
-resulting connection will be used to send multicast messages, and that the Preconnection will
-support Initiate() but not Listen(). Any Connections created this way are send-only.
-
-A Rendezvous() call on Preconnections containing group addresses results in an
-EstablishmentError as described in {{rendezvous}}.
+To use multicast, a Preconnection is first created with the Local/Remote Endpoint
+specifying the ASM/SSM multicast group and destination port number.
```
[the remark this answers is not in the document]
tfpauly:
```suggestion
specifying the Any-source multicast (ASM) or Source-specific multicast (SSM) multicast group and destination port number.
```

draft-ietf-taps-interface.md
```diff
@@ -737,20 +737,60 @@ early binding when required, for example with some Network Address Translator
...
-that are accepted from this Listener are receive-only.
-
-Similarly, specifying a multicast group address on the Remote Endpoint will indicate that the
-resulting connection will be used to send multicast messages, and that the Preconnection will
-support Initiate() but not Listen(). Any Connections created this way are send-only.
-
-A Rendezvous() call on Preconnections containing group addresses results in an
-EstablishmentError as described in {{rendezvous}}.
+To use multicast, a Preconnection is first created with the Local/Remote Endpoint
+specifying the _any source multicast (ASM)_ / _source specific multicast (SSM)_ multicast group and destination port number.
```
tfpauly, line 741:
```suggestion
specifying the any-source multicast (ASM) or source-specific multicast (SSM) multicast group and destination port number.
```

draft-ietf-taps-interface.md
```diff
@@ -737,20 +737,60 @@ early binding when required, for example with some Network Address Translator
...
+specifying the any-source multicast (ASM) or source-specific multicast (SSM) multicast group and destination port number.
+
+Calling Initiate() on that Preconnection creates a Connection that can be
+used to send messages to the multicast group. The Connection object that is
+created will support Send() but not Receive(). Any Connections created this
+way are send-only, and do not join the multicast group. The resulting
+Connection will have a Local Endpoint indicating the local interface to
+which the connection is bound and a Remote Endpoint indicating the
+multicast group.
+
```
csperkins, line 750: add "The following API calls can be used to configure a `Preconnection` before calling `Initiate()`:

draft-ietf-taps-interface.md
```diff
@@ -737,20 +737,60 @@ early binding when required, for example with some Network Address Translator
...
+To use multicast, a Preconnection is first created with the Local/Remote Endpoint
+specifying the any-source multicast (ASM) or source-specific multicast (SSM) multicast group and destination port number.
+
+Calling Initiate() on that Preconnection creates a Connection that can be
+used to send messages to the multicast group. The Connection object that is
+created will support Send() but not Receive(). Any Connections created this
+way are send-only, and do not join the multicast group. The resulting
+Connection will have a Local Endpoint indicating the local interface to
+which the connection is bound and a Remote Endpoint indicating the
+multicast group.
```
mwelzl, line 749:
```suggestion
multicast group.

The following API calls can be used to configure a Preconnection before calling Initiate():
```

mwelzl (approved):

PR #1058, opened 2022-08-09
Clone causes a ConnectionReceived event.

### LINKED ISSUE #1049: Does Connection.clone() cause a ConnectionReceived<Connection> event at the peer?
If `Connection.clone()` is called, is an unsolicited `ConnectionReceived<Connection>` event generated at the peer? I assume so, but can't find it documented.

mwelzl:
Closes #1049.

I added text to ConnectionReceived about this. I didn't additionally explicitly state this in the text about clone to avoid complicating things - since similarly, the text describing Initiate() doesn't talk about ConnectionReceived.

### DIFF
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -1613,10 +1613,10 @@ action is performed on the Listener object.
 Listener -> ConnectionReceived<Connection>
 ~~~
 
-The ConnectionReceived Event occurs when a Remote Endpoint has established a
+The ConnectionReceived Event occurs when a Remote Endpoint has established or cloned ((see {{groups}}) a
 transport-layer connection to this Listener (for Connection-oriented
 transport protocols), or when the first Message has been received from the
-Remote Endpoint (for Connectionless protocols), causing a new Connection to be
+Remote Endpoint (for Connectionless protocols or streams of a multi-streaming transport), causing a new Connection to be
 created. The resulting Connection is contained within the ConnectionReceived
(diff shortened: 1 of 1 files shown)


britram (approved):

draft-ietf-taps-interface.md
```diff
@@ -1613,10 +1613,10 @@ action is performed on the Listener object.
 Listener -> ConnectionReceived<Connection>
 ~~~
 
-The ConnectionReceived Event occurs when a Remote Endpoint has established a
+The ConnectionReceived Event occurs when a Remote Endpoint has established or cloned ((see {{groups}}) a
```
tfpauly, line 1616: Since the remote endpoint may not be using TAPS, I wonder if there's a better way to say this than that they "cloned" the connection — they are opening a connection or a new stream. Maybe we just say that?

tfpauly (changes requested):

draft-ietf-taps-interface.md
csperkins: "…or cloned (e.g., by creating a new stream in a multi-stream transport; see {{groups}}), …" maybe? Otherwise, looks good.

csperkins (approved):

draft-ietf-taps-interface.md
mwelzl:
```suggestion
The ConnectionReceived Event occurs when a Remote Endpoint has established or cloned ((see {{groups}}) a
```

PR #1059, opened 2022-08-09
Fixing property names and their references

### LINKED ISSUE #1040: Refer to all properties by their name, not the section title
Examples:

> A connection can not be used
>   for sending if the connection was created with the Selection Property
>   `Direction of Communication` set to `unidirectional receive` or if a Message
>   marked as `Final` was sent over this connection.
> 
should be:
> A connection can not be used
>   for sending if the connection was created with the Selection Property
>   `direction` set to `unidirectional receive` or if a Message
>   marked as `Final` was sent over this connection.
> 

, and:

> The Transport Services API provides two properties to allow a sender
> to signal the relative priority of data transmission: the Priority Message
> Property {{msg-priority}}, and the Connection Priority Connection Property
> {{conn-priority}}.

should be:

> The Transport Services API provides two properties to allow a sender
> to signal the relative priority of data transmission: the `msgPriority`
> Message Property {{msg-priority}},
> and the `connPriority` Connection Property
> {{conn-priority}}.

...and many more. Thoughts?  This is a large and painful effort, so I'm checking first  :)

### LINKED ISSUE #1046: Turn all property names into CamlCase
for consistency; e.g., a few now have dashes instead. Make sure to fix all references to these names.

mwelzl:
Closes #1040. Closes #1046.

I also fixed some capitalization errors in passing.

### DIFF
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -2616,7 +2616,7 @@ Type:
 Default:
 : false
 
-If true, safelyReplayable specifies that a Message is safe to send to the Remote Endpoint
+If true, `safelyReplayable` specifies that a Message is safe to send to the Remote Endpoint
 more than once for a single Send Action. It marks the data as safe for
 certain 0-RTT establishment techniques, where retransmission of the 0-RTT data
 may cause the remote application to receive the Message multiple times.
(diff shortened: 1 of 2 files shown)


draft-ietf-taps-interface.md
```diff
@@ -2616,16 +2616,16 @@ Type:
 Default:
 : false
 
-If true, Safely Replayable specifies that a Message is safe to send to the Remote Endpoint
+If true, safelyReplayable specifies that a Message is safe to send to the Remote Endpoint
```
tfpauly, line 2619:
```suggestion
If true, `safelyReplayable` specifies that a Message is safe to send to the Remote Endpoint
```

tfpauly (approved):

PR #1060, opened 2022-09-19
Describe metadata-only messages

### LINKED ISSUE #1030: Review by CA - Is the content of messages always bytes?
* Is the content of messages always bytes? I certainly got that
  impression, but didn't see it explicitly either. (The "protocols that
  expose byte-streams" have that property visible in their name). When
  framers are used a lot, I can envision that at some point a message
  might not use data bytes at all any more, as all information in the
  underlying message is being dissected into protocol specific
  properties.

tfpauly:
Closes #1030

### DIFF
--- a/draft-ietf-taps-impl.md
+++ b/draft-ietf-taps-impl.md
@@ -518,7 +518,8 @@ For protocols that expose byte-streams, the only delineation provided by the pro
 Protocols that provide the framing (such as length-value protocols, or protocols that use delimiters) may support Message sizes that do not fit within a single datagram. Each Message for framing protocols corresponds to a single frame, which may be sent either as a complete Message in the underlying protocol, or in multiple parts.
 
 Messages themselves generally consist of bytes passed in the messageData parameter intended to be processed at an application layer. However, Message objects presented through the API
-can carry associated Message Properties. When these are Protocol Specific Properties, they can include metadata that exists separately from a byte
+can carry associated Message Properties passed through the messageContext parameter.
+When these are Protocol Specific Properties, they can include metadata that exists separately from a byte
 encoding. For example, these Properties can include name-value pairs of information, like HTTP header fields. In such cases, Messages might be "empty",
 insofar as they contain zero bytes in the messageData parameter, but can still include data in the messageContext that is interpreted by the Protocol Stack.
 
--- a/draft-ietf-taps-impl.md
+++ b/draft-ietf-taps-impl.md
@@ -517,10 +517,10 @@ For protocols that expose byte-streams, the only delineation provided by the pro
 
 Protocols that provide the framing (such as length-value protocols, or protocols that use delimiters) may support Message sizes that do not fit within a single datagram. Each Message for framing protocols corresponds to a single frame, which may be sent either as a complete Message in the underlying protocol, or in multiple parts.
 
-Messages themselves generally consist of bytes intended to be processed at an application layer. However, Message objects presented through the API
+Messages themselves generally consist of bytes passed in the messageData parameter intended to be processed at an application layer. However, Message objects presented through the API
 can carry associated Message Properties. When these are Protocol Specific Properties, they can include metadata that exists separately from a byte
 encoding. For example, these Properties can include name-value pairs of information, like HTTP header fields. In such cases, Messages might be "empty",
-insofar as they contain zero bytes of raw data, but can still include metadata that is interpreted by the Protocol Stack.
+insofar as they contain zero bytes in the messageData parameter, but can still include data in the messageContext that is interpreted by the Protocol Stack.
 
 ## Sending Messages
 
--- a/draft-ietf-taps-impl.md
+++ b/draft-ietf-taps-impl.md
@@ -517,6 +517,11 @@ For protocols that expose byte-streams, the only delineation provided by the pro
 
 Protocols that provide the framing (such as length-value protocols, or protocols that use delimiters) may support Message sizes that do not fit within a single datagram. Each Message for framing protocols corresponds to a single frame, which may be sent either as a complete Message in the underlying protocol, or in multiple parts.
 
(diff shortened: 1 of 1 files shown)


draft-ietf-taps-impl.md
```diff
@@ -517,6 +517,11 @@ For protocols that expose byte-streams, the only delineation provided by the pro
 
 Protocols that provide the framing (such as length-value protocols, or protocols that use delimiters) may support Message sizes that do not fit within a single datagram. Each Message for framing protocols corresponds to a single frame, which may be sent either as a complete Message in the underlying protocol, or in multiple parts.
 
+Messages themselves generally consist of bytes intended to be process at an application layer. However, Message objects presented through the API
```
mwelzl, line 520:
```suggestion
Messages themselves generally consist of bytes intended to be processed at an application layer. However, Message objects presented through the API
```

mwelzl:
I think this is good, but it would also need a change to the API, where sending or receiving a Message without data is not currently an option. Concretely, I think:

- the messageData parameter should be optional in the send call in Section 9.2: https://ietf-tapswg.github.io/api-drafts/draft-ietf-taps-interface.html#name-sending-data
- There should be text there explaining, similar to the proposed text here for the implementation draft, why this makes sense
- Similarly, the messageData parameter should be optional in the Received and ReceivedPartial events, see:  https://ietf-tapswg.github.io/api-drafts/draft-ietf-taps-interface.html#name-receive-events

Sorry for phrasing this as a request - in principle, I could make these changes myself but don't want to create confusion with a separate PR on the same matter, and I'm not sure I'd get the wording right either.

mwelzl (changes requested):
I think this is good, but it would also need a change to the API, where sending or receiving a Message without data is not currently an option. Concretely, I think:

the messageData parameter should be optional in the send call in Section 9.2: https://ietf-tapswg.github.io/api-drafts/draft-ietf-taps-interface.html#name-sending-data
There should be text there explaining, similar to the proposed text here for the implementation draft, why this makes sense
Similarly, the messageData parameter should be optional in the Received and ReceivedPartial events, see: https://ietf-tapswg.github.io/api-drafts/draft-ietf-taps-interface.html#name-receive-events
Sorry for phrasing this as a request - in principle, I could make these changes myself but don't want to create confusion with a separate PR on the same matter, and I'm not sure I'd get the wording right either.

tfpauly:
@mwelzl I think I disagree that this needs an API change, but rather a note that a messageData object can be empty. At least in the data objects we work with, they can support 0-length data blobs.

mwelzl:
@tfpauly I agree, you're right, this is better.

abrunstrom (approved):
This text looks fine to me, but as discussed some complementing text about the possibility of a messageData object being empty will be needed in the API draft.

mwelzl (approved):
LGTM.  @abrunstrom says that text will be needed in the API draft to state that Messages can be empty. I'm not sure this is really necessary to add there; it wouldn't hurt, but it also doesn't get in the way of merging this PR.

draft-ietf-taps-impl.md
```diff
@@ -517,6 +517,11 @@ For protocols that expose byte-streams, the only delineation provided by the pro
 
 Protocols that provide the framing (such as length-value protocols, or protocols that use delimiters) may support Message sizes that do not fit within a single datagram. Each Message for framing protocols corresponds to a single frame, which may be sent either as a complete Message in the underlying protocol, or in multiple parts.
 
+Messages themselves generally consist of bytes intended to be processed at an application layer. However, Message objects presented through the API
+can carry associated Message Properties. When these are Protocol Specific Properties, they can include metadata that exists separately from a byte
+encoding. For example, these Properties can include name-value pairs of information, like HTTP header fields. In such cases, Messages might be "empty",
+insofar as they contain zero bytes of raw data, but can still include metadata that is interpreted by the Protocol Stack.
```
gorryfair, line 523: is "raw" the correct word here? Is this better as /they contain zero bytes of raw data/they contain no data to be processed at the application layer/

tfpauly:
```suggestion
insofar as they contain zero bytes in the messageData parameter, but can still include data in the messageContext that is interpreted by the Protocol Stack.
```

draft-ietf-taps-impl.md
```diff
@@ -517,6 +517,11 @@ For protocols that expose byte-streams, the only delineation provided by the pro
 
 Protocols that provide the framing (such as length-value protocols, or protocols that use delimiters) may support Message sizes that do not fit within a single datagram. Each Message for framing protocols corresponds to a single frame, which may be sent either as a complete Message in the underlying protocol, or in multiple parts.
 
+Messages themselves generally consist of bytes intended to be processed at an application layer. However, Message objects presented through the API
```
tfpauly, line 520:
```suggestion
Messages themselves generally consist of bytes passed in the messageData parameter intended to be processed at an application layer. However, Message objects presented through the API
```

draft-ietf-taps-impl.md
```diff
@@ -517,6 +517,11 @@ For protocols that expose byte-streams, the only delineation provided by the pro
 
 Protocols that provide the framing (such as length-value protocols, or protocols that use delimiters) may support Message sizes that do not fit within a single datagram. Each Message for framing protocols corresponds to a single frame, which may be sent either as a complete Message in the underlying protocol, or in multiple parts.
 
+Messages themselves generally consist of bytes passed in the messageData parameter intended to be processed at an application layer. However, Message objects presented through the API
+can carry associated Message Properties. When these are Protocol Specific Properties, they can include metadata that exists separately from a byte
```
tfpauly, line 521:
```suggestion
can carry associated Message Properties passed through the messageContext parameter.
When these are Protocol Specific Properties, they can include metadata that exists separately from a byte
```

PR #1062, opened 2022-09-19
Remove mention of ICMP errors for multicast receive

### LINKED ISSUE #1032: Review by CA - Which ICMP to forward for Multicast?
(More out of curiosity, with no expectation that this would find its
  way into the text unless you think I've hit a bug): In UDP Multicast
  Receive, which kind of ICMP notifications can come in there that would
  constitute a ConnectionError -- given that this transport does not
  send (and would thus usually not trigger any)?

tfpauly:
Closes #1032


mwelzl (approved):
LGTM

gorryfair (approved):
Seems like we do not loose anything by removing this. The change WFM.

PR #1063, opened 2022-09-19
Align framer parameter names with API

### LINKED ISSUE #1028: Review by CA - Final message property IsEndOfMessage event?
Framers: What is the difference between the Final message property and
  the separate IsEndOfMessage event argument?

tfpauly:
Closes #1028

### DIFF
--- a/draft-ietf-taps-impl.md
+++ b/draft-ietf-taps-impl.md
@@ -629,8 +629,8 @@ properties of the Connection or the network being used that may influence how
 to frame Messages.
 
 ~~~
-MessageFramer -> Start<Connection>
-MessageFramer -> Stop<Connection>
+MessageFramer -> Start<connection>
+MessageFramer -> Stop<connection>
 ~~~
 
 When a Message Framer generates a `Start` event, the framer implementation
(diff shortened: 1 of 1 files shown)


mwelzl:
Capitalization seems to be a problem for us. We need to make things uniform. At first, I reacted to seeing "connection" as parameters in this PR, because we use the term "Connection" to refer to a TAPS-provided connection (as opposed to a connection of the underlying system), and so, de-capitalizing the term here may look better, but it is also somewhat imprecise.

This said, we do the same with other capitalized terms too, here and in the API draft - e.g., "messageContext".  Should we generally agree to begin terms with a lowercase letter when they are used as parameters in code examples?  I think we have this in most places, though at least the API draft has several counterexamples... e.g., the server example code:

Preconnection := NewPreconnection(LocalSpecifier,
                                  TransportProperties,
                                  SecurityParameters)

(..)

mwelzl (approved):
I think this looks ok (and thanks for doing this!), but see also my comment about making things uniform wrt. capitalization: if we merge this PR, we should also do one that does the same with the API draft.

draft-ietf-taps-impl.md
```diff
@@ -639,20 +639,20 @@ its `Ready` event. This allows the implementation to communicate control data to
 Remote Endpoint that can be used to parse Messages.
 
 ~~~
-MessageFramer.MakeConnectionReady(Connection)
+MessageFramer.MakeConnectionReady(connection)
```
abrunstrom, line 642: Not directly related to parameter names, but perhaps we should add "The framer implementation can indicate that it has finished with this"  **before** the above line. Similar to what we do for the stop event text below.

abrunstrom (approved):
Thanks LGTM! Although I noticed another nit that can perhaps be fixed in the same PR even if not directly related to the parameter names. Regarding the capitalization I think we can merge this PR and then have a separate issue for the capitalization as we will need to fix this either way.

draft-ietf-taps-impl.md
```diff
@@ -639,20 +639,20 @@ its `Ready` event. This allows the implementation to communicate control data to
...
 ~~~
 
 Similarly, when a Message Framer generates a `Stop` event, the framer implementation has the opportunity to write some final data or clear up its local state before the `Closed` event is delivered to the Application. The framer implementation can indicate that it has finished with this.
```
tfpauly, line 645:
```suggestion
Similarly, when a Message Framer generates a `Stop` event, the framer implementation has the opportunity to write some final data or clear up its local state before the `Closed` event is delivered to the Application. The framer implementation can indicate that it has finished with this call.
```

draft-ietf-taps-impl.md
```diff
@@ -639,20 +639,20 @@ its `Ready` event. This allows the implementation to communicate control data to
 Remote Endpoint that can be used to parse Messages.
```
tfpauly, line 639:
```suggestion
Remote Endpoint that can be used to parse Messages.

Once the framer implementation has completed its setup or handshake, it can indicate to
the application that it is ready to handling data with this call.
```

draft-ietf-taps-impl.md
tfpauly: Added!

PR #1065, opened 2022-09-21
Reference v6 address considerations for temp local property

### LINKED ISSUE #1053: FG: Temporary Local Address
Adding this issue from Fernando via the TAPS list:
> 6.2.13. Use Temporary Local Address Name: useTemporaryLocalAddress Type: Preference Default: Avoid for Listeners and Rendezvous
> Connections. Prefer for other Connections. This property allows the
> application to express a preference for the use of temporary local
> addresses, sometimes called "privacy" addresses [RFC8981]. Temporary
> addresses are generally used to prevent linking connections over time
> when a stable address, sometimes called "permanent" address, is not
> needed. There are some caveats to note when specifying this property.
> First, if an application Requires the use of temporary addresses, the
> resulting Connection cannot use IPv4, because temporary addresses do
> not exist in IPv4. Second, temporary local addresses might involve
> trading off privacy for performance. For instance, temporary
> addresses can interfere with resumption mechanisms that some
> protocols rely on to reduce initial latency.

The terminology employed in this subsection seems a bit confusing, or at least deviates from what we normally use in v6-circles. I suggest taking a look at RFC7721 -- certainly not flawless, but it was 6man's shot to try to agree on terminology. For example, I don't think we use "permanent" and "stable" interchangeably.

e.g. the "Local" in "useTemporaryLocalAddress" might be misleading to some. -- e.g. could be interpreted as related with link-locals or ULAs.

Aside, I believe there's much more to dig regarding address properties.
e.g., we have given it a shot at: https://www.ietf.org/id/draft-gont-v6ops-ipv6-addressing-considerations-02.html

e.g., the choice of ipv6 addresses has more than one dimension... and there are tradeoffs involved, which IMHO deserve discussion.

And aside from the regular address systems normally configure, if one were to specify an API one would probably want to provide a mechanism for apps to be able to request sole/single-use addresses (.e.g., one address per app or container).

britram:
fixes #1053

### DIFF
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -1284,7 +1284,9 @@ application Requires the use of temporary addresses, the resulting Connection
 cannot use IPv4, because temporary addresses do not exist in IPv4. Second,
 temporary local addresses might involve trading off privacy for performance.
 For instance, temporary addresses can interfere with resumption mechanisms
-that some protocols rely on to reduce initial latency.
+that some protocols rely on to reduce initial latency. See 
+{?I-D.gont-v6ops-ipv6-addressing-considerations} for more discussion on the
+trade-offs involved with this preference.
 
 ### Multipath Transport {#multipath-mode}
 
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -1284,7 +1284,7 @@ application Requires the use of temporary addresses, the resulting Connection
 cannot use IPv4, because temporary addresses do not exist in IPv4. Second,
 temporary local addresses might involve trading off privacy for performance.
 For instance, temporary addresses can interfere with resumption mechanisms
-that some protocols rely on to reduce initial latency. See 
+that some protocols rely on to reduce initial latency. See
 {?I-D.gont-v6ops-ipv6-addressing-considerations} for more discussion on the
 trade-offs involved with this preference.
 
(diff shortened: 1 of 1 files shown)


draft-ietf-taps-interface.md
```diff
@@ -1284,7 +1284,9 @@ application Requires the use of temporary addresses, the resulting Connection
 cannot use IPv4, because temporary addresses do not exist in IPv4. Second,
 temporary local addresses might involve trading off privacy for performance.
 For instance, temporary addresses can interfere with resumption mechanisms
-that some protocols rely on to reduce initial latency.
+that some protocols rely on to reduce initial latency. See
```
britram, line 1287:
```suggestion
that some protocols rely on to reduce initial latency. See
```

gorryfair (approved):
Seems fine to me. It might be worth checking this ID continues to progress, as we near final publication.

gorryfair (approved):

draft-ietf-taps-interface.md
```diff
@@ -1284,7 +1284,9 @@ application Requires the use of temporary addresses, the resulting Connection
...
+{?I-D.gont-v6ops-ipv6-addressing-considerations} for more discussion on the
```
tfpauly, line 1288: I'm a bit concerned about adding a reference to an individual draft here. Could we find a better reference for temporary addresses?

gorryfair: IT could BE ENOUGH to just add a ref: "For instance, temporary addresses (e.g. RFC8981) that can interfere with resumption mechanisms?

tfpauly: Yes, I'd prefer that ref.

mwelzl (changes requested):
I agree with the comments of @tfpauly and @gorryfair

tfpauly (approved):

PR #1066, opened 2022-09-22
Introduce racing more clearly

### LINKED ISSUE #1020: Implementation Draft Review: Section 4.1 - Define Racing earlier to include Staggered/Failover scenarios
Raising this issue as part of my review of the the Implementation Draft.

To a reader who understands "protocol racing" in the "Happy Eyeballs" sense of near-simultaneous connection establishment attempts, parts of Section 4.1 are confusing. It appears as if "real racing" actually only occurs between Derived Endpoints and not e.g., among Alternate Paths, where we instead have "failover after timeout":

Section 4.1.1.2
>  Attempts should be started on one path (e.g., a specific interface), and then successively on other paths (or interfaces) after delays based on expected path round-trip-time or other available metrics.

However, in Section 4.1.1.3 "Protocol Options", the "rank[ing] and attempt[ing]" of options is also referred to as "racing":
> Implementations that support racing protocols and protocol options should maintain a history of which protocols and protocol options successfully established, on a per-network and per-endpoint basis [...]. This information can influence future racing decisions to prioritize or prune branches.

It seems clear that this kind of information is only useful when connection attempts are not in fact "raced" (nearly) in parallel, but are only conducted in succession, where it makes sense to rank the candidates accordingly.

This confusion is only cleared up in Section 4.3, where it says:

>  There are three different approaches to racing the attempts for different nodes of the connection establishment tree:
> 1. Simultaneous
> 2. Staggered
> 3. Failover

So it appears as if (in partial departure from the Happy Eyeballs RFC) the term "racing" here is meant to include staggered or failover scenarios, somewhat undermining the original metaphor.

Suggestions:
- Explain the extended scope of the "racing" term earlier
- Perhaps replace some uses the term "racing" with "failover" or even just "ranking"

tfpauly:
Closes #1020

### DIFF
--- a/draft-ietf-taps-impl.md
+++ b/draft-ietf-taps-impl.md
@@ -166,9 +166,9 @@ It is expected that the database of system policies and the method of looking up
 
 The process of establishing a network connection begins when an application expresses intent to communicate with a Remote Endpoint by calling Initiate,  at which point the Preconnection object contains all constraints or requirements the application has configured for the connection. The establishment process can be considered complete once there is at least one Protocol Stack that has completed any required setup to the point that it can transmit and receive the application's data.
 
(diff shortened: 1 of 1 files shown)


abrunstrom (approved):
LGTM, thanks!

PR #1067, opened 2022-09-22
Editorial updates for implementation

### LINKED ISSUE #1048: Review of "Implementing Interfaces to Transport Services"
# Conclusion of my Review of Implementing Interfaces to Transport Services
_NB: This reviewer may be unreasonably biased against unnecessary mention of event-driven interaction patterns._

**Viewed in concert with the other pre-requisite TAPS documents, the text of this draft is quite readable and represents a useful collection of implementation considerations for various components of a transport services system.**

My biggest point of concern is the fluctuating level of detail between the different Sections, leading to speculations about who might have written what. Some Sections remain quite general and avoid making too many assumptions about the underlying software architecture. Others discuss a specific implementation approach in some detail. In particular **Section 6** (Message Framers) stands out from the rest of the text. Language that relates to event-driven interaction patterns is almost exclusively found in this Section, where different types of events are "generated", "delivered", "received" or "handled" by and between various components. In the rare cases where events are mentioned in the other Sections, they are often associated with other verbs like "notify", "cause", "provoke" or even "call" (see, e.g., **Section 4.5** or **Section 8**). Indeed, the text in most of these other Sections would require only very minor edits to avoid any reference to an event-driven architecture altogether (examples below). It is interesting that the same can not easily be said of Section 6, alone. 

**Overall, the text could benefit from an edit pass that normalizes language and (use of) terms and adds a few more references to related documents.** Other than that, I have few issues with it. Below are a few remarks and suggestions to cherry-pick from, as usual.

## Possible minor edits relating to event-system language

Quoting from **4.4.1. Determining Successful Establishment**:

> To avoid the issue of TLS possibly failing, the implementation should not generate a Ready event for the Connection until TLS is established.

Not only is the term "Ready event" used without any sort of context or introduction, it is the first point in the text where familiarity with a specific event type is implied. The intent of the whole sentence could be much better conveyed without this terminological baggage. Suggestion:

> The implementation should not report the Connection as ready for use to the application before the TLS handshake, which might still fail, has also been successfully completed.

From later in the section, quoting: 
> [T]he transport system should notify the application with an InitiateError event. An InitiateError event should also be generated in case the transport system finds no usable candidates to race.

While it is reasonable to mention the specific error type here (although a reference to the document where it is defined is still missing), wouldn't the following suffice?

> [T]he transport system should report an InitiateError to the application. An InitiateError should also be raised in case the transport system finds no usable candidates to race.

Quoting from **4.5. Establishing multiplexed connections**:

> Therefore, when the Initiate action of a Transport Services API is called without Messages being handed over, it cannot be guaranteed that the Remote Endpoint will have any way to know about this, and hence a passive endpoint's ConnectionReceived event might not be called until data is received. Instead, calling the ConnectionReceived event could be delayed until the first Message arrives.

Shouldn't events rather be "emitted" or "triggered" instead of (ambiguously) "called"?

It is gratifying to observe that **Sections 4.6** and **4.7** on the other hand entirely avoid specific references to an event-based architecture, instead speaking of, e.g., "passing" Connection objects to the application. 

Quoting from **6. Implementing Message Framers**:

> This section describes one possible API for defining Message Framers as an example.

Given that this section is the only place in the text where concepts and language centered around an event-based implementation are consistently used, it would be appropriate to point this out in its introduction:

> This section describes one possible API for defining Message Framers in an event-based implementation as an example.

## Other minor editing suggestions and remarks

Quoting **4.4.1. Determining Successful Establishment**:

> Implementations may select the criteria by which a leaf node is considered to be successfully connected differently on a per-protocol basis.

The adverb "differently" is too far removed from "select", becoming ambiguous and could be wrongly understood to modify "connected" instead. Suggestion:

> On a per-protocol basis, implementations may select different criteria by which a leaf node is considered to be successfully connected.

Quoting **4.6. Handling connectionless protocols**:

> To mitigate this, an application can use a Message Framer Section 6 on top of a connectionless protocol to only mark a specific connection attempt as ready when some data has been received, or after some application-level handshake has been performed.

The reference to the section is ungrammatical.

About **5. Implementing Sending and Receiving Data**: 

Might it not be helpful to conclude the introduction with some examples: UDP Datagrams, TCP streams, HTTP responses?

About **5.1.1. Message Properties**:

The list of Message Properties is given without any sort of context or introduction. What's the story here, is this an exhaustive list or just a list of those properties that can be adjusted to modify the send and receive behavior?

Quoting **5.2. Sending Messages**:

> This requires the implementation to understand where messages end, either via a supplied or because the top-level protocol in the established Protocol Stack preserves message boundaries. 

That sentence is exceedingly hard to parse.

Quoting **5.3. Handling of data for fast-open protocols**:

> If the application has 0-RTT data to send during handshake(s), it needs to provide this data before the handshakes have begun.
> When racing, this means that the data should be provided before the process of connection establishment has begun.

Possibly intentional redundant wording, but unfortunately distracting.

tfpauly:
Closes #1048


draft-ietf-taps-impl.md
```diff
@@ -463,11 +463,11 @@ Keeping additional connections is generally not recommended, because those attem
...
+For Protocol Stacks with multiple handshakes, the decision becomes more nuanced. If the Protocol Stack involves both TLS and TCP, an implementation could determine that a leaf node is connected after the TCP handshake is complete, or it can wait for the TLS handshake to complete as well. The benefit of declaring completion when the TCP handshake finishes, and thus stopping the race for other branches of the tree, is reduced burden on the network and Remote Endpoints from further connection attempts that are likely to be abandoned. On the other hand, by waiting until the TLS handshake is complete, an implementation avoids the scenario in which a TCP handshake completes quickly, but TLS negotiation is either very slow or fails altogether in particular network conditions or to a particular endpoint. To avoid the issue of TLS possibly failing, the implementation should not generate a Ready event for the Connection until TLS handshake is complete.
```
abrunstrom, line 468: Should this not be "until **the** TLS handshake..."

draft-ietf-taps-impl.md
```diff
@@ -511,18 +511,20 @@ Protocols that provide multiplexing of streams into a single four-tuple can list
...
+Protocols that provide the framing (such as length-value protocols, or protocols that use delimiters like HTTP/1.1) may support Message sizes that do not fit within a single datagram. Each Message for framing protocols corresponds to a single frame, which may be sent either as a complete Message in the underlying protocol, or in multiple parts.
 
 ## Sending Messages
 
 The effect of the application sending a Message is determined by the top-level protocol in the established Protocol Stack. That is, if the top-level protocol provides an abstraction of framed messages over a connection, the receiving application will be able to obtain multiple Messages on that connection, even if the framing protocol is built on a byte-stream protocol like TCP.
 
 ### Message Properties {#msg-properties}
 
+The API allows various properties to be associated with each Message, which should be implemented as discussed below.
```
abrunstrom, line 526: Should we add a reference after "The API" or do we already have the API document so frequently referenced that it is not needed?

abrunstrom (approved):
Thanks for taking care of the edits! Just added a nit and a question. 

I assume you deliberately skipped the comment from the issue about event-based terminology in Section 6, not thinking it was needed as we reference the API draft for the terminology?

draft-ietf-taps-impl.md
philsbln: As we could have some endpoints that fail the security properties, it may be useful to delay the Ready event until the TLS handshake completes.

philsbln (approved):
LGMT

draft-ietf-taps-impl.md
mwelzl:
```suggestion
For Protocol Stacks with multiple handshakes, the decision becomes more nuanced. If the Protocol Stack involves both TLS and TCP, an implementation could determine that a leaf node is connected after the TCP handshake is complete, or it can wait for the TLS handshake to complete as well. The benefit of declaring completion when the TCP handshake finishes, and thus stopping the race for other branches of the tree, is reduced burden on the network and Remote Endpoints from further connection attempts that are likely to be abandoned. On the other hand, by waiting until the TLS handshake is complete, an implementation avoids the scenario in which a TCP handshake completes quickly, but TLS negotiation is either very slow or fails altogether in particular network conditions or to a particular endpoint. To avoid the issue of TLS possibly failing, the implementation should not generate a Ready event for the Connection until the TLS handshake is complete.
```

draft-ietf-taps-impl.md
mwelzl: I'd say the latter. The abstract and introduction make it clear which API this is about, and we mention it so often in this document, with and without a reference...

PR #1069, opened 2022-09-26
Update draft-ietf-taps-arch.md

gorryfair:
Clarified one use of word "API" comes just after discussion of the socket API. Tried to fix specialised features to say they define additional requirements. Minor changes to wording, with no intended changes to substance.

### DIFF
--- a/draft-ietf-taps-arch.md
+++ b/draft-ietf-taps-arch.md
@@ -184,7 +184,7 @@ This API also delivers events regarding the lifetime of a connection and changes
 
 Using asynchronous events allows for a more natural interaction model when establishing connections and transferring data. Events in time more closely reflect the nature of interactions over networks, as opposed to how the Socket API represent network resources as file system objects that may be temporarily unavailable.
 
-Separate from events, callbacks are also provided for asynchronous interactions with the API not directly related to events on the network or network interfaces.
+Separate from events, callbacks are also provided for asynchronous interactions with the Transport Services API, which are not directly related to events on the network or network interfaces.
 
 ## Data Transfer Using Messages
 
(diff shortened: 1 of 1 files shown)


draft-ietf-taps-arch.md
```diff
@@ -184,7 +184,7 @@ This API also delivers events regarding the lifetime of a connection and changes
 
 Using asynchronous events allows for a more natural interaction model when establishing connections and transferring data. Events in time more closely reflect the nature of interactions over networks, as opposed to how the Socket API represent network resources as file system objects that may be temporarily unavailable.
 
-Separate from events, callbacks are also provided for asynchronous interactions with the API not directly related to events on the network or network interfaces.
+Separate from events, callbacks are also provided for asynchronous interactions with the Transport Services API, which are not directly related to events on the network or network interfaces.
```
mwelzl, line 187:
```suggestion
Separate from events, callbacks are also provided for asynchronous interactions with the Transport Services API that are not directly related to events on the network or network interfaces.
```

This may seem like terrible nit-picking... and for a non-native speaker, I'm venturing far into native-speaker land!

...but: doesn't ", which" give the impression of explaining further that asynchronous interactions with the Transport Services API (in general) are not directly related to events on the network or network interfaces?  Which would be confusing, and not what we're trying to say here.

Please feel very free to ignore this comment!!

mwelzl (approved):
LGTM but I made one very small comment

draft-ietf-taps-arch.md
```diff
@@ -239,7 +239,7 @@ There are applications that will need to control fine-grained details of transpo
...
+Other specialized features, however, can also be strictly required by an application and thus furthr constrain the set of protocols that can be used. For example, if an application requires support for automatic handover or failover for a connection, only protocol stacks that provide this feature are eligible to be used, e.g., protocol stacks that include a multipath protocol or a protocol that supports connection migration. A Transport Services API needs to allow applications to define such requirements and constrain the options available to a Transport Services implementation. Since such options are not part of the core/common features, it will generally be simple for an application to modify its set of constraints and change the set of allowable protocol features without changing the core implementation.
```
tfpauly, line 242:
```suggestion
Other specialized features, however, can also be strictly required by an application and thus further constrain the set of protocols that can be used. For example, if an application requires support for automatic handover or failover for a connection, only protocol stacks that provide this feature are eligible to be used, e.g., protocol stacks that include a multipath protocol or a protocol that supports connection migration. A Transport Services API needs to allow applications to define such requirements and constrain the options available to a Transport Services implementation. Since such options are not part of the core/common features, it will generally be simple for an application to modify its set of constraints and change the set of allowable protocol features without changing the core implementation.
```

draft-ietf-taps-arch.md
```diff
@@ -496,7 +497,7 @@ Connection establishment attempts for a set of candidates may be performed simul
 
 ### Separating Connection Contexts {#conn-context}
 
-By default, stored properties of the implementation, such as cached protocol state, cached path state, and heuristics, may be shared (e.g. across multiple connections in an application). This provides efficiency and convenience for the application, since the Transport Services system can automatically optimize behavior.
+A TAPS system may be default share the stored properties of the implementation, such as cached protocol state, cached path state, and heuristics (e.g. sharing across multiple connections in an application). This provides efficiency and convenience for the application, since the Transport Services system can automatically optimize behavior.
```
tfpauly, line 500:
```suggestion
A Transport Services implementation can by default share stored properties across Connections within an application, such as cached protocol state, cached path state, and heuristics. This provides efficiency and convenience for the application, since the Transport Services system can automatically optimize behavior.
```

tfpauly (approved):

Issue #608, opened 2020-07-17
Add properties to implementation protocol mappings


tfpauly:
See discussion on #605

abrunstrom:
Provide a complete template that covers all functions/properties from the API. Indicate how the template should be used (e.g. leave out the parts that are not relevant for a particular protocol or include all fields and mark fields that are not relevant as such.

tfpauly:
I'm wondering if this is really something that's in scope that we should cover. We do have the message properties section that explains how to implement various properties, but I think adding each property to every protocol mapping will be a huge amount of text that will be difficult to get exactly right.

abrunstrom:
I agree, we have added additional properties since we opened this issue to check the mapping and make sure we have a complete coverage. We should still be consistent in how we treat things though, and perhaps add a sentence that we do not describe all the properties for instance. I have made a pass through to check the mapping from the API to impl. draft for all the protocol mappings as a basis for making things consistent. Here is the summary.
- We cover the same actions and events for all the protocols with the exceptions that UDP Multicast does not cover the ConnectionError event and SCTP covers the Priority connection property. And MPTCP and UDP-lite are special as they refer back to the base protocols.
- There is one terminology inconsistency where the impl. draft refers to InitiateError and this has been generalized to EstablishmentError in the API
- Unless I missed something, there are only two actions that can be made on a connection that are not covered, addRemote and removeRemote. This is probably fine as they do not mean much for most of the protocols (or we could add them to be complete). We should perhaps mention how they relate to MPTCP though if we leave them out.
- We also do not cover the Preconnection.Rendezvous action. Again this may be fine as it is a bit special (or we can mention it for consistency.)
- For the events, we cover a subset with main focus on the connection related events. The Closed Event and some events related to connection setup are missing. As we cover ConnectionError we should perhaps cover the Closed event as well? We do not cover any of the Send or Receive events which seems fine.
- In relation to Connection and Message properties, we do not cover any of them except for the Connection priority covered for SCTP and we talk about some properties in connection to UDP-lite and MPTCP. SCTP also has support for partial reliability and unordered messages so not sure why the priority part was singled out and added. @mwelzl any particular reasoning behind this part? Possibly we should remove the priority part for SCTP or discuss it in running text in a similar fashion as for UDP-lite and MPTCP
- We may want to update the UDP-lite description to also say something about what properties it supports over UDP. The current description sounds like it is a subset.

mwelzl:
[...]

PR #1075, opened 2022-10-10
Terminology alignment and consistent coverage of actions and events

abrunstrom:
This PR covers steps 1 and 2 in #608. When aligning the error events for UDP multicast I also noticed that the text for UDP was not correct as soft errors do not generate ConnectionError events so also attempted to fix this.

### DIFF
--- a/draft-ietf-taps-impl.md
+++ b/draft-ietf-taps-impl.md
@@ -1140,11 +1140,6 @@ ConnectionReceived:
 Clone:
 : Calling `Clone` on an SCTP association creates a new Connection object and assigns it a new stream id in accordance with the stream id assignment procedure described above. If there are not enough streams available, ADD_STREAM.SCTP must be called.
 
-Priority (Connection):
-: When this value is changed, or a Message with Message Property `msgPriority` is sent, and there are multiple
-Connection objects assigned to the same SCTP association,
-CONFIGURE_STREAM_SCHEDULER.SCTP is called to adjust the priorities of streams in the SCTP association.
-
 Send:
 : SEND.SCTP. Message Properties such as `msgLifetime` and `msgOrdered` map to parameters of this primitive.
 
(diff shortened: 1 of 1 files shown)


draft-ietf-taps-impl.md
```diff
@@ -1011,7 +1011,7 @@ Close:
 : Calling `Close` on a UDP Connection (ABORT.UDP(-Lite)) releases the local port reservation.
 
 Abort:
-: Calling `Abort` on a UDP Connection (ABORT.UDP(-Lite)) is identical to calling `Close`.
+: Calling `Abort` on a UDP Connection (ABORT.UDP(-Lite)) is identical to calling `Close`except that a ConnectionErrror Event rather than a Closed Event will be sent by the Connection.
```
tfpauly, line 1014:
```suggestion
: Calling `Abort` on a UDP Connection (ABORT.UDP(-Lite)) is identical to calling `Close`, except that the Connection will send a ConnectionError Event rather than a Closed Event.
```

draft-ietf-taps-impl.md
```diff
@@ -1160,6 +1158,9 @@ Calling `CloseGroup` calls CLOSE.SCTP, closing all Connections in the SCTP assoc
 AbortGroup:
 Calling `AbortGroup` calls ABORT.SCTP, immediately closing all Connections in the SCTP association.
 
+In addition to the API mappings described above, when there are multiple Connection objects assigned to the same SCTP association, SCTP can support Connection properties such as `connPriority`and `connScheduler` where CONFIGURE_STREAM_SCHEDULER.SCTP can be called to adjust the priorities of streams in the SCTP association.
+
+
```
tfpauly, line 1163:
```suggestion
```

tfpauly:
@abrunstrom what's the reason for having Abort cause an error, rather than just moving to closed?

abrunstrom:
@tfpauly This comes from the API draft, which says:

> Abort terminates a Connection without delivering any remaining Messages. This action does not affect any other Connection that is entangled with this one in a Connection Group. When the Abort Action has finished, the Connection will send a ConnectionError Event, indicating local Abort as a reason.

tfpauly:
Great, thank you!

tfpauly (approved):

PR #1098, opened 2023-01-19
Be more specific on namespace

### LINKED ISSUE #1092: Be more specific on namespace
https://github.com/ietf-tapswg/api-drafts/blob/9266fbe3b55966edaf4013e755d46ae6f0fed8de/draft-ietf-taps-interface.md?plain=1#L551-L553

are all the current protocols defined their Namespace in the respective RFCs? can we put a example reference? can we be more detailed on "SHOULD be defined in an RFC" actually mean?

mwelzl:
Closes #1092. This implements the change I suggested in #1092, and it also replaces "Protocol Specific Properties" with "protocol-specific Properties".

### DIFF
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -553,6 +553,7 @@ form \[\<Namespace>.\]\<PropertyName\>.
   value, in a protocol-specific property called `tcp.userTimeoutValue` (see {{tcp-uto}}).
 - Vendor or implementation specific properties MUST use a string identifying
   the vendor or implementation as the Namespace.
+- For IETF protocols, the name of a Protocol-specific Property SHOULD be specified in a IETF document published in the RFC Series.
 
 Namespaces for each of the keywords provided in the IANA protocol numbers registry
 (see https://www.iana.org/assignments/protocol-numbers/protocol-numbers.xhtml) are reserved
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -548,8 +548,9 @@ form \[\<Namespace>.\]\<PropertyName\>.
 
 - The Namespace component MUST be empty for well-known, generic properties, i.e., for
   properties that are not specific to a protocol and are defined in an RFC.
-- Protocol-specific Properties MUST use the protocol acronym as the Namespace, e.g.,
-  `tcp` for TCP-specific Transport Properties as in {{tcp-uto}}.
+- Protocol-specific Properties MUST use the protocol acronym as the Namespace (e.g., a 
+  `tcp` Connection could support a TCP-specific Transport Property, such as the user timeout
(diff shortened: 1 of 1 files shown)


philsbln (approved):
LGMT

gorryfair:
The example could be even more useful perhaps:
/Protocol-specific Properties MUST use the protocol acronym as the Namespace (e.g., a
   `tcp` connection could support a TCP-specific Transport Property, such as {{tcp-uto}}, in a protocol-specific parameter called `tcp.uto`./

mwelzl:
Done, in line with your suggestion. The wording is slightly different for several reasons: really using `tcp.uto` would be confusing because that particular property doesn't exist (there are, if you will, sub-properties). The {{tcp-uto}} reference expands as "Section ...", so using "such as" doesn't work; it's not a "protocol-specific parameter" but a "protocol-specific property".  Oh and the connection is a Connection  :-)

gorryfair:
Specific suggestion you could start from: "For IETF protocols, the name of a Protocol-specific Property SHOULD be specified in a IETF document published in the RFC Series."

gorryfair:
P.S. I would also be happy to continue with another recommendation, for additional transport properties that was proposed which says something like add it to the RFC.

draft-ietf-taps-interface.md
```diff
@@ -548,15 +548,15 @@ form \[\<Namespace>.\]\<PropertyName\>.
...
 - The Namespace component MUST be empty for well-known, generic properties, i.e., for
   properties that are not specific to a protocol and are defined in an RFC.
-- Protocol Specific Properties MUST use the protocol acronym as the Namespace, e.g.,
-  `tcp` for TCP specific Transport Properties. For IETF protocols, property
-  names under these namespaces SHOULD be defined in an RFC.
+- Protocol-specific Properties MUST use the protocol acronym as the Namespace (e.g., a
+  `tcp` Connection could support a TCP-specific Transport Property, such as the user timeout
+  value, in a protocol-specific property called `tcp.userTimeoutValue` (see {{tcp-uto}}).
 - Vendor or implementation specific properties MUST use a string identifying
   the vendor or implementation as the Namespace.
```
philsbln, line 556: Clickable version of @gorryfair's suggestion
```suggestion
- For IETF protocols, the name of a Protocol-specific Property SHOULD be specified in a IETF document published in the RFC Series.

```

draft-ietf-taps-interface.md
```diff
@@ -548,15 +548,16 @@ form \[\<Namespace>.\]\<PropertyName\>.
...
+- For IETF protocols, the name of a Protocol-specific Property SHOULD be specified in a IETF document published in the RFC Series.
```
mwelzl, line 556:
```suggestion
- For IETF protocols, the name of a Protocol-specific Property SHOULD be specified in an IETF document published in the RFC Series.
```

Issue #1082, opened 2023-01-13
Should there be no requirement on letting the user of the TAPS system about a special feature not applied to the Connection?


zaheduzzaman:
https://github.com/ietf-tapswg/api-drafts/blob/9266fbe3b55966edaf4013e755d46ae6f0fed8de/draft-ietf-taps-arch.md?plain=1#L240

If an application wishes to use a special feature then if might need to adjust itself if that special feature is not applicable to the Connection Object it gets/ gets created for. Should there be no requirement(at least recommendation) on getting this information back? 

The application can ask a specialized feature in a way that it gets what it wants or the communication MUST fail. In such case the should be a requirement on the application to be specific about it or?

philsbln:
The API solves this in two ways:
- We have a preference-type that allows the application to explicitly state that it requires/prefers/does-not-care/wants-to-avoid/prohibits each feature.
- After the establishment, the application can query each feature and gets a boolean back whether it is present or not.
Should we add this as a forward reference here or at least hint at the solution?

PR #1103, opened 2023-01-23
Closes Issue #1082 by hinting at the API gestalt

philsbln:


draft-ietf-taps-arch.md
```diff
@@ -241,6 +241,8 @@ A specialized feature could be needed by an application only when using a specif
...
 
+To control these specialized features, the application can declare its preference – whether the presence of a specific feature is prohibited, should be avoided, can be ignored, is preferred, or is required in the pre establishment phase. An implementation of A Transport Services API would honor this preference and allow the application to query the availability of each specialized feature after a successful establishment.
```
mwelzl, line 244:
```suggestion
To control these specialized features, the application can declare its preference – whether the presence of a specific feature is prohibited, should be avoided, can be ignored, is preferred, or is required in the Pre-Establishment phase. An implementation of a Transport Services API would honor this preference and allow the application to query the availability of each specialized feature after a successful establishment.
```

mwelzl (approved):

tfpauly (approved):

PR #1105, opened 2023-01-23
Mention future minset additions

### LINKED ISSUE #1081: Discussion about impact on the architecture when the minset expands
https://github.com/ietf-tapswg/api-drafts/blob/9266fbe3b55966edaf4013e755d46ae6f0fed8de/draft-ietf-taps-arch.md?plain=1#L226

Assuming there would be more protocols in the future and the minset functionalities will expand via extension or amendments. It would important to indicate if this architecture should is also extensible or it only supports on the minimum set referred. Definitely we haven't seen the future yet, but it would be important to express if we have considered such expansion of minimum functionalities.

tfpauly:
Closes #1081

### DIFF
--- a/draft-ietf-taps-arch.md
+++ b/draft-ietf-taps-arch.md
@@ -223,7 +223,7 @@ The normative requirements described in this section allow Transport Services AP
 
 ## Provide Common APIs for Common Features
 
-Any functionality that is common across multiple transport protocols SHOULD be made accessible through a unified set of calls using the Transport Services API. As a baseline, any Transport Services API SHOULD allow access to the minimal set of features offered by transport protocols {{?RFC8923}}. If that minimal set is updated and expanded in the future, the Transport Services API ought to be extended to match, if necessary.
+Any functionality that is common across multiple transport protocols SHOULD be made accessible through a unified set of calls using the Transport Services API. As a baseline, any Transport Services API SHOULD allow access to the minimal set of features offered by transport protocols {{?RFC8923}}. If that minimal set is updated or expanded in the future, the Transport Services API ought to be extended to match.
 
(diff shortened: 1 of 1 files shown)


draft-ietf-taps-arch.md
```diff
@@ -223,7 +223,7 @@ The normative requirements described in this section allow Transport Services AP
 
 ## Provide Common APIs for Common Features
 
-Any functionality that is common across multiple transport protocols SHOULD be made accessible through a unified set of calls using the Transport Services API. As a baseline, any Transport Services API SHOULD allow access to the minimal set of features offered by transport protocols {{?RFC8923}}.
+Any functionality that is common across multiple transport protocols SHOULD be made accessible through a unified set of calls using the Transport Services API. As a baseline, any Transport Services API SHOULD allow access to the minimal set of features offered by transport protocols {{?RFC8923}}. If that minimal set is updated and expanded in the future, the Transport Services API ought to be extended to match, if necessary.
```
mwelzl, line 226:
```suggestion
Any functionality that is common across multiple transport protocols SHOULD be made accessible through a unified set of calls using the Transport Services API. As a baseline, any Transport Services API SHOULD allow access to the minimal set of features offered by transport protocols {{?RFC8923}}. If that minimal set is updated and expanded in the future, the Transport Services API ought to be extended to match.
```

I just find the "if necessary"  a bit too much ... "if it happens, it ought to, if necessary". I'm not even sure what "if necessary" means here... when would doing that become necessary, and when not? What would make it necessary - compatibility, business reasons?

mwelzl (approved):
LGTM modulo a nit

draft-ietf-taps-arch.md
gorryfair: also: I think it's /updated **or** extended/ not both.

tfpauly:
```suggestion
Any functionality that is common across multiple transport protocols SHOULD be made accessible through a unified set of calls using the Transport Services API. As a baseline, any Transport Services API SHOULD allow access to the minimal set of features offered by transport protocols {{?RFC8923}}. If that minimal set is updated or expanded in the future, the Transport Services API ought to be extended to match.
```

tfpauly: I originally meant "if necessary" to say that the minset change might not require an API change, but I can remove.

PR #1106, opened 2023-01-23
Clarify protocol stack equivalence

### LINKED ISSUE #1083: Several issues in section 3.3
https://github.com/ietf-tapswg/api-drafts/blob/9266fbe3b55966edaf4013e755d46ae6f0fed8de/draft-ietf-taps-arch.md?plain=1#L246

Here it seems we are defining "equivalent" twice. I would rather write - 

>Equivalent Protocol Stacks are defined as stacks that can provide the same Transport Properties and interface expectations as requested by the application. and the equivalent stacks can be swapped or raced. 

Also in the example - it is not clear that swapping and racing in important, however, application provided properties are.

Apart from this I also think the title and content is mismatching. I found the last paragraph talking about swapping and racing focusing on security. Here we are imposing requirements on swapping and racing. These are very important for the security consideration and to me should get its own title and section separated from selection of equivalence.

tfpauly:
Closes #1083


mwelzl (approved):
LGTM!

gorryfair (approved):
Looks good

PR #1109, opened 2023-01-31
Normatively forbid multiple identifiers per endpoint.

### LINKED ISSUE #1093: Should we use normative language for restricting multiple identifiers on a endpoint?
https://github.com/ietf-tapswg/api-drafts/blob/9266fbe3b55966edaf4013e755d46ae6f0fed8de/draft-ietf-taps-interface.md?plain=1#L720-L721

This seems to me as a strong claim without using normative language. Shall we use normative language here?

mwelzl:
Closes #1093.


gorryfair (commented):
I like the change. Is the English clunky, could it be:

An Endpoint MUST NOT be configured with multiple identifiers of a same type.

mwelzl:
I'm happy to do this!  But while we're at it, isn't the "a" a bit awkward?  Wouldn't this read even better as:

An Endpoint MUST NOT be configured with multiple identifiers of the same type.

tfpauly (approved):

Issue #1091, opened 2023-01-13
Some minor comments on architecture document


zaheduzzaman:
These comments are good to be addressed to improve the document, however, they can be done as part of IETF last call as well

> Please put forward ref to major concepts that are later defined in the document when using earlier than defined. This would avoid unnecessary confusions and misinterpretation. There are already good examples of such forward reference in the document. The ask here is to be consisted with that. 

> "the nature of the interaction over networks" has been one of the motivations for asynchronous event driven TAPS API , I think it would be great to provide some details of the nature of the interaction over networks so that readers are on the same page.

> Can we mention in section 2.2 that a packet might content more than one messages? This will make the 2nd bullet of the benefits more appealing.

> The last paragraph of section 2.2 would be better to split into the benefit bullets.

> Can we define the term like "protocol racing"?  

> Section 4.1 : looking at the architecture figure - "state is torn down in the transport" should this be connection instead of transport?

> Section 4.1.2 : I would describe the object first and then the transport properties. In the current arrangement when properties are described it mentions Connection which is described just later. Also transport properties here are kind of repetition of what we have in Section 3.1.

> Section 4.1.7 : a ref to the "half-close" concept would be helpful, as we are not defining it.

gorryfair:
Fixes proposed in PRs #1110, #1111, #1112, #1113, #1114

PR #1110, opened 2023-02-15
A glossary?

gorryfair:
This provides a glossary. Each glossary item is made from shortening the laster definition in the architecture document. It is intended to avoid a reader being overwhelmed by the new vocabulary. Intended to help address point 1 of  #1091.

### DIFF
--- a/draft-ietf-taps-arch.md
+++ b/draft-ietf-taps-arch.md
@@ -495,7 +495,7 @@ The following categories of events can be delivered to an application:
 
 A Connection Group is a set of Connections that shares properties and caches. A Connection Group represents state for managing Connections within a single application, and does not require end-to-end protocol signaling. For multiplexing transport protocols, only Connections within the same Connection Group are allowed to be multiplexed together.
 
-When the API clones an existing Connection, this adds a new Connection to the Connection Group. A change to one of the Connection Properties on any Connection in the Connection Group automatically changes the Connection Property for all others. All Connections in a Connection Group share the same set of Connection Properties except for the Connection Priority. These Connection Properties are said to be entangled.
+The API allows a Connection to be created from another Connection.  This adds the new Connection to the Connection Group. A change to one of the Connection Properties on any Connection in the Connection Group automatically changes the Connection Property for all others. All Connections in a Connection Group share the same set of Connection Properties except for the Connection Priority. These Connection Properties are said to be entangled.
 
 For multiplexing transport protocols, only Connections within the same Connection Group are allowed to be multiplexed together. Passive Connections can also be added to a Connection Group, e.g., when a Listener receives a new Connection that is just a new stream of an already active multi-streaming protocol
 instance.
--- a/draft-ietf-taps-arch.md
+++ b/draft-ietf-taps-arch.md
@@ -113,6 +113,7 @@ they appear in all capitals, as shown here.
 This subsection provides a brief gloassary of key terms related to the architecture. These terms ar defined in the relevent sections of this document.
 
 - Cached State: The state and history that the implementation keeps for each set of associated Endpoints that have been used previously.
+- Clone: A copy of a Connection that forms a part of a Connection Group.
 - Connection: An object that can be used to send and receive messages.
 - Connection Group: A set of Connections that shares properties and caches.
 - Connection Property: A Transport Property that can be used to configure protocol-specific options and control per-connection behavior of a Transport Services implementation.
--- a/draft-ietf-taps-arch.md
+++ b/draft-ietf-taps-arch.md
@@ -118,7 +118,7 @@ This subsection provides a glossary of key terms related to the Transport Servic
 - Clone: A Connection that was created from another Connection, and forms a part of a Connection Group.
 - Connection: Shared state of two or more Endpoints that persists across Messages that are transmitted and received between these Endpoints {{?RFC8303}}.
 - Connection Group: A set of Connections that shares properties and caches.
-- Connection Property: A Transport Property that can be used to configure protocol-specific options and control per-connection behavior of a Transport Services implementation.
+- Connection Property: A Transport Property that controls per-Connection behavior of a Transport Services implementation.
 - Endpoint: An identifier for one side of a transport connection.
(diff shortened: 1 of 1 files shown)


mwelzl:
To ensure consistency, I suggest that you look for the definitions that we already have agreed on and published in prior RFCs, and replace the ones for the same words that you already have with the ones from these RFCs.

RFC 8095 defines: Transport Feature, Transport Service, Transport Protocol, Application
RFC 8303 defines, in addition: Transport Protocol Component, Transport Service Instance, Endpoint, Connection, Primitive, Event, Parameter, Socket, Transport Address
RFC 8304 defines, in addition: Application-specific knowledge, End System, Connection Group

mwelzl (changes requested):
Many thanks for doing this!  This must have been a lot of work. However, for consistency with prior work, I would suggest to replace the descriptions of already defined terms with the definitions that are already published in earlier TAPS RFCs. (I remember that a lot of effort has initially gone into debating terminology... plus, these published definitions are already agreed upon)

gorryfair:
Now updated also by looking for definitions in other published TAPS RFCs.

draft-ietf-taps-arch.md
```diff
@@ -108,6 +108,46 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT",
...
+- Protocol Stack: A set of Protocol Instances that are used together to establish connectivity or send and receive Messages.
+- Racing: The attempt to select between multiple Protocol Stacks based on the Selection and Connection Properties communicated by the application, along with any security parameters.
+- Remote Endpoint: A representation of the application's identifier for a peer that can participate in a transport connection.
+- Rendezvous: The action of establishing a peer-to-peer connection with a Remote Endpoint.
+- Security Parameters: Parameters that define an application's requirements for authentication and encryption on a Connection.
+- Server: The peer responsible for responding to a session initiation.
+- Socket: The combination of a destination IP address and a destination port number {{?RFC8303}}..
+- System Policy: The input from an operating system or other global preferences that can constrain or influence how an implementation will gather candidate paths and Protocol Stacks and race the candidates during connection establishment.
```
mwelzl, line 144:
```suggestion
- System Policy: The input from an operating system or other global preferences that can constrain or influence how an implementation will gather candidate paths and Protocol Stacks and race the candidates during Connection establishment.
```

I keep fixing "connection" to "Connection" because in some cases I feel it is an important semantic difference. E.g., to me, "connection establishment" is the TCP handshake (in case of TCP), whereas "Connection establishment" is what a Transport System does after Initiate() is called, including racing etc.

Now, if we want to be super picky, we should probably also write "candidate Paths" here, since we're also writing "Protocol Stacks"...  maybe it would be better to write all of them in lower case though, as there won't be a misunderstanding about what they are anyway?

draft-ietf-taps-arch.md
```diff
@@ -108,6 +108,46 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT",
...
+- Selection Property: A Transport Property that can set to influence the selection of paths between the Local and Remote Endpoints.
```
mwelzl, line 145:
```suggestion
- Selection Property: A Transport Property that can be set to influence the selection of paths between the Local and Remote Endpoints.
```

mwelzl (changes requested):
Many small things... but again many thanks for doing this, this is very useful!

draft-ietf-taps-arch.md
```diff
@@ -108,6 +108,46 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT",
...
+This subsection provides a brief gloassary of key terms related to the architecture. These terms ar defined in the relevent sections of this document.
+
+- Application: An entity that uses the transport layer for end-to-end delivery data across the network {{?RFC8095}}.
+- Cached State: The state and history that the implementation keeps for each set of associated Endpoints that have been used previously.
+- Client: The peer responsible for initiating a session.
+- Clone: A copy of a Connection that forms a part of a Connection Group.
+- Connection: Shared state of two or more Endpoints that persists across Messages that are transmitted and received between these Endpoints {{?RFC8303}}.
+- Connection Group: A set of Connections that shares properties and caches.
+- Connection Property: A Transport Property that can be used to configure protocol-specific options and control per-connection behavior of a Transport Services implementation.
+- Endpoint: An identifier for one side of a transport connection.
```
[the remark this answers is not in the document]
gorryfair: Unsure. This is what I read... will fix /endpoint/

draft-ietf-taps-arch.md
```diff
@@ -108,6 +108,48 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT",
...
+- Primitive: A function call that is used to locally communicate between an application and an endpoint, which is
+related to one or more Transport Features {{?RFC8303}}.
+- Protocol Instance: A single instance of one protocol, including any state necessary to establish connectivity or send and receive Messages.
+- Protocol Stack: A set of Protocol Instances that are used together to establish connectivity or send and receive Messages.
+- Racing: The attempt to select between multiple Protocol Stacks based on the Selection and Connection Properties communicated by the application, along with any security parameters.
+- Remote Endpoint: A representation of the application's identifier for a peer that can participate in establishing a Connection.
+- Rendezvous: The action of establishing a peer-to-peer Connection with a Remote Endpoint.
+- Security Parameters: Parameters that define an application's requirements for authentication and encryption on a Connection.
+- Server: The peer responsible for responding to a Connection initiation.
+- Socket: The combination of a destination IP address and a destination port number {{?RFC8303}}..
```
mwelzl, line 145:
```suggestion
- Socket: The combination of a destination IP address and a destination port number {{?RFC8303}}.
```
Perhaps you missed the suggestion: I fixed a double full stop.

mwelzl (approved):
Many thanks!  Such a large effort!

gorryfair:
I will close and then merge this in a new PR - can work out how to resolve the conflict via the web interface:-(

PR #1111, opened 2023-02-15
Messages and Packets

gorryfair:
Proposal to address review comment: Can we mention in section 2.2 that a packet might content more than one messages? This will make the 2nd bullet of the benefits more appealing.  The last paragraph of section 2.2 would be better to split into the benefit bullets.


draft-ietf-taps-arch.md
```diff
@@ -223,7 +223,10 @@ The Transport Services API represents data as messages, so that it more closely
 * the ability to control reliability, which messages to retransmit when there is packet loss, and how best to make use of the data that arrived;
 * the ability to automatically assign messages and connections to underlying transport connections to utilize multi-streaming and pooled connections.
 
-Allowing applications to interact with messages is backwards-compatible with existing protocols and APIs because it does not change the wire format of any protocol. Instead, it gives the protocol stack additional information to allow it to make better use of modern transport services, while simplifying the application's role in parsing data. For protocols which natively use a streaming abstraction, framers ({{datatransfer}}) bridge the gap between the two abstractions.
+This can have several benefits:
```
mwelzl, line 226: This is strange now. I suggest to look at the rich diff and show text above this PR as well: just above, there is already a bullet list of possible benefits.

draft-ietf-taps-arch.md
```diff
@@ -444,7 +447,7 @@ The diagram below provides a high-level view of the actions and events during th
 
 ### Data Transfer Objects and Actions {#datatransfer}
 
-* Message: A Message object is a unit of data that can be represented as bytes that can be transferred between two endpoints over a transport connection. The bytes within a Message are assumed to be ordered. If an application does not care about the order in which a peer receives two distinct spans of bytes, those spans of bytes are considered independent Messages.
+* Message: A Message object is a unit of data that can be represented as bytes that can be transferred between two endpoints over a transport connection. The bytes within a Message are assumed to be ordered. If an application does not care about the order in which a peer receives two distinct spans of bytes, those spans of bytes are considered independent Messages. Messages are sent in the payload of IP packet. One packet can carry one or more message or parts of a message.
```
mwelzl, line 450:
```suggestion
* Message: A Message object is a unit of data that can be represented as bytes that can be transferred between two endpoints over a transport connection. The bytes within a Message are assumed to be ordered. If an application does not care about the order in which a peer receives two distinct spans of bytes, those spans of bytes are considered independent Messages. Messages are sent in the payload of IP packet. One packet can carry one or more Messages or parts of a Message.
```

mwelzl (changes requested):
See my in line comments... a nit towards the bottom, but more importantly, I think you introduced an oddity by repeating the bullet list style just below an already existing one.

draft-ietf-taps-arch.md
gorryfair: You're correct - fixed one part to mess up another, I think I see how it should have been, will return to that

mwelzl (changes requested):
LGTM. But careful!  Something strange happened here. You seem to have the glossary here, too?  With typos etc. that we're fixing in the other PR.  One thing that could work would be to merge this one first, and then hope that the other one fixes these things...  but I think it would be better to remove the whole glossary stuff from this PR!

Either way, let's please avoid that we fix things in the other PR, and then overwrite parts of the glossary with this one!!

gorryfair:
Based on Main and merged tecxt - rather than using this PR

Issue #1107, opened 2023-01-30
Protocol option examples are hard!!!


zaheduzzaman:
https://github.com/ietf-tapswg/api-drafts/blob/9a98252f1b9c790a17df01f1ed9be458bfb2ea3d/draft-ietf-taps-impl.md?plain=1#L302-L317

We have given example but we haven't really defined how
 the format should be written or how one should interpret in the
example. Here "Any Stream" means any protocol that supports
streaming ... but it would take quit some time to get to that point . It is also confusing that an "Any" is used with comma (,) and another is without a
 comma... 

To understand the example correctly, I suggest we either keep the examples on to a level where this is already described or describe how we are formatting the expression in this document as we described tree
structure with numeric ( for example 1 , 1,1 ).

abrunstrom:
Perhaps we can just write HTTP and Stream without the Any part? Alternatively we could add a specific interface, e.g Wi-Fi, to avoid the first Any. And then also clarify the description of the example and what is raced. 

The notation is described earlier in the text, so I assume that is not the problem?

zaheduzzaman:
If you propose the changes then I can perhaps understand it better about the modified example.

philsbln:
@zaheduzzaman does PR #1119 solve the issue for you?

PR #1119, opened 2023-02-28
Clarify protocol examples in implemenation

philsbln:
Fixes Issue #1107

### DIFF
--- a/draft-ietf-taps-impl.md
+++ b/draft-ietf-taps-impl.md
@@ -213,7 +213,7 @@ The example aggregate connection attempt above can be drawn as a tree by groupin
   +====================+  +====================+  +======================+
 ~~~~~~~~~~
 
-The rest of this section will use a notation scheme to represent this tree. The parent (or trunk) node of the tree will be represented by a single integer, such as "1". ("1" is used assuming that this is the first connection made by the system; future connections created by the application would allocate numbers in an increasing manner.) Each child of that node will have an integer that identifies it, from 1 to the number of children. That child node will be uniquely identified by concatenating its integer to it's parents identifier with a dot in between, such as "1.1" and "1.2". Each node will be summarized by a tuple of three elements: endpoint, path (labeled here by interface), and protocol. The above example can now be written more succinctly as:
+The rest of this section will use a notation scheme to represent this tree. The parent (or trunk) node of the tree will be represented by a single integer, such as "1". ("1" is used assuming that this is the first connection made by the system; future connections created by the application would allocate numbers in an increasing manner.) Each child of that node will have an integer that identifies it, from 1 to the number of children. That child node will be uniquely identified by concatenating its integer to it's parents identifier with a dot in between, such as "1.1" and "1.2". Each node will be summarized by a tuple of three elements: endpoint, path (labeled here by interface), and protocol. In protocol stacks, the layers are separated by '/' and ordered top-down. The above example can now be written more succinctly as:
 
 ~~~~~~~~~~
 1 [www.example.com:80, Any, TCP]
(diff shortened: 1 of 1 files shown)


draft-ietf-taps-impl.md
```diff
@@ -289,31 +289,31 @@ Differences in possible protocol compositions and options can also provide a bra
 This approach is commonly used for connections with optional proxy server configurations. A single connection might have several options available: an HTTP-based proxy, a SOCKS-based proxy, or no proxy. As above, these options should be ranked based on preference, system policy, and performance and attempted in succession.
 
 ~~~~~~~~~~
-1 [www.example.com:80, Any, HTTP/TCP]
-  1.1 [192.0.2.8:80, Any, HTTP/HTTP Proxy/TCP]
-  1.2 [192.0.2.7:10234, Any, HTTP/SOCKS/TCP]
-  1.3 [www.example.com:80, Any, HTTP/TCP]
-    1.3.1 [192.0.2.1:80, Any, HTTP/TCP]
+1 [www.example.com:80, Wi-Fi, HTTP/TCP]
```
tfpauly, line 292: These examples of "Any" had meant "no interface preference". Changing them all to require Wi-Fi seems wrong?

draft-ietf-taps-impl.md
```diff
@@ -213,7 +213,7 @@ The example aggregate connection attempt above can be drawn as a tree by groupin
...
+The rest of this section will use a notation scheme to represent this tree. The parent (or trunk) node of the tree will be represented by a single integer, such as "1". ("1" is used assuming that this is the first connection made by the system; future connections created by the application would allocate numbers in an increasing manner.) Each child of that node will have an integer that identifies it, from 1 to the number of children. That child node will be uniquely identified by concatenating its integer to it's parents identifier with a dot in between, such as "1.1" and "1.2". Each node will be summarized by a tuple of three elements: endpoint, path (labeled here by interface), and protocol. In protocol stacks, the layers are separated by '/' and ordered top-down. The above example can now be written more succinctly as:
```
tfpauly, line 216: Where is the '/' used?

tfpauly (changes requested):

draft-ietf-taps-impl.md
philsbln: Before reading #1107 I had argued the same, but I guess the "Any" easily distracts from the protocol as its meaning is not obvious on first sight. Having anything non-changing in the examples solves that issue.
As an Alternative, I would suggest replacing all occurrences of `Any` in the path column with `AnyPath` 
Does this work four you @tfpauly?

philsbln: Before reading #1107 I had argued the same, but I guess the "Any" easily distracts from the protocol as its meaning is not obvious on first sight. Having anything non-changing in the examples solves that issue.
As an Alternative, I would suggest replacing all occurrences of `Any` in the path column with `AnyPath` 
Does this work four you @tfpauly?

draft-ietf-taps-impl.md
philsbln: in the protocol examples below

draft-ietf-taps-impl.md
abrunstrom: An alternative would be to explain the complete example, mentioning also the endpoint and the meaning of Any for the path. But I am fine with AnyPath.

draft-ietf-taps-impl.md
```diff
@@ -289,31 +289,31 @@ Differences in possible protocol compositions and options can also provide a bra
...
 
 Another example is racing SCTP with TCP:
 
 ~~~~~~~~~~
-1 [www.example.com:80, Any, Any Stream]
-  1.1 [www.example.com:80, Any, SCTP]
-    1.1.1 [192.0.2.1:80, Any, SCTP]
-  1.2 [www.example.com:80, Any, TCP]
-    1.2.1 [192.0.2.1:80, Any, TCP]
+1 [www.example.com:80, Wi-Fi, Stream+Reliable]
```
abrunstrom, line 312: We have a transport property profile described in the API draft called reliable-inorder-stream. Should we use that rather than Stream+Reliable? Just a thought, I am also fine with Stream+Reliable.

abrunstrom (approved):
Thanks for the PR. I put two comments as input. I am fine merging it once Tommy's questions are resolved.

draft-ietf-taps-impl.md
tfpauly: Hm, I didn't see see that? Can you point to a specific line?

draft-ietf-taps-impl.md
tfpauly: I think I'd prefer using "Any Path" (written out with a space). But that looks good.

draft-ietf-taps-impl.md
tfpauly: +1 to using "reliable-inorder-stream"

draft-ietf-taps-impl.md
philsbln: Sure … 292-296, 302-306

draft-ietf-taps-impl.md
philsbln: looks good. After editing, I needed to adjust to lower case for readability and change at a few additional places for consistency.

draft-ietf-taps-impl.md
philsbln: done

draft-ietf-taps-impl.md
tfpauly: Oh, I see now. thanks!

tfpauly (approved):

PR #1122, opened 2023-03-07
More text on framers in -arch

### LINKED ISSUE #1096: Placement of section 9.1.2 and Section 11 of Interface document
This I would like us to discuss.

Well I know we have discussed this before. As I am reading the Interface document after a long while and having read it along with the architecture document, I kind of strongly feel that the architecture document lack of description on framer and the states. These are mentioned and pictured in the architecture document but haven't been clear there. Hence, I kind of feel that we need to move this two sections to architecture document. Please discuss so that I can understand the reasoning of not doing so.

mwelzl:
Closes #1096 

(sorry that it took me so long)

This uses some text that is also in the interface draft.


draft-ietf-taps-arch.md
```diff
@@ -487,7 +487,7 @@ a connection.
...
 
-* Framer: A Framer is a data translation layer that can be added to a Connection to define how application-layer Messages are transmitted over a Protocol Stack. This is particularly relevant when using a protocol that otherwise presents unstructured streams, such as TCP.
+* Framer: A Framer is a data translation layer that can be added to a Connection. Framers allow extending a Connection's protocol stack to define how to encapsulate or encode outbound Messages, and how to decapsulate or decode inbound data into Messages. This allows message boundaries to be preserved when using a Connection object, even with a protocol that otherwise presents unstructured streams, such as TCP. This is designed based on the fact that many of the current application protocols evolved over TCP, which does not provide message boundary preservation, and since many of these protocols require message boundaries to function, each application layer protocol has defined its own framing. For example, when an HTTP application sends and receives HTTP messages over a byte-stream transport, it must parse the boundaries of HTTP messages from the stream of bytes.
```
mwelzl, line 490:
```suggestion
* Framer: A Framer is a data translation layer that can be added to a Connection. Framers allow extending a Connection's protocol stack to define how to encapsulate or encode outbound Messages, and how to decapsulate or decode inbound data into Messages. In this way, message boundaries can be preserved when using a Connection object, even with a protocol that otherwise presents unstructured streams, such as TCP. This is designed based on the fact that many of the current application protocols evolved over TCP, which does not provide message boundary preservation, and since many of these protocols require message boundaries to function, each application layer protocol has defined its own framing. For example, when an HTTP application sends and receives HTTP messages over a byte-stream transport, it must parse the boundaries of HTTP messages from the stream of bytes.
```

tfpauly (approved):
Looks great!

Issue #1124, opened 2023-03-10
Let's be clear the glossary is a glossary


gorryfair:
BT: "I’m not convinced of the value of the glossary of terms, as it forwardrefs basically the entire document, and some of the definitions are so generic as to be potentially confusing (e.g. Event), so I think putting these up front reduces the document’s approachability and readability.  I understand that some people like these (and other standards orgs make them mandatory, hi ETSI), so I won’t fight it."

"Adding a sentence making it clear it’s safe to skip the glossary if you’re going to read the rest of the document would address my approachability concern completely. 😄"

PR #1125, opened 2023-03-10
An attempt to craft a sentence for Issue #1124

gorryfair:
Feel free to approve or propose changes to improve this sentence in Arch.

### DIFF
--- a/draft-ietf-taps-arch.md
+++ b/draft-ietf-taps-arch.md
@@ -110,7 +110,7 @@ they appear in all capitals, as shown here.
 
 ## Glossary of Key Terms
 
-This subsection provides a glossary of key terms related to the Transport Services architecture. It provides a simple description of key terms that are later defined in this document.
+This subsection provides a glossary of key terms related to the Transport Services architecture. It provides a short description of key terms that are later defined in this document.
 
 - Application: An entity that uses the transport layer for end-to-end delivery of data across the network {{?RFC8095}}.
 - Cached State: The state and history that the implementation keeps for each set of associated Endpoints that have been used previously.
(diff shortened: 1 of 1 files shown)


draft-ietf-taps-arch.md
```diff
@@ -110,7 +110,7 @@ they appear in all capitals, as shown here.
 
 ## Glossary of Key Terms
 
-This subsection provides a glossary of key terms related to the Transport Services architecture.
+This subsection provides a glossary of key terms related to the Transport Services architecture. It provides a simple description of key terms that are later defined in this document.
```
abrunstrom, line 113: Would "summary" be better than "simple description"?

gorryfair: Would removing "simple" be a useful step in this direction?

abrunstrom: Yes, "simple" was the word that triggered my response, as I did not find that it read well. I am fine with just removing "simple". If you want another word, perhaps we could use "short" instead of "simple", but just removing "simple" is the easiest.

gorryfair: I have tried /simple/short/ at your suggestion.

draft-ietf-taps-arch.md
```diff
@@ -110,7 +110,7 @@ they appear in all capitals, as shown here.
 
 ## Glossary of Key Terms
 
-This subsection provides a glossary of key terms related to the Transport Services architecture.
+This subsection provides a glossary of key terms related to the Transport Services architecture. It provides a short description of key terms that are later defined in this document.
```
britram, line 113: retroactive lgtm :)

PR #1127, opened 2023-03-29
Two normative language changes

mwelzl:
According to:
https://mailarchive.ietf.org/arch/msg/taps/BrKFfpcnwbTneXfhGAqvdnnaa4Q/

As a small change to Gorry's second proposal, this uses:
"The design of the API MUST NOT permit an Endpoint to be configured with multiple identifiers of the same type."
instead of:
"The design of the API MUST NOT permit an Endpoint to set multiple identifiers of the same type."
because it's not the Endpoint itself that is actually doing anything.


gorryfair (approved):
looks fine to me

PR #1130, opened 2023-04-24
Another nit

seanturner:
There's four of them right so it's "... these properties ..."


mwelzl (approved):

gorryfair (approved):
Approved, and I have already changed /some Remote Endpoint/ to /a Remote Endpoint/

PR #1157, opened 2023-05-13
Connection termination update

### LINKED ISSUE #1149: Connection termination
8.  Implementing Connection Termination

I think this section needs substantial revision.  As far as I can
tell, the core of it is "Once you send a close on an SCTP connection,
any data in flight from the remote end will be lost, so in TAPS, the
semantics of a reliable, bidirectional, byte-stream connection are
that once the application calls Close, any data in flight from the
remote end MAY be lost."  As a design choice, I would argue against it
(and that the implementation cover over that SCTP behavior), but I
assume the authors know better than I do.

However, it's critically important that this is stated in the API
definition, as the application writer might not expect that property,
and certainly can't be expected to read the implementation
documentation to discover that it doesn't get quite as good service as
TCP provides.

Once that is done, this section can be reduced to mentioning that
because the API permits the loss of tail-end data, a reliable,
bidirectional, byte-stream connection can be mapped directly to an
SCTP stream.

***

mwelzl:
Closes #1149.

### DIFF
--- a/draft-ietf-taps-impl.md
+++ b/draft-ietf-taps-impl.md
@@ -799,7 +799,7 @@ The strategy to do so is implementation-specific, but should be consistent with 
 
 # Implementing Connection Termination
 
-For Close (provoking a Finished event) and Abort (provoking a ConnectionError event), it
+For Close (which leads to a Finished event) and Abort (which leads to a ConnectionError event),
 is desirable to be informed when a peer closes or aborts a
 Connection. Whether this is possible depends on the underlying protocol, and no guarantees
 can be given. When a Connection is mapped to an SCTP stream, the transport system can use the stream reset procedure to cause a Finish event upon a Close action from the peer {{NEAT-flow-mapping}}.
--- a/draft-ietf-taps-impl.md
+++ b/draft-ietf-taps-impl.md
@@ -799,39 +799,10 @@ The strategy to do so is implementation-specific, but should be consistent with 
 
 # Implementing Connection Termination
 
-With TCP, when an application closes a connection, this
-means that it has no more data to send (but expects all data that has been
-handed over to be reliably delivered). However, with TCP only, "close" does
-not mean that the application will stop receiving data. This is related to TCP's ability to
-support half-closed connections.
-
-SCTP is an example of a protocol that does not support such half-closed connections.
-Hence, with SCTP, the meaning of "close" is stricter: an application has no more data
-to send (but expects all data that has been handed over to be reliably delivered), and will
-also not receive any more data.
-
-Implementing a protocol independent transport system means that the exposed
-semantics must be the strictest subset of the semantics of all supported protocols.
-Hence, as is common with all reliable transport protocols, after a Close action, the
-application can expect to have its reliability requirements honored regarding the data
-provided to the Transport Services API, but it cannot expect to be able to read any
-more data after calling Close.
-
-Abort differs from Close only in that no guarantees are given regarding any data
-that the application sent to the Transport Services API before calling Abort.
-
-As explained in {{establish-mux}}, when a new stream is multiplexed on an already
-existing connection of a Transport Protocol Instance, there is no need for a connection
-establishment procedure. Because the Connections that are offered by a Transport Services implementation
-can be implemented as streams that are multiplexed on a transport protocol's connection,
-it can therefore not be guaranteed an Initiate action from one endpoint
-provokes a ConnectionReceived event at its peer.
-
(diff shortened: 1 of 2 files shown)


draft-ietf-taps-interface.md
```diff
@@ -3262,6 +3262,9 @@ for a Message handed over before calling Close, the Closed Event will signify
 that this Message has indeed been delivered. This Action does not affect any other Connection
 in the same Connection Group.
 
+Applications MUST NOT assume that it can receive any further data on a Connection
```
tfpauly, line 3265:
```suggestion
An application MUST NOT assume that it can receive any further data on a Connection
```

draft-ietf-taps-impl.md
```diff
@@ -799,39 +799,10 @@ The strategy to do so is implementation-specific, but should be consistent with
...
-can be implemented as streams that are multiplexed on a transport protocol's connection,
-it can therefore not be guaranteed an Initiate action from one endpoint
-provokes a ConnectionReceived event at its peer.
-
-For Close (provoking a Finished event) and Abort (provoking a ConnectionError event), the
-same logic applies: while it is desirable to be informed when a peer closes or aborts a
-Connection, whether this is possible depends on the underlying protocol, and no guarantees
-can be given. With SCTP, the transport system can use the stream reset procedure to cause
-a Finish event upon a Close action from the peer {{NEAT-flow-mapping}}.
+For Close (provoking a Finished event) and Abort (provoking a ConnectionError event), it
```
tfpauly, line 802:
```suggestion
For Close (which leads to a Finished event) and Abort (which leads to a ConnectionError event),
```

draft-ietf-taps-impl.md
```diff
@@ -799,39 +799,10 @@ The strategy to do so is implementation-specific, but should be consistent with
...
+is desirable to be informed when a peer closes or aborts a
+Connection. Whether this is possible depends on the underlying protocol, and no guarantees
+can be given. When a Connection is mapped to an SCTP stream, the transport system can use the stream reset procedure to cause a Finish event upon a Close action from the peer {{NEAT-flow-mapping}}.
```
tfpauly, line 805:
```suggestion
can be given. For example, when a Connection is mapped to an SCTP stream, the transport system can use the stream reset procedure to cause a Finish event upon a Close action from the peer {{NEAT-flow-mapping}}.
```

draft-ietf-taps-impl.md
```diff
@@ -799,39 +799,10 @@ The strategy to do so is implementation-specific, but should be consistent with
...
-it can therefore not be guaranteed an Initiate action from one endpoint
-provokes a ConnectionReceived event at its peer.
-
-For Close (provoking a Finished event) and Abort (provoking a ConnectionError event), the
-same logic applies: while it is desirable to be informed when a peer closes or aborts a
-Connection, whether this is possible depends on the underlying protocol, and no guarantees
-can be given. With SCTP, the transport system can use the stream reset procedure to cause
-a Finish event upon a Close action from the peer {{NEAT-flow-mapping}}.
+For Close (provoking a Finished event) and Abort (provoking a ConnectionError event), it
+is desirable to be informed when a peer closes or aborts a
```
tfpauly, line 803:
```suggestion
the application might find it useful to be informed when a peer closes or aborts a
```

tfpauly (changes requested):

draft-ietf-taps-impl.md
```diff
@@ -799,39 +799,10 @@ The strategy to do so is implementation-specific, but should be consistent with
...
-
-For Close (provoking a Finished event) and Abort (provoking a ConnectionError event), the
-same logic applies: while it is desirable to be informed when a peer closes or aborts a
-Connection, whether this is possible depends on the underlying protocol, and no guarantees
-can be given. With SCTP, the transport system can use the stream reset procedure to cause
-a Finish event upon a Close action from the peer {{NEAT-flow-mapping}}.
+For Close (which leads to a Finished event) and Abort (which leads to a ConnectionError event),
+the application might find it useful to be informed when a peer closes or aborts a
+Connection. Whether this is possible depends on the underlying protocol, and no guarantees
+can be given. When an underlying transport connection such as SCTP supports multi-streaming, the transport system can use a stream reset procedure to cause a Finish event upon a Close action from the peer {{NEAT-flow-mapping}}.
```
mwelzl, line 805:
```suggestion
can be given. When an underlying transport connection supports multi-streaming (such as SCTP), the transport system can use a stream reset procedure to cause a Finish event upon a Close action from the peer {{NEAT-flow-mapping}}.
```

tfpauly (approved):

PR #1159, opened 2023-05-13
Discussion text on receiving: min => max

### LINKED ISSUE #1143: Receiving Messages: specify min or max bytes?
5.2.  Receiving Messages

   If the top-level protocol only
   supports a byte-stream and no framers were supported, the application
   can control the flow of received data by specifying the minimum
   number of bytes of Message content it wants to receive at one time.

Naively, I would expect this sentence to say "specifying the maximum
number of bytes", but perhaps the sentence as written is also correct.
Perhaps the application can specify both?  Please check against the
API definition and update if necessary.

***

mwelzl:
Closes #1143.

Both a minimum and a maximum can be specified. The preceding sentences clarify the minimum, so this is likely meant to discuss the maximum. I removed the condition of the top-level protocol only supplying a byte-stream because the application shouldn't normally have to be aware of what the top level protocol is.


draft-ietf-taps-impl.md
```diff
@@ -567,7 +567,7 @@ Since sending a Message may involve a context switch between the application and
...
+Similar to sending, receiving a Message is determined by the top-level protocol in the established Protocol Stack. The main difference with receiving is that the size and boundaries of the Message are not known beforehand. The application can communicate in its Receive action the parameters for the Message, which can help the Transport Services implementation know how much data to deliver and when. For example, if the application only wants to receive a complete Message, the implementation should wait until an entire Message (datagram, stream, or frame) is read before delivering any Message content to the application. This requires the implementation to understand where messages end, either via a supplied Message Framer or because the top-level protocol in the established Protocol Stack preserves message boundaries. The application can also control the flow of received data by specifying the maximum number of bytes of Message content it wants to receive at one time.
```
tfpauly, line 570:
```suggestion
Similar to sending, receiving a Message is determined by the top-level protocol in the established Protocol Stack. The main difference with receiving is that the size and boundaries of the Message are not known beforehand. The application can communicate in its Receive action the parameters for the Message, which can help the Transport Services implementation know how much data to deliver and when. For example, if the application only wants to receive a complete Message, the implementation should wait until an entire Message (datagram, stream, or frame) is read before delivering any Message content to the application. This requires the implementation to understand where messages end, either via a supplied Message Framer or because the top-level protocol in the established Protocol Stack preserves message boundaries. The application can also control the flow of received data by specifying the minimum and maximum number of bytes of Message content it wants to receive at one time.
```

tfpauly (changes requested):

PR #1163, opened 2023-05-15
Address OPSDIR management comment - fix #1153

### LINKED ISSUE #1153: Operations and Management Considerations section
From the OPSDIR review:

The document lacks any operational/manageability considerations. I am unsure
how much of the guidance from RFC 5706 would be relevant for API architecture.
I would urge authors to think if any operational/manageability considerations
need to be highlighted here or in the other documents. For instance (feel free
to disregard if they do not make sense) -

- Any guidelines to control and configure the API layer or the Implementations?
Anything on configuring System Policy?

- Any guidance on the co-existence of Socket API as well as the new Transport
API? I don't see an issue but you are the expert!

- Are there any changes in how would application locate issues with transport
sessions? In the event handling is there a need to also highlight errors for
instance?

britram:
Add an Operations and Management Considerations section pointing out hooks for ops and management (and declaring specifications for them to be out of scope, because implementation dependent).

Add a new Coexistence subsection to make it explicit that we intend for this architecture to exist alongside / atop the Socket API.

### DIFF
--- a/draft-ietf-taps-arch.md
+++ b/draft-ietf-taps-arch.md
@@ -359,7 +359,9 @@ The following diagram summarizes the top-level concepts in the architecture and 
 {: #fig-abstractions title="Concepts and Relationships in the Transport Services Architecture"}
 
 The Transport Services Implementation includes the Cached State and System Policy.
-The System Policy provides input from an operating system or other global preferences that can constrain or influence how an implementation will gather Candidate Paths and Protocol Stacks and race the candidates when establishing a Connection.
+
+The System Policy provides input from an operating system or other global preferences that can constrain or influence how an implementation will gather Candidate Paths and Protocol Stacks and race the candidates when establishing a Connection. As the details of System Policy configuration and enforcement are largely platform- and implementation- dependent, and do not affect application-level interoperability, the Transport Services API {{?I-D.ietf-taps-interface}} does not specify an interface for reading or writing System Policy.
+
 The Cached State is the state and history that the implementation keeps for each set of associated endpoints that have previously been used.
 
 ## Transport Services API Concepts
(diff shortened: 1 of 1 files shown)


draft-ietf-taps-arch.md
```diff
@@ -577,6 +581,30 @@ RFC-EDITOR: Please remove this section before publication.
...
+is the System Policy, which constrains and influences, at the host
+level, how an implementation will gather Candidate Paths and Protocol Stacks
+and race the candidates when establishing a Connection. As the details of
+System Policy configuration and enforcement are largely platform- and
+implementation- dependent, and do not affect application-level interoperability,
+the Transport Services API {{?I-D.ietf-taps-interface}} does not specify
+an interface for reading or writing System Policy.
+
+The Transport Services API specifies both synchronous and asynchronous methods
+for reporting connection errors to the application, as the application is the most
```
mwelzl, line 601:
```suggestion
for reporting errors to the application, as the application is the most
```

mwelzl: I wanted to suggest "Connection" and then understood that these could be more general (hence "connection"), but they can also be related to Messages...  all in all, the word "connection" here is unnecessary and can cause confusion, IMO.

mwelzl (changes requested):
LGTM, many thanks!  I only suggest one nit fix

draft-ietf-taps-arch.md
britram: hah, i thought this already said "connection errors" 😆. applied.

mwelzl:
We want to move this somewhere else.

tfpauly (approved):

PR #1164, opened 2023-05-15
Note that the architecture forbids downgrade

### LINKED ISSUE #1151: Downgrade avoidance
12.1.  Considerations for Candidate Gathering

   Implementations should avoid downgrade attacks that allow network
   interference to cause the implementation to select less secure, or
   entirely insecure, combinations of paths and protocols.

12.2.  Considerations for Candidate Racing

   Implementations should ensure that all options have equivalent
   security properties to avoid incentivizing attacks.

For 12.1, of course implementations should use all "downgrade
avoidance" techniques that are specified for each protocol in the
protocol's standards.  But more thought needs to be done about the
situation where the application specifies allowing a set of protocols
which, taken as a whole, has a downgrade problem.  There are only two
solutions:  (1) TAPS allows the application to specify a group of
protocols with unequal security properties; in which case, the
application shouldn't expect to get more security than the least
secure protocol in the group.  (2) TAPS forbids the application to
specify a group of protocols with unequal security properties and
enforces that condition.  Which obtains depends on the API definition,
but the implementation has no leeway in either case, and this document
ought to state the situation clearly.

***

britram:
Fixes #1151

### DIFF
--- a/draft-ietf-taps-impl.md
+++ b/draft-ietf-taps-impl.md
@@ -1146,7 +1146,7 @@ This document has no actions for IANA.
 
 ## Considerations for Candidate Gathering
 
-The Security Considerations of the TAPS Architecture {{I-D.ietf-taps-arch}} forbids gathering and racing with protocol stacks that do not have equivalent security properties. Therefore, implementations need to avoid downgrade attacks that allow network interference to cause the implementation to select less secure, or entirely insecure, combinations of paths and protocols.
+The Security Considerations of the Transport Services Architecture {{I-D.ietf-taps-arch}} forbids gathering and racing with protocol stacks that do not have equivalent security properties. Therefore, implementations need to avoid downgrade attacks that allow network interference to cause the implementation to select less secure, or entirely insecure, combinations of paths and protocols.
 
 ## Considerations for Candidate Racing
 
--- a/draft-ietf-taps-impl.md
+++ b/draft-ietf-taps-impl.md
@@ -1146,7 +1146,7 @@ This document has no actions for IANA.
 
 ## Considerations for Candidate Gathering
 
-Implementations should avoid downgrade attacks that allow network interference to cause the implementation to select less secure, or entirely insecure, combinations of paths and protocols.
+As described in the Security Considerations of {{I-D.ietf-taps-arch}}, TAPS forbids applications from specifying protocol stacks with inequivalent security properties for gathering and racing. Implementations must avoid downgrade attacks that allow network interference to cause the implementation to select less secure, or entirely insecure, combinations of paths and protocols.
 
 ## Considerations for Candidate Racing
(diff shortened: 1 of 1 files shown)


draft-ietf-taps-impl.md
```diff
@@ -1146,7 +1146,7 @@ This document has no actions for IANA.
 
 ## Considerations for Candidate Gathering
 
-Implementations should avoid downgrade attacks that allow network interference to cause the implementation to select less secure, or entirely insecure, combinations of paths and protocols.
+As described in the Security Considerations of {{I-D.ietf-taps-arch}}, TAPS forbids applications from specifying protocol stacks with inequivalent security properties for gathering and racing. Implementations must avoid downgrade attacks that allow network interference to cause the implementation to select less secure, or entirely insecure, combinations of paths and protocols.
```
gorryfair, line 1149: Can we try: "The Security Considerations of the TAPS Architecture {{I-D.ietf-taps-arch}} forbids gathering and racing with protocol stacks that do not have equivalent security properties. Therefore, implementations need to..."

gorryfair (commented):
Suggested new text?

draft-ietf-taps-impl.md
britram, line 1149:
```suggestion
The Security Considerations of the TAPS Architecture {{I-D.ietf-taps-arch}} forbids gathering and racing with protocol stacks that do not have equivalent security properties. Therefore, implementations need to avoid downgrade attacks that allow network interference to cause the implementation to select less secure, or entirely insecure, combinations of paths and protocols.
```

britram: yep that's better

draft-ietf-taps-impl.md
```diff
@@ -1146,7 +1146,7 @@ This document has no actions for IANA.
 
 ## Considerations for Candidate Gathering
 
-Implementations should avoid downgrade attacks that allow network interference to cause the implementation to select less secure, or entirely insecure, combinations of paths and protocols.
+The Security Considerations of the TAPS Architecture {{I-D.ietf-taps-arch}} forbids gathering and racing with protocol stacks that do not have equivalent security properties. Therefore, implementations need to avoid downgrade attacks that allow network interference to cause the implementation to select less secure, or entirely insecure, combinations of paths and protocols.
```
tfpauly, line 1149:
```suggestion
The Security Considerations of the Transport Services Architecture {{I-D.ietf-taps-arch}} forbids gathering and racing with protocol stacks that do not have equivalent security properties. Therefore, implementations need to avoid downgrade attacks that allow network interference to cause the implementation to select less secure, or entirely insecure, combinations of paths and protocols.
```

tfpauly (approved):

PR #1165, opened 2023-05-16
One change to normative, for consistency

### LINKED ISSUE #1146: What if an app sets a property before conn. establishment is complete?
7.  Implementing Connection Management

   If an error is encountered in setting a property (for example, if the
   application tries to set a TCP-specific property on a Connection that
   is not using TCP), the action should fail gracefully.  The
   application may be informed of the error, but the Connection itself
   should not be terminated.

An important situation is if the application sets a property before
connection establishment is complete.  In that case, the
implementation cannot tell whether the property is applicable to the
protocol that will eventually be chosen or not.  So either the
implementation needs to reject the operation -- this requires that the
API clearly separates "general" from "protocol specific" properties,
allowing the former to be modified during connection establishment but
not the latter -- or the implementation has to store the property
value for the potential future use of the protocol -- this requires
that the implementation store *all* property values, even the ones
that are not applicable to all protocols.  In either case, sec. 7
paras. 2 and 3 don't quite describe the situation, and the API
definition needs to make clear which of these situations applies.

***

mwelzl:
We already use MUST NOT for Init() and Listen(), so it's probably better to do the same for Rendezvous.

Closes #1146.

### DIFF
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -1796,8 +1796,8 @@ transport-layer connection is established; for Connectionless transports,
 it occurs when the first Message is received from the Remote Endpoint. The
 resulting Connection is contained within the RendezvousDone<> Event, and is
 ready to use as soon as it is passed to the application via the Event.
-Changes made to a Preconnection after Rendezvous() has been called do
-not have any effect on existing Connections.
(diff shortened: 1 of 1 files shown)


tfpauly (approved):

csperkins:
Agree this change is good.

PR #1166, opened 2023-05-16
MUXed listeners

### LINKED ISSUE #1141: Implementing Listeners for MUXed protocols
4.7.3.  Implementing listeners for Multiplexed Protocols

   If the
   abstraction of Connection presented to the application is mapped to
   the multiplexed stream, then the Listener should deliver new
   Connection objects in the same way for either case.

What controls the condition "If the ... Connection is mapped to
[i.e. represents] a multiplexed stream ..."?  It seems to me that
there is either a blanket rule (e.g. SCTP streams are always presented
as individual Connections) or there is a setting in the Listener.  In
either case, this is described somewhere in the TAPS API definition,
and a reference should be given here.

   The
   implementation should allow the application to introspect the
   Connection Group marked on the Connections to determine the grouping
   of the multiplexing.

The meaning of "should allow" is not clear.  I would expect that the
API is specified to set Connection Groups for all sets of Connections
that share a multiplexed lower-level protocol, and the API has a
defined mechanism for accessing the Connection Group of a Connection.

For some multiplexed protocols (e.g. QUIC), once the connection is
established, either end can initiate new streams.  This means that the
TAPS client may need a way to provide Listener service.  Does the TAPS
API define how this happens?  And if so, have you checked that the
various passages of this document that refer to Listeners cover
client-end Listeners correctly?

***

mwelzl:
Closes #1141.

This is a bit different than we discussed - see my longer comment in #1141.

### DIFF
--- a/draft-ietf-taps-impl.md
+++ b/draft-ietf-taps-impl.md
@@ -509,7 +509,7 @@ Connectionless protocols such as UDP and UDP-lite generally do not provide the s
 
 ### Implementing listeners for Multiplexed Protocols
 
-Protocols that provide multiplexing of streams into a single four-tuple can listen both for entirely new connections (a new HTTP/2 stream on a new TCP connection, for example) and for new sub-connections (a new HTTP/2 stream on an existing connection). If the abstraction of Connection presented to the application is mapped to the multiplexed stream, then for either case the Listener ought to deliver the new Connection objects in the same way. The implementation should allow the application to introspect the Connection Group marked on the Connections to determine the grouping of the multiplexing.
+Protocols that provide multiplexing of streams can listen for entirely new connections and for new sub-connections (streams of an already existing connection). A new stream arrival of an already existing connection must be presented to the application as a new Connection. This new Connection must be grouped with all other Connections that are multiplexed via the same protocol. 
 
 # Implementing Sending and Receiving Data
 
--- a/draft-ietf-taps-impl.md
+++ b/draft-ietf-taps-impl.md
@@ -509,7 +509,7 @@ Connectionless protocols such as UDP and UDP-lite generally do not provide the s
 
 ### Implementing listeners for Multiplexed Protocols
 
-Protocols that provide multiplexing of streams can listen for entirely new connections and for new sub-connections (streams of an already existing connection). A new stream arrival of an already existing connection must be presented to the application as a new Connection. This new Connection must be grouped with all other Connections that are multiplexed via the same protocol.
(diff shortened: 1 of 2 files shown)


draft-ietf-taps-impl.md
```diff
@@ -509,7 +509,7 @@ Connectionless protocols such as UDP and UDP-lite generally do not provide the s
 
 ### Implementing listeners for Multiplexed Protocols
 
-Protocols that provide multiplexing of streams into a single four-tuple can listen both for entirely new connections (a new HTTP/2 stream on a new TCP connection, for example) and for new sub-connections (a new HTTP/2 stream on an existing connection). If the abstraction of Connection presented to the application is mapped to the multiplexed stream, then for either case the Listener ought to deliver the new Connection objects in the same way. The implementation should allow the application to introspect the Connection Group marked on the Connections to determine the grouping of the multiplexing.
+Protocols that provide multiplexing of streams can listen for entirely new connections and for new sub-connections (streams of an already existing connection). A new stream arrival of an already existing connection must be presented to the application as a new Connection. This new Connection must be grouped with all other Connections that are multiplexed via the same protocol.
```
mwelzl, line 512:
```suggestion
Protocols that provide multiplexing of streams can listen for entirely new connections and for new sub-connections (streams of an already existing connection). A new stream arrival on an already existing connection must be presented to the application as a new Connection. This new Connection must be grouped with all other Connections that are multiplexed via the same protocol.
```

draft-ietf-taps-impl.md
```diff
@@ -509,7 +509,7 @@ Connectionless protocols such as UDP and UDP-lite generally do not provide the s
 
 ### Implementing listeners for Multiplexed Protocols
 
-Protocols that provide multiplexing of streams into a single four-tuple can listen both for entirely new connections (a new HTTP/2 stream on a new TCP connection, for example) and for new sub-connections (a new HTTP/2 stream on an existing connection). If the abstraction of Connection presented to the application is mapped to the multiplexed stream, then for either case the Listener ought to deliver the new Connection objects in the same way. The implementation should allow the application to introspect the Connection Group marked on the Connections to determine the grouping of the multiplexing.
+Protocols that provide multiplexing of streams can listen for entirely new connections and for new sub-connections (streams of an already existing connection). A new stream arrival on an already existing connection must be presented to the application as a new Connection. This new Connection must be grouped with all other Connections that are multiplexed via the same protocol.
```
tfpauly, line 512:
```suggestion
Protocols that provide multiplexing of streams can listen for entirely new connections as well as for new sub-connections (streams of an already existing connection). A new stream arrival on an existing connection is presented to the application as a new Connection. This new Connection is grouped with all other Connections that are multiplexed via the same protocol.
```

tfpauly (approved):

PR #1167, opened 2023-05-16
handling muxed conns. Closes #1139.

### LINKED ISSUE #1139: Handling MUXed connections
4.5.  Establishing multiplexed connections

   Multiplexing several Connections over a single underlying transport
   connection requires that the Connections to be multiplexed belong to
   the same Connection Group (as is indicated by the application using
   the Clone call).  When the underlying transport connection supports
   multi-streaming, the Transport Services System can map each
   Connection in the Connection Group to a different stream.  Thus, when
   the Connections that are offered to an application by the Transport
   Services API are multiplexed, the Transport Services implementation
   can establish a new Connection by simply beginning to use a new
   stream of an already established transport Connection and there is no
   need for a connection establishment procedure.  This, then, also
   means that there may not be any "establishment" message (like a TCP
   SYN), but the application can simply start sending or receiving.
   Therefore, when the Initiate action of a Transport Services API is
   called without Messages being handed over, it cannot be guaranteed
   that the Remote Endpoint will have any way to know about this, and
   hence a passive endpoint's ConnectionReceived event might not be
   delivered until data is received.  Instead, delivering the
   ConnectionReceived event could be delayed until the first Message
   arrives.

This should be clarified, I think.  The first part is fine:

   Multiplexing several Connections over a single underlying transport
   connection requires that the Connections to be multiplexed belong to
   the same Connection Group (as is indicated by the application using
   the Clone call).  When the underlying transport connection supports
   multi-streaming, the Transport Services System can map each
   Connection in the Connection Group to a different stream.

The next part is not quite correct:

   Thus, when
   the Connections that are offered to an application by the Transport
   Services API are multiplexed, the Transport Services implementation
   can establish a new Connection by simply beginning to use a new
   stream of an already established transport Connection and there is no
   need for a connection establishment procedure.  This, then, also
   means that there may not be any "establishment" message (like a TCP
   SYN), but the application can simply start sending or receiving.

What is really going on is that the protocol has two layers, a lower
one that creates the connection over which streams are multiplexed and
a higher one that is the stream which is multiplexed.  What is being
described is when the upper layer has no explicit handshake to
establish a new stream.  But that is just another case of the
situation with UDP discussed in sec. 4.4.1 para. 1.  I think a better
phrasing would be as follows (and probably reads better as a separate
paragraph):

   Thus, when the Connections that are offered to an application by
   the Transport Services API are multiplexed, the Transport Services
   implementation can establish a new Connection by using a new stream
   of an already established transport Connection.  Effectively, the
   streams are an additional protocol layer on top of the transport
   connection, and for many such there is no explicit connection
   establishment procedure for the new stream prior to sending data on
   it.  In this case, the same considerations apply to determining
   stream establishment as apply to establishing a UDP connection, as
   discussed in section 4.4.1.

The final part seems to be correct:

   Therefore, when the Initiate action of a Transport Services API is
   called without Messages being handed over, it cannot be guaranteed
   that the Remote Endpoint will have any way to know about this, and
   hence a passive endpoint's ConnectionReceived event might not be
   delivered until data is received.  Instead, delivering the
   ConnectionReceived event could be delayed until the first Message
   arrives.

but (1) it could be clarified, and (2) it applies generally to
non-handshake protocols, and so probably should be relocated to
e.g. sec. 4.4.1 between paragraphs 1 and 2, where we first discuss the
nuances of handshake-less protocols.  Perhaps these adjustments to the
wording:

   When the Initiate action of a Transport Services API is
   called without Messages being handed over, depending on the
   protocols involved, it is not guaranteed that the Remote
   Endpoint will be notified of this, and hence a passive
   endpoint's application may not receive a ConnectionReceived event
   until it receives the first Message the connection.

***

mwelzl:
Pretty much following all suggestions in #1139. I only didn't add the sentence about "Effectively, the streams are an additional protocol layer on top of the transport..." because I don't think this contributes much.

### DIFF
--- a/draft-ietf-taps-impl.md
+++ b/draft-ietf-taps-impl.md
@@ -485,7 +485,6 @@ establishment procedure for the new stream prior to sending data on it (e.g., wi
 considerations apply to determining stream establishment as apply to establishing a UDP connection, as
 discussed in {{determining-successful-establishment}}.
 
-a new stream of an already established transport Connection and there is no need for a connection establishment
 procedure. This, then, also means that there may not
 be any "establishment" message (like a TCP SYN), but the application can simply start sending
 or receiving.
--- a/draft-ietf-taps-impl.md
+++ b/draft-ietf-taps-impl.md
@@ -485,7 +485,6 @@ establishment procedure for the new stream prior to sending data on it (e.g., wi
 considerations apply to determining stream establishment as apply to establishing a UDP connection, as
 discussed in {{determining-successful-establishment}}.
 
-Thus, when the Connections that are offered to an application by the Transport Services API are multiplexed,
 the Transport Services implementation can establish a new Connection by simply beginning to use
 a new stream of an already established transport Connection and there is no need for a connection establishment
 procedure. This, then, also means that there may not
--- a/draft-ietf-taps-impl.md
+++ b/draft-ietf-taps-impl.md
@@ -463,7 +463,7 @@ New connection attempts that involve transmitting data on the network ought not 
 An implementation could choose to let certain handshakes and negotiations complete to gather metrics that influence future connections.
 Keeping additional connections is generally not recommended, because those attempts were slower to connect and may exhibit less desirable properties.
 
-### Determining Successful Establishment {#establish-determine}
+### Determining Successful Establishment {#determining-successful-establishment}
 
 On a per-protocol basis, implementations may select different criteria by which a leaf node is considered to be successfully connected. If the only protocol being used is a transport protocol with a clear handshake, like TCP, then the obvious choice is to declare that node "connected" when the last packet of the three-way handshake has been received. If the only protocol being used is an connectionless protocol, like UDP, the implementation may consider the node fully "connected" the moment it determines a route is present, before sending any packets on the network, see further {{connectionless-racing}}.
 
@@ -485,7 +485,7 @@ the Transport Services implementation can establish a new Connection by using
 a new stream of an already established transport connection. For such streams, there is often no explicit connection
 establishment procedure for the new stream prior to sending data on it (e.g., with SCTP). In this case, the same
 considerations apply to determining stream establishment as apply to establishing a UDP connection, as
-discussed in {{establish-determine}}.
+discussed in {{determining-successful-establishment}}.
 
 Thus, when the Connections that are offered to an application by the Transport Services API are multiplexed,
 the Transport Services implementation can establish a new Connection by simply beginning to use
--- a/draft-ietf-taps-impl.md
+++ b/draft-ietf-taps-impl.md
@@ -485,7 +485,7 @@ establishment procedure for the new stream prior to sending data on it (e.g., wi
 considerations apply to determining stream establishment as apply to establishing a UDP connection, as
 discussed in {{determining-successful-establishment}}.
 
-procedure. This, then, also means that there may not
+This, then, also means that there may not
 be any "establishment" message (like a TCP SYN), but the application can simply start sending
 or receiving.
 
--- a/draft-ietf-taps-impl.md
+++ b/draft-ietf-taps-impl.md
@@ -482,7 +482,7 @@ Multiplexing several Connections over a single underlying transport connection r
 
 Thus, when the Connections that are offered to an application by the Transport Services API are multiplexed,
 the Transport Services implementation can establish a new Connection by using
-a new stream of an already established transport connection. For such streams, there is often no explicit connection
+For such streams, there is often no explicit connection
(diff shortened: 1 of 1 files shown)


draft-ietf-taps-impl.md
```diff
@@ -463,26 +463,37 @@ New connection attempts that involve transmitting data on the network ought not
...
 If all of the leaf nodes fail to connect during racing, i.e. none of the configurations that satisfy all requirements given in the Transport Properties actually work over the available paths, then the transport system should report an EstablishmentError to the application. An EstablishmentError event should also be generated in case the transport system finds no usable candidates to race.
 
 ## Establishing multiplexed connections {#establish-mux}
 
 Multiplexing several Connections over a single underlying transport connection requires that the Connections to be multiplexed belong to the same Connection Group (as is indicated by the application using the Clone call). When the underlying transport connection supports multi-streaming, the Transport Services System can map each Connection in the Connection Group to a different stream.
+
+
+Thus, when the Connections that are offered to an application by the Transport Services API are multiplexed,
+the Transport Services implementation can establish a new Connection by using
+a new stream of an already established transport connection. For such streams, there is often no explicit connection
```
mwelzl, line 485:
```suggestion
For such streams, there is often no explicit connection
```

draft-ietf-taps-impl.md
```diff
@@ -463,26 +463,34 @@ New connection attempts that involve transmitting data on the network ought not
...
 Multiplexing several Connections over a single underlying transport connection requires that the Connections to be multiplexed belong to the same Connection Group (as is indicated by the application using the Clone call). When the underlying transport connection supports multi-streaming, the Transport Services System can map each Connection in the Connection Group to a different stream.
-Thus, when the Connections that are offered to an application by the Transport Services API are multiplexed,
+
+
+For such streams, there is often no explicit connection
+establishment procedure for the new stream prior to sending data on it (e.g., with SCTP). In this case, the same
+considerations apply to determining stream establishment as apply to establishing a UDP connection, as
+discussed in {{determining-successful-establishment}}.
+
 the Transport Services implementation can establish a new Connection by simply beginning to use
```
mwelzl, line 488:
```suggestion
```

draft-ietf-taps-impl.md
```diff
@@ -463,26 +463,32 @@ New connection attempts that involve transmitting data on the network ought not
...
-the Transport Services implementation can establish a new Connection by simply beginning to use
-a new stream of an already established transport Connection and there is no need for a connection establishment
+
+
+For such streams, there is often no explicit connection
+establishment procedure for the new stream prior to sending data on it (e.g., with SCTP). In this case, the same
+considerations apply to determining stream establishment as apply to establishing a UDP connection, as
+discussed in {{determining-successful-establishment}}.
+
 procedure. This, then, also means that there may not
```
mwelzl, line 488:
```suggestion
This, then, also means that there may not
```

draft-ietf-taps-impl.md
```diff
@@ -463,26 +463,32 @@ New connection attempts that involve transmitting data on the network ought not
...
-the Transport Services implementation can establish a new Connection by simply beginning to use
-a new stream of an already established transport Connection and there is no need for a connection establishment
-procedure. This, then, also means that there may not
+
+
+For such streams, there is often no explicit connection
+establishment procedure for the new stream prior to sending data on it (e.g., with SCTP). In this case, the same
+considerations apply to determining stream establishment as apply to establishing a UDP connection, as
+discussed in {{determining-successful-establishment}}.
+
```
mwelzl, line 487:
```suggestion
```

draft-ietf-taps-impl.md
```diff
@@ -463,26 +463,30 @@ New connection attempts that involve transmitting data on the network ought not
 An implementation could choose to let certain handshakes and negotiations complete to gather metrics that influence future connections.
 Keeping additional connections is generally not recommended, because those attempts were slower to connect and may exhibit less desirable properties.
 
-### Determining Successful Establishment
+### Determining Successful Establishment {#determining-successful-establishment}
 
 On a per-protocol basis, implementations may select different criteria by which a leaf node is considered to be successfully connected. If the only protocol being used is a transport protocol with a clear handshake, like TCP, then the obvious choice is to declare that node "connected" when the last packet of the three-way handshake has been received. If the only protocol being used is an connectionless protocol, like UDP, the implementation may consider the node fully "connected" the moment it determines a route is present, before sending any packets on the network, see further {{connectionless-racing}}.
 
+When the Initiate action of a Transport Services API is called without Messages being handed over, depending on the
```
tfpauly, line 470:
```suggestion
When the Initiate action is called without any Messages being sent at the same time, depending on the
```

tfpauly (approved):

PR #1168, opened 2023-05-17
This makes a few sentences clearer to address #1134.

### LINKED ISSUE #1134: Lack of clarity wrt. Properties in Connection Groups
2.  Implementing Connection Objects

   The properties held by a
   Connection or Listener are independent of other connections that are
   not part of the same Connection Group.

First, "connections" should be "Connections".

Indeed, it appears that the properties are independent of connections
in the same Connection Group, too, as properties are inherited from
the Preconnection and then fixed thereafter.

But I think my issue is what is meant by "properties".  I suspect that
the API model has two sorts of properties, one sort are the ones
inherited from the Preconnection and one that are set by the process
of connecting (e.g., the local ephemeral port), and another sort that
are not from the Preconnection but are shared among Connections in a
Connection Group.  If so, this distinction needs to be made explicit.
But it's still not clear to me how these connection properties can be
changed once as Connection is established, even by another Connection
in a Connection Group.

   Once Initiate has been called, the Selection Properties and Endpoint
   information are immutable (i.e, an application is not able to later
   modify Selection Properties on the original Preconnection object).
   Listener objects are created with a Preconnection, at which point
   their configuration should be considered immutable by the
   implementation.

I had some difficulty reading this paragraph.  After thinking about
it, it seems that all of it is about listening, but that is not made
clear at the beginning, and I assumed it was continuing the discussion
of connection establishment in the previous paragraph.

Within that context, sec. 2 states that when a Listener is created,
all applicable properties are copied from the Precondition, and no
later changes to the Precondition change the Listener.  But the first
sentence says that the Precondition object cannot be changed, which
makes no sense.

***

mwelzl:
Closes #1134.

### DIFF
--- a/draft-ietf-taps-impl.md
+++ b/draft-ietf-taps-impl.md
@@ -130,7 +130,7 @@ The Connection Properties held by a Connection or Listener are independent of ot
 
 Connection establishment is only a local operation for a connectionless protocols, which serves to simplify the local send/receive functions and to filter the traffic for the specified addresses and ports {{?RFC8085}} (for example using UDP or UDP-Lite transport without a connection handshake procedure).
 
-Once Initiate has been called, the Selection Properties and Endpoint information of the created Connection are immutable (i.e, an application is not able to later modify these properties of a Connection by manipulating the original Preconnection object).
+Once Initiate has been called, the Selection Properties and Endpoint information of the created Connection are immutable (i.e, an application is not able to later modify the properties of a Connection by manipulating the original Preconnection object).
 Listener objects are created with a Preconnection, at which point their configuration should be considered immutable by the implementation. The process of listening is described in {{listen}}.
 
 # Implementing Pre-Establishment


draft-ietf-taps-impl.md
```diff
@@ -126,11 +126,11 @@ Preconnection objects should be implemented as bundles of properties that an app
...
-The properties held by a Connection or Listener are independent of other connections that are not part of the same Connection Group.
+The Connection Properties held by a Connection or Listener are independent of other Connections that are not part of the same Connection Group.
 
 Connection establishment is only a local operation for a connectionless protocols, which serves to simplify the local send/receive functions and to filter the traffic for the specified addresses and ports {{?RFC8085}} (for example using UDP or UDP-Lite transport without a connection handshake procedure).
 
-Once Initiate has been called, the Selection Properties and Endpoint information are immutable (i.e, an application is not able to later modify Selection Properties on the original Preconnection object).
+Once Initiate has been called, the Selection Properties and Endpoint information of the created Connection are immutable (i.e, an application is not able to later modify these properties of a Connection by manipulating the original Preconnection object).
```
tfpauly, line 133:
```suggestion
Once Initiate has been called, the Selection Properties and Endpoint information of the created Connection are immutable (i.e, an application is not able to later modify the properties of a Connection by manipulating the original Preconnection object).
```

tfpauly (approved):

PR #1172, opened 2023-05-19
Improve introduction of message framers

### LINKED ISSUE #1145: Message framer clarifications
6.  Implementing Message Framers

Of course, a Message Framer is just another protocol layered on top of
a protocol stack, and I think it would be helpful if that was stated
early in this section.

But there are a lot of details of the situation that I think you want
to be clearer about.  Does the TAPS specification define and require
support of a particular set of Message Framers?  Applications will
generally take the defensive stance of not using Message Framers that
aren't required to be present.

Also, it's not clear to me to what degree this discussion is connected
to the API.  In particular, does the TAPS API define how a Message
Framer interacts with the rest of the implementation?  In that case,
the details in this section are normative.  But if the idea is that
TAPS does not define such an API but that a TAPS implementation would
want to provide an API for "user" Message Framers, then this section
is a generic discussion of implementation considerations.  Either way,
the intention should be stated clearly at the beginning.

E.g.

   [...] these are ways for
   applications or application frameworks to define their own Message
   parsing to be included within a Connection's Protocol Stack.

suggests that allowing applications to specify their own custom
Message Framers is a part of the TAPS API definition, whereas

   This section describes one possible API for defining
   Message Framers, as an example.

suggests that there is no definition of such a facility (and so it is
an optional, non-standardized, extension in any implementation that
does have it).

***

tfpauly:
Closes #1145

### DIFF
--- a/draft-ietf-taps-impl.md
+++ b/draft-ietf-taps-impl.md
@@ -596,7 +596,7 @@ It is also possible for Protocol Stacks within a particular leaf node to use 0-R
 Message Framers are functions that define
 simple transformations between application Message data and raw transport
 protocol data. Generally, a Message Framer implements a simple
-application protocol that can either be provided by the Transport Service
+application protocol that can either be provided by the Transport Services
 implementation or be customized by the application.
 A Framer can encapsulate or encode outbound Messages,
 decapsulate or decode inbound data into Messages, and implement parts of
--- a/draft-ietf-taps-impl.md
+++ b/draft-ietf-taps-impl.md
@@ -595,7 +595,10 @@ It is also possible for Protocol Stacks within a particular leaf node to use 0-R
 
 Message Framers are functions that define
 simple transformations between application Message data and raw transport
-protocol data. A Framer can encapsulate or encode outbound Messages,
+protocol data. Generally, a Message Framer implements a simple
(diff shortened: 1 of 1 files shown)


draft-ietf-taps-impl.md
```diff
@@ -595,7 +595,10 @@ It is also possible for Protocol Stacks within a particular leaf node to use 0-R
 
 Message Framers are functions that define
 simple transformations between application Message data and raw transport
-protocol data. A Framer can encapsulate or encode outbound Messages,
+protocol data. Generally, a Message Framer implements a simple
+application protocol that can either be provided by the Transport Service
+implementation or be customized by the application.
```
mwelzl, line 600:
```suggestion
implementation or by the application. It is optional for Transport Services system implementations to provide Message Framers: the specification {{I-D.ietf-taps-interface}} does not prescribe any particular Message Framers to be implemented.
```
1) this suggests "by the application" because "customized" sounds like the application could only adjust an offered Framer, not entirely implement one (as it would).
2) the rest is a proposal to make it clear that no framers must be shipped with a TAPS system... to better address #1145.

mwelzl (commented):
Two small suggestions

PR #1173, opened 2023-05-19
Update 0-RTT text

### LINKED ISSUE #1144: 0-RTT stuff
5.3.  Handling of data for fast-open protocols

The text needs to clarify how 0-RTT data is handled by the API.
Naively, I would expect that 0-RTT data must be supplied with the
connection request, since it is sent with the handshake, after all.
However, sec. 5.3 para. 3 sent. 2 suggests that 0-RTT data that is
provided to the API needs to be marked with specific properties that
all 0-RTT would have to be marked with, implying that the API is
provided with the data in a separate call from the connection request
call.  How 0-RTT data is provided in API call(s) should be made clear
at the beginning of para. 3, as that is the background for everything
following.

   An implementation can set this property
   according to the protocols that it will race based on the given
   Selection Properties when the application requests to establish a
   connection.

How is the zeroRttMsgMaxLen value derived from the zeroRttMsgMaxLen
properties of the individual protocols that will potentially be used by
connection establishment?  I'd guess that it is the minimum of them,
but in that case, if both a fast-open and a non-fast-open protocol
might be used, then zeroRttMsgMaxLen will be 0 and the application
will never get the benefit of fast-open.  OTOH, if zeroRttMsgMaxLen is
> 0, what happens if it is the non-fast-open protocol that succeeds?

In any case, if zeroRttMsgMaxLen is returned by the connection
establishment API request, how does the application provide the 0-RTT
data before the first connection request is sent?

In another aspect, the set of protocols that will be raced will only
be determined *during* the connection establishment process.  E.g. the
usable protocols may depend on the host addresses.  There is no way in
general for the API call to know which protocols might be used unless
either the API call finishes after the needed information is known, or
if the API call assumes that all protocols the implementation supports
might be used.

   It is also possible for Protocol Stacks within a particular leaf node
   to use 0-RTT handshakes without any safely replayable application
   data if a protocol in the stack has idempotent handshake data to
   send.  For example, TCP Fast Open could use a Client Hello from TLS
   as its 0-RTT data, shortening the cumulative handshake time.

This should be clarified.  The data in question isn't 0-RTT data from
the *application's* perspective, and so that name shouldn't be applied
here without qualification.  Perhaps

   It is also possible for a protocol implementation in a Protocol
   Stack for a particular leaf node to use 0-RTT data in a lower-level
   protocol to carry its own handshakes; in such cases, whether the
   data involved is safely replayable is determined by the protocol
   implementation that generates the handshake data, not the
   application (which does not generate the data).  For example, TCP
   Fast Open could use a Client Hello from TLS as its 0-RTT data,
   shortening the cumulative handshake time.

***

tfpauly:
Closes #1144

Actually removes zeroRttMsgMaxLen. I'm getting more convinced it's not useful, and it's certainly not part of what we've implemented thus far.


mwelzl:
But... how does an application know how large such a message can be?  I assume that you just give a static recommendation for the max length of such a message - which isn't much different from a simple implementation of this function that just returns that number.

tfpauly:
It doesn't really matter how long it can be — the implementation will only send what it needs to send as 0-RTT data anyhow, and send the rest later.

mwelzl:
I understand, and I'm convinced. I guess the only point of getting this max len would be for an application to make a decision like: "if I can only put 800 bytes in there, not 1000, I'll skip these extra 200 and send them on a later message, because I don't want to sacrifice 0-RTT". This is really rather complex decision making / optimization for an application to make... quite unlikely that someone would want to do that, indeed.

mwelzl (approved):
LGTM but I'd like to see more positive ACKs than just mine.

csperkins:
The implementation has to assume a conservatively low path MTU anyway, so I'm not sure specifying a protocol specific values gains much.

gorryfair:
So I am unsure that we need this at all?

tfpauly:
@gorryfair right, hence why I am removing it.

PR #1175, opened 2023-06-03
Resolves #1140 Handling connectionless protocols

### LINKED ISSUE #1140: Handling connectionless protocols
4.6.  Handling connectionless protocols

The nuances of connectionless protocols also are discussed in
sec. 4.4.1 para. 1 and the part of the paragraph discussed above.  It
may be an improvement to gather all that information into this
section.

Also, these considerations apply to any handshake-less protocol.
E.g. there probably is no guarantee that a server will accept another
stream on a multiplexed connection.  So it may be worth introducing
the class "handshake-less protocol" explicitly.

   To mitigate this, an
   application can use a Message Framer (Section 6) on top of a
   connectionless protocol to only mark a specific connection attempt as
   ready when some data has been received, or after some application-
   level handshake has been performed.

Of course, a Message Framer is just another protocol layered on top.
In this instance, the point is that it presents to the application a
protocol with handshake but it uses a protocol without a handshake.
You might want to state that explicitly.

***

abrunstrom:
introducing the class "handshake-less protocol" does not seem necessary to me, so I have only tried to make the text a bit more explicit to deal with the feedback.


tfpauly (approved):
Looks great!

PR #1178, opened 2023-06-04
Fix DNS nits

### LINKED ISSUE #1177: DNS nits
From DNSdir review (Reviewer: Peter van Dijk):

In 4.1.1.1, "... send DNS queries for both A (IPv4) and AAAA (IPv6) records
...", this assumes no other address types will exist in the future. RFC8499
uses "address records" to refer to A/AAAA and any future variants implicitly,
pointing out that RFC2181 informally says "(A, AAAA, etc)". I'm not aware of a
more useful formal reference than 8499, so perhaps something like ".. send DNS
queries for address records, which currently means sending A (IPv4) and AAAA
(IPv6) queries ..."?

While ietf-taps-interface mentions SRV records briefly, this document does not.
An SRV record set could cause a single RemoteSpecifier to map to multiple
names. HTTPS and SVCB records can also cause indirection to multiple names (and
even transport indications). I do not think the document should go into detail
on all of these, but, as with address records, perhaps it can mention that DNS
lookups can involve an extra layer of indirection, and provide SRV, HTTPS and
SVCB as current examples.

In 4.3.2, "should follow the Happy Eyeballs algorithm described in [RFC8305]."
- should this be a "SHOULD"? (I noticed a few other lowercase "should"s as
well.)

In 9.1, ".. resolution answers (A and AAAA queries, for example)" - the queries
are not cached, (part of) their responses is cached. Perhaps ".. resolution
answers (from A and AAAA queries, for example"

Should 12.2 mention DNSSEC? I can very much understand it if you say no :-)

tfpauly:
Closes #1177

### DIFF
--- a/draft-ietf-taps-impl.md
+++ b/draft-ietf-taps-impl.md
@@ -247,7 +247,7 @@ There are three types of branching from a parent node into one or more child nod
 
 If a connection originally targets a single endpoint, there may be multiple endpoints of different types that can be derived from the original. This creates an ordered list of the derived endpoints according to application preference, system policy and expected performance.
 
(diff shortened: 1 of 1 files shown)


mwelzl (approved):
LGTM!  Thanks!

abrunstrom (approved):
Looks great, thanks!

PR #1198, opened 2023-06-05
Clarify multiple remote endpoints

### LINKED ISSUE #1188: s/SHOULD/MUST regarding multiple remote endpoints

Section 6:

5th para: why SHOULD and not MUST? Specifically, when is it OK for
multiple remote endpoints to represent different services?

tfpauly:
Closes #1188 

Attempt at closing the issue, based on my interpretation. Happy to have a different direction too.

### DIFF
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -651,7 +651,7 @@ can be resolved to a server reflexive address for a Preconnection used to
 make a peer-to-peer `Rendezvous`.
 
 If more than one Remote Endpoint is specified on the Preconnection, then
-all the Remote Endpoints on the Preconnection SHOULD represent the same
+all the Remote Endpoints on the Preconnection should represent the same
 service, to the extent that the application and the Transport Services
 system can validate that the Remote Endpoints are indeed the same service.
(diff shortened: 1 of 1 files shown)


draft-ietf-taps-interface.md
```diff
@@ -652,7 +652,9 @@ make a peer-to-peer `Rendezvous`.
 
 If more than one Remote Endpoint is specified on the Preconnection, then
 all the Remote Endpoints on the Preconnection SHOULD represent the same
```
mwelzl, line 654:
```suggestion
all the Remote Endpoints on the Preconnection should represent the same
```
AS per this suggestion, I now think that the solution is to de-capitalise the SHOULD ... it's not something that the local Transport Services system has any control over?!   The "to the extent ..." thing is okay, I think it doesn't really change anything. It's the SHOULD that was indeed confusing. On the server side, I'd say MUST. On the client side, one doesn't have control.

tfpauly:
@mwelzl please check, applied your change

mwelzl (approved):

britram (approved):

PR #1199, opened 2023-06-06
Clarify behaviour of multicast

### LINKED ISSUE #1190: Multicast confusion in 1st paragraph of section 6.1.1

# 6.1.1

The three options:
* Initiate() - send only
* Listen() - receive only
* Rendezvous() - send and receive
are well described.

One source of confusion for me though is the very first para.  Does it
apply to all three scenarios?  The sentence starts with the very generic
"To use multicast [...]" so it'd seem so, but then reading the second
para it looks like that preamble is specific to Initiate().

Could you please clarify?

csperkins:
Closes #1190


mwelzl (approved):
Looks great, thanks!

PR #1201, opened 2023-06-06
Downgrade (most) normative references

### LINKED ISSUE #1197: Fewer normative references?

I was a bit surprised that the TCP, SCTP and QUIC RFCs were not
referenced.  But more generally, I am unsure about the criteria used to
tell normative references from informative.  To me, BCP14 and
draft-ietf-taps-arch are the only two normative things, the rest could
be made informative.

britram:
Fix #1197


mwelzl (approved):
LGTM, many thanks!

tfpauly (approved):

PR #1202, opened 2023-06-06
Make implementation freedom clearer

### LINKED ISSUE #1181: Be clearer about implementation freedom

The API objects can be quite complex, and there is no explicit
validation on construction.  Question: is the assumption that
implementations will do the same?  Or are they are free to, for example,
use, say, a builder pattern to provide a safer experience for their
users?

britram:
fix #1181

### DIFF
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -586,9 +586,10 @@ There is no interoperability benefit in tightly defining how the API is
 presented to application programmers across diverse platforms. However,
 maintaining the "shape" of the abstract API across different platforms reduces
 the effort for programmers who learn to use the Transport Services API to then
-apply their knowledge to another platform.
-
-We therefore make the following recommendations:
+apply their knowledge to another platform. That said, implementations have
+significant freedom in presenting this API to programmers, balancing the
+conventions of the protocol with the shape of the API. We make the following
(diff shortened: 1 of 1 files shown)


tfpauly (approved):

draft-ietf-taps-interface.md
```diff
@@ -586,9 +586,10 @@ There is no interoperability benefit in tightly defining how the API is
 presented to application programmers across diverse platforms. However,
 maintaining the "shape" of the abstract API across different platforms reduces
 the effort for programmers who learn to use the Transport Services API to then
-apply their knowledge to another platform.
-
-We therefore make the following recommendations:
+apply their knowledge to another platform. That said, implementations have
+significant freedom in presenting this API to programmers, balancing the
+conventions of the protocol with the shape of the API. We make the following
+recommendations to strike the right balance:
```
mwelzl, line 592:
```suggestion
recommendations:
```
This would feel slightly better to me without the "strike the right balance" because, first, who are we to state that we know what the right balance is... and, second, it repeats "balancing" in the previous sentence. I think it's a great update but better by just saying "We make the following recommendations:"

mwelzl (changes requested):
The tiniest of suggestions... feel free to merge with or without accepting it.

PR #1203, opened 2023-06-07
Security parameters fixes

### LINKED ISSUE #1184: HTTPS + raw PK

HTTPS with raw public keys is not a great match ;-)
What about using a PK cert rather than raw PK?

britram:
Add certificates and clean up callbacks from GenArt review.

fix #1184


mwelzl (approved):
LGTM, thanks!

tfpauly (approved):

PR #1205, opened 2023-06-07
GENART Section 4 editorial rollup

### LINKED ISSUE #1187: Editorial things - in section 4 (and its subsections)

The five paragraphs are slightly patchwork-y.  For example, the second
para fluctuates from Selection to Connection and then again to Selection
props.  

I suggest grouping more tightly around the different properties:

NEW

4.  Transport Properties

   Each application using the Transport Services API declares its
   preferences for how the Transport Services system should operate.
   This is done by using Transport Properties, as defined in
   [I-D.ietf-taps-arch], at each stage of the lifetime of a connection.

   Transport Properties are divided into Selection, Connection, and
   Message Properties.

   Selection Properties (see Section 6.2) can only be set during pre-
   establishment.  They are only used to specify which paths and
   protocol stacks can be used and are preferred by the application.
   Calling Initiate on a Preconnection creates an outbound Connection or
   a Listener, and the Selection Properties remain readable from the
   Connection or Listener, but become immutable.  Selection Properties
   can be set on Preconnections, and the effect of Selection Properties
   can be queried on Connections and Messages.

   Connection Properties (see Section 8.1) are used to inform decisions
   made during establishment and to fine-tune the established
   connection.  They can be set during pre-establishment and they may be
   changed later.  Connection Properties can be set on Connections and
   Preconnections; when set on Preconnections, they act as an initial
   default for the resulting Connections.

   Message Properties (see Section 9.1.3) control the behavior of the
   selected protocol stack(s) when sending Messages.  Message Properties
   can be set on Messages, Connections, and Preconnections; when set on
   the latter two, they act as an initial default for the Messages sent
   over those Connections.

   Note that configuring Connection Properties and Message Properties on
   Preconnections is preferred over setting them later.  Early
   specification of Connection Properties allows their use as additional
   input to the selection process.  Protocol-specific Properties, which
   enable configuration of specialized features of a specific protocol,
   see Section 3.2 of [I-D.ietf-taps-arch], are not used as an input to
   the selection process, but only support configuration if the
   respective protocol has been selected.

If you find the proposal above OK, I can PR this:
https://github.com/thomas-fossati/taps-api-drafts/tree/s4-reflow

I am not sure the following is entirely correct:

  "Calling Initiate on a Preconnection creates an outbound Connection or
   a Listener"

since a Listener is created with a call to Listen(), isn't it?

# 4.1

In this sentence:

  "alphanumeric strings in which words may be separated by hyphens."

It is not clear what is meant by "words"?  Also, s/hyphens/hyphens and
underscores"?

Note: wouldn't allowing only underscores as separators help uniformity
across languages?

In this sentence:

  "[...] and are defined in an RFC."

I am not sure why being defined in an RFC is relevant here?  Maybe
adding an example would help to understand the rationale..

The sentence:

  "Avoid using any [...]"

should probably use BCP14 language.

# 4.2

I got confused by the use of "Ignore" to mean "I have no preference, do
whatever".  Not a real problem, just a bit of cognitive dissonance.

britram:
fix #1187

### DIFF
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -519,48 +519,64 @@ Connection.Close()
 # Transport Properties {#transport-properties}
 
 Each application using the Transport Services API declares its preferences
-for how the Transport Services system should operate. This is done by using Transport Properties, as defined in
-{{!I-D.ietf-taps-arch}}, at each stage of the lifetime of a Connection.
+for how the Transport Services system should operate. This is done by using
+Transport Properties, as defined in {{!I-D.ietf-taps-arch}}, at each stage
+of the lifetime of a Connection.
 
 Transport Properties are divided into Selection, Connection, and Message
-Properties. Selection Properties (see {{selection-props}}) can only be set during pre-establishment. They are only used to specify which paths and Protocol Stacks can be used and are preferred by the application.
-Although Connection Properties (see {{connection-props}}) can be set during pre-establishment, they may be changed later. They are used to inform decisions made during establishment and to fine-tune the established Connection. Calling `Initiate` on a Preconnection creates an outbound Connection or a Listener, and the Selection Properties remain readable from the Connection or Listener, but become immutable.
+Properties. 
 
-The behavior of the selected Protocol Stack(s) when
-sending Messages is controlled by Message Properties (see {{message-props}}).
+Selection Properties (see {{selection-props}}) can only be set
+during pre-establishment. They are only used to specify which paths and
+Protocol Stacks can be used and are preferred by the application.
+Calling `Initiate` on a Preconnection creates an outbound Connection, 
+and the Selection Properties remain readable from the
+Connection, but become immutable. Selection Properties
+can be set on Preconnections, and the effect of Selection Properties
+can be queried on Connections and Messages.
 
-Selection Properties can be set on Preconnections, and the effect of
-Selection Properties can be queried on Connections and Messages.
-Connection Properties can be set on Connections and Preconnections;
-when set on Preconnections, they act as an initial default for the
-resulting Connections. Message Properties can be set on Messages,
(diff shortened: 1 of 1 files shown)


draft-ietf-taps-interface.md
```diff
@@ -519,48 +519,64 @@ Connection.Close()
...
-sending Messages is controlled by Message Properties (see {{message-props}}).
-
-Selection Properties can be set on Preconnections, and the effect of
-Selection Properties can be queried on Connections and Messages.
-Connection Properties can be set on Connections and Preconnections;
-when set on Preconnections, they act as an initial default for the
-resulting Connections. Message Properties can be set on Messages,
-Connections, and Preconnections; when set on the latter two, they act as
-an initial default for the Messages sent over those Connections.
+Properties.
```
britram, line 527:
```suggestion
Properties.
```

draft-ietf-taps-interface.md
```diff
@@ -519,48 +519,64 @@ Connection.Close()
...
-when set on Preconnections, they act as an initial default for the
-resulting Connections. Message Properties can be set on Messages,
-Connections, and Preconnections; when set on the latter two, they act as
-an initial default for the Messages sent over those Connections.
+Properties. 
+
+Selection Properties (see {{selection-props}}) can only be set
+during pre-establishment. They are only used to specify which paths and
+Protocol Stacks can be used and are preferred by the application.
+Calling `Initiate` on a Preconnection creates an outbound Connection,
```
britram, line 532:
```suggestion
Calling `Initiate` on a Preconnection creates an outbound Connection,
```

mwelzl (approved):
LGTM, thanks!

I note that the comment on "Ignore" was not taken onboard ("ignored"?  :-)  ), and I agree with this decision - we had broad agreement on these names for the requirement levels long ago.

PR #1236, opened 2023-09-04
Nits, closes #1232.

### LINKED ISSUE #1232: Very minor editorial nits
From the INT review by Tatuya Jinmei:

Very minor editorial nits follow (those are likely to be caught by the RFC
editor, but just in case)

- Section 3: s/a server/a server;/
   *  through listening on the Preconnection (i.e., passively opening,
      as in a server Section 7.2),
- Section 6: s/of they/or they/
   a multi-homed host, of they might correspond to local interfaces and
- Section 7.1: s/action Section 9.2.5/action (Section 9.2.5)/
   Connection establishment and transmission of the first Message can be
   combined in a single action Section 9.2.5.
- Section 8.1.4: s/packets Section 6.2/packets (Section 6.2)/
   A Transport Services API can request a protocol that supports sending
   keep alive packets Section 6.2.10.  If this property is Numeric, it
- Section 8.1.11.3: s/It/It is/
   can receive.  It specified as the number of bytes.
- Section 9.1.2.2: is this sentence perhaps incomplete?
   In order to set
   these properties, the add and get actions on the MessageContext.  To
- Section 9.1.3.9: s/prefered/preferred/
   avoid network layer fragmentation when transport segmentation is
   prefered.
- Section 9.1.3.10 s/endpount/endpoint/
   source fragmentation of Messages.  When running over IPv4, setting
   this property to true will result in a sending endpount setting the
- Section 13: s/whith/with/
   objects allows an application to determine whith protocol(s) and
- Section 13: s/Tranport/Transport/
   path(s) are in use.  A Tranport Services system SHOULD provide a

mwelzl:
Closes #1232


draft-ietf-taps-interface.md
```diff
@@ -2348,7 +2348,7 @@ Type:
 : Integer (non-negative)
 
 This property represents the maximum Message size that an application can receive.
-It specified as the number of bytes.
+It is specified as the number of bytes.
```
gorryfair, line 2351: Better /It is/This is/

gorryfair (approved):
Looks like could be improved in one place, see comment, and then seems ready.

philsbln (approved):
LGMT

mwelzl:
> LGMT

Looking good, me too?   :-D

draft-ietf-taps-interface.md
mwelzl, line 2351:
```suggestion
This is specified as the number of bytes.
```

PR #1237, opened 2023-09-04
A better Introduction

gorryfair:
One of the art-art review comments notes that we don't even hint at racing and cloning early in the document, and we ought to.


draft-ietf-taps-arch.md
```diff
@@ -524,7 +532,7 @@ The following categories of events can be delivered to an application:
 
 ### Connection Groups
 
-A Connection Group is a set of Connections that shares Connection Properties and cached state generated by protocols.
+A Connection Group is a set of Connections that shares Connection Properties and Cached State generated by protocols.
```
philsbln, line 535: I don't think we have a cached state objet here … so why capitalize?

gorryfair: reverted - good either way

draft-ietf-taps-arch.md
```diff
@@ -76,7 +76,17 @@ This document describes an architecture for exposing transport protocol features
...
 
-The goal of the Transport Services architecture is to provide a flexible and reusable architecture that provides a common interface for transport protocols. As applications adopt this interface, they will benefit from a wide set of transport features that can evolve over time, and ensure that the system providing the interface can optimize its behavior based on the application requirements and network conditions, without requiring changes to the applications. This flexibility enables faster deployment of new features and protocols. It can also support applications by offering racing mechanisms (attempting multiple IP addresses, protocols, or network paths in parallel), which otherwise need to be implemented in each application separately (see {{racing}}).
+The goal of the Transport Services architecture is to provide a flexible
+and a reusable architecture that provides a common interface for transport protocols. An application uses the Transport Services system through an abstract Connection.
```
mwelzl, line 80:
```suggestion
and reusable architecture that provides a common interface for transport protocols. An application uses the Transport Services system through an abstract Connection.
```

(to avoid the impression that there are two architectures)

mwelzl: I'm also not sure about the capitalization - this is before the "glossary of key terms", so to a reader, the use of capital-C "Connection" that early in the document may be weird. Maybe:

```suggestion
and reusable architecture that provides a common interface for transport protocols. An application uses the Transport Services system through an abstract Connection (we use capitalization to distinguish these from the underlying connections of, e.g., TCP).
```

draft-ietf-taps-arch.md
```diff
@@ -76,7 +76,17 @@ This document describes an architecture for exposing transport protocol features
...
+This provides
+flexible connection establishment allowing an application to request or require a set of properties.
+The system selects one or more candidate each with equivalent protocol stacks.
+This set of candidates can be raced to identify
```
mwelzl, line 84: I would complain that it's not yet explained what racing is - but I think you missed that we have already addressed this comment, at least partially, by explaining racing in the next paragraph ("It can also support applications by offering racing mechanisms..."). So this is just a duplication here.

draft-ietf-taps-arch.md
```diff
@@ -76,7 +76,17 @@ This document describes an architecture for exposing transport protocol features
...
+an optimal combination of transport protocol instance such as TCP, UDP, or another transport, together with  configuration of parameters and
+interfaces. The Connection represents an object that, once established, can be used to send and receive messages.
+A Connection can be created from another Connection, by cloning, and then forms a part of a Connection Group that share properties.
```
mwelzl, line 87: Something's grammatically weird here, with plural "share" and singular "Group". Perhaps:
```suggestion
A Connection can be created from another Connection, by cloning, and then forms a part of a Connection Group whose Connections share properties.
```

draft-ietf-taps-arch.md
```diff
@@ -314,7 +322,7 @@ A Transport Services system makes decisions that select protocols and interfaces
 
 # Transport Services Architecture and Concepts {#concepts}
 
-This section and the remainder of this document describe the architecture non-normatively. The concepts defined in this document are intended primarily for use in the documents and specifications that describe the Transport Services system. This includes the architecture, the Transport Services API and the associated Transport Services implementation. While the specific terminology can be used in some implementations, it is expected that there will remain a variety of terms used by running code.
+This section of the document describe the architecture non-normatively. The concepts defined in this document are intended primarily for use in the documents and specifications that describe the Transport Services system. This includes the architecture, the Transport Services API and the associated Transport Services implementation. While the specific terminology can be used in some implementations, it is expected that there will remain a variety of terms used by running code.
```
mwelzl, line 325:
```suggestion
This section of the document describes the architecture non-normatively. The concepts defined in this document are intended primarily for use in the documents and specifications that describe the Transport Services system. This includes the architecture, the Transport Services API and the associated Transport Services implementation. While the specific terminology can be used in some implementations, it is expected that there will remain a variety of terms used by running code.
```

mwelzl (commented):
A few suggestions...

draft-ietf-taps-arch.md
gorryfair: OK, well ... I guess we can avoid racing here, and only say the candidates are used

mwelzl: Good idea. Here it goes:
```suggestion
This set of candidates is used to identify
```

mwelzl (commented):
One fix left

draft-ietf-taps-arch.md
```diff
@@ -76,7 +76,18 @@ This document describes an architecture for exposing transport protocol features
...
+The goal of the Transport Services architecture is to provide a flexible
+and reusable architecture that provides a common interface for transport protocols. An application uses the Transport Services system through an abstract Connection (we use capitalization to distinguish these from the underlying connections of, e.g., TCP).
+This provides
+flexible connection establishment allowing an application to request or require a set of properties.
+
+As applications adopt this interface, they will benefit from a wide set of transport features that can evolve over time, and ensure that the system providing the interface can optimize its behavior based on the application requirements and network conditions, without requiring changes to the applications. This flexibility enables faster deployment of new features and protocols.
```
mwelzl, line 84:
```suggestion
As applications adopt this interface, they will benefit from a wide set of transport features that can evolve over time, and ensure that the system providing the interface can optimize its behavior based on the application requirements and network conditions, without requiring changes to the applications. This flexibility enables faster deployment of new features and protocols.
```

This might fix the compilation problem

mwelzl (approved):
lgtm !

Issue #1224, opened 2023-08-28
What about DNSSEC or DoH?


mwelzl:
From the review by Éric Vyncke, INT AD:

## Section 4.1.3

Do the security parameters include DNS resolution security parameters ? E.g.,
mandatory use of DNSSEC or DoH?

gorryfair:
I think this ought to be more cleae. To me, an equivalent stack (i.e. a candidate for racing) needs to have equivalent security properties. If any property is required, then it must be present to be offered as a candidate. The Transport Services system does not dictate what the default system policy is, but a default could require a specific security property.  Is this a good start at a short PR?

Issue #1217, opened 2023-08-28
Some items in section 2


mwelzl:
From the review by Éric Vyncke, INT AD:

## Section 2

Should 'DNS' be included in `system-provided stub resolver` ?

Figure 1 & 2 are nice but, please, add a references to them in the text.

In `it describes how implementations can use multiple IP addresses` isn't it
hidden usually to the application ?

PR #1251, opened 2023-09-05
More on Equivalence

gorryfair:
Includes:
Properties allowed for use of DNSSEC or DoH? #1224 Some items in section 2 #1217 (adding the word DNS, ref to figure 1,2)

### DIFF
--- a/draft-ietf-taps-arch.md
+++ b/draft-ietf-taps-arch.md
@@ -123,7 +123,7 @@ This subsection provides a glossary of key terms related to the Transport Servic
 - Connection Group: A set of Connections that share properties and caches.
 - Connection Property: A Transport Property that controls per-Connection behavior of a Transport Services implementation.
 - Endpoint: An identifier for one side of a Connection (local or remote), such as a hostname or URL.
-- Equivalent Protocol Stacks: Protocol Stacks that can be safely swapped or raced in parallel during establishment of a Connection.
+- Equivalent Protocol Stack: A Protocol Stack that can be safely swapped or raced in parallel during establishment of a Connection.
 - Event: A primitive that is invoked by an endpoint {{?RFC8303}}.
 - Framer: A data translation layer that can be added to a Connection to define how application-layer Messages are transmitted over a Protocol Stack.
 - Local Endpoint: A representation of the application's identifier for itself that it uses for a Connection.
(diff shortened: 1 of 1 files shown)


draft-ietf-taps-arch.md
```diff
@@ -123,7 +123,7 @@ This subsection provides a glossary of key terms related to the Transport Servic
 - Connection Group: A set of Connections that share properties and caches.
 - Connection Property: A Transport Property that controls per-Connection behavior of a Transport Services implementation.
 - Endpoint: An identifier for one side of a Connection (local or remote), such as a hostname or URL.
-- Equivalent Protocol Stacks: Protocol Stacks that can be safely swapped or raced in parallel during establishment of a Connection.
+- Equivalent Protocol Stack: A Protocol Stack that can be safely swapped or raced in parallel during establishment of a Connection.
```
mwelzl, line 126: I'm against this part of the change. Equivalence is only defined for > 1 entities - you can't say "this protocol stack is equivalent" without saying "equivalent to what?".

So, I suggest to undo, here:
```suggestion
- Equivalent Protocol Stacks: Protocol Stacks that can be safely swapped or raced in parallel during establishment of a Connection.
```

draft-ietf-taps-arch.md
```diff
@@ -300,11 +300,18 @@ The following two examples show non-equivalent Protocol Stacks:
...
+A Transport Services implementation can race different security
+protocols, e.g., if the application explicitly specifies that it considers them equivalent.
+A Transport Services implementation SHOULD only race Protocol Stacks where the transport security protocols within the stacks are identical.
+To ensure that security protocols are not incorrectly swapped, a Transport Services implementation MUST only select Protocol Stacks that meet application requirements ({{?RFC8922}}).
+A Transport Services implementation MUST NOT automatically fall back from secure protocols to insecure protocols, or to weaker versions of secure protocols.
+A Transport Services implementation MAY allow applications to explicitly specify which versions of a protocol ought to be permitted, e.g., to allow a minimum version of TLS 1.2 in case TLS 1.3 is not available.
+
+A Transport Services implementation MAY specify security properties relating to how the system operates (e.g., requirements, prohibitions, and preferences for the use of DNSSEC or DoH).
```
mwelzl, line 314:
```suggestion
A Transport Services implementation MAY specify security properties relating to how the system operates (e.g., requirements, prohibitions, and preferences for the use of DNS Security Extensions (DNSSEC) or DNS over HTTPS (DoH)).
```

mwelzl (commented):
Looks good to me, but see my two suggestions

PR #1257, opened 2023-09-05
A nit in an example IPv6 address

### LINKED ISSUE #1231: IPv6 address format nit
From the INT review by Tatuya Jinmei:

Mostly a pure editorial matter, I suggest removing the leading 0 from '0a'
in 2001:db8:4920:e29d:a420:7461:7073:0a (that appears in Sections 6.1 and
6.1.4) as it's redundant.

mwelzl:
Closes #1231

I really know nothing about the common way to write IPv6 addresses; someone who does, should just merge this very simple PR.

### DIFF
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -740,7 +740,7 @@ RemoteSpecifier.WithIPAddress(192.0.2.21)
 ~~~
 
 ~~~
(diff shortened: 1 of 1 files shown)


mwelzl:
... and: will replacing "0a" with "a" make this a valid RFC 5952 address?  (see the comment by Éric Vyncke in his review)

philsbln:
Yip, addresses are now as RFC 5952 suggested (remove all leading zeros, replace longes series of series of zeros between ":" with "::", use lower-case)

PR #1269, opened 2023-09-05
Should remote endpoint include mcast group?

gorryfair:
We need to be careful, if we say a host, ought we say a host or a multicast group?


draft-ietf-taps-arch.md
```diff
@@ -149,7 +149,7 @@ This subsection provides a glossary of key terms related to the Transport Servic
 - Protocol Instance: A single instance of one protocol, including any state necessary to establish connectivity or send and receive Messages.
 - Protocol Stack: A set of Protocol Instances that are used together to establish connectivity or send and receive Messages.
 - Racing: The attempt to select between multiple Protocol Stacks based on the Selection and Connection Properties communicated by the application, along with any Security Parameters.
-- Remote Endpoint: A representation of the application's identifier for a peer that can participate in establishing a Connection.
+- Remote Endpoint: A representation of the application's identifier for a peer (or multicast group) that can participate in establishing a Connection.
```
csperkins, line 152: Agree we should mention multicast here. Maybe "…for a peer that can participate in establishing a Connection or for a multicast group to which the application will send data"?

csperkins (approved):

draft-ietf-taps-arch.md
gorryfair: now in the PR

PR #1270, opened 2023-09-05
One Receive Event per Receive call

### LINKED ISSUE #1254: Many, many events upon reception
From the review by Robert Wilton

(2) p 23, sec 4.1.6.  Event Handling

  *  Message Received: Delivers received Message content to the
     application, based on a Receive action.  This can include an error
     if the Receive action cannot be satisfied due to the Connection
     being closed.

Perhaps this is already answered by the other documents, but is it always the
case that every message requires a separate event/notification to the
application?  E.g., I can imagine that under some high performance/throughput
scenarios where there are large number of messages being received that the
application may only want a single notification that there are one or more
messages waiting to be processed, upon which it pulls some/all of them off a
receive queue, and only receives another "message received/waiting" event if a
message is received when the queue is empty, or the queue is not empty after a
subset of the messages have been dequeued.

mwelzl:
Closes #1254

### DIFF
--- a/draft-ietf-taps-arch.md
+++ b/draft-ietf-taps-arch.md
@@ -531,7 +531,7 @@ The following categories of events can be delivered to an application:
 
 * Connection Received: Signals to an application that a given Listener has received a Connection.
 
-* Message Received: Delivers received Message content to the application, based on a `Receive` action. To allow an application to limit the occurrence of such events, each call to `Receive` will be paired with a single `Receive` event. This can include an error if the `Receive` action cannot be satisfied, e.g. due to the Connection being closed.
+* Message Received: Delivers received Message content to the application, based on a `Receive` action. To allow an application to limit the occurrence of such events, each call to `Receive` will be paired with a single `Receive` event. This can include an error if the `Receive` action cannot be satisfied, e.g., due to the Connection being closed.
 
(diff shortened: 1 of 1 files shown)


gorryfair (approved):
Good with a comma after e.g., according to US style.

draft-ietf-taps-arch.md
```diff
@@ -531,7 +531,7 @@ The following categories of events can be delivered to an application:
 
 * Connection Received: Signals to an application that a given Listener has received a Connection.
 
-* Message Received: Delivers received Message content to the application, based on a `Receive` action. This can include an error if the `Receive` action cannot be satisfied due to the Connection being closed.
+* Message Received: Delivers received Message content to the application, based on a `Receive` action. To allow an application to limit the occurrence of such events, each call to `Receive` will be paired with a single `Receive` event. This can include an error if the `Receive` action cannot be satisfied, e.g. due to the Connection being closed.
```
mwelzl, line 534:
```suggestion
* Message Received: Delivers received Message content to the application, based on a `Receive` action. To allow an application to limit the occurrence of such events, each call to `Receive` will be paired with a single `Receive` event. This can include an error if the `Receive` action cannot be satisfied, e.g., due to the Connection being closed.
```

gorryfair (approved):

tfpauly (approved):

PR #1274, opened 2023-09-06
A tiny fix to one sentence

### LINKED ISSUE #1273: A nit in section 2.1
From the review by Warren Kumari:

2: S 2.1.  Event-Driven API
This paragraph compares and contrasts the Socket API and the Transport Services
API, but in the paragraph starting with "For example, an application first
issues a call to receive new data from the connection.", it is unclear under
which paradigm you are meaning. I'd suggest: "For example, when using the
Transport Services API, an application first ..."

mwelzl:
Closes #1273


philsbln (approved):
works

PR #1277, opened 2023-09-06
Endpoint, Endpoint identifier, and specifier

gorryfair:
Three similar things ought to represent specific aspects.

### DIFF
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -496,7 +496,7 @@ Connection -> Received<messageDataResponse, messageContext>
 // Trickle ICE, then add them to the Connection:
 Connection.AddRemote(NewRemoteCandidates)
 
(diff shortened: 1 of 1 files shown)


draft-ietf-taps-interface.md
```diff
@@ -672,7 +672,7 @@ the Preconnection is used to `Listen` for incoming Connections.
 At least one Local Endpoint and one Remote Endpoint MUST be specified if a
 peer-to-peer `Rendezvous` is to occur based on the Preconnection.
 
-If more than one Local Endpoint is specified on a Preconnection, then all
+If more than one Local Endpoint indentifier is specified on a Preconnection, then all
```
tfpauly, line 675:
```suggestion
If more than one Local Endpoint identifier is specified on a Preconnection, then all
```

britram (approved):

PR #1288, opened 2023-09-08
A Transport Services System

gorryfair:
This document has been edited many ways, and this PR proposes a change in title and refinement in the use of the word 'architecture' to focus on the system specification, rather than the architectural principles. It does not intend to change the system in any way, this is intended as a language revision, to make it slightly less philosophical.

PR #1306, opened 2023-09-08
A title change to The Transport Service System?

gorryfair:
Do not merge this before discussion.

This specific PR relies upon another PR  #1288  to untangle 'architecture' 'implementation' and 'system'.  This other PR undos a mangling generated by many years of editing but that anyway needs to be resolved.

### DIFF
--- a/draft-ietf-taps-arch.md
+++ b/draft-ietf-taps-arch.md
@@ -1,5 +1,5 @@
 ---
-title: An Architecture for Transport Services
+title: The Transport Service System
 abbrev: TAPS Architecture
 docname: draft-ietf-taps-arch-latest
 date:


mirjak:
This looks good to me but you changed finally everything but not the title...

gorryfair:
Indeed, this was the big cross-check, but with small gain :-)
The title change is one click further away in #1306, please review (and improve). I thought for the title, people ought to make a specific decision to change. If this all works out, there'll be minor PRs to the other 2 drafts also.

mwelzl (approved):

tfpauly:
I'm not sure about this change in approach — I think it still is, largely, an architecture, or at least the design and principles behind the abstract interface. Just saying "system" seems a bit too vague to me.

zaheduzzaman:
(with my AD hat off) I am not super convinced that this title change is necessary or is the right thing to do. what problem is this title change solving?

gorryfair:
So, when I started working on a document entitled architecture - I had in mind something different to what now Eric has in mind for a document with this title. I see why Eric suggests a name change, and I agree. This document has requirements, it specifies the components you need for TAPS - specifically focussing on the API, and the transport services system implementation. The word implementation here isn't the same as "implementing TAPS ID" but I can live with that. Together they are called the "Transport Services System". 

Eric also suggested something like "transport services Architecture and Requirements" - which I agree might be another way to match the reader expectations and the document content, but I preferred this myself.

abrunstrom:
If we want to change the title I think "Transport Services Architecture and Requirements" is a bit more clear than "Transport Services System". The latter sounds a bit vague to me as well.

gorryfair:
Overtaken by events - see new PR

tfpauly:
See #1381

PR #1318, opened 2023-09-11
Create IANA registry for Tansport Property Namespace 

### LINKED ISSUE #1275: How are IETF namespaces to be differentiated?
This issue arose while trying to determine the case of TCP in the specs, but isn't related to that. Issue #1271.

This concerns how an IETF WG could create a namespace for properties associated with an IETF protocol. The original idea was that we just copied entries from an IANA registry to give us a set of namespaces for each IETF protocol. However, we never actually said how we would avoid a vendor name matching an IETF protocol, or for that matter what happens when a new protocol emerges (as QUIC did). The only way I can see to make this manageable for a consistent parameter set for an app, is to declare the format of IETF namespace at least different to the vendor space. The two namespaces can be treated the same in the transport system, this is about allocating the names.

That would mean that the format of the two types of namespaces needs to differ in some way. We could use an initial capital letter to differentiate IETF namespaces (most IETF  protocols and extensions do use capital letters in the acronyms). That might be less language independent, so an alternative proposal is to prefix vendor namespaces with a special character (this could be underscore, or anything more appropriate). I think we need to do this, what are thoughts on using underscore before vendor namespaces?

philsbln:
Proposal how to close #1275

### DIFF
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -3426,15 +3426,13 @@ Registry Name: Transport Property Namespaces
 Registration Procedure:
 
 - IETF Review
-- First Come First Served for vendor/implementation specific entries strating with an underscore "_"
+- First Come First Served for vendor/implementation specific entries; These entries MUST start with an underscore "_" and provide a stable URL as reference for the values within the namespace.
 
 The initial values in the "Transport Property Namespaces" registry are:
 
-   | namespace  | Protocol                              | Reference       |
+   | Namespace  | Protocol                              | Reference       |
    |------------|---------------------------------------|-----------------|
-   | udp        | User Datagram Protocol                | This Document   |
    | tcp        | Transmission Control Protocol         | This Document   |
-   | sctp       | Stream Transmission Control Protocol  | This Document   |
 {: #tab-iana-taps-namespaces title="Inital Transport Property Namespaces"}
 
 Later versions of this document may create an IANA registry for generic transport property names (see {{property-names}}).
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -583,9 +583,9 @@ form \[\<Namespace>.\]\<PropertyName\>.
 - Protocol-specific Properties MUST use the protocol acronym as the Namespace (e.g., a
   `tcp` Connection could support a TCP-specific Transport Property, such as the user timeout
   value, in a Protocol-specific Property called `tcp.userTimeoutValue` (see {{tcp-uto}})).
-- Vendor or implementation specific properties MUST use a string identifying
-  the vendor or implementation as the Namespace.
-- For IETF protocols, the name of a Protocol-specific Property SHOULD be specified in an IETF document published in the RFC Series after IETF review.
+- Vendor or implementation specific properties MUST be placed in a Namespace starting with the underscore `_` character
+   and SHOULD use a string identifying the vendor or implementation.
+- For IETF protocols, the name of a Protocol-specific Property SHOULD be specified in an IETF document published in the RFC Series after IETF review and registered in the "Transport Property Namespaces" registry.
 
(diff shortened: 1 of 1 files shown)


draft-ietf-taps-interface.md
```diff
@@ -3419,8 +3419,25 @@ The Transport Services API  provides the following guarantees about the ordering
...
+- IETF Review
+- First Come First Served for vendor/implementation specific entries strating with an underscore "_"
+
+The initial values in the "Transport Property Namespaces" registry are:
+
+   | namespace  | Protocol                              | Reference       |
+   |------------|---------------------------------------|-----------------|
+   | udp        | User Datagram Protocol                | This Document   |
+   | tcp        | Transmission Control Protocol         | This Document   |
+   | sctp       | Stream Transmission Control Protocol  | This Document   |
```
gorryfair, line 3437: Ought we to add UDP-Lite?

gorryfair (commented):
Some suggestions

draft-ietf-taps-interface.md
philsbln: That is a good question … we actually only specify protocol specific properties for TCP and do not explicitly state that there are none for UDP and SCTP. 
So either remove UDP and SCTP form the list or add UDP-Lite and add sentence that there are no protocol specific properties because all tuneable they are mapped from generic properties.

gorryfair: I'd say: Either all the protocols we specify, or only the ones we touch :-)

philsbln: Fixed – only include the ones we touch.

draft-ietf-taps-interface.md
```diff
@@ -583,9 +583,9 @@ form \[\<Namespace>.\]\<PropertyName\>.
 - Protocol-specific Properties MUST use the protocol acronym as the Namespace (e.g., a
   `tcp` Connection could support a TCP-specific Transport Property, such as the user timeout
   value, in a Protocol-specific Property called `tcp.userTimeoutValue` (see {{tcp-uto}})).
-- Vendor or implementation specific properties MUST use a string identifying
-  the vendor or implementation as the Namespace.
-- For IETF protocols, the name of a Protocol-specific Property SHOULD be specified in an IETF document published in the RFC Series after IETF review.
+- Vendor or implementation specific properties MUST be placed in a Namespace starting with the underscore `_` character
+   and SHOULD use a string identifying the vendor or implementation.
+- For IETF protocols, the name of a Protocol-specific Property SHOULD be specified in an IETF document published in the RFC Series after IETF review and registered in the "Transport Property Namespaces" registry.
```
gorryfair, line 589: Add for clarity here: An IETF protocol Namespace does not start with an underscore character.

gorryfair (changes requested):
Ready I think, but we could add a sentence which I think could help and does not harm, see suggestion.

philsbln:
I guess this one is ready now. Let's discuss on the interim whether to registry or not and i will finalise accordingly.

draft-ietf-taps-interface.md
```diff
@@ -583,9 +583,10 @@ form \[\<Namespace>.\]\<PropertyName\>.
...
+  An IETF protocol Namespace does not start with an underscore character.
```
tfpauly, line 589:
```suggestion
- Vendor or implementation specific properties MUST be placed in a Namespace starting with the underscore `_` character
   and SHOULD use a string identifying the vendor or implementation.
- For IETF protocols, the name of a Protocol-specific Property SHOULD be specified in an IETF document published in the RFC Series after IETF review.
  An IETF protocol Namespace does not start with an underscore character.
```

mwelzl:
Consensus at interim don't add a registry yet, but keep some of the other text (the underscore, ..).
A registry can be created by a future RFC.

martinduke:
I guess I dont't konw why the registry is a big deal, but barring a really unlikely naming collision  I don't think this will be catastrophic.

PR #1380, opened 2023-09-24
Fix word rendezvousing and purposes in API

### LINKED ISSUE #1379: Rendezvouzzzzz

NITS:

Wouldn't the American spelling be "rendezvouzing" instead of "rendezvousing" :-)

gorryfair:
Closes #1379 when merged

### DIFF
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -305,6 +305,8 @@ connection-oriented. Connections can be created from Preconnections in three way
 - by listening on the Preconnection (i.e., passively opening, as in a server; {{listen}}),
 - or by a rendezvous for the Preconnection (i.e., peer to peer establishment; {{rendezvous}}).
 
+if the application requires this or
+
 Once a Connection is established, data can be sent and received on it in the form of
 Messages. The API supports the preservation of message boundaries both
 via explicit Protocol Stack support, and via application support through a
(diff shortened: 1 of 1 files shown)


draft-ietf-taps-interface.md
```diff
@@ -3476,7 +3475,7 @@ the goal of a Transport Services system is to support
 such mechanisms as a generic service within the transport layer. This enables applications to more dynamically
 benefit from innovations and new protocols in the transport, although it reduces transparency of the
 underlying communication actions to the application itself. The Transport Services API is designed such that protocol and path selection
-can be limited to a small and controlled set if required by the application for functional or security purposes. Further,
+can be limited to a small and controlled set if required by the application to perform a function or to provide security. Further,
```
philsbln, line 3478: I stumble over the new text here and would suggest to keep the old one.

philsbln (changes requested):
The first two changes work for me and make the text clearer, the third one not.

draft-ietf-taps-interface.md
```diff
@@ -303,7 +303,7 @@ connection-oriented. Connections can be created from Preconnections in three way
 
 - by initiating the Preconnection (i.e., actively opening, as in a client; {{initiate}}),
 - through listening on the Preconnection (i.e., passively opening, as in a server; {{listen}}),
-- or rendezvousing on the Preconnection (i.e., peer to peer establishment; {{rendezvous}}).
+- or a rendezvous for the Preconnection (i.e., peer to peer establishment; {{rendezvous}}).
```
mwelzl, line 306:
```suggestion
- or via a rendezvous for the Preconnection (i.e., peer to peer establishment; {{rendezvous}}).
```

I think this needs a word such as "via" (or "through"?) because it's an item list, and all items in the list begin with "by..." or "through..." or something like that.

mwelzl (changes requested):
A truly tiny change request  :)

draft-ietf-taps-interface.md
mwelzl: FWIW, I think the very latest version ("if the application requires this or ...") is better than both the original and the update that was discussed here.

philsbln (approved):
Yip … works for me now

draft-ietf-taps-interface.md
gorryfair: Will fix to use "by" in each case to be uniform

gorryfair:
resolved thanks for review comments

PR #1381, opened 2023-09-24
2nd attempt at a Title Change

gorryfair:
This PR seeks to resolve Eric's note that the arch is more than an architecture because it includes the normative requirements

### DIFF
--- a/draft-ietf-taps-arch.md
+++ b/draft-ietf-taps-arch.md
@@ -1,5 +1,5 @@
 ---
-title: An Architecture for Transport Services
+title: An Architecture and Requirements for Transport Services
 abbrev: TAPS Architecture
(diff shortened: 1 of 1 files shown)


mwelzl (commented):
Works for me (as not even a co-author); whether this addresses the concern, or helps, I don't know... this should be discussed at the interim IMO.

draft-ietf-taps-arch.md
```diff
@@ -1,5 +1,5 @@
 ---
-title: An Architecture for Transport Services
+title: An Architecture and Requirements for Transport Services
```
tfpauly, line 2:
```suggestion
title: Architecture and Requirements for Transport Services
```

Issue #1349, opened 2023-09-15
MSS and PMTUD


mwelzl:
From the review by Lars Eggert, COMMENT part:

### Section 8.1.11.1, paragraph 3
```
    This property, if applicable, represents the maximum Message size
    that can be sent without incurring network-layer fragmentation at the
    sender.  It is specified as the number of bytes.  It exposes a value
    to the application based on the Maximum Packet Size (MPS) as
    described in Datagram PLPMTUD [RFC8899].  This can allow a sending
    stack to avoid unwanted fragmentation at the network-layer or
    segmentation by the transport layer.
```
Is this supposed to vary as PMTUD happens?

gorryfair:
Yes? What do we need to say... read this parameter to find out ...?

mwelzl:
I think we just should say "yes" and say so, with a sentence. Proposed in #1383.

PR #1383, opened 2023-09-25
Fix MPS text

gorryfair:


draft-ietf-taps-interface.md
```diff
@@ -2327,10 +2327,14 @@ Type:
...
+It is specified as a number of bytes and is less than or equal to the
+Maximum Message Size on Send.
+It exposes a readable value to the application
 based on the Maximum Packet Size (MPS) as described in Datagram PLPMTUD {{?RFC8899}}.
-This can allow a sending stack to avoid unwanted fragmentation at the
-network-layer or segmentation by the transport layer.
+This value allows a sending stack to avoid unwanted fragmentation at the
+network-layer or segmentation by the transport layer before
+choosing the message size and/or after a `SendError` occurs indicating
+an attempt to send a Message that is too large.
```
mwelzl, line 2337: Is this needed? It seems to me that this just made the sentence harder to read.

draft-ietf-taps-interface.md
```diff
@@ -2327,10 +2327,14 @@ Type:
 
 This property, if applicable, represents the maximum Message size that can be
 sent without incurring network-layer fragmentation at the sender.
-It is specified as a number of bytes. It exposes a value to the application
+It is specified as a number of bytes and is less than or equal to the
+Maximum Message Size on Send.
+It exposes a readable value to the application
 based on the Maximum Packet Size (MPS) as described in Datagram PLPMTUD {{?RFC8899}}.
```
mwelzl, line 2333: To address #1349, I'd suggest the following change (added sentence), here:
```suggestion
based on the Maximum Packet Size (MPS) as described in Datagram PLPMTUD {{?RFC8899}}. This means that the value of this property can change over time (it can be different before and after executing Datagram PLPMTUD).
```

mwelzl (changes requested):
I made suggestions... perhaps this addresses #1349, then?

draft-ietf-taps-interface.md
gorryfair: At some point you ought too explain when to use this?

draft-ietf-taps-interface.md
gorryfair: What does execute DPLPMTUD mean? Is this crisper?

/as described in Datagram PLPMTUD {{?RFC8899}}./. The value of this property can change over time (and can be updated by Datagram PLPMTUD  {{?RFC8899}})./

PR #1385, opened 2023-09-26
Nits

### LINKED ISSUE #1355: nits

All comments below are about very minor potential issues that you may choose to
address in some way - or ignore - as you see fit. Some were flagged by
automated tools (via https://github.com/larseggert/ietf-reviewtool), so there
will likely be some false positives. There is no need to let me know what you
did with these suggestions.

#### Section 9.1.3.9, paragraph 4
```
-    prefered.
+    preferred.
+          +
```

#### Section 9.1.3.10, paragraph 3
```
-    this property to true will result in a sending endpount setting the
-                                                        ^
+    this property to true will result in a sending endpoint setting the
+                                                        ^
```

### Outdated references

Reference `[RFC8229]` to `RFC8229`, which was obsoleted by `RFC9329` (this may
be on purpose).

#### Section 1.1, paragraph 2
```
ming language, e.g. UTF-8. If the intend of this API is to be language indep
                                 ^^^^^^
```
The word "intend" is a verb. Did you mean the noun "intent"?

#### Section 2, paragraph 2
```
the kind of transport, can be bi-directional or unidirectional, and that ca
                              ^^^^^^^^^^^^^^
```
This word is normally spelled as one.

#### Section 6.1.2, paragraph 3
```
tener := Preconnection.Listen() Join an Source-Specific Multicast group in re
                                    ^^
```
Use "a" instead of "an" if the following word doesn't start with a vowel sound,
e.g. "a sentence", "a university".

#### Section 8, paragraph 9
```
lgorithm); to prefer immediate acknowledgment from the peer endpoint when su
                              ^^^^^^^^^^^^^^
```
Do not mix variants of the same word ("acknowledgment" and "acknowledgement")
within a single text.

#### Section 8.1.6, paragraph 11
```
onal Message Context, which allows to add Message Properties, identify Send
                                  ^^^^^^
```
Did you mean "adding"? Or maybe you should add a pronoun? In active voice,
"allow" + "to" takes an object, usually a pronoun.

#### Section 8.1.10, paragraph 4
```
In these cases, they add an additional transformation to further encode or e
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
This phrase might be redundant. Consider either removing or replacing the
adjective "additional".

#### Section 9.3.2.3, paragraph 3
```
paths and are not necessarily privacy sensitive. Still, some information cou
                             ^^^^^^^^^^^^^^^^^
```
This word is normally spelled with a hyphen.

#### Section 9.3.2.3, paragraph 3
```
l, some information could be privacy sensitive because it might reveal usage
                            ^^^^^^^^^^^^^^^^^
```
This word is normally spelled with a hyphen.

#### Section 14, paragraph 5
```
cts (see Section 6.2) that are pre-configured with frequently used sets of p
                              ^^^^^^^^^^^^^^
```
Do not mix variants of the same word ("pre-configure" and "preconfigure")
within a single text.

mwelzl:
Closes #1355.

### DIFF
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -959,7 +959,7 @@ port on a named local interface:
    Listener := Preconnection.Listen()
 ~~~
 
-Join an Source-Specific Multicast group in receive-only mode, bound to a known
+Join a Source-Specific Multicast group in receive-only mode, bound to a known
 port on a named local interface:
 
 ~~~
@@ -2449,7 +2449,7 @@ to communicate the boundaries of the data being transferred.
 
 ## Messages and Framers {#msg}
 
-Each Message has an optional Message Context, which allows to add Message Properties, identify `Send` events related to a specific Message or to inspect meta-data related to the Message sent. Framers can be used to extend or modify the Message data with additional information that can be processed at the receiver to detect message boundaries.
+Each Message has an optional Message Context, which allows adding add Message Properties, identify `Send` events related to a specific Message or to inspect meta-data related to the Message sent. Framers can be used to extend or modify the Message data with additional information that can be processed at the receiver to detect message boundaries.
 
 
 ### Message Contexts {#msg-ctx}
@@ -2497,7 +2497,7 @@ Then, the Message Framer can intercept all calls to `Send` or `Receive`
 on a Connection to add Message semantics, in addition to interacting with
 the setup and teardown of the Connection. A Framer can start sending data
 before the application sends data if the framing protocol requires a prefix
-or handshake (see {{?RFC8229}} for an example of such a framing protocol).
+or handshake (see {{?RFC9329}} for an example of such a framing protocol).
 
 ~~~~~~~~~~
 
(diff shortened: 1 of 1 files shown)


mwelzl:
"one NIT" - what?

draft-ietf-taps-interface.md
```diff
@@ -2449,7 +2449,7 @@ to communicate the boundaries of the data being transferred.
 
 ## Messages and Framers {#msg}
 
-Each Message has an optional Message Context, which allows to add Message Properties, identify `Send` events related to a specific Message or to inspect meta-data related to the Message sent. Framers can be used to extend or modify the Message data with additional information that can be processed at the receiver to detect message boundaries.
+Each Message has an optional Message Context, which allows adding add Message Properties, identify `Send` events related to a specific Message or to inspect meta-data related to the Message sent. Framers can be used to extend or modify the Message data with additional information that can be processed at the receiver to detect message boundaries.
```
gorryfair, line 2452: /adding add/adding/

mwelzl:
```suggestion
Each Message has an optional Message Context, which allows adding Message Properties, identify `Send` events related to a specific Message or to inspect meta-data related to the Message sent. Framers can be used to extend or modify the Message data with additional information that can be processed at the receiver to detect message boundaries.
```

PR #1386, opened 2023-09-27
Endpoints and Endpoint Identifiers. Closes #1278.

### LINKED ISSUE #1278: Bring use of "endpoint" in line with new definition from arch
Once the "endpoint" PRs have landed. This is all based upon #1272

mwelzl:
Closes #1278.

Please take a close look even though it's long and painful; this isn't actually trivial.

For example: multipath protocols deal with adding or removing addresses, which are Endpoint Identifiers. However, the API specifies an "Endpoint", not an "Endpoint Identifier", because the Endpoint can be represented in various ways (IP address, name, ... - these things are "Endpoint Identifiers"). So, in the API, the text about adding or removing endpoints should indeed talk about endpoints (and it does!), not endpoint identifiers.

Then again, all the racing stuff in the implementation draft is really about Endpoint Identifiers, right?

See... not trivial.

**NOTE:** the changes to -arch are just capitalization.


philsbln:
Before giving a review here, let's make sure that we are on the same page with definitions and don't mess up the terminology:
- An Endpoint is an Object representing a thingy I want to talk to (ignoring whether this is just one IoT-device, a container on a single host or a CND tenant hosting some content)
- An Endpoint Identifier is a hostname or address representing that Endpoint in order to initiate a communication.
- There is not such a thing as "adding endpoints" to a connection … I can just add additional identifiers
- When Multi-path protocols add addresses internally, they are actually adding alternative means to reach same endpoint that might or might not be usable as an endpoint identifier, but adding them through the API is identical to adding an additional identifier for the same endpoint (sic!).
- Racing is not about endpoint identifiers, but about candidates instances of the same Endpoint where each endpoint identifier may actually represent several instances of that endpoint. In the end, typically one of these instances will be chosen.

So that means candidate instance of an Endpoint != Endpoint?

mwelzl:
Thanks @philsbln !  I changed your item list into a numbered list so it's easier to talk about.

I agree with items 1 (but I think that this definition needs to be broader) and 2.

Item 3 is already where things become complicated... at first, when doing this update, I changed "NewRemoteEndpoint" to "NewRemoteEndPointIdentifier" etc. etc. ... but then I realized, the identifier is: a name+port, or an IPv4 address + port, or an IPv6 address + port... so, in a sense, the way we use "EndPoint" in the API it's a slightly more abstract thing.

So, I think, the way we've been using it, and "endpoint" is really:
a) what you say in item 1 (a "thingy" that we talk to), but also
b) a representation of that thing in our interface (just a tiny bit more abstract than an address).

Let's see if we agree about this first, or if this should change.

philsbln:
Hi @mwelzl,

I agree with your assessment so far. 

In my opinion, based on a stringent, terminology, NewRemoteEndpoint() creates an Endpoint Identifier.
Now comes the trade-off between a stringent terminology and a usable API – no one likes Java-Style MyFooWindowDecoratorAdpterProxy like names). 

Therefore I think it is justifiable to call the Endpoint Identifier Object just "Endpoint" as the real Endpoint is an abstract thingy somewhat within the Preconnection/Connection.

Is it fine ti review based on that?

gorryfair:
How about EndpointID ... that is not much longer than Endpoint?

philsbln:
@gorryfair In principle 👍🏻 but I fear most developers associate ID with something uniq / that has a bijective association which isn't true for the Endpoint Identifier

mwelzl:
@philsbln it seems we agree... yes, reviewing with this as a basis would be fine, thanks!

Just for clarification, with the understanding that we agree that an "Endpoint" can either be a thing that we talk to, or really what should probably be called the "Endpoint Identifier Object", regarding the other points above:

3. no, we add "Endpoints".
4. multipath protocols also add "Endpoints", which can be specified by various types of Endpoint Identifiers (e.g., an IPv4 address or a name),
5. racing is another tricky thing in this context....  of course, when we talk about sending packets to an "Endpoint", we send them to the "thingy", not the address, but most of the text on racing in the implementation draft is about building tables and how to sort them... where, really, the tables have to contain something quite concrete: Endpoint Identifiers!  So I changed many of these occurrences into "Endpoint Identifier".

philsbln:
So for 5) … would not be "Endpoints Candidate" the better term here?

mwelzl:
I don't know... it adds another term?  E.g., just to take an example from the text:

```
Aggregate [Endpoint Identifier: www.example.com:80] [Interface: Any]   [Protocol: TCP]
 |-> [Endpoint Identifier: 192.0.2.1:80]       [Interface: Wi-Fi] [Protocol: TCP]
 |-> [Endpoint Identifier: 192.0.2.1:80]       [Interface: LTE]   [Protocol: TCP]
 |-> [Endpoint Identifier: 2001:DB8::1.80]     [Interface: LTE]   [Protocol: TCP]
```

We could replace all these with "Endpoint Candidate" - but they ARE all Endpoint Identifiers at the same time, so ... it won't make things clearer to use an extra term here?

Please, if you look through this and you find occurrences where "Endpoint candidate" (perhaps not with capital C, as it's not a new term that we define - we do talk about candidates, and Endpoint candidates in this text already) would be a better fit, make a change suggestion.  It really wasn't easy to always know precisely what the right term should be - I did my best.

philsbln:
I will try my very best … but unsure whether I can make it tomorrow or on Wednesday after my vacation

draft-ietf-taps-impl.md
```diff
@@ -316,23 +316,23 @@ Another example is racing SCTP with TCP:
...
+Branch types ought to occur in a specific order relative to one another to avoid creating leaf nodes with invalid or incompatible settings. In the example above, it would be invalid to branch for derived Endpoint Identifers (the DNS results for www.example.com) before branching between interface paths, since there are situations when the results will be different across networks due to private names or different supported IP versions. Implementations need to be careful to branch in a consistent order that results in usable leaf nodes whenever there are multiple branch types that could be used from a single node.
 
 This document recommends the following order of operations for branching:
 
 1. Network Paths
 2. Protocol Options
-3. Derived Endpoints
+3. Derived Endpoint Identifers
```
philsbln, line 329:
```suggestion
3. Derived Endpoints
```

draft-ietf-taps-impl.md
```diff
@@ -316,23 +316,23 @@ Another example is racing SCTP with TCP:
...
+Protocol options are next checked in order. Whether or not a set of protocols, or protocol-specific options, can successfully connect is generally not dependent on which specific IP address is used. Furthermore, the Protocol Stacks being attempted may influence or altogether change the Endpoint Identifers being used. Adding a proxy to a connection's branch will change the Endpoint Identifer to the proxy's IP address or hostname. Choosing an alternate protocol may also modify the ports that should be selected.
 
-Branching for derived endpoints is the final step, and may have multiple layers of derivation or resolution, such as DNS service resolution and DNS hostname resolution.
+Branching for derived Endpoint Identifers is the final step, and may have multiple layers of derivation or resolution, such as DNS service resolution and DNS hostname resolution.
```
philsbln, line 335:
```suggestion
Branching for derived Endpoints is the final step, and may have multiple layers of derivation or resolution, such as DNS service resolution and DNS hostname resolution.
```

draft-ietf-taps-impl.md
```diff
@@ -380,31 +380,31 @@ The available protocols and paths on a specific system and in a specific context
...
 
 ### Gathering Endpoint Candidates
 
-Both Local and Remote Endpoint Candidates must be discovered during connection establishment.  To support Interactive Connectivity Establishment (ICE) {{?RFC8445}}, or similar protocols that involve out-of-band indirect signalling to exchange candidates with the Remote Endpoint, it is important to query the set of candidate Local Endpoints, and provide the Protocol Stack with a set of candidate Remote Endpoints, before the Local Endpoint attempts to establish connections.
+Both Local and Remote Endpoint Identifer Candidates must be discovered during connection establishment.  To support Interactive Connectivity Establishment (ICE) {{?RFC8445}}, or similar protocols that involve out-of-band indirect signalling to exchange candidates with the Remote Endpoint, it is important to query the set of candidate Local Endpoints, and provide the Protocol Stack with a set of candidate Remote Endpoint Identifers, before the Local Endpoint attempts to establish connections.
```
philsbln, line 387:
```suggestion
Both Local and Remote Endpoint Candidates must be discovered during connection establishment.  To support Interactive Connectivity Establishment (ICE) {{?RFC8445}}, or similar protocols that involve out-of-band indirect signalling to exchange candidates with the Remote Endpoint, it is important to query the set of candidate Local Endpoints, and provide the Protocol Stack with a set of candidate Remote Endpoints, before the Local Endpoint attempts to establish connections.
```

draft-ietf-taps-impl.md
```diff
@@ -380,31 +380,31 @@ The available protocols and paths on a specific system and in a specific context
...
 
 Gathering Local Endpoints is primarily a local operation, although it might involve exchanges with a STUN server to derive server reflexive Local Endpoints, or with a TURN server or other relay to derive relayed Local Endpoints.  However, it does not involve communication with the Remote Endpoint.
 
 #### Remote Endpoint Candidates
 
-The Remote Endpoint is typically a name that needs to be resolved into a set of possible addresses that can be used for communication.  Resolving the Remote Endpoint is the process of recursively performing such name lookups, until fully resolved, to return the set of candidates for the Remote Endpoint of this Connection.
+The Remote Endpoint Identifer is typically a name that needs to be resolved into a set of possible addresses that can be used for communication.  Resolving the Remote Endpoint Identifer is the process of recursively performing such name lookups, until fully resolved, to return the set of candidates for the Remote Endpoint of this Connection.
```
philsbln, line 399:
```suggestion
The Remote Endpoint Identifer is typically a name that needs to be resolved into a set of possible addresses that can be used for communication.  Resolving the Remote Endpoint is the process of recursively performing such name lookups, until fully resolved, to return the set of candidates for the Remote Endpoint of this Connection.
```
Resolving/deriving endpoints sounds somewhat wrong … prefer endpoint here

draft-ietf-taps-impl.md
```diff
@@ -380,31 +380,31 @@ The available protocols and paths on a specific system and in a specific context
...
+How this resolution is done will depend on the type of the Remote Endpoint, and can also be specific to each Local Endpoint.  A common case is when the Remote Endpoint Identifer is a DNS name, in which case it is resolved to give a set of IPv4 and IPv6 addresses representing that name.  Some types of Remote Endpoint Identifers might require more complex resolution.  Resolving the Remote Endpoint Identifer for a peer-to-peer connection might involve communication with a rendezvous server, which in turn contacts the peer to gain consent to communicate and retrieve its set of candidate Local Endpoint Identifers, which are returned and form the candidate remote addresses for contacting that peer.
```
philsbln, line 401:
```suggestion
How this resolution is done will depend on the type of the Remote Endpoint, and can also be specific to each Local Endpoint.  A common case is when the Remote Endpoint Identifer is a DNS name, in which case it is resolved to give a set of IPv4 and IPv6 addresses representing that name.  Some types of Remote Endpoint Identifers might require more complex resolution.  Resolving the Remote Endpoint for a peer-to-peer connection might involve communication with a rendezvous server, which in turn contacts the peer to gain consent to communicate and retrieve its set of candidate Local Endpoint, which are returned and form the candidate remote addresses for contacting that peer.
```

draft-ietf-taps-impl.md
```diff
@@ -380,31 +380,31 @@ The available protocols and paths on a specific system and in a specific context
...
 
-Resolving the Remote Endpoint is not a local operation.  It will involve a directory service, and can require communication with the Remote Endpoint to rendezvous and exchange peer addresses.  This can expose some or all of the candidate Local Endpoints to the Remote Endpoint.
+Resolving the Remote Endpoint Identifer is not a local operation.  It will involve a directory service, and can require communication with the Remote Endpoint to rendezvous and exchange peer addresses.  This can expose some or all of the candidate Local Endpoint Identifers to the Remote Endpoint.
```
philsbln, line 403:
```suggestion
Resolving the Remote Endpoint is not a local operation.  It will involve a directory service, and can require communication with the Remote Endpoint to rendezvous and exchange peer addresses.  This can expose some or all of the candidate Local Endpoints to the Remote Endpoint.
```

draft-ietf-taps-impl.md
```diff
@@ -1150,7 +1150,7 @@ See {{fastopen}} for security considerations around racing with 0-RTT data.
...
 
-Since results from the network can determine how a connection attempt tree is built, such as when DNS returns a list of resolved endpoints, it is possible for the network to cause an implementation to consume significant on-device resources. Implementations should limit the maximum amount of state allowed for any given node, including the number of child nodes, especially when the state is based on results from the network.
+Since results from the network can determine how a connection attempt tree is built, such as when DNS returns a list of resolved Endpoint Identifers, it is possible for the network to cause an implementation to consume significant on-device resources. Implementations should limit the maximum amount of state allowed for any given node, including the number of child nodes, especially when the state is based on results from the network.
```
philsbln, line 1153:
```suggestion
Since results from the network can determine how a connection attempt tree is built, such as when DNS returns a list of resolved Endpoints, it is possible for the network to cause an implementation to consume significant on-device resources. Implementations should limit the maximum amount of state allowed for any given node, including the number of child nodes, especially when the state is based on results from the network.
```

philsbln (changes requested):
Thanks @mwelzl for doing this tedious work!
The PR does the right thing in most places, the only criticism I have is that I would prefer not to talk about derived/resolved endpoint identifiers as they are hidden anyway, so better talk about endpoints here.

draft-ietf-taps-impl.md
mwelzl:
```suggestion
How this resolution is done will depend on the type of the Remote Endpoint, and can also be specific to each Local Endpoint.  A common case is when the Remote Endpoint Identifer is a DNS name, in which case it is resolved to give a set of IPv4 and IPv6 addresses representing that name.  Some types of Remote Endpoint Identifers might require more complex resolution.  Resolving the Remote Endpoint for a peer-to-peer connection might involve communication with a rendezvous server, which in turn contacts the peer to gain consent to communicate and retrieve its set of candidate Local Endpoints, which are returned and form the candidate remote addresses for contacting that peer.
```

mwelzl: Agreed, but just to fix the plural

mwelzl:
Many thanks @philsbln !  I accepted all your suggestions. To nitpick, I'm a *little* bit unsure about this "resolving" thing, since one does resolve addresses, and I think in racing we do deal with the actual addresses and not more abstract entities? but maybe it's my misunderstanding... but I'm ok with it, I think it's understandable either way.

draft-ietf-taps-impl.md
```diff
@@ -492,15 +492,15 @@ However, this can cause a problem if a specific peer is not reachable over the n
...
 
 If the Selection Properties allow multiple protocols to be used for listening, and the implementation supports it, the Listener object should support receiving inbound connections for each eligible protocol on each eligible path.
 
 ### Implementing Listeners for Connected Protocols
 
-Connected protocols such as TCP and TLS-over-TCP have a strong mapping between the Local and Remote Endpoints (four-tuple) and their protocol connection state. These map into Connection objects. Whenever a new inbound handshake is being started, the Listener should generate a new Connection object and pass it to the application.
+Connected protocols such as TCP and TLS-over-TCP have a strong mapping between the Local and Remote Identifers (four-tuple) and their protocol connection state. These map into Connection objects. Whenever a new inbound handshake is being started, the Listener should generate a new Connection object and pass it to the application.
```
tfpauly, line 503:
```suggestion
Connected protocols such as TCP and TLS-over-TCP have a strong mapping between the Local and Remote Endpoint Identifers (four-tuple) and their protocol connection state. These map into Connection objects. Whenever a new inbound handshake is being started, the Listener should generate a new Connection object and pass it to the application.
```

draft-ietf-taps-impl.md
```diff
@@ -779,11 +779,11 @@ errors, the API will deliver them to the application as `SoftError` events. Thes
 
 ## Pooled Connection {#pooled-connections}
 
-For applications that do not need in-order delivery of Messages, the Transport Services implementation may distribute Messages of a single Connection across several underlying transport connections or multiple streams of multi-streaming connections between endpoints, as long as all of these satisfy the Selection Properties.
+For applications that do not need in-order delivery of Messages, the Transport Services implementation may distribute Messages of a single Connection across several underlying transport connections or multiple streams of multi-streaming connections between Endpoints, as long as all of these satisfy the Selection Properties.
```
tfpauly, line 782:
```suggestion
For applications that do not need in-order delivery of Messages, the Transport Services implementation may distribute Messages of a single Connection across several underlying transport connections or multiple streams of multi-streaming connections between endpoints, as long as all of these satisfy the Selection Properties.
```

draft-ietf-taps-impl.md
```diff
@@ -813,14 +813,14 @@ can be given. When an underlying transport connection supports multi-streaming (
 Beyond a single Connection's lifetime, it is useful for an implementation to keep state and history. This cached
 state can help improve future Connection establishment due to re-using results and credentials, and favoring paths and protocols that performed well in the past.
 
-Cached state may be associated with different endpoints for the same Connection, depending on the protocol generating the cached content.
-For example, session tickets for TLS are associated with specific endpoints, and thus should be cached based on a connection's
-hostname endpoint (if applicable). However, performance characteristics of a path are more likely tied to the IP address
+Cached state may be associated with different Endpoints for the same Connection, depending on the protocol generating the cached content.
+For example, session tickets for TLS are associated with specific Endpoints, and thus should be cached based on a connection's
```
tfpauly, line 817:
```suggestion
For example, session tickets for TLS are associated with specific endpoints, and thus should be cached based on a connection's
```

draft-ietf-taps-impl.md
```diff
@@ -813,14 +813,14 @@ can be given. When an underlying transport connection supports multi-streaming (
...
+hostname Endpoint Identifer (if applicable). However, performance characteristics of a path are more likely tied to the IP address
 and subnet being used.
 
 ## Protocol state caches
 
-Some protocols will have long-term state to be cached in association with endpoints. This state often has some time after which
+Some protocols will have long-term state to be cached in association with Endpoints. This state often has some time after which
```
tfpauly, line 823:
```suggestion
Some protocols will have long-term state to be cached in association with endpoints. This state often has some time after which
```

draft-ietf-taps-impl.md
```diff
@@ -865,8 +865,8 @@ For example, the mapping of the `Send` function for TCP applies to Connections i
 
 Each protocol has a notion of Connectedness. Possible values for Connectedness are:
 
-- Connectionless. Connectionless protocols do not establish explicit state between endpoints, and do not perform a handshake during Connection establishment.
-- Connected. Connected (also called "connection-oriented") protocols establish state between endpoints, and perform a handshake during connection establishment. The handshake may be 0-RTT to send data or resume a session, but bidirectional traffic is required to confirm connectedness.
+- Connectionless. Connectionless protocols do not establish explicit state between Endpoints, and do not perform a handshake during Connection establishment.
```
tfpauly, line 868:
```suggestion
- Connectionless. Connectionless protocols do not establish explicit state between endpoints, and do not perform a handshake during Connection establishment.
```

draft-ietf-taps-impl.md
```diff
@@ -1150,7 +1150,7 @@ See {{fastopen}} for security considerations around racing with 0-RTT data.
...
 
-Since results from the network can determine how a connection attempt tree is built, such as when DNS returns a list of resolved endpoints, it is possible for the network to cause an implementation to consume significant on-device resources. Implementations should limit the maximum amount of state allowed for any given node, including the number of child nodes, especially when the state is based on results from the network.
+Since results from the network can determine how a connection attempt tree is built, such as when DNS returns a list of resolved Endpoints, it is possible for the network to cause an implementation to consume significant on-device resources. Implementations should limit the maximum amount of state allowed for any given node, including the number of child nodes, especially when the state is based on results from the network.
```
tfpauly, line 1153:
```suggestion
Since results from the network can determine how a connection attempt tree is built, such as when DNS returns a list of resolved endpoints, it is possible for the network to cause an implementation to consume significant on-device resources. Implementations should limit the maximum amount of state allowed for any given node, including the number of child nodes, especially when the state is based on results from the network.
```

draft-ietf-taps-interface.md
```diff
@@ -123,7 +123,7 @@ of new protocols and protocol features without requiring changes to the
 applications. The specified API follows the Transport Services architecture
 by providing asynchronous, atomic transmission of messages. It is intended to replace the
 BSD sockets API as the common interface to the
-transport layer, in an environment where endpoints could select from
+transport layer, in an environment where Endpoints could select from
```
tfpauly, line 126:
```suggestion
transport layer, in an environment where endpoints could select from
```

draft-ietf-taps-interface.md
```diff
@@ -134,7 +134,7 @@ This document specifies an abstract application programming interface (API) that
 the high-level Transport Services architecture defined in
 {{!I-D.ietf-taps-arch}}. A Transport Services system supports
 asynchronous, atomic transmission of messages over transport protocols and
-network paths dynamically selected at runtime, in environments where an endpoint
+network paths dynamically selected at runtime, in environments where an Endpoint
```
tfpauly, line 137:
```suggestion
network paths dynamically selected at runtime, in environments where an endpoint
```

draft-ietf-taps-interface.md
```diff
@@ -405,7 +405,7 @@ TrustCallback := NewCallback({
 })
 SecurityParameters.SetTrustVerificationCallback(TrustCallback)
 
-// Specifying a local endpoint is optional when using Initiate
+// Specifying a local Endpoint is optional when using Initiate
```
tfpauly, line 408:
```suggestion
// Specifying a Local Endpoint is optional when using Initiate
```

draft-ietf-taps-interface.md
```diff
@@ -497,8 +497,8 @@ Connection -> Received<messageDataResponse, messageContext>
 // Trickle ICE, then add them to the Connection:
 Connection.AddRemote(NewRemoteCandidates)
 
-// On a PathChange<> events, resolve the local endpoint identifiers to
-// see if a new local endpoint has become available and, if
+// On a PathChange<> events, resolve the local Endpoint Identifiers to
```
tfpauly, line 500:
```suggestion
// On a PathChange<> events, resolve the Local Endpoint Identifiers to
```

draft-ietf-taps-interface.md
```diff
@@ -497,8 +497,8 @@ Connection -> Received<messageDataResponse, messageContext>
...
+// see if a new local Endpoint has become available and, if
```
tfpauly, line 501:
```suggestion
// see if a new Local Endpoint has become available and, if
```

draft-ietf-taps-interface.md
```diff
@@ -510,7 +510,7 @@ if ResolvedLocal has changed:
   // ResolvedLocal list to peer via signalling channel
   ...
 
-  // Add the new local endpoints to the Connection:
+  // Add the new local Endpoints to the Connection:
```
tfpauly, line 513:
```suggestion
  // Add the new Local Endpoints to the Connection:
```

draft-ietf-taps-interface.md
```diff
@@ -692,25 +692,25 @@ service.
...
 (see {{initiate}}) will internally resolve that name to a list of concrete
-endpoints. Specifying multiple Remote Endpoints on a Preconnection allows
+Endpoint Identifers. Specifying multiple Remote Endpoints on a Preconnection allows
 applications to override this for more detailed control.
 
 If Message Framers are used (see {{framing}}), they MUST be added to the
 Preconnection during pre-establishment.
 
-## Specifying Endpoints {#endpointspec}
+## Specifying Endpoints {#Endpointspec}
```
tfpauly, line 701:
```suggestion
## Specifying Endpoints {#endpointspec}
```

draft-ietf-taps-interface.md
```diff
@@ -692,25 +692,25 @@ service.
...
 
 The Transport Services API uses the Local Endpoint and Remote Endpoint objects
-to refer to the endpoints of a Connection. Endpoints can be created
+to refer to the Endpoints of a Connection. Endpoints can be created
```
tfpauly, line 704:
```suggestion
to refer to the endpoints of a Connection. Endpoints can be created
```

draft-ietf-taps-interface.md
```diff
@@ -753,15 +753,15 @@ LocalSpecifier.WithInterface("en0")
...
 For example, an Endpoint object cannot specify two IP addresses. Two separate IP addresses
 are represented as two Endpoint objects. If a Preconnection specifies a Remote
 Endpoint with a specific IP address set, it will only establish Connections to
-that IP address. If, on the other hand, a Remote Endpoint identifier specifies a hostname
+that IP address. If, on the other hand, a Remote Endpoint specifies a hostname
 but no addresses, the Connection can perform name resolution and attempt
 using any address derived from the original hostname of the Remote Endpoint.
 Note that multiple Remote Endpoints can be added to a Preconnection, as discussed
-in {{add-endpoints}}.
+in {{add-Endpoints}}.
```
tfpauly, line 764:
```suggestion
in {{add-endpoints}}.
```

draft-ietf-taps-interface.md
```diff
@@ -1046,7 +1046,7 @@ between multiple local interfaces that are connected to different access
 networks.
 
 When additional information (such as Provisioning Domain (PvD) information
-{{?RFC7556}}) is available about the networks over which an endpoint can operate,
+{{?RFC7556}}) is available about the networks over which an Endpoint can operate,
```
tfpauly, line 1049:
```suggestion
{{?RFC7556}}) is available about the networks over which an endpoint can operate,
```

draft-ietf-taps-interface.md
```diff
@@ -1338,7 +1338,7 @@ such as metered or unmetered network access. If an application needs to prohibit
 metered interfaces, this should be specified via Provisioning Domain attributes
 (see {{prop-pvd}}) or another specific property.
 
-Note that this property is not used to specify an interface scope zone for a particular endpoint. {{ifspec}} provides details about how to qualify endpoint candidates on a per-interface basis.
+Note that this property is not used to specify an interface scope zone for a particular Endpoint. {{ifspec}} provides details about how to qualify Endpoint candidates on a per-interface basis.
```
tfpauly, line 1341:
```suggestion
Note that this property is not used to specify an interface scope zone for a particular Endpoint. {{ifspec}} provides details about how to qualify endpoint candidates on a per-interface basis.
```

draft-ietf-taps-interface.md
```diff
@@ -1427,7 +1427,7 @@ Passive:
 : The Connection will support the use of multiple paths if the Remote Endpoint requests it.
 
 The policy for using multiple paths is specified using the separate `multipathPolicy` property, see {{multipath-policy}} below.
-To enable the peer endpoint to initiate additional paths towards a local address other than the one initially used, it is necessary to set the `advertisesAltaddr` property (see {{altaddr}} below).
+To enable the peer Endpoint to initiate additional paths towards a local address other than the one initially used, it is necessary to set the `advertisesAltaddr` property (see {{altaddr}} below).
```
tfpauly, line 1430:
```suggestion
To enable the peer endpoint to initiate additional paths towards a local address other than the one initially used, it is necessary to set the `advertisesAltaddr` property (see {{altaddr}} below).
```

draft-ietf-taps-interface.md
```diff
@@ -1447,7 +1447,7 @@ Default:
 : False
 
 This property specifies whether alternative addresses, e.g., of other interfaces, should be advertised to the
-peer endpoint by the Protocol Stack. Advertising these addresses enables the peer-endpoint to establish additional connectivity, e.g., for Connection migration or using multiple paths.
+peer Endpoint by the Protocol Stack. Advertising these addresses enables the peer Endpoint to establish additional connectivity, e.g., for Connection migration or using multiple paths.
```
tfpauly, line 1450:
```suggestion
peer endpoint by the Protocol Stack. Advertising these addresses enables the peer endpoint to establish additional connectivity, e.g., for Connection migration or using multiple paths.
```

draft-ietf-taps-interface.md
```diff
@@ -1585,8 +1585,8 @@ SecurityParameters.Set(cached-session-lifetime-seconds, 3600)
 ~~~
 
 Connections that use Transport Services SHOULD use security in general. However, for
-compatibility with endpoints that do not support transport security protocols (such
-as a TCP endpoint that does not support TLS), applications can initialize their
+compatibility with Endpoints that do not support transport security protocols (such
+as a TCP Endpoint that does not support TLS), applications can initialize their
```
tfpauly, line 1589:
```suggestion
as a TCP endpoint that does not support TLS), applications can initialize their
```

draft-ietf-taps-interface.md
```diff
@@ -1690,7 +1690,7 @@ Connection -> EstablishmentError<reason?>
 An `EstablishmentError` occurs either when the set of transport properties and security
 parameters cannot be fulfilled on a Connection for initiation (e.g., the set of
 available Paths and/or Protocol Stacks meeting the constraints is empty) or
-reconciled with the Local and/or Remote Endpoints; when the remote endpoint specifier
+reconciled with the Local and/or Remote Endpoints; when a remote Endpoint
```
tfpauly, line 1693:
```suggestion
reconciled with the Local and/or Remote Endpoints; when a remote Endpoint Identifier
```

draft-ietf-taps-interface.md
```diff
@@ -1783,10 +1783,10 @@ Endpoint candidates to the Remote Endpoint candidates.
 
 If there are multiple Local Endpoints or Remote Endpoints configured, then
 initiating a `Rendezvous` action will systematically probe the reachability
-of those endpoint candidates following an approach such as that used in
+of those Endpoint candidates following an approach such as that used in
```
tfpauly, line 1786:
```suggestion
of those endpoint candidates following an approach such as that used in
```

draft-ietf-taps-interface.md
```diff
@@ -1933,11 +1933,11 @@ the `connScheduler` property ({{conn-scheduler}}).
 See {{priority-in-taps}} for more.
 
 
-## Adding and Removing Endpoints on a Connection {#add-endpoints}
+## Adding and Removing Endpoints on a Connection {#add-Endpoints}
```
tfpauly, line 1936:
```suggestion
## Adding and Removing Endpoints on a Connection {#add-endpoints}
```

draft-ietf-taps-interface.md
```diff
@@ -1933,11 +1933,11 @@ the `connScheduler` property ({{conn-scheduler}}).
...
 
 Transport protocols that are explicitly multipath aware are expected to automatically
 manage the set of Remote Endpoints that they are communicating with, and the paths to
-those endpoints. A `PathChange<>` event, described in {{conn-path-change}}, will be
+those Endpoints. A `PathChange<>` event, described in {{conn-path-change}}, will be
```
tfpauly, line 1940:
```suggestion
those endpoints. A `PathChange<>` event, described in {{conn-path-change}}, will be
```

draft-ietf-taps-interface.md
```diff
@@ -1971,7 +1971,7 @@ switch to a new path if other reachable Remote Endpoints exist, or the
 connection might abort.
 
 Similarly, the `AddLocal` and `RemoveLocal` actions can be used to add
-and remove local endpoints to/from a Connection.
+and remove local Endpoints to/from a Connection.
```
tfpauly, line 1974:
```suggestion
and remove Local Endpoints to/from a Connection.
```

draft-ietf-taps-interface.md
```diff
@@ -2197,7 +2197,7 @@ The following values are valid for the capacity profile:
   and efficient use of the available capacity when sending on this Connection. This can be
   used by the system to disable the coalescing of multiple small Messages into
   larger packets (Nagle's algorithm); to prefer immediate acknowledgment from
-  the peer endpoint when supported by the underlying transport; and so on.
+  the peer Endpoint when supported by the underlying transport; and so on.
```
tfpauly, line 2200:
```suggestion
  the peer endpoint when supported by the underlying transport; and so on.
```

draft-ietf-taps-interface.md
```diff
@@ -2440,7 +2440,7 @@ Connection -> SoftError<>
 
 This event notifies the application when at least one of the paths underlying a Connection has changed. Changes occur
 on a single path when the PMTU changes as well as when multiple paths are used
-and paths are added or removed, the set of local endpoints changes, or a handover has been performed.
+and paths are added or removed, the set of local Endpoints changes, or a handover has been performed.
```
tfpauly, line 2443:
```suggestion
and paths are added or removed, the set of local endpoints changes, or a handover has been performed.
```

draft-ietf-taps-interface.md
```diff
@@ -2866,7 +2866,7 @@ to not provide segmentation of Messages larger than the
 maximum size permitted by the network layer, and also
 to avoid network-layer source fragmentation of Messages.
 When running over IPv4, setting this property to
-true will result in a sending endpoint setting the
+true will result in a sending Endpoint setting the
```
tfpauly, line 2869:
```suggestion
true will result in a sending endpoint setting the
```

draft-ietf-taps-interface.md
```diff
@@ -3275,7 +3275,7 @@ to the `final` property that may be marked on a sent Message, see {{msg-final}}.
 
 Some transport protocols and peers do not support signaling of the `final` property.
 Applications therefore should not rely on receiving a Message marked Final to know
-that the sending endpoint is done sending on a Connection.
+that the sending Endpoint is done sending on a Connection.
```
tfpauly, line 3278:
```suggestion
that the sending endpoint is done sending on a Connection.
```

draft-ietf-taps-interface.md
```diff
@@ -3462,7 +3462,7 @@ This is the case, for example, when the Transport Services system also executes
 TURN or ICE are used to establish connectivity, if protocols or paths are raced, or if a path fails and
 fallback or re-establishment is supported in the Transport Services system. Applications should take special
 care when using 0-RTT session resumption (see {{prop-0rtt}}), as early data sent across multiple paths during
-connection establishment may reveal information that can be used to correlate endpoints on these paths.
+connection establishment may reveal information that can be used to correlate Endpoints on these paths.
```
tfpauly, line 3465:
```suggestion
connection establishment may reveal information that can be used to correlate endpoints on these paths.
```

tfpauly (changes requested):

mwelzl:
Many thanks @tfpauly for fixing all my exaggerated capitalizations!   I approved all your suggestions.

All reviewers - are we good now???  Has this converged?

tfpauly (approved):

Issue #1352, opened 2023-09-15
Clarify rcv backpressure


mwelzl:
From the review by Lars Eggert, COMMENT part:

### Section 9.3.2, paragraph 1
```
    Each call to Receive will be paired with a single Receive event,
    which can be a success or an error.  This allows an application to
    provide backpressure to the transport stack when it is temporarily
    not ready to receive Messages.
```
Isn't the act of not reading and the resulting buffer growth already
providing this backpressure?

mwelzl:
Does this mean: as long as an event is not "consumed" (the application doesn't read anything when it occurs), no further receive events should fire?

To me, this is an interesting idea. Indeed it should make things simpler... why would an application not want to consume receive events in sequence? Even when accepting out-of-order data, or streams, the application needs to first receive messages in order to understand more about them (read sequence numbers or whatever).

mwelzl:
Thinking more, this would require us to specify deeper how the Receive event is supposed to work. Data are just handed over - e.g. in NEATPy, a function is called with a message object as a parameter; what the application then does with that parameter is its own business. "Freeing the buffer" is ... garbage collection?  and not visible to the TAPS system anymore.

This is somewhat similar to handing over a pointer to a memory block in C, and saying "this block is your problem now, here's the pointer and here's the length, better free it later". All such implementations are possible, I suppose... and if this is how it's done, then the buffer as seen by the TAPS system is always empty immediately after the callback is done, and Lars' back-pressure idea doesn't really work out that way.

mwelzl:
.... and hence, I suggest to close without action. Thoughts?

PR #1387, opened 2023-09-27
Wording of receiver backpressure.

### LINKED ISSUE #1283: Make receive backpressure clearer

p 70, sec 9.3.2.  Receive Events

```
  Each call to Receive will be paired with a single Receive event,
  which can be a success or an error.  This allows an application to
  provide backpressure to the transport stack when it is temporarily
  not ready to receive Messages.
```

The backpressure aspect wasn't entirely clear to me.  So, if an application can
handle receiving multiple receive events at the same time, it can make multiple
calls to receive without waiting for, or processing, any receive events?  Is
that how the application controls the backpressure?

Also:

How does the transport stack know that the application has finished with the
memory holding the message in the receive event, or is the expectation that the
receive message contents is logically always allocated on the heap and it is
responsibility of the receiver to free the memory once it is done with it?

mwelzl:
Closes #1283.
May become unnecessary though, depending on what we think of #1352.

### DIFF
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -3133,9 +3133,10 @@ as a receiver preference for Message reordering.
 
 ### Receive Events {#receive-events}
 
-Each call to `Receive` will be paired with a single `Receive` event, which can be a success
-or an error. This allows an application to provide backpressure to the transport stack
-when it is temporarily not ready to receive Messages.
+Each call to `Receive` will be paired with a single `Receive` event. This allows an application
+to provide backpressure to the transport stack when it is temporarily not ready to receive Messages.
+For example, an application that will later be able to handle multiple receive events at the same time
+can make multiple calls to `Receive` without waiting for, or processing, any receive events.
 
 The Transport Services API should allow the application to correlate which call to `Receive` resulted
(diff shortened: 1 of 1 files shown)


gorryfair:
I agree with the NEW text - HOWEVER, I am not sure this describes back pressure yet. I suspect we ought to talk also about the opposite case, e.g.,:

"An application that is temporarily unable to process received events for a connection, could leave these pending actions in the queue . This leadis to a build-up of unread data, which, in turn, could result in back pressure on the sender by the transport protocol (e.g., restricting transmission of new data by the sender using the TCP rand or QUIC flow credit mechanisms).

mwelzl:
So this is about a more local form of backpressure - as the text says, the application can give backpressure to the stack below, to limit the number of events that it will see. In my opinion, that's already clear enough?

As for bringing in flow control (for protocols that do it) - In my opinion, discussing this here just makes it more complicated?

gorryfair:
I would prefer if you hinted at what "provide backpressure to the transport stack" means to the transport services implementation, although quite agree if one tries to give protocol examples, it's too much....

gorryfair (approved):
Seems good enough to me, we should merge this!

tfpauly (approved):

PR #1388, opened 2023-09-27
s/Collection/Set

### LINKED ISSUE #1280: s/Collection/Set or Bag?

p 6, sec 1.1.  Terminology and Notation

  *  Array: Denoted []Type, an instance takes a value for each of zero
     or more elements in a sequence of the given Type.  An array may be
     of fixed or variable length.
  *  Collection: An unordered grouping of one or more values of the
     same type.

Perhaps calling it a Set may be better than Collection (if duplicates are not
allowed) or a Bag (if duplicates are allowed).

mwelzl:
Closes #1280.


gorryfair (approved):
OK: I prefer this to the other proposa.

gorryfair:
"set" is the term Iused in arch anyway.

PR #1390, opened 2023-09-29
A statement about supporting monitoring

### LINKED ISSUE #1279: Add a (stronger?) statement about monitoring?

As per the architecture doc, I think that it is great that you are defining
a new transport API.  I note that this API doesn't really include any standard
APIs or structures to monitor the state of the transport sessions for a given
application (i.e., API user).  E.g., how many connections are currently open,
total number of connections (since library was initialized), number of errored
transport connections, drops, mtu issues, flow rates, etc.  I think that with
some of the changes to the Internet architecture (e.g., QUIC to cite one
obvious example), it reduces the ability for network operators to monitor and
debug network issues between applications.  A potential corollary of this is
that a lot more debug and diagnostics information will need to be made
available to applications in a common way to allow application support staff,
and users of those applications to better understand where in the network
issues and failures are happening.  It would seem unreasonable for me to hold a
discuss on this document for what might be a lot of work and discussion that
could take a long time to resolve but I hope that the authors and WG will
consider whether there is further useful future work required in additional
RFCs.

mwelzl:
Closes #1279

### DIFF
--- a/draft-ietf-taps-arch.md
+++ b/draft-ietf-taps-arch.md
@@ -337,6 +337,15 @@ It is important to note that neither the Transport Services API {{?I-D.ietf-taps
 
 A Transport Services system makes decisions that select protocols and interfaces. In normal use, a given version of a Transport Services system SHOULD result in consistent protocol and interface selection decisions for the same network conditions given the same set of Properties. This is intended to provide predictable outcomes to the application using the API.
 
+## Support Monitoring
+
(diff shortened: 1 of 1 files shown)


gorryfair (approved):
Benign, but in the correct direction and likely sufficient, so think this should be merged.

gorryfair (approved):
Seems very good.

PR #1394, opened 2023-10-02
Make it clearer that Messages can be spread across multiple packets

### LINKED ISSUE #1292: Message can be spread across multiple packets
From the review by John Scudder:

### Section 4.1.5, payload of IP packet... or packets?

The first bullet has "Messages are sent in the payload of IP packet. One packet
can carry one or more Messages or parts of a Message." The first sentence
either needs an indefinite article on "IP packet" or for "payload" and "packet"
to be plural. I think probably you mean "payloads of IP packets" since in
general (as the second sentence shows) a message might be split across two or
more IP packets.

Also, is it even necessary to say that messages are IP payloads? What else
would they be?

mwelzl:
... with a rather minimalistic change.

Closes #1292


gorryfair (approved):
Seems Ok

PR #1397, opened 2023-10-02
Explain errors for message properties

### LINKED ISSUE #1336: Contradicting Connection / Message properties

### Section 9.1.3, paragraph 9
```
    If a Message Property contradicts a Connection Property, and if this
    per-Message behavior can be supported, it overrides the Connection
    Property for the specific Message.  For example, if reliability is
    set to Require and a protocol with configurable per-Message
    reliability is used, setting msgReliable to false for a particular
    Message will allow this Message to be sent without any reliability
    guarantees.  Changing the msgReliable Message Property is only
    possible for Connections that were established enabling the Selection
    Property perMsgReliability.
```
How is this going to work in the opposite case, i.e., if an unreliable
connection was set up and then some messages are passed in which
require reliable transmission? At connection open, the stack may have
chosen a transport protocol that cannot support reliable
transmission? (Also, I think this general principle of message >
connection has other issues for other properties.)

tfpauly:
Closes #1336

### DIFF
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -2647,7 +2647,10 @@ with configurable per-Message reliability is used, setting
 allow this Message to be sent without any reliability guarantees. Changing
 the `msgReliable` Message Property is only possible for
 Connections that were established enabling the Selection Property
-`perMsgReliability`.
+`perMsgReliability`. If the contradicting Message Property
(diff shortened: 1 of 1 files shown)


gorryfair (approved):
Seems like what we discussed

mwelzl (approved):

PR #1398, opened 2023-10-02
Hostnames SHOULD be FQDNs

### LINKED ISSUE #1360: Hostname vs FQDN

Can WithHostname be unqualified? (God I hope not!)
If not, can the call WithHostname be renamed to WithFQDN ?
If it can be unqualified, how does one deal with credentials
and identifiers, eg with a 'search domain' containing multiple
domains, RemoteSpecifier.WithHostname("mail") could end up on
either [mail.example.com](http://mail.example.com/) or [mail.example.net](http://mail.example.net/) (or heaven forbid,
the .mail TLD)

tfpauly:
Closes #1360


mwelzl (approved):
LGTM, thanks

PR #1399, opened 2023-10-02
Add ALPN parameter

### LINKED ISSUE #1362: ALPN

I don't see a WithALPN method for setting the ALPN.

tfpauly:
Closes #1362

### DIFF
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -1550,6 +1550,13 @@ SecurityParameters := NewSecurityParameters()
 
 Security configuration parameters and sample usage follow:
 
+- Application-layer protocol negotiation (ALPN) values: Used to indicate which application-layer protocols
+are negotiated by the security protocol layer. See {{!ALPN=RFC7301}} for definition of the ALPN field.
+
+~~~
+SecurityParameters.Set(alpn, "h2")
+~~~
+
 - Local identity, certificates, and private keys: Used to perform private key operations and prove one's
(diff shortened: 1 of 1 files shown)


philsbln:
Hi @tfpauly … I agree that this is the most obvious solution, but on a more abstract level, wouldn't it be advisable to make the ALPN part of the remote endpoint?

draft-ietf-taps-interface.md
```diff
@@ -1550,6 +1550,13 @@ SecurityParameters := NewSecurityParameters()
 
 Security configuration parameters and sample usage follow:
 
+- Application-layer protocol negotiation (ALPN) values: Used to indicate which application-layer protocols
```
mwelzl, line 1553:
```suggestion
- Application-layer protocol negotiation (ALPN) values: used to indicate which application-layer protocols
```

draft-ietf-taps-interface.md
```diff
@@ -1550,6 +1550,13 @@ SecurityParameters := NewSecurityParameters()
...
+are negotiated by the security protocol layer. See {{!ALPN=RFC7301}} for definition of the ALPN field.
```
tfpauly, line 1554:
```suggestion
are negotiated by the security protocol layer. See {{!ALPN=RFC7301}} for definition of the ALPN field. Note that the Transport Services System can provide ALPN values automatically, based on
the protocols being used, if not explicitly specified by the application.
```

PR #1401, opened 2023-10-05
Make UTO more clearly a TCP example

### LINKED ISSUE #1303: UTO is not TCP-specific
From the review by Martin Duke:

(S8.2) UTO is not TCP-specific! QUIC does it too. Maybe this should be a
generic property?

gorryfair:
We provide UTO here as an example of something that is offered by a specific transport. It is true that QUIC also includes a UTO - but we don't have a QUIC spec yet, and anyway other protocols may also define transport-specific options.  I suggest this is sufficiently niche that the example prevails - when you tear doen a QUIC connection might be different to a TCP connection, because of differences in transport. I therefore propose some minor teaks and think we ought not to say about QUIC here.
Will close #1303 if merged


mwelzl (approved):

mwelzl:
See #1303 for discussion

gorryfair:
Let's merge this.

PR #1402, opened 2023-10-10
Property names aren't case sensitive

### LINKED ISSUE #1338: Property names

### Section 4.1, paragraph 1
```
    Transport Properties are referred to by property names.  For the
    purposes of this document, these names are alphanumeric strings in
    which the following characters are allowed: lowercase letters a-z,
    uppercase letters A-Z, digits 0-9, the hyphen -, and the underscore
    _. These names serve two purposes:
```
This allows property names that are indistinguishable from integers.
Is that a bug or a feature? Also, are they case-sensitive?

mwelzl:
Closes #1338

### DIFF
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -565,7 +565,7 @@ the respective protocol has been selected.
 
 Transport Properties are referred to by property names. These names are alphanumeric strings in which the following
 characters are allowed: lowercase letters `a-z`, uppercase letters `A-Z`,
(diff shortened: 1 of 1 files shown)


mwelzl:
Note, before merging and therewith closing #1338, let's make sure we agree about the discussion in #1338.

tfpauly (approved):

gorryfair (approved):
This is what I understood!

philsbln (approved):
Good to make that explicit

PR #1404, opened 2023-10-11
Make Transport Property Names more strict (and case-insenstitive)

### LINKED ISSUE #1338: Property names

philsbln:
alternative to #1402
closes #1338

### DIFF
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -563,11 +563,7 @@ the respective protocol has been selected.
 
 ## Transport Property Names {#property-names}
 
-Transport Properties are referred to by property names. These names are
-case-insensitive alphanumeric strings in which the following characters are allowed:
-letters `a-z` / `A-Z`, digits `0-9`, the hyphen `-`.
-The underscore `_` is only allowed at the first position (with the exception of transposing into snake_case, see below).
-They must start with a letter or an underscore. These names serve two purposes:
+Transport Properties are referred to by property names, represented as case-insensitive strings. These names serve two purposes:
 
 - Allowing different components of a Transport Services implementation to pass Transport
   Properties, e.g., between a language frontend and a policy manager,
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -563,9 +563,11 @@ the respective protocol has been selected.
 
 ## Transport Property Names {#property-names}
 
-Transport Properties are referred to by property names. These names are alphanumeric strings in which the following
-characters are allowed: lowercase letters `a-z`, uppercase letters `A-Z`,
-digits `0-9`, the hyphen `-`, and the underscore `_`. These names serve two purposes:
+Transport Properties are referred to by property names. These names are 
+case-insensitive alphanumeric strings in which the following characters are allowed:
+letters `a-z` / `A-Z`, digits `0-9`, the hyphen `-`. 
+The underscore `_` is only allowed at the first position. 
+They must start with a letter or an underscore. These names serve two purposes:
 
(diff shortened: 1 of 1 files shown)


draft-ietf-taps-interface.md
```diff
@@ -563,9 +563,11 @@ the respective protocol has been selected.
 
 ## Transport Property Names {#property-names}
 
-Transport Properties are referred to by property names. These names are alphanumeric strings in which the following
-characters are allowed: lowercase letters `a-z`, uppercase letters `A-Z`,
-digits `0-9`, the hyphen `-`, and the underscore `_`. These names serve two purposes:
+Transport Properties are referred to by property names. These names are
+case-insensitive alphanumeric strings in which the following characters are allowed:
+letters `a-z` / `A-Z`, digits `0-9`, the hyphen `-`.
+The underscore `_` is only allowed at the first position.
```
mwelzl, line 569:
```suggestion
```

I don't understand the point of this rule, and it even contradicts what you say below about snake_case. The next sentence, "They must start with a letter or an underscore" seems good enough.

draft-ietf-taps-interface.md
```diff
@@ -595,6 +597,8 @@ for Protocol-specific Properties and MUST NOT be used for vendor or implementati
 Terms listed as keywords as in the protocol numbers registry SHOULD be avoided as any part of a vendor- or
 implementation-specific property name.
 
+Though Transport Property Names are case-insensitive, it is recommended to use camelCase to improve readability.
+Implementations may transpose Transport Property Names into snake_case or PascalCase to blend into the language environment.
```
mwelzl, line 601: I note that this is pretty strict... we don't know how styles of future languages will be?  But I like it nevertheless. If we allow "any" kind of transposition, we make porting applications harder.

mwelzl (changes requested):
One suggestion

draft-ietf-taps-interface.md
mwelzl: Aha, I see...
```suggestion
The underscore `_` is only allowed at the first position (with the exception of transposing into snake_case, see below).
```

mwelzl (approved):

draft-ietf-taps-interface.md
```diff
@@ -563,9 +563,11 @@ the respective protocol has been selected.
 
 ## Transport Property Names {#property-names}
 
-Transport Properties are referred to by property names. These names are alphanumeric strings in which the following
-characters are allowed: lowercase letters `a-z`, uppercase letters `A-Z`,
-digits `0-9`, the hyphen `-`, and the underscore `_`. These names serve two purposes:
+Transport Properties are referred to by property names. These names are
+case-insensitive alphanumeric strings in which the following characters are allowed:
+letters `a-z` / `A-Z`, digits `0-9`, the hyphen `-`.
```
mwelzl, line 568:
```suggestion
letters `a-z` / `A-Z`, digits `0-9`, the hyphen `-` and the underscore `_`.
```

tfpauly (changes requested):

gorryfair (approved):
I like this line of approach - subject to addressing current review comments

mwelzl:
I think this minimalistic and much less strict angle reflects what was agreed on at the interim. About time to merge it, I believe.

PR #1406, opened 2023-10-11
Clarify rate bounds

### LINKED ISSUE #1347: Bounds on send and receive rate

### Section 8.1.8, paragraph 4
```
    Numeric values of these properties specify an upper-bound rate that a
    transfer is not expected to exceed (even if flow control and
    congestion control allow higher rates), and/or a lower-bound rate
    below which the application does not deem it will be useful.  These
    are specified in bits per second.  The enumerated value Unlimited
    indicates that no bound is specified.
```
Over what timescale, or is this supposed to apply to instantaneous
bandwidth? Because an app can control how much data over time to feed
into a connection, but cannot usually know what send rate that
results in.

mwelzl:
Closes #1347

### DIFF
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -2281,9 +2281,12 @@ Default:
 
 Numeric values of these properties specify an upper-bound rate that a transfer is not expected to
 exceed (even if flow control and congestion control allow higher rates), and/or a
-lower-bound rate below which the application does not deem
-it will be useful. These are specified in bits per second.
-The enumerated value `Unlimited` indicates that no bound is specified.
+lower-bound application-layer rate below which the application does not deem
+it will be useful. These rate values are meant at the application layer, i.e. not considering header overheads
+from protocols used by the Transport Services system. The values are specified in bits per second,
+and assumed to be measured over one-second time intervals. E.g., specifying a maxSendRate of X bits per second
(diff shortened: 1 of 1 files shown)


draft-ietf-taps-interface.md
```diff
@@ -2281,9 +2281,12 @@ Default:
 
 Numeric values of these properties specify an upper-bound rate that a transfer is not expected to
 exceed (even if flow control and congestion control allow higher rates), and/or a
-lower-bound rate below which the application does not deem
-it will be useful. These are specified in bits per second.
-The enumerated value `Unlimited` indicates that no bound is specified.
+lower-bound application-layer rate below which the application does not deem
+it will be useful. These rate values are meant at the application layer, i.e. not considering header overheads
+from protocols used by the Transport Services system. The values are specified in bits per second,
```
tfpauly, line 2286:
```suggestion
it will be useful. These rate values are measured at the application layer, i.e. not considering header overheads
from protocols used by the Transport Services system. The values are specified in bits per second,
```

tfpauly (approved):

draft-ietf-taps-interface.md
gorryfair: Added a small delta to the suggestion from Tommy

gorryfair (approved):
Heading in the correct direction, but added a small delta to TP's suggestion.

PR #1409, opened 2023-10-11
Replacing "if then" type clauses with tuples

### LINKED ISSUE #1346: How to express "Full Coverage"

### Section 8.1.1, paragraph 3
```
    Type:  Integer (non-negative) or Full Coverage

    Default:  Full Coverage
```
How is "Full Coverage" expressed?

mwelzl:
Closes #1346

This also corrects usage of true without apostrophes, and adjusts the text of interface and pvd to talk about "sets of tuples" instead of just a tuple  (their type *is* a set, but the text wasn't in line with that)

### DIFF
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -1323,8 +1323,8 @@ or categories of interfaces it wants to `Require`, `Prohibit`, `Prefer`, or
 selection to that single interface, and often leads to less flexible and resilient
 connection establishment.
 
-In contrast to other Selection Properties, this property is a tuple of an
-(Enumerated) interface identifier and a preference, and can either be
+In contrast to other Selection Properties, this property is a set of
+(Enumerated) interface identifier and preference tuples, and can either be
 implemented directly as such, or for making one preference available for each
 interface and interface type available on the system.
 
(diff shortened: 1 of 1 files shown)


draft-ietf-taps-interface.md
```diff
@@ -1323,8 +1323,8 @@ or categories of interfaces it wants to `Require`, `Prohibit`, `Prefer`, or
 selection to that single interface, and often leads to less flexible and resilient
 connection establishment.
 
-In contrast to other Selection Properties, this property is a tuple of an
-(Enumerated) interface identifier and a preference, and can either be
+In contrast to other Selection Properties, this property is a set of
+(Enumerated) interface identifier and preference tuples, and can either be
```
philsbln, line 1327: This took me three times reading it… 
```suggestion
In contrast to other Selection Properties, this property is a set of tuples,
where each tuple consists if an (Enumerated) interface identifier and preference tuples.
It and can either be
```

draft-ietf-taps-interface.md
```diff
@@ -1367,8 +1367,8 @@ Provisioning Domain (PvD) or categories of PVDs it wants to
 consistent sets of network properties that might be more specific than network
 interfaces {{?RFC7556}}.
 
-As with interface instances and types, this property is a tuple of an (Enumerated)
-PvD identifier and a preference, and can either be implemented directly as such,
+As with interface instances and types, this property is a set of (Enumerated)
+PvD identifier and a preference tuples, and can either be implemented directly as such,
```
philsbln, line 1371:
```suggestion
As with interface instances and types, this property is a set of tuples of (Enumerated)
PvD identifier and a preference, and can either be implemented directly as such,
```

mwelzl:
Alternative proposed in #1410  (I personally also like #1410 more)

draft-ietf-taps-interface.md
mwelzl: Thanks for this suggestion. I agree that this is better - I'll do it in a separate PR.

PR #1410, opened 2023-10-12
Dealing with "if then" type clauses, try 2  (alternative to #1409)

### LINKED ISSUE #1346: How to express "Full Coverage"

mwelzl:
Closes #1346


draft-ietf-taps-interface.md
```diff
@@ -2086,7 +2086,9 @@ to the application. The application is responsible for handling
 any corruption within the non-protected part of the Message {{?RFC8085}}.
 A special value of 0 means that a received packet might also have a zero checksum field,
 and the enumerated value `Full Coverage` means
-that the entire Message needs to be protected by a checksum.
+that the entire Message needs to be protected by a checksum. A Transport Services API
+could express `Full Coverage` in various ways -- e.g., via an additional Boolean
+parameter of a function call.
```
philsbln, line 2091: Still sounds a little heavy handed…
```suggestion
that the entire Message needs to be protected by a checksum. An implementation
is supposed to express `Full Coverage` in an environment-typical way, e.g., as a Union type or special value.
```

philsbln (changes requested):
I still don't like the additional Boolean-Parameter as is looks heavy handed and does not allow for multiple special values…

draft-ietf-taps-interface.md
```diff
@@ -2124,7 +2125,9 @@ Default:
 If this property is Numeric, it specifies how long to wait before deciding that an active Connection has
 failed when trying to reliably deliver data to the Remote Endpoint. Adjusting this property
 will only take effect when the underlying stack supports reliability. If this property has the enumerated
-value `Disabled`, it means that no timeout is scheduled.
+value `Disabled`, it means that no timeout is scheduled. A Transport Services API
+could express `Disabled` in various ways -- e.g., via an additional Boolean
+parameter of a function call.
```
philsbln, line 2130:
```suggestion
could express `Disabled` in an environment-typical way, e.g., as a Union type or special value.
```

draft-ietf-taps-interface.md
```diff
@@ -2144,7 +2147,9 @@ the Local Endpoint sends a keep-alive packet to the Remote Endpoint. Adjusting t
 will only take effect when the underlying stack supports sending keep-alive packets.
 Guidance on setting this value for connection-less transports is
 provided in {{?RFC8085}}.
-A value greater than the Connection timeout ({{conn-timeout}}) or the enumerated value `Disabled` will disable the sending of keep-alive packets.
+A value greater than the Connection timeout ({{conn-timeout}}) or the enumerated value `Disabled` will disable the sending of keep-alive packets. A Transport Services API
+could express `Disabled` in various ways -- e.g., via an additional Boolean
+parameter of a function call.
```
philsbln, line 2152:
```suggestion
could express `Disabled` in an environment-typical way, e.g., as a Union type or special value.
```

draft-ietf-taps-interface.md
```diff
@@ -2284,6 +2289,8 @@ exceed (even if flow control and congestion control allow higher rates), and/or
 lower-bound rate below which the application does not deem
 it will be useful. These are specified in bits per second.
 The enumerated value `Unlimited` indicates that no bound is specified.
+A Transport Services API could express `Unlimited` in various ways -- e.g.,
+via an additional Boolean parameter of a function call.
```
philsbln, line 2293:
```suggestion
A Transport Services API could express `Unlimited` in an environment-typical way, e.g., as a Union type or special value.
```

draft-ietf-taps-interface.md
```diff
@@ -2300,7 +2307,9 @@ If this property is Numeric, it controls the number of Connections that can be a
 a peer as new members of the Connection's group. Similar to `SetNewConnectionLimit`,
 this limits the number of `ConnectionReceived` events that will occur, but constrained
 to the group of the Connection associated with this property. For a multi-streaming transport,
-this limits the number of allowed streams.
+this limits the number of allowed streams.  A Transport Services API
+could express `Unlimited` in various ways -- e.g., via an additional Boolean
+parameter of a function call.
```
philsbln, line 2312:
```suggestion
could express `Unlimited` in an environment-typical way, e.g., as a Union type or special value.
```

draft-ietf-taps-interface.md
```diff
@@ -2344,7 +2353,9 @@ based on the Maximum Packet Size (MPS). The value of this property can change ov
 This value allows a sending stack to avoid unwanted fragmentation at the
 network-layer or segmentation by the transport layer before
 choosing the message size and/or after a `SendError` occurs indicating
-an attempt to send a Message that is too large.
+an attempt to send a Message that is too large.  A Transport Services API
+could express `Not applicable` in various ways -- e.g., by letting a querying
+function call fail.
```
philsbln, line 2358:
```suggestion
could express `Not applicable` in an environment-typical way, e.g., as a Union type or special value.
```

philsbln (changes requested):

draft-ietf-taps-interface.md
mwelzl: Wait wait, this one is different!  It's a read-only parameter, where "not applicable" is returned... do you want to say "Union type or special value" here?  Ok with me but just checking... I thought a failing query function is easy enough?

philsbln: you are right, it's different. But the proposed variant "letting a querying function call fail." is also no really good solution. I also prefer as a Union type or special value here.

tfpauly (approved):

PR #1419, opened 2023-10-12
Prioritization: don't talk about "performance"

### LINKED ISSUE #1351: What does worse performance mean?

### Section 9.2.6, paragraph 2
```
    A Transport Services system gives no guarantees about how its
    expression of relative priorities will be realized.  However, the
    Transport Services system will seek to ensure that performance of
    relatively-prioritized Connections and Messages is not worse with
    respect to those Connections and Messages than an equivalent
    configuration in which all prioritization properties are left at
    their defaults.
```
Without defining what is meant by "performance" in this case, it's not
possible to know what "not worse" means. (Or how a transport system
might seek to ensure that.)

mwelzl:
Closes #1351


gorryfair (approved):
Fine.

tfpauly (approved):

PR #1420, opened 2023-10-12
"Initial Message will be Safely Replayable" instead of "0-RTT ..."

### LINKED ISSUE #1345: 0-RTT property name

### Section 6.2.5, paragraph 0
```
 6.2.5.  Use 0-RTT Session Establishment with a Safely Replayable Message
```
This is an awfully specific preference compared to all the others.
Maybe make it a bit less specific by calling it "Initial Message will
be Safely Replayable" without mentioning a specific use case such as
0-RTT? (Or maybe even define a "replayable" message flag, and only do
0-RTT if that is set on the first message.)

mwelzl:
Closes #1345

### DIFF
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -1205,7 +1205,7 @@ Default:
 This property specifies whether the application wishes to use a transport
 protocol that can ensure that data is received by the application at the Remote Endpoint in the same order as it was sent.
 
-### Use 0-RTT Session Establishment with a Safely Replayable Message {#prop-0rtt}
+### Initial Message will be Safely Replayable {#prop-0rtt}
 
 Name:
 : zeroRttMsg
@@ -1217,8 +1217,8 @@ Default:
 : Ignore
 
 This property specifies whether an application would like to supply a Message to
-the transport protocol before connection establishment that will then be
-reliably transferred to the other side before or during connection
+the transport protocol before Connection establishment that will then be
+reliably transferred to the other side before or during Connection
 establishment. This Message can potentially be received multiple times (i.e.,
 multiple copies of the Message data could be passed to the Remote Endpoint).
 See also {{msg-safelyreplayable}}.


philsbln:
I think it gets a weird twist with this PR … If it was a Boolean "Initial message is safely replayable", it would fit perfectly, but the preference we gave here does two things at once:

- Signal that it is safely replayable
- Request aggressive transport initiation

So either we should split it, convert it to a boolean, or find a better title like "Aggressive session establishment with safely replayable inital message"

gorryfair:
I'm not so sure we need to split this: the transport system can decide to automate the decision to use 0-RTT (or similar). The app can already control whether the system decides to set the boolean. I don't think we need to expose more control here.

mwelzl:
Actually, here, I disagree. This is a statement on the properties of the data that is handed over - the necessary application-specific information; clearly, whenever such things are stated, they have consequences in the system below. Here, the explaining text says that such data can be handed over before or during connection establishment.

This is similar to choosing unreliable transfer - where a name like "give me a protocol that doesn't require reliability and hence will operate more lightweight and somehow faster, because i don't need reliability" would just be unnecessary, even if we'd shorten the phrase. Enough to say "unreliable", and, IMHO, similarly, here, enough to say "safely replayable".

draft-ietf-taps-interface.md
```diff
@@ -1217,8 +1217,8 @@ Default:
 : Ignore
 
 This property specifies whether an application would like to supply a Message to
-the transport protocol before connection establishment that will then be
-reliably transferred to the other side before or during connection
+the transport protocol before Connection establishment that will then be
+reliably transferred to the other side before or during Connection
```
tfpauly, line 1221:
```suggestion
the transport protocol before connection establishment that will then be
reliably transferred to the other side before or during connection
```

tfpauly (approved):

draft-ietf-taps-interface.md
mwelzl: Okay, okay   :-)

mwelzl:
Decided not to do this and leave the name as it is

PR #1421, opened 2023-10-17
app SHOULD NOT assume that ignoring events is always safe

### LINKED ISSUE #1307: s/should not / SHOULD NOT
From the review by Erik Kline:
### S3

* "The application should not assume that ignoring events..."
 s/should not/SHOULD NOT/?

mwelzl:
Closes #1307

In a PR where some "should not" became "ought not", this was chosen to be "ought not", but I think Erik is right in #1307 about this one, SHOULD NOT seems better to me.


gorryfair (approved):
I agree this is good advice to the application designer

PR #1425, opened 2023-10-17
List of endpoint identifiers: types

### LINKED ISSUE #1373: List of endpoint identifiers should contain types for all

Section 6.1

```
       An Endpoint object can be configured with the following identifiers:
       * Hostname (string):
```

This starts with listing endpoint identifiers with types (eg string and 16-bit
integers) but then stops specifying their types further down. I guess WithService(),
WithIPAddress() and WithInterface take a (string)

mwelzl:
Closes #1373

### DIFF
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -738,13 +738,13 @@ RemoteSpecifier.WithHostName("example.com")
 RemoteSpecifier.WithPort(443)
 ~~~
 
-- Service (an identifier that maps to a port; either the name of a well-known service, or a DNS SRV service name to be resolved):
+- Service (an identifier string that maps to a port; either the name of a well-known service, or a DNS SRV service name to be resolved):
 
 ~~~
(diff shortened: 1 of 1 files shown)


draft-ietf-taps-interface.md
```diff
@@ -738,13 +738,13 @@ RemoteSpecifier.WithHostName("example.com")
...
 
-- Service (an identifier that maps to a port; either the name of a well-known service, or a DNS SRV service name to be resolved):
+- Service (an identifier string that maps to a port; either the name of a well-known service, or a DNS SRV service name to be resolved):
 
 ~~~
 RemoteSpecifier.WithService("https")
 ~~~
 
-- IP address (IPv4 or IPv6 address):
+- IP address (IPv4 or IPv6 address strings):
```
tfpauly, line 747:
```suggestion
- IP address (IPv4 or IPv6 address, as byte arrays or strings):
```

PR #1429, opened 2023-10-17
Clarify that Listen creates a Listener from a Preconnection

### LINKED ISSUE #1363: Preconnection vs Listener

The arch document talks about Preconnection, Connection and Listener
but this document in Section 3 places Listener as a Preconnection.

britram:
Fixes #1363


tfpauly (approved):

TateGunning:
It would make sense that every button could be listened too from connect..

PR #1430, opened 2023-10-17
Disambiguate security properties per rdd review

### LINKED ISSUE #1286: Clarify ambiguity in security properties (section 6.3 DISCUSS from Roman Danyliw)
For security properties in section 6.3:

- Define the highest usable level of abstraction for these parameters
- Reiterate the principle that the exact set of values for available enumerations and the exact data formats for each security parameter are platform-dependent.

### LINKED ISSUE #1358: Single credentials for multiple protocols?

Does this document cause encouraging of re-using of key pairs for
different protocols (eg TLS and IKEv2, or TLS 1.2 and QUIC). This
might have security implications (eg using RSA as RSA-PKCS1.5 as
well as RSA-PSS)

### LINKED ISSUE #1359: Identity and server-certificate forces same use over different protocols?

The current API kind of forces re-use of credentials for different protocols, eg:

```
SecurityParameters.Set(server-certificate, myCertificate)

       // Specifying a Remote Endpoint is optional when using Listen
       Preconnection := NewPreconnection(LocalSpecifier,
                                 TransportProperties,
                                 SecurityParameters)
```

Let's say that a server only supports TLS 1.2 with RSA-PKCS-1.5 and TLS
1.3 with RSASSA-PSS. We shouldn't use 1 RSA key from 1 certificate for
these two different RSA operations. Would this mean we could never tell
the TAPS API "TLS 1.2 or 1.3 is fine". Maybe that is okay and we should
pick an keypair that works the same in 1.2 and 1.3, but we don't know
what the future will hold.

If we could set multiple identities, TAPS could pick the appropriate one.
eg:
```
       SecurityParameters.Set(identity, myIdentity)
       SecurityParameters.Set(server-certificate, []myCertificate)
```

(and possibly multiple identities too?)

### LINKED ISSUE #1367: Identity

```
       SecurityParameters.Set(identity, myIdentity)
```

I assume myIdentity would be an array of Type and Value? Eg type=fqdn or type=RDN ?

### LINKED ISSUE #1370: What is "credentials"?

What is "credentials" (especially because previously we have myCertificate,
so it is unclear if this now contains myKey and if so, where in the previous
example myKey was conveyed to TAPS.

### LINKED ISSUE #1375: How to set remote peer's identity or certificate

SecurityParameters.Set() seems to allow to set our identiy and our certificate,
but not the remote peer's identity or certificate? For example, one might want
to pin a remote certificate and not just rely on a WithHostname() identifier
being present as subjectAltname on a certificate.

### LINKED ISSUE #1389: Parameters take algorithms as a parameter
I stumbled over this, and can't grok it. The last sentence makes it weirdly recursive:

```
- Supported algorithms: Used to restrict what parameters are used by underlying transport security protocols.
When not specified, these algorithms should use known and safe defaults for the system. Parameters include:
ciphersuites, supported groups, and signature algorithms. These parameters take a collection of supported algorithms as parameter.
```

I'd fix if I could but I'm lost  :-)

britram:
fixes #1286.
fixes #1359.
fixes #1375.
fixes #1389.
fixes #1358 as a side-effect.
fixes #1367 as a side-effect.
fixes #1370 as a side-effect.

### DIFF
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -1556,25 +1556,25 @@ are inherited during cloning (see {{groups}}).
 
 This document specifies an abstract API, which may appear to conflict with the need
 for security parameters to be unambiguous in order to reduce the chances
-of implementation errors. The Transport Services system SHOULD provide reasonable, 
+of implementation errors. The Transport Services system SHOULD provide reasonable,
 secure defaults for each enumerated security parameter, such that users of the system
-only need to specify parameters required to establish a secure connection 
+only need to specify parameters required to establish a secure connection
 ('server-certificate', 'client-certificate'). Specifying specific security parameters
-from enumerated values (e.g., specific ciphersuites) may constrain the Transport 
+from enumerated values (e.g., specific ciphersuites) may constrain the Transport
 Protocols that can be selected during connection establishment.
 
 Except as noted below, as with the rest of the TAPS system, exact names of parameters and/or
 values of enumerations (e.g., ciphersuites) used in the security parameters are system
-and implementation specific, and should be chosen to follow the principle of least 
+and implementation specific, and should be chosen to follow the principle of least
 surprise for users of the platform / language environment in question.
 
 ### Specifying Security Parameters on a Preconnection
 
-Common security parameters such as TLS ciphersuites are known to implementations. 
-Clients should use common safe defaults for these values whenever possible. 
+Common security parameters such as TLS ciphersuites are known to implementations.
+Clients should use common safe defaults for these values whenever possible.
 However, as discussed in {{?RFC8922}}, many transport security protocols require specific
 security parameters and constraints from the client at the time of configuration and
-actively during a handshake. 
+actively during a handshake.
 
 The configuration parameters need to be specified in the pre-connection phase
 and are created as follows:
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -1555,12 +1555,12 @@ of Connection establishment. Similar to Transport Properties, both parameters an
 are inherited during cloning (see {{groups}}).
 
 This document specifies an abstract API, which may appear to conflict with the need
-for security parameters to be unambiguous in order to reduce the chances
-of implementation errors. The Transport Services system SHOULD provide reasonable,
+for security parameters to be unambiguous, intended to reduce the chances
+of implementation errors. The Transport Services System SHOULD provide reasonable,
 secure defaults for each enumerated security parameter, such that users of the system
 only need to specify parameters required to establish a secure connection
 ('server-certificate', 'client-certificate'). Specifying specific security parameters
-from enumerated values (e.g., specific ciphersuites) may constrain the Transport
+from enumerated values (e.g., specific ciphersuites) might constrain the Transport
 Protocols that can be selected during connection establishment.
 
 Except as noted below, as with the rest of the Transport Services API, exact names of parameters and/or
@@ -1571,7 +1571,7 @@ surprise for users of the platform / language environment in question.
 ### Specifying Security Parameters on a Preconnection
 
 Common security parameters such as TLS ciphersuites are known to implementations.
-Clients should use common safe defaults for these values whenever possible.
+Applications should use common safe defaults for these values whenever possible.
 However, as discussed in {{?RFC8922}}, many transport security protocols require specific
 security parameters and constraints from the client at the time of configuration and
 actively during a handshake.
@@ -1640,7 +1640,7 @@ SecurityParameters.Set(cached-session-lifetime-seconds, 3600)
 out-of-band. Each pre-shared keying material is associated with some identity
 that typically identifies its use or has some protocol-specific meaning to the
 Remote Endpoint. Note that use of a pre-shared key will tend to select a single
-security protocol and therefore a single underlying protocol stack directly.
+security protocol, and therefore directly select a single underlying protocol stack.
 
 ~~~
 SecurityParameters.Set(pre-shared-key, key, myIdentity)
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -1596,6 +1596,12 @@ SecurityParameters.Set(server-certificate, myCertificateBundle[])
 SecurityParameters.Set(client-certificate, myCertificateBundle[])
 ~~~
 
+- One or more certificate chains to use as pinned server certificates, such that connection will fail if the presented server certificate does not match one of the  supplied pinned certificates. The form and format of the certificate chain is implementation-specific 
+
+~~~
+SecurityParameters.Set(pinned-server-certificate, yourCertificateChain[])
+~~~
+
 - Application-layer protocol negotiation (ALPN) values: used to indicate which
   application-layer protocols are negotiated by the security protocol layer.
   See {{!ALPN=RFC7301}} for definition of the ALPN field. Note that the Transport
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -1585,10 +1585,12 @@ SecurityParameters := NewSecurityParameters()
 
 Security configuration parameters and sample usage follow:
 
-- One or more certificate bundles identifying the Local Endpoint, whether as a server
-  certificate or a client certificate. Multiple bundles may be provided to allow selection
-  among different protocol stacks that may require differently formatted bundles. The form and format of the certificate bundle is implementation-specific. Note that if the private keys associated
-  with a bundle are not available, e.g., since they are stored in hardware
+- One or more certificate bundles identifying the Local Endpoint, whether
+  as a server certificate or a client certificate. Multiple bundles may
+  be provided to allow selection among different protocol stacks that may
+  require differently formatted bundles. The form and format of the
+  certificate bundle is implementation-specific. Note that if the private
+  keys associated with a bundle are not available, e.g., since they are stored in hardware
   security modules (HSMs), handshake callbacks must be used. See below for details.
 
 ~~~
@@ -1596,7 +1598,10 @@ SecurityParameters.Set(server-certificate, myCertificateBundle[])
 SecurityParameters.Set(client-certificate, myCertificateBundle[])
 ~~~
 
-- One or more certificate chains to use as pinned server certificates, such that connection will fail if the presented server certificate does not match one of the  supplied pinned certificates. The form and format of the certificate chain is implementation-specific 
+- One or more certificate chains to use as pinned server
+  certificates, such that connection will fail if the presented server
+  certificate does not match one of the  supplied pinned certificates.
+  The form and format of the certificate chain is implementation-specific.
 
 ~~~
 SecurityParameters.Set(pinned-server-certificate, yourCertificateChain[])
@@ -1612,9 +1617,9 @@ SecurityParameters.Set(pinned-server-certificate, yourCertificateChain[])
 SecurityParameters.Set(alpn, "h2")
 ~~~
 
-- Supported algorithms: Used to restrict what parameters are used by underlying transport security protocols.
-When not specified, these algorithms should use known and safe defaults for the system. Parameters include:
-ciphersuites, supported groups, and signature algorithms. These parameters take a collection of supported algorithms as parameter.
+- Supported algorithms, ciphersuites, groups: Used to restrict what parameters
+  are used by underlying transport security protocols. When not specified, '
+  these algorithms should use known and safe defaults for the system. 
 
 ~~~
 SecurityParameters.Set(supported-group, "secp256r1")
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -1563,7 +1563,7 @@ only need to specify parameters required to establish a secure connection
 from enumerated values (e.g., specific ciphersuites) may constrain the Transport
 Protocols that can be selected during connection establishment.
 
-Except as noted below, as with the rest of the TAPS system, exact names of parameters and/or
+Except as noted below, as with the rest of the Transport Services API, exact names of parameters and/or
 values of enumerations (e.g., ciphersuites) used in the security parameters are system
 and implementation specific, and should be chosen to follow the principle of least
 surprise for users of the platform / language environment in question.
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -1619,7 +1619,7 @@ SecurityParameters.Set(alpn, "h2")
 
 - Supported algorithms, ciphersuites, groups: Used to restrict what parameters
   are used by underlying transport security protocols. When not specified, '
-  these algorithms should use known and safe defaults for the system. 
+  these algorithms should use known and safe defaults for the system.
 
 ~~~
 SecurityParameters.Set(supported-group, "secp256r1")
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -1554,14 +1554,30 @@ Security parameters and callbacks are partitioned based on their place in the li
 of Connection establishment. Similar to Transport Properties, both parameters and callbacks
 are inherited during cloning (see {{groups}}).
 
-### Specifying Security Parameters on a Pre-Connection
+This document specifies an abstract API, which may appear to conflict with the need
+for security parameters to be unambiguous in order to reduce the chances
+of implementation errors. The Transport Services system SHOULD provide reasonable, 
+secure defaults for each enumerated security parameter, such that users of the system
+only need to specify parameters required to establish a secure connection 
+('server-certificate', 'client-certificate'). Specifying specific security parameters
+from enumerated values (e.g., specific ciphersuites) may constrain the Transport 
+Protocols that can be selected during connection establishment.
 
-Common security parameters such as TLS ciphersuites are known to implementations. Clients should
-use common safe defaults for these values whenever possible. However, as discussed in
-{{?RFC8922}}, many transport security protocols require specific
+Except as noted below, as with the rest of the TAPS system, exact names of parameters and/or
+values of enumerations (e.g., ciphersuites) used in the security parameters are system
+and implementation specific, and should be chosen to follow the principle of least 
+surprise for users of the platform / language environment in question.
+
+### Specifying Security Parameters on a Preconnection
+
+Common security parameters such as TLS ciphersuites are known to implementations. 
+Clients should use common safe defaults for these values whenever possible. 
+However, as discussed in {{?RFC8922}}, many transport security protocols require specific
 security parameters and constraints from the client at the time of configuration and
-actively during a handshake. These configuration parameters need to be specified in the
-pre-connection phase and are created as follows:
+actively during a handshake. 
+
+The configuration parameters need to be specified in the pre-connection phase
+and are created as follows:
 
 ~~~
 SecurityParameters := NewSecurityParameters()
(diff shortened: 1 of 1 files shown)


draft-ietf-taps-interface.md
```diff
@@ -1554,38 +1554,55 @@ Security parameters and callbacks are partitioned based on their place in the li
...
+Except as noted below, as with the rest of the TAPS system, exact names of parameters and/or
+values of enumerations (e.g., ciphersuites) used in the security parameters are system
+and implementation specific, and should be chosen to follow the principle of least 
+surprise for users of the platform / language environment in question.
+
+### Specifying Security Parameters on a Preconnection
+
+Common security parameters such as TLS ciphersuites are known to implementations. 
+Clients should use common safe defaults for these values whenever possible. 
+However, as discussed in {{?RFC8922}}, many transport security protocols require specific
```
tfpauly, line 1575:
```suggestion
This document specifies an abstract API, which may appear to conflict with the need
for security parameters to be unambiguous in order to reduce the chances
of implementation errors. The Transport Services system SHOULD provide reasonable,
secure defaults for each enumerated security parameter, such that users of the system
only need to specify parameters required to establish a secure connection
('server-certificate', 'client-certificate'). Specifying specific security parameters
from enumerated values (e.g., specific ciphersuites) may constrain the Transport
Protocols that can be selected during connection establishment.

Except as noted below, as with the rest of the TAPS system, exact names of parameters and/or
values of enumerations (e.g., ciphersuites) used in the security parameters are system
and implementation specific, and should be chosen to follow the principle of least
surprise for users of the platform / language environment in question.

### Specifying Security Parameters on a Preconnection

Common security parameters such as TLS ciphersuites are known to implementations.
Clients should use common safe defaults for these values whenever possible.
However, as discussed in {{?RFC8922}}, many transport security protocols require specific
```

draft-ietf-taps-interface.md
```diff
@@ -1554,38 +1554,55 @@ Security parameters and callbacks are partitioned based on their place in the li
...
 security parameters and constraints from the client at the time of configuration and
-actively during a handshake. These configuration parameters need to be specified in the
-pre-connection phase and are created as follows:
+actively during a handshake. 
+
+The configuration parameters need to be specified in the pre-connection phase
+and are created as follows:
```
tfpauly, line 1580:
```suggestion
actively during a handshake.

The configuration parameters need to be specified in the pre-connection phase
and are created as follows:
```

draft-ietf-taps-interface.md
```diff
@@ -1554,38 +1554,55 @@ Security parameters and callbacks are partitioned based on their place in the li
...
+This document specifies an abstract API, which may appear to conflict with the need
+for security parameters to be unambiguous in order to reduce the chances
+of implementation errors. The Transport Services system SHOULD provide reasonable,
+secure defaults for each enumerated security parameter, such that users of the system
+only need to specify parameters required to establish a secure connection
+('server-certificate', 'client-certificate'). Specifying specific security parameters
+from enumerated values (e.g., specific ciphersuites) may constrain the Transport
+Protocols that can be selected during connection establishment.
+
+Except as noted below, as with the rest of the TAPS system, exact names of parameters and/or
```
tfpauly, line 1566:
```suggestion
Except as noted below, as with the rest of the Transport Services API, exact names of parameters and/or
```

draft-ietf-taps-interface.md
```diff
@@ -1554,38 +1554,55 @@ Security parameters and callbacks are partitioned based on their place in the li
...
-the protocols being used, if not explicitly specified by the application.
+- A certificate bundle identifying the Local Endpoint, whether as a server
+  certificate or a client certificate. Note that if the private keys associated
+  with a bundle are not available, e.g., since they are stored in hardware
+  security modules (HSMs), handshake callbacks must be used. See below for details.
 
 ~~~
-SecurityParameters.Set(alpn, "h2")
+SecurityParameters.Set(server-certificate, myCertificateBundle)
+SecurityParameters.Set(client-certificate, myCertificateBundle)
```
britram, line 1595:
```suggestion
SecurityParameters.Set(server-certificate, myCertificateBundle[])
SecurityParameters.Set(client-certificate, myCertificateBundle[])
```

draft-ietf-taps-interface.md
```diff
@@ -1554,38 +1554,55 @@ Security parameters and callbacks are partitioned based on their place in the li
...
 
 Security configuration parameters and sample usage follow:
 
-- Application-layer protocol negotiation (ALPN) values: used to indicate which application-layer protocols
-are negotiated by the security protocol layer. See {{!ALPN=RFC7301}} for definition of the ALPN field. Note that the Transport Services System can provide ALPN values automatically, based on
-the protocols being used, if not explicitly specified by the application.
+- A certificate bundle identifying the Local Endpoint, whether as a server
+  certificate or a client certificate. Note that if the private keys associated
+  with a bundle are not available, e.g., since they are stored in hardware
+  security modules (HSMs), handshake callbacks must be used. See below for details.
```
britram, line 1591:
```suggestion
- One or more certificate bundles identifying the Local Endpoint, whether as a server
  certificate or a client certificate. Multiple bundles may be provided to allow selection
  among different protocol stacks that may require differently formatted bundles. The form and format of the certificate bundle is implementation-specific. Note that if the private keys associated
  with a bundle are not available, e.g., since they are stored in hardware
  security modules (HSMs), handshake callbacks must be used. See below for details.
```

draft-ietf-taps-interface.md
```diff
@@ -1554,38 +1554,56 @@ Security parameters and callbacks are partitioned based on their place in the li
...
+  certificate or a client certificate. Multiple bundles may be provided to allow selection
+  among different protocol stacks that may require differently formatted bundles. The form and format of the certificate bundle is implementation-specific. Note that if the private keys associated
+  with a bundle are not available, e.g., since they are stored in hardware
+  security modules (HSMs), handshake callbacks must be used. See below for details.
 
 ~~~
-SecurityParameters.Set(alpn, "h2")
+SecurityParameters.Set(server-certificate, myCertificateBundle[])
+SecurityParameters.Set(client-certificate, myCertificateBundle[])
 ~~~
```
britram, line 1597:
```suggestion
~~~

- One or more certificate chains to use as pinned server certificates, such that connection will fail if the presented server certificate does not match one of the  supplied pinned certificates. The form and format of the certificate chain is implementation-specific 

~~~
SecurityParameters.Set(pinned-server-certificate, yourCertificateChain[])
~~~
```

draft-ietf-taps-interface.md
```diff
@@ -1554,66 +1554,98 @@ Security parameters and callbacks are partitioned based on their place in the li
...
-SecurityParameters.Set(signature-algorithm, "ecdsa_secp256r1_sha256")
+SecurityParameters.Set(alpn, "h2")
 ~~~
 
-- Pre-Shared Key import: Used to install pre-shared keying material established
-out-of-band. Each pre-shared keying material is associated with some identity that typically identifies
-its use or has some protocol-specific meaning to the Remote Endpoint.
+- Supported algorithms, ciphersuites, groups: Used to restrict what parameters
+  are used by underlying transport security protocols. When not specified, '
+  these algorithms should use known and safe defaults for the system.
```
britram, line 1622:
```suggestion
  these algorithms should use known and safe defaults for the system.
```

draft-ietf-taps-interface.md
```diff
@@ -1554,66 +1554,98 @@ Security parameters and callbacks are partitioned based on their place in the li
 of Connection establishment. Similar to Transport Properties, both parameters and callbacks
 are inherited during cloning (see {{groups}}).
 
-### Specifying Security Parameters on a Pre-Connection
-
-Common security parameters such as TLS ciphersuites are known to implementations. Clients should
-use common safe defaults for these values whenever possible. However, as discussed in
-{{?RFC8922}}, many transport security protocols require specific
+This document specifies an abstract API, which may appear to conflict with the need
+for security parameters to be unambiguous in order to reduce the chances
```
gorryfair, line 1558: /in order to reduce/intended to reduce/

draft-ietf-taps-interface.md
```diff
@@ -1554,66 +1554,98 @@ Security parameters and callbacks are partitioned based on their place in the li
...
+of implementation errors. The Transport Services system SHOULD provide reasonable,
+secure defaults for each enumerated security parameter, such that users of the system
+only need to specify parameters required to establish a secure connection
+('server-certificate', 'client-certificate'). Specifying specific security parameters
+from enumerated values (e.g., specific ciphersuites) may constrain the Transport
+Protocols that can be selected during connection establishment.
```
gorryfair, line 1564: possible /may/MAY/

draft-ietf-taps-interface.md
```diff
@@ -1554,66 +1554,98 @@ Security parameters and callbacks are partitioned based on their place in the li
...
+Except as noted below, as with the rest of the Transport Services API, exact names of parameters and/or
+values of enumerations (e.g., ciphersuites) used in the security parameters are system
+and implementation specific, and should be chosen to follow the principle of least
+surprise for users of the platform / language environment in question.
+
+### Specifying Security Parameters on a Preconnection
+
+Common security parameters such as TLS ciphersuites are known to implementations.
+Clients should use common safe defaults for these values whenever possible.
+However, as discussed in {{?RFC8922}}, many transport security protocols require specific
```
gorryfair, line 1575: /Clients/ - is that applictaion, the APi, the transport services system?  I was not sure.

draft-ietf-taps-interface.md
```diff
@@ -1554,66 +1554,98 @@ Security parameters and callbacks are partitioned based on their place in the li
...
+
 SecurityParameters.Set(cached-session-lifetime-seconds, 3600)
 ~~~
 
+- Pre-Shared Key import: Used to install pre-shared keying material established
+out-of-band. Each pre-shared keying material is associated with some identity
+that typically identifies its use or has some protocol-specific meaning to the
+Remote Endpoint. Note that use of a pre-shared key will tend to select a single
+security protocol and therefore a single underlying protocol stack directly.
+
```
gorryfair, line 1644: /therefore a single underlying protocol stack directly/therefore directly select a specific underlying protocol stack/

draft-ietf-taps-interface.md
```diff
@@ -1554,66 +1554,98 @@ Security parameters and callbacks are partitioned based on their place in the li
...
 are inherited during cloning (see {{groups}}).
 
-### Specifying Security Parameters on a Pre-Connection
-
-Common security parameters such as TLS ciphersuites are known to implementations. Clients should
-use common safe defaults for these values whenever possible. However, as discussed in
-{{?RFC8922}}, many transport security protocols require specific
+This document specifies an abstract API, which may appear to conflict with the need
+for security parameters to be unambiguous in order to reduce the chances
+of implementation errors. The Transport Services system SHOULD provide reasonable,
```
gorryfair, line 1559: /Transport Services system/Transport Services System/

gorryfair (changes requested):

draft-ietf-taps-interface.md
britram:
```suggestion
for security parameters to be unambiguous, intended to reduce the chances
```

draft-ietf-taps-interface.md
britram: no, but may/might.

draft-ietf-taps-interface.md
britram:
```suggestion
of implementation errors. The Transport Services System SHOULD provide reasonable,
```

draft-ietf-taps-interface.md
```diff
@@ -1554,66 +1554,98 @@ Security parameters and callbacks are partitioned based on their place in the li
...
+secure defaults for each enumerated security parameter, such that users of the system
+only need to specify parameters required to establish a secure connection
+('server-certificate', 'client-certificate'). Specifying specific security parameters
+from enumerated values (e.g., specific ciphersuites) may constrain the Transport
```
britram, line 1563:
```suggestion
from enumerated values (e.g., specific ciphersuites) might constrain the Transport
```

draft-ietf-taps-interface.md
```diff
@@ -1554,66 +1554,98 @@ Security parameters and callbacks are partitioned based on their place in the li
...
+
+Except as noted below, as with the rest of the Transport Services API, exact names of parameters and/or
+values of enumerations (e.g., ciphersuites) used in the security parameters are system
+and implementation specific, and should be chosen to follow the principle of least
+surprise for users of the platform / language environment in question.
+
+### Specifying Security Parameters on a Preconnection
+
+Common security parameters such as TLS ciphersuites are known to implementations.
+Clients should use common safe defaults for these values whenever possible.
```
britram, line 1574:
```suggestion
Applications should use common safe defaults for these values whenever possible.
```

draft-ietf-taps-interface.md
britram: it's the app, editing.

draft-ietf-taps-interface.md
```diff
@@ -1554,66 +1554,98 @@ Security parameters and callbacks are partitioned based on their place in the li
...
 SecurityParameters.Set(max-cached-sessions, 16)
+
 SecurityParameters.Set(cached-session-lifetime-seconds, 3600)
 ~~~
 
+- Pre-Shared Key import: Used to install pre-shared keying material established
+out-of-band. Each pre-shared keying material is associated with some identity
+that typically identifies its use or has some protocol-specific meaning to the
+Remote Endpoint. Note that use of a pre-shared key will tend to select a single
+security protocol and therefore a single underlying protocol stack directly.
```
britram, line 1643:
```suggestion
security protocol, and therefore directly select a single underlying protocol stack.
```

draft-ietf-taps-interface.md
```diff
@@ -1554,66 +1554,98 @@ Security parameters and callbacks are partitioned based on their place in the li
...
-SecurityParameters.Set(alpn, "h2")
+SecurityParameters.Set(server-certificate, myCertificateBundle[])
+SecurityParameters.Set(client-certificate, myCertificateBundle[])
 ~~~
 
-- Local identity, certificates, and private keys: Used to perform private key operations and prove one's
-identity to the Remote Endpoint. (Note, if private keys are not available, e.g., since they are
-stored in hardware security modules (HSMs), handshake callbacks must be used. See below for details.)
+- One or more certificate chains to use as pinned server
+  certificates, such that connection will fail if the presented server
```
mwelzl, line 1602:
```suggestion
  certificates, such that connecting will fail if the presented server
```
Or "a connection", or even "a call to Inititate()" ...

draft-ietf-taps-interface.md
```diff
@@ -1554,66 +1554,98 @@ Security parameters and callbacks are partitioned based on their place in the li
...
-SecurityParameters.Set(supported-group, "secp256r1")
-SecurityParameters.Set(ciphersuite, "TLS_AES_128_GCM_SHA256")
-SecurityParameters.Set(signature-algorithm, "ecdsa_secp256r1_sha256")
+SecurityParameters.Set(alpn, "h2")
 ~~~
 
-- Pre-Shared Key import: Used to install pre-shared keying material established
-out-of-band. Each pre-shared keying material is associated with some identity that typically identifies
-its use or has some protocol-specific meaning to the Remote Endpoint.
+- Supported algorithms, ciphersuites, groups: Used to restrict what parameters
```
mwelzl, line 1620:
```suggestion
- Supported groups, ciphersuites, algorithms: Used to restrict what parameters
```
I think this list should be in the same order as the three example commands below. You could also go the other way round - change the order of commands. I don't think it matters, but it's a bit weird to have this in a different sequence here.

draft-ietf-taps-interface.md
```diff
@@ -1554,66 +1554,98 @@ Security parameters and callbacks are partitioned based on their place in the li
...
+  are used by underlying transport security protocols. When not specified, '
```
mwelzl, line 1621:
```suggestion
  are used by underlying transport security protocols. When not specified,
```

mwelzl (changes requested):
Really good IMO, many thanks!  I double checked all the issues (since this closes so many), and confirm that they are indeed all addressed by this PR.

My suggestions are just a few nits.

britram:
@gorryfair @mwelzl are we good to merge?

gorryfair (commented):
A little outside my expertise, but I read this as consistent with TAPS, so editorially good.

mwelzl (approved):
LGTM, many thanks!

draft-ietf-taps-interface.md
```diff
@@ -1554,66 +1554,98 @@ Security parameters and callbacks are partitioned based on their place in the li
...
-
-Common security parameters such as TLS ciphersuites are known to implementations. Clients should
-use common safe defaults for these values whenever possible. However, as discussed in
-{{?RFC8922}}, many transport security protocols require specific
+This document specifies an abstract API, which may appear to conflict with the need
+for security parameters to be unambiguous, intended to reduce the chances
+of implementation errors. The Transport Services System SHOULD provide reasonable,
+secure defaults for each enumerated security parameter, such that users of the system
+only need to specify parameters required to establish a secure connection
+('server-certificate', 'client-certificate'). Specifying specific security parameters
```
tfpauly, line 1562: Are these examples? Generally, you don't need to set _both_.

draft-ietf-taps-interface.md
```diff
@@ -1554,66 +1554,98 @@ Security parameters and callbacks are partitioned based on their place in the li
 of Connection establishment. Similar to Transport Properties, both parameters and callbacks
 are inherited during cloning (see {{groups}}).
 
-### Specifying Security Parameters on a Pre-Connection
-
-Common security parameters such as TLS ciphersuites are known to implementations. Clients should
-use common safe defaults for these values whenever possible. However, as discussed in
-{{?RFC8922}}, many transport security protocols require specific
+This document specifies an abstract API, which may appear to conflict with the need
```
tfpauly, line 1557:
```suggestion
This document specifies an abstract API, which could appear to conflict with the need
```

draft-ietf-taps-interface.md
```diff
@@ -1554,66 +1554,98 @@ Security parameters and callbacks are partitioned based on their place in the li
...
+for security parameters to be unambiguous, intended to reduce the chances
```
tfpauly, line 1558: This comma seems a bit odd here

draft-ietf-taps-interface.md
```diff
@@ -1554,66 +1554,98 @@ Security parameters and callbacks are partitioned based on their place in the li
...
+of implementation errors. The Transport Services System SHOULD provide reasonable,
+secure defaults for each enumerated security parameter, such that users of the system
+only need to specify parameters required to establish a secure connection
+('server-certificate', 'client-certificate'). Specifying specific security parameters
+from enumerated values (e.g., specific ciphersuites) might constrain the Transport
+Protocols that can be selected during connection establishment.
+
+Except as noted below, as with the rest of the Transport Services API, exact names of parameters and/or
+values of enumerations (e.g., ciphersuites) used in the security parameters are system
+and implementation specific, and should be chosen to follow the principle of least
```
tfpauly, line 1568: SHOULD or ought to?

draft-ietf-taps-interface.md
```diff
@@ -1554,66 +1554,98 @@ Security parameters and callbacks are partitioned based on their place in the li
...
+surprise for users of the platform / language environment in question.
+
+### Specifying Security Parameters on a Preconnection
+
+Common security parameters such as TLS ciphersuites are known to implementations.
+Applications should use common safe defaults for these values whenever possible.
```
tfpauly, line 1574: should/SHOULD?

draft-ietf-taps-interface.md
tfpauly: Or do we mean local cert?

draft-ietf-taps-interface.md
```diff
@@ -1554,66 +1554,98 @@ Security parameters and callbacks are partitioned based on their place in the li
...
-- Application-layer protocol negotiation (ALPN) values: used to indicate which application-layer protocols
-are negotiated by the security protocol layer. See {{!ALPN=RFC7301}} for definition of the ALPN field. Note that the Transport Services System can provide ALPN values automatically, based on
-the protocols being used, if not explicitly specified by the application.
+- One or more certificate bundles identifying the Local Endpoint, whether
+  as a server certificate or a client certificate. Multiple bundles may
+  be provided to allow selection among different protocol stacks that may
+  require differently formatted bundles. The form and format of the
+  certificate bundle is implementation-specific. Note that if the private
+  keys associated with a bundle are not available, e.g., since they are stored in hardware
+  security modules (HSMs), handshake callbacks must be used. See below for details.
```
tfpauly, line 1594: must/MUST/need to

draft-ietf-taps-interface.md
```diff
@@ -1554,66 +1554,98 @@ Security parameters and callbacks are partitioned based on their place in the li
...
 other policies.
 
 ~~~
 SecurityParameters.Set(max-cached-sessions, 16)
+
 SecurityParameters.Set(cached-session-lifetime-seconds, 3600)
 ~~~
 
+- Pre-Shared Key import: Used to install pre-shared keying material established
+out-of-band. Each pre-shared keying material is associated with some identity
```
tfpauly, line 1640:
```suggestion
out-of-band. Each instance of pre-shared keying material is associated with some identity
```

draft-ietf-taps-interface.md
britram:
```suggestion
and implementation specific, and ought to be chosen to follow the principle of least
```

draft-ietf-taps-interface.md
britram:
```suggestion
Applications SHOULD use common safe defaults for these values whenever possible.
```

draft-ietf-taps-interface.md
britram:
```suggestion
  security modules (HSMs), handshake callbacks are necessary. See below for details.
```

draft-ietf-taps-interface.md
```diff
@@ -1554,66 +1554,98 @@ Security parameters and callbacks are partitioned based on their place in the li
...
 are inherited during cloning (see {{groups}}).
 
-### Specifying Security Parameters on a Pre-Connection
-
-Common security parameters such as TLS ciphersuites are known to implementations. Clients should
-use common safe defaults for these values whenever possible. However, as discussed in
-{{?RFC8922}}, many transport security protocols require specific
+This document specifies an abstract API, which could appear to conflict with the need
+for security parameters to be unambiguous, intended to reduce the chances
+of implementation errors. The Transport Services System SHOULD provide reasonable,
```
britram, line 1559:
```suggestion
for security parameters to be unambiguous. The Transport Services System SHOULD provide reasonable,
```

draft-ietf-taps-interface.md
britram: yeah, this was a vestigial edit, fixed.

tfpauly (approved):

PR #1434, opened 2023-10-27
Where do interface names come from?

### LINKED ISSUE #1331: How does an application learn allowed/available interface names?

### Section 6.1, paragraph 13
```
    *  Interface name (string), e.g., to qualify link-local or multicast
       addresses (see Section 6.1.2 for details):

    LocalSpecifier.WithInterface("en0")
```
How does an application learn which interface name strings are
allowed/available on the system? The API doesn't seem to provide this
information. Also, based on the examples, there seem to be special
names such as "any" - where are those defined and how are conflicts
with names used on the system avoided?

mwelzl:
Closes #1331

Note: at the interim, we discussed that we could also point at Preconnection.Resolve() as a way to get interface names. I decided against this, however, due to the text that was added there by PR #947: "Note that the set of Local Endpoints returned by Resolve might or might not contain information about all possible local interfaces; it is valid only for a Rendezvous happening at the same time as the resolution. Care ought to be taken in using these values in any other context."

=> This seems to specifically warn against getting interface names for any other purpose with Resolve().

### DIFF
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -757,6 +757,10 @@ RemoteSpecifier.WithIPAddress(2001:db8:4920:e29d:a420:7461:7073:a)
 LocalSpecifier.WithInterface("en0")
 ~~~
 
+Systems usually offer means to obtain a list of available local interfaces; such means ought to
+be usable by an application to obtain the string to be supplied here. The specifics of these
+means depend on the implementation.
+
 Note that an IPv6 address specified with a scope zone ID (e.g. `fe80::2001:db8%en0`)
 is equivalent to `WithIPAddress` with an unscoped address and `WithInterface ` together.
 
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -757,9 +757,8 @@ RemoteSpecifier.WithIPAddress(2001:db8:4920:e29d:a420:7461:7073:a)
 LocalSpecifier.WithInterface("en0")
 ~~~
 
-Systems usually offer means to obtain a list of available local interfaces; such means ought to
-be usable by an application to obtain the string to be supplied here. The specifics of these
-means depend on the implementation.
+Systems usually offer means to obtain a list of available local interfaces; when available, these allow the application to obtain the interface name string to supply to this function.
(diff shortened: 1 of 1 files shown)


philsbln:
I understand that we want a generic solution, but why not add something like ```getifaddrs()```?
@tfpauly how does Network.framework to that?

tfpauly:
We have a separate object to enumerate interfaces, https://developer.apple.com/documentation/network/nwpathmonitor. I'm not sure we want to standardize that here, though. It's not really a transport-api centric thing.

mwelzl:
Hm... so can this be merged?

draft-ietf-taps-interface.md
```diff
@@ -757,6 +757,10 @@ RemoteSpecifier.WithIPAddress(2001:db8:4920:e29d:a420:7461:7073:a)
 LocalSpecifier.WithInterface("en0")
 ~~~
 
+Systems usually offer means to obtain a list of available local interfaces; such means ought to
```
britram, line 760: is this a should?

britram (approved):

draft-ietf-taps-interface.md
britram: nah, ought to is ok.

draft-ietf-taps-interface.md
```diff
@@ -757,6 +757,10 @@ RemoteSpecifier.WithIPAddress(2001:db8:4920:e29d:a420:7461:7073:a)
...
+be usable by an application to obtain the string to be supplied here. The specifics of these
+means depend on the implementation.
```
tfpauly, line 762:
```suggestion
Systems usually offer means to obtain a list of available local interfaces; when available, these allow the application to obtain the interface name string to supply to this function.
The details of the system means to enumerate local interfaces are platform-specific.
```

csperkins:
I'm not understanding why we don't use `Rendezvous()` here?

>"Note that the set of Local Endpoints returned by Resolve might or might not contain information about all possible local interfaces; it is valid only for a Rendezvous happening at the same time as the resolution. Care ought to be taken in using these values in any other context."

Sure, but so remove the last sentence. All the rest are true of every mechanism to find local interfaces.

PR #1436, opened 2023-10-31
Explain reason for separate multicast specifiers

### LINKED ISSUE #1343: Why didn't we overload WithIPAddress for multicast?

### Section 6.1.1, paragraph 4
```
    RemoteSpecifier.WithMulticastGroupIP(GroupAddress)
```
It's a design choice to not overload `WithIPAddress` here and instead
define `WithMulticastGroupIP`. I'd have overloaded, in order to avoid
the need for error handling if the wring type of IP address is passed
in.

philsbln:
Closes #1343

### DIFF
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -793,6 +793,8 @@ used to send messages to the multicast group, receive messages from
 the group, or, for an any-source multicast group, to both send and
 receive messages.
 
+Note that the Transport Services API has separate specifier calls for multicast groups to avoid introducing filter properties for single-source multicast and the confusion traditions overloading of the unicast specifiers in other systems causes.   
+
 Calling `Initiate` on that Preconnection creates a Connection that can be
 used to send Messages to the multicast group. The Connection object that is
 created will support `Send` but not `Receive`. Any Connections created this


draft-ietf-taps-interface.md
```diff
@@ -793,6 +793,8 @@ used to send messages to the multicast group, receive messages from
 the group, or, for an any-source multicast group, to both send and
 receive messages.
 
+Note that the Transport Services API has separate specifier calls for multicast groups to avoid introducing filter properties for single-source multicast and the confusion traditions overloading of the unicast specifiers in other systems causes.   
+
```
gorryfair, line 797: doesn't read quite correct to me:
/and seeks to  avoid the confusion caused in other systems by overloading the unicast specifiers./

gorryfair: text changed, thanks

gorryfair (approved):
Looks good

mwelzl (commented):
I shouldn't even review this, I understand far too little of it.

This said "qualify link-local or addresses" reads a bit odd to me... perhaps "IP addresses" ?

I'm genuinely confused ... it seems that the old text *did* overload the call, and now we don't, while the comment in the issue asks for the opposite?  Also, these separate calls... would it be useful to explicitly point at them? and what is this "confusion caused in other systems", I'm not sure I like our API spec criticizing "other systems"...

gorryfair:
The example I think is for link local addresses - removed the "or"

philsbln:
@mwelzl Yes, the comment ask to overload it as the BSD Socket API does and it looks simpler at first glance. 
However, unless you are a multicast black-belt, the question wether the multicast group goes in src or dst and how to specify the source for SSN is confusing and only makes sense if you think about how the API evolved. 
I remember @GrumpyOldTroll and me confusing each other and looking up how others' are doing it several times before coming up with the current API. Overloading with something similar can go bad easily…

mwelzl:
I was confused because it seemed to me that it *was* overloaded... anyway, this looks ok to me but perhaps we should avoid saying "confusing caused in other systems"...

draft-ietf-taps-interface.md
```diff
@@ -793,6 +793,8 @@ used to send messages to the multicast group, receive messages from
 the group, or, for an any-source multicast group, to both send and
 receive messages.
 
+Note that the Transport Services API has separate specifier calls for multicast groups to avoid introducing filter properties for single-source multicast and seeks to avoid the confusion caused in other systems by overloading the unicast specifiers.
```
mwelzl, line 796:
```suggestion
Note that the Transport Services API has separate specifier calls for multicast groups to avoid introducing filter properties for single-source multicast and seeks to avoid confusion that can be caused by overloading the unicast specifiers.
```

mwelzl (approved):
FWIW...

PR #1437, opened 2023-11-01
Three new read-only Connection properties

### LINKED ISSUE #1377: Define words related to Connection states

```
      The Connection state, which can be one of the following:
       Establishing, Established, Closing, or Closed.
```

I think the text in Section 8 should more clearly show the property names if the
goal is to have different implementations use the identical name. Eg in this case,
why not write: The Connection state ("state"). The next two entries are similarly
lacking a clear keyword to use:

```
       Whether the Connection can be used to send data.

       Whether the Connection can be used to receive data.
```

eg. why not define words for these to implementations will use the same words,
in this case perhaps ReadySend and ReadyRcv ?
Writing that now, perhaps "state" should be "State" then ?

mwelzl:
Closes #1377

### DIFF
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -2372,7 +2372,7 @@ Name:
 Type:
 : Enumeration
 
-This property informs about the current state of the Connection. Possible values are: `Establishing`, `Established`, `Closing` or `Closed`.
+This property informs about the current state of the Connection. Possible values are: `Establishing`, `Established`, `Closing` or `Closed`; for more details on Connection state, see {{state-ordering}}.
 
 
 #### Can Send Data {#conn-send-data}
@@ -3444,7 +3444,7 @@ Connection -> ConnectionError<reason?>
 ~~~
 
 
-# Connection State and Ordering of Operations and Events
+# Connection State and Ordering of Operations and Events {#state-ordering}
 
 This Transport Services API is designed to be independent of an implementation's
 concurrency model.  The details of how exactly actions are handled, and how
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -2055,15 +2055,17 @@ Depending on the status of the Connection, the queried Connection
 Properties will include different information:
 
 * The Connection state, which can be one of the following:
-  Establishing, Established, Closing, or Closed.
+  Establishing, Established, Closing, or Closed (see {{conn-state}}).
 
-* Whether the Connection can be used to send data. A Connection can not be used
+* Whether the Connection can be used to send data (see {{conn-send-data}}).
+  A Connection can not be used
   for sending if the Connection was created with the Selection Property
   `direction` set to `unidirectional receive` or if a Message
   marked as `Final` was sent over this Connection. See also {{msg-final}}.
 
-* Whether the Connection can be used to receive data. A Connection cannot be
-  used for reading if the Connection was created with the Selection Property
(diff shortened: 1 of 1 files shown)


philsbln (approved):
LGMT

draft-ietf-taps-interface.md
```diff
@@ -3408,7 +3444,7 @@ Connection -> ConnectionError<reason?>
 ~~~
 
 
-# Connection State and Ordering of Operations and Events
+# Connection State and Ordering of Operations and Events {#state-ordering}
```
abrunstrom, line 3447: For context I read the text below this modification. It is stated that 

> Each transition of Connection state is associated with one of more events:

But this does not seem correct, there is no event associated with the transition to Closing described in the text or Figure. Searching the doc for references to Closing, it does not seem to be used?

draft-ietf-taps-interface.md
```diff
@@ -2361,6 +2364,39 @@ estimates). Note that this property could degrade Connection performance.
 
 The following generic Connection Properties are read-only, i.e. they cannot be changed by an application.
 
+#### Connection state {#conn-state}
+
+Name:
+: state
```
abrunstrom, line 2370: Suggest expanding the name to connState.

draft-ietf-taps-interface.md
```diff
@@ -2079,7 +2081,8 @@ Properties will include different information:
   Protocol Selection Properties that the application specified on the
   Preconnection.
 
-* For Connections that are Established: information concerning the
+* For Connections that are Established: Transport Services system implementations
+  ought to provide information concerning the
```
abrunstrom, line 2085: Was this update meant to only improve readability or also change the meaning? It now sounds more optional, which may be fine as I am not sure the system will always have this information. Just wanted to check that this was an intentional modification.

abrunstrom (commented):
The updates look fine, I added only smaller comments, but reading the surrounding text the description of the Connection state does not look correct to me, see comment. If you agree we should either fix it here or close this and open a new issue.

draft-ietf-taps-interface.md
mwelzl: Just readability: I did this because before, it was a list of things that "are" provided - but the rest of the list is now concrete, assigned to a property, whereas this item isn't (it's more general). I think that, vague as this is, indeed we couldn't make it a "must"... so maybe now it's a bit more explicit that this is indeed optional, which I think is ok.

draft-ietf-taps-interface.md
mwelzl, line 2370:
```suggestion
: connState
```

draft-ietf-taps-interface.md
mwelzl: Thanks for reading this!  I must admit, I didn't check - there seems to be a discrepancy here, to be fixed!

mwelzl:
@abrunstrom regarding your concern: I think this is just a matter of:
s/Each transition of Connection state is associated with one of more events:/Some Connection state transitions are associated with events:

(side note, "one of more" seems like a nit, this should probably have been "one or more").

I thought about creating Events for *all* transitions, but I think the transitions to Establishing and Closing are unnecessary to couple with an Event, since this happens upon an Action being issued by the application. E.g., immediately upon calling "Close", the Connection should be in "Closing" state.

I'll make a separate PR for this.

tfpauly (approved):

PR #1443, opened 2023-11-01
What about zero Integer or Numeric values?

### LINKED ISSUE #1433: Zero values for some properties are invalid or unspecified
This is a super nit, but the definitions in Interface for Numeric and Integer specify that they take "positive or negative ... values". Although IEEE754 has the concept of positive and negative zero, this idea is not relevant to any of the TAPS specification. Maybe useful to "fix" these definitions to be inclusive of zero.

In working on #1428, I had to do a double-take to see whether we discuss zero anywhere. In doing so, I noticed that zero is implicitly valid for all of our Numeric and Integer options (as they all specify "non-negative"), but only some of them specify what a zero value means for that property.

In some cases, a value of zero is nonsensical. For example, the only logical (to me) behavior for a zero value for `connTimeout` and `keepAliveTimeout` would be for those fields to be `Disabled`. However, `Disabled` is implicitly an option for an implicitly enumerated type, and is distinct from the zero value. These two properties ought to say "Integer (positive)" or explicitly define a zero value as erroneous.

Further cases:
 * What does zero mean for [Bounds on Send or Receive Rate](https://datatracker.ietf.org/doc/html/draft-ietf-taps-interface#name-bounds-on-send-or-receive-r)?
 * Zero seems nonsensical for [Group Connection Limit](https://datatracker.ietf.org/doc/html/draft-ietf-taps-interface#name-group-connection-limit)
 * [Maximum Message Size Before Fragmentation or Segmentation](https://datatracker.ietf.org/doc/html/draft-ietf-taps-interface#name-maximum-message-size-before) should be above zero or have the zero value specified.
 * [Maximum Message Size on Send](https://datatracker.ietf.org/doc/html/draft-ietf-taps-interface#name-maximum-message-size-on-sen) should be above zero or have the zero value specified.
 * [Maximum Message Size on Receive](https://datatracker.ietf.org/doc/html/draft-ietf-taps-interface#name-maximum-message-size-on-rec)  should be above zero or have the zero value specified.
 * [Advertised User Timeout](https://datatracker.ietf.org/doc/html/draft-ietf-taps-interface#name-advertised-user-timeout) should be at least 1 per [RFC5482](https://datatracker.ietf.org/doc/html/rfc5482#section-3.1) ("Thus, the UTO option allows hosts to exchange user timeout values from 1 second to over 9 hours at a granularity of seconds, and from 1 minute to over 22 days at a granularity of minutes.") (Sadly, this RFC neither defines nor precludes zero.)
 * [Lifetime](https://datatracker.ietf.org/doc/html/draft-ietf-taps-interface#name-lifetime) seems to have a valid use-case for zero that should be explained.

I have attempted to be comprehensive, but I am not sure this is every case.

mwelzl:
Closes #1433

### DIFF
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -214,8 +214,8 @@ implementing platform.
 We also make use of the following basic types:
 
 - Boolean: Instances take the value `true` or `false`.
-- Integer: Instances take positive or negative integer values, or the value 0.
-- Numeric: Instances take positive or negative real number values, or the value 0.
+- Integer: Instances take integer values.
+- Numeric: Instances take real number values.
 - String: Instances are represented in UTF-8.
 - IP Address: An IPv4 or IPv6 address {{?RFC5952}}.
 - Enumeration: A family of types in which each instance takes one of a fixed,
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -214,8 +214,8 @@ implementing platform.
 We also make use of the following basic types:
 
 - Boolean: Instances take the value `true` or `false`.
-- Integer: Instances take positive or negative integer values.
-- Numeric: Instances take positive or negative real number values.
+- Integer: Instances take positive or negative integer values, or the value 0.
+- Numeric: Instances take positive or negative real number values, or the value 0.
 - String: Instances are represented in UTF-8.
 - IP Address: An IPv4 or IPv6 address {{?RFC5952}}.
 - Enumeration: A family of types in which each instance takes one of a fixed,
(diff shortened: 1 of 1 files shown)


britram (approved):

draft-ietf-taps-interface.md
```diff
@@ -214,8 +214,8 @@ implementing platform.
 We also make use of the following basic types:
 
 - Boolean: Instances take the value `true` or `false`.
-- Integer: Instances take positive or negative integer values.
-- Numeric: Instances take positive or negative real number values.
+- Integer: Instances take positive or negative integer values, or the value 0.
+- Numeric: Instances take positive or negative real number values, or the value 0.
```
britram, line 218:
```suggestion
- Integer: Instances take integer values.
- Numeric: Instances take real number values.
```

britram: we probably shouldn't create a normative defintion of "integer" and "real" in a transport area document...

tfpauly (approved):

Issue #1428, opened 2023-10-17
What are implications of keepAlive and keepaliveTimeout on stacked protocols with keepalives?


dhobsd:
When a protocol stack contains multiple protocols that support keepalive messages (e.g. h2 on TCP), where are keepalives enabled?

When one would like different keepalive policies for the different bits of the stack, how is this specified?

How does one specify TCP keepalive retry policies? (Typically TCP keepalives retry some number of times before giving up.)

gorryfair:
That's an interesting but sometime complex question. The stack has to work-out what to do, to me the API helps by providing some additional context, that can be used - not a specific behaviour. Do you have a proposal for more that we can say?

dhobsd:
To answer part of my own question:

> How does one specify TCP keepalive retry policies? (Typically TCP keepalives retry some number of times before giving up.)

This would rely on setting `connTimeout`.

Regarding what more to say, I think this depends on answers to some questions that I think are kind of ambiguous. I have recommendations on direction at the end of this comment.

The [text for `keepAlive`](https://datatracker.ietf.org/doc/html/draft-ietf-taps-interface#section-6.2.10) says:

> Note that when supported, the system will use the default period for generation of the keep alive-packets.

This seems [somewhat at odds with TCP](https://datatracker.ietf.org/doc/html/rfc9293#name-tcp-keep-alives):

> If keep-alives are included, the application MUST be able to turn them on or off for each TCP connection (MUST-24), and they MUST default to off (MUST-25).

If I'm a networking software author, per the above, how do I express a preference for protocols that have their own keepalive timeouts while also expressing a preference for enabling or disabling TCP keepalives?

If I'm a TAPS system implementer, and I get a `Preconnection` where `keepAlive` is set to `Prohibit`, but `keepAliveTimeout` is not specified, am I allowed to choose HTTP/2 (and not participate in `PING`)? Or anything on top of TCP? What if I'm implementing a transport on top of multiple TCP connections?

If I'm a TAPS system implementer, and I get a `Preconnection` where `keepAlive` is set to `Require`, am I allowed to choose HTTP/1.1 as long as the underlying TCP connection can have keepalive enabled?

I can allow folks to solve this problem with non-standard connection properties, but given the `MUST` in TCP for application authors, I feel like we probably ought to specify how and where these options take effect. I get that we have options for setting UTO, but this is distinct from TCP keepalives. [UTO says]:

[...]

PR #1444, opened 2023-11-01
keepAliveTimeout: introduce `System default`

mwelzl:
This partially addresses #1428
Please see the discussion there for the rationale behind this.

### DIFF
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -2160,10 +2160,10 @@ Name:
 : keepAliveTimeout
 
 Type:
-: Numeric (non-negative) or `Disabled`
+: Numeric (non-negative) or `Disabled` or `System default`
 
 Default:
-: `Disabled`
+: `System default`
 
 A Transport Services API can request a protocol that supports sending keep alive packets ({{keep-alive}}).
 If this property is Numeric, it specifies the maximum length of time an idle Connection (one for which no transport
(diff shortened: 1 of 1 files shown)


draft-ietf-taps-interface.md
```diff
@@ -2160,10 +2160,10 @@ Name:
 : keepAliveTimeout
 
 Type:
-: Numeric (non-negative) or `Disabled`
+: Numeric (non-negative) or `Disabled` or `System default`
```
philsbln, line 2163: I guess having a single-word value makes this more portable…

```suggestion
: Numeric (non-negative) or `Disabled` or `Default`
```

draft-ietf-taps-interface.md
```diff
@@ -2160,10 +2160,10 @@ Name:
...
 
 Default:
-: `Disabled`
+: `System default`
```
philsbln, line 2166:
```suggestion
: `Default`
```

draft-ietf-taps-interface.md
```diff
@@ -2173,7 +2173,9 @@ will only take effect when the underlying stack supports sending keep-alive pack
 Guidance on setting this value for connection-less transports is
 provided in {{?RFC8085}}.
 A value greater than the Connection timeout ({{conn-timeout}}) or the enumerated value `Disabled` will disable the sending of keep-alive packets. A Transport Services API
-could express `Disabled` in an environment-typical way, e.g., as a Union type or special value.
+could express `Disabled` and `System default` in an environment-typical way, e.g., as a Union type or special value. Choosing `System default` is a hint to the Transport Services system that it should
```
philsbln, line 2176:
```suggestion
could express `Disabled` and `Default` in an environment-typical way, e.g., as a Union type or special value. Choosing `Default` is a hint to the Transport Services system that it should
```

philsbln (approved):
Works for me except that I really prefer a single-word here

mwelzl:
How about SysDefault instead of Default? It's a single word, yet avoids the somewhat ugly:
"Default: default"

draft-ietf-taps-interface.md
```diff
@@ -2160,10 +2160,10 @@ Name:
 : keepAliveTimeout
 
 Type:
-: Numeric (non-negative) or `Disabled`
+: Numeric (non-negative) or `Disabled` or `SysDefault`
 
 Default:
-: `Disabled`
+: `SysDefault`
```
tfpauly, line 2166: The default for keepalives should be off.

tfpauly:
I think we should close as not to be fixed

tfpauly (changes requested):
Let's discuss and maybe close

philsbln (changes requested):
After discussion, I second @tfpauly

PR #1445, opened 2023-11-03
Section 11: not all state transitions have events

mwelzl:


abrunstrom (commented):
The text update looks fine. 

The only strange thing now is that nowhere in the documents is it mentioned how or when you enter the states Establishing or Closing. So you have to guess how you get to those states. In general I guess we do not describe the state transitions anywhere, other than describing the events that are happening when they occur in the section in question. Not sure if this is a problem, but it is a bit odd to have a state like Closing that is never mentioned anywhere except for when listing the states.

tfpauly (approved):

PR #1452, opened 2023-11-09
Explain stacked layers of keep alive protocols

### LINKED ISSUE #1428: What are implications of keepAlive and keepaliveTimeout on stacked protocols with keepalives?

tfpauly:
Closes #1428

### DIFF
--- a/draft-ietf-taps-impl.md
+++ b/draft-ietf-taps-impl.md
@@ -780,7 +780,7 @@ in storage common to all protocols, and notify the Protocol Stack as a whole whe
 For Protocol-specific Properties, such as the User Timeout that applies to TCP, the Transport Services implementation only needs to update the relevant protocol instance.
 
 Some Connection Properties might apply to multiple protocols within a Protocol Stack. Depending on the specific property,
-it might be appropriate to apply the property across these multiple protocols simultaneously, or else only apply to one.
+it might be appropriate to apply the property across multiple protocols simultaneously, or else only apply it to one protocol.
 In general, the Transport Services implementation should allow the protocol closest to the application to interpret
 Connection Properties, and potentially modify the set of Connection Properties passed down to the next protocol in the
 stack. For example, if the application has requested to use keepalives with the `keepAlive` property, and the Protocol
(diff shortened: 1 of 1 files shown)


gorryfair (approved):
I think this follows what we talked about. I also think it is a very valuable example of how the stack deals with a property that might inform multiple layers - saying that a specific property does not infer that the implementation makes the same request at each layer. Thanks for doing this.

draft-ietf-taps-impl.md
```diff
@@ -776,9 +776,18 @@ Connection Properties. A Connection can also generate error events in the form o
...
+in storage common to all protocols, and notify the Protocol Stack as a whole whenever the properties have been modified by the application. {{!RFC8303}} and {{!RFC8304}} offer guidance on how to do this for TCP, MPTCP, SCTP, UDP and UDP-Lite; see {{specific-protocol-considerations}} for a description of a back-tracking method to find the relevant protocol primitives using these documents.
 For Protocol-specific Properties, such as the User Timeout that applies to TCP, the Transport Services implementation only needs to update the relevant protocol instance.
 
+Some Connection Properties might apply to multiple protocols within a Protocol Stack. Depending on the specific property,
+it might be appropriate to apply the property across these multiple protocols simultaneously, or else only apply to one.
```
abrunstrom, line 783: "these multiple protocols" sounds like all of the protocols in the protocol stack to me, but it could be multiple, but still a subset of the protocols in the stack so perhaps:

... apply the property across multiple protocols simultaneously, or else only apply it to one protocol.

abrunstrom (approved):
thanks, looks good! made a small comment.

draft-ietf-taps-impl.md
tfpauly:
```suggestion
it might be appropriate to apply the property across multiple protocols simultaneously, or else only apply it to one protocol.
```

mwelzl (approved):
This is really great, in my opinion, and appears to say exactly what needs to be said. It makes me wonder in retrospect why I couldn't wrap my head this.

PR #1453, opened 2023-11-13
Clarify Rendezvous behaviour and who's in charge of ICE

### LINKED ISSUE #1300: Who's in charge of ICE ?  (the barkeeper, ha ha)
From the review by Martin Duke:

(S7.3) This section confused me as to whether ICE is done by the Transport
Service (paragraph 4), the app (paragraph 8), or both. (I think it's both.)
Maybe state this more explicitly?

csperkins:
Closes #1300

### DIFF
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -1848,23 +1848,22 @@ while also simultaneously trying to establish a Connection from the Local
 Endpoint candidates to the Remote Endpoint candidates.
 
 If there are multiple Local Endpoints or Remote Endpoints configured, then
-initiating a `Rendezvous` action will systematically probe the reachability
+initiating a `Rendezvous` action will cause the Transport Service
+Implementation to systematically probe the reachability
 of those endpoint candidates following an approach such as that used in
 Interactive Connectivity Establishment (ICE) {{?RFC8445}}.
 
-If the endpoints are suspected to be behind a NAT, `Rendezvous` can be
-initiated using Local Endpoints that support a method of discovering NAT
-bindings such as Session Traversal Utilities for NAT (STUN) {{?RFC8489}} or
-Traversal Using Relays around NAT (TURN) {{?RFC8656}}.  In this case, the
-Local Endpoint will resolve to a mixture of local and server reflexive
-addresses. The `Resolve` action on the Preconnection can be used to
-discover these bindings:
+If the endpoints are suspected to be behind a NAT, and the Local Endpoint
+supports a method of discovering NAT bindings, such as Session Traversal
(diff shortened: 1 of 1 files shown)


britram (approved):

tfpauly (approved):

draft-ietf-taps-interface.md
```diff
@@ -1848,23 +1848,22 @@ while also simultaneously trying to establish a Connection from the Local
 Endpoint candidates to the Remote Endpoint candidates.
 
 If there are multiple Local Endpoints or Remote Endpoints configured, then
-initiating a `Rendezvous` action will systematically probe the reachability
+initiating a `Rendezvous` action will cause the Transport Service
```
mwelzl, line 1851:
```suggestion
initiating a `Rendezvous` action will cause the Transport Services
```

draft-ietf-taps-interface.md
```diff
@@ -1848,23 +1848,22 @@ while also simultaneously trying to establish a Connection from the Local
...
+Implementation to systematically probe the reachability
```
mwelzl, line 1852:
```suggestion
implementation to systematically probe the reachability
```

mwelzl: Actually, we more commonly call it the "Transport Services system" instead of "implementation... but this is nitpicking, I believe.

draft-ietf-taps-interface.md
```diff
@@ -1887,9 +1886,10 @@ The set of Remote Endpoint candidates are then configured onto the Preconnection
 Preconnection.AddRemote([]RemoteEndpoint)
 ~~~
 
-The `Rendezvous` action can be initiated once both the Local Endpoint
-candidates and the Remote Endpoint candidates retrieved from the peer via
-the signalling channel have been added to the Preconnection.
+The `Rendezvous` action is initiated, and causes the Transport Services
+Implementation to begin connectivity checks, once the application has
```
mwelzl, line 1890:
```suggestion
implementation to begin connectivity checks, once the application has
```

As above: perhaps s/implementation/system ?  but maybe I'm just being too picky

mwelzl (changes requested):
The tiniest of nits. Else this is really good, a lot clearer about who does what indeed!

draft-ietf-taps-interface.md
csperkins: We split System into Implementation + API, no?

mwelzl: Ah, I see, Implementation is capitalized in -arch but not in -impl and mixed in api... I'll fix this with a separate PR then.

mwelzl (approved):

PR #1454, opened 2023-11-13
useTemporaryLocalAddress should be Prefer for Rendezvous

### LINKED ISSUE #1312: Temporary local address: why avoid for rendezvous?
From the review by Erik Kline:

### S6.2.13

* Why should the preference be Avoid for Rendezvous use cases?  It seems
 to me like many Rendezvous uses are not necessarily long-lived and so
 might be Preference None (Ignore), or even Prefer.

csperkins:
Closes #1312

### DIFF
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -1414,7 +1414,7 @@ Type:
 : Preference
 
 Default:
-: Avoid for Listeners and Rendezvous Connections. Prefer for other Connections.
+: Avoid for Listeners. Prefer for other Connections.
 
 This property allows the application to express a preference for the use of
 temporary local addresses, sometimes called "privacy" addresses {{?RFC8981}}.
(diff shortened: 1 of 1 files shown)


draft-ietf-taps-interface.md
```diff
@@ -1427,6 +1427,10 @@ temporary local addresses might involve trading off privacy for performance.
 For instance, temporary addresses (e.g., {{?RFC8981}}) can interfere with resumption mechanisms
 that some protocols rely on to reduce initial latency.
 
+The default is Prefer for Rendezvous since the expectation is that peer-to-peer
+connections are short-lived. Listeners, that are expected to be long-lived public
+services default to avoiding temporary local addresses.
```
abrunstrom, line 1432: If Listeners and Rendezvous are mentioned explicitly, should Initiate also be mentioned? 
I thinkit is the only one left.

britram: IMO the default for Initiate is probably a matter of system policy, so I'm fine with the text as is.

PR #1456, opened 2023-11-13
Use Resolve() to find local interfaces

### LINKED ISSUE #1331: How does an application learn allowed/available interface names?

csperkins:
Clarifies the limitations of using `Resolve()` on a `Preconnection` and makes it clear that it can be used to find local interfaces. Removes the need for a non-standard interface for this when we have a standard one.

Closes #1331

### DIFF
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -758,8 +758,8 @@ RemoteSpecifier.WithIPAddress(2001:db8:4920:e29d:a420:7461:7073:a)
 LocalSpecifier.WithInterface("en0")
 ~~~
 
-Systems usually offer means to obtain a list of available local interfaces; when available, these allow the application to obtain the interface name string to supply to this function.
-The details of the system means to enumerate local interfaces are platform-specific.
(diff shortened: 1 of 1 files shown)


britram (approved):

mwelzl:
See my answer in #1331 - there's text already there in the API draft that we should at least make sure we're not contradicting.   (the proposed wording in this PR doesn't seem to contradict it - but the text explains why we're not proposing to use Resolve() for anything but Rendezvous)

csperkins:
I thought I fixed it - is there more I missed?

mwelzl:
Ahhh... I'm an idiot, the text I'm referring to *is* part of this PR!  Sorry, answered in a hurry yesterday. All is well !

PR #1457, opened 2023-11-14
s/Transport Services implementation/Transport Services Implementation

mwelzl:
This wasn't uniform: almost everywhere, the -arch draft has "Transport Services Implementation" (and since it's -arch, I assume that this is the right style) whereas the -impl draft has only "Transport Services implementation" and the usage is mixed in -api.


gorryfair (approved):
WFM

britram (approved):

tfpauly (approved):

PR #1458, opened 2023-12-13
s/SHOULD/MUST, addressing a DISCUSS item from Lars Eggert

mwelzl:
see https://mailarchive.ietf.org/arch/msg/taps/1or3_qkM3GeJM2rYBOj3IvxbO1E/


gorryfair (approved):
~Good change!

Issue #1460, opened 2024-01-02
Remaining comments from Paul Wouters


mwelzl:
New comments in the COMMENT part of the review from Paul Wouters:
_(Note, he mentioned that he'll move his Ballot to Yes once we've published an updated draft)_

```
I'm not sure that the model really expands from netflows to IP flows (or TOR)
in the future.

I'm not sure the Security Considerations warn enough about re-using the same
credentials with different protocols/auth mechanisms. (eg TLS and IKEv2, or TLS
1.2 and QUIC, or using RSA as RSA-PKCS1.5 as well as RSA-PSS)

Unqualified Hostname vs FQDN seems a security risk punted mostly to the
application.
```

mwelzl:
I think that the second and third paragraph call for expanding the security considerations section some more. About the first, I'm not sure what to do?

mwelzl:
Our email answer to the first item:

>We think that the architecture described in draft-ietf-taps-arch allows for such things (IP transport or the TOR suggestion). One way of scoping to TOR or IPsec or a set of proxies with the interface that we have is by scoping to a PvD that provides those capabilities; specifying a PvD is supported by our interface.
>
>Other than that, the interface document does not specify bindings for doing such things, and we see such extensions as a possibility for the future. Our consensus was also that creating an IANA registry can be done at such a later time (if needed, i.e. in case TAPS really does become a big success).

Is there anything else we should or could do here?

mwelzl:
The one comment left standing and unaddressed is:
```
I'm not sure that the model really expands from netflows to IP flows (or TOR)
in the future.
```
...and I don't think there's anything that's not potentially very disruptive to the entire API surface that we can do about this.

On this basis, I suggest to close this issue.

PR #1461, opened 2024-01-06
Security Considerations: a first stab at #1460

mwelzl:
#1460 raises three points:

1) "I'm not sure that the model really expands from netflows to IP flows (or TOR)
in the future."  => not addressed here.

2) re-using the same credentials with different protocols/auth mechanisms. => I think the fourth paragraph of the -arch Security Considerations speaks to this; instead of repeating it, I thought it would be better to point at this document (similar to how the implementation draft does it), as its security considerations should apply to all of the API draft anyway.

3) not using an FQDN punts a security risk mostly to the application: I more or less copied the wording about the no-FQDN security risk from section 6.1, it seems fair to repeat this here as a warning.

### DIFF
--- a/draft-ietf-taps-impl.md
+++ b/draft-ietf-taps-impl.md
@@ -1154,7 +1154,7 @@ This document has no actions for IANA.
 
 # Security Considerations
 
-{{I-D.ietf-taps-arch}} outlines general security consideration and requirements for any system that implements the Transport Services architecture. {{I-D.ietf-taps-interface}} provides further discussion on security and privacy implications of the Transport Services API. This document provides additional guidance on implementation specifics for the Transport Services API and as such the security considerations in both of these documents apply. The next two subsections discuss further considerations that are specific to mechanisms specified in this document.
+{{I-D.ietf-taps-arch}} provides general security consideration and requirements for any system that implements the Transport Services architecture. {{I-D.ietf-taps-interface}} provides further discussion on security and privacy implications of the Transport Services API. This document provides additional guidance on implementation specifics for the Transport Services API and as such the security considerations in both of these documents apply. The next two subsections discuss further considerations that are specific to mechanisms specified in this document.
 
 ## Considerations for Candidate Gathering
 
--- a/draft-ietf-taps-interface.md
+++ b/draft-ietf-taps-interface.md
@@ -3556,6 +3556,8 @@ in {{security-parameters}}. It does not recommend use (or disuse) of specific
 algorithms or protocols. Any API-compatible transport security protocol ought to work in a Transport Services system.
 Security considerations for these protocols are discussed in the respective specifications.
 
+{{I-D.ietf-taps-arch}} outlines general security considerations and requirements for any system that implements the Transport Services architecture. These include recommendations of relevance to the API, e.g. regarding the use of keying material.
+
 The described API is used to exchange information between an application and the Transport Services system. While
 it is not necessarily expected that both systems are implemented by the same authority, it is expected
 that the Transport Services Implementation is either provided as a library that is selected by the application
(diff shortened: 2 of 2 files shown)


draft-ietf-taps-interface.md
```diff
@@ -3556,6 +3556,8 @@ in {{security-parameters}}. It does not recommend use (or disuse) of specific
 algorithms or protocols. Any API-compatible transport security protocol ought to work in a Transport Services system.
 Security considerations for these protocols are discussed in the respective specifications.
 
+{{I-D.ietf-taps-arch}} outlines general security considerations and requirements for any system that implements the Transport Services architecture. These include recommendations of relevance to the API, e.g. regarding the use of keying material.
```
gorryfair, line 3559: /outlines/provides/ ? (i didn't much like "outlines `...` requirements"

draft-ietf-taps-interface.md
```diff
@@ -3598,6 +3600,7 @@ of potentially limited scope for alternate path discovery during Connection
 establishment, as well as potential additional information leakage about
 application interest when used with a resolution method (such as DNS without
 TLS) which does not protect query confidentiality.
+Names used with the Transport Services API SHOULD be fully-qualified domain names (FQDNs); not providing an FQDN will result in the Transport Services Implementation needing to resolve using DNS search domains, which might lead to inconsistent or unpredictable behavior.
```
gorryfair, line 3603: /needing to resolve/needing it to be resolved/

gorryfair (changes requested):
These look like improvements, see suggestions on wording.

draft-ietf-taps-interface.md
mwelzl: Side note, this particular wording, from @tfpauly, exists in two places.
I think "needing it to be resolved" makes the sentence structure a little complex... how about:
"needing to use DNS search domains for name resolution" ?

draft-ietf-taps-interface.md
mwelzl: ok by me (I don't see the problem, but I'm not a native speaker). This applies to the -impl draft as well, I'll propose a fix to both.

draft-ietf-taps-impl.md
```diff
@@ -1154,7 +1154,7 @@ This document has no actions for IANA.
 
 # Security Considerations
 
-{{I-D.ietf-taps-arch}} outlines general security consideration and requirements for any system that implements the Transport Services architecture. {{I-D.ietf-taps-interface}} provides further discussion on security and privacy implications of the Transport Services API. This document provides additional guidance on implementation specifics for the Transport Services API and as such the security considerations in both of these documents apply. The next two subsections discuss further considerations that are specific to mechanisms specified in this document.
+{{I-D.ietf-taps-arch}} provides general security consideration and requirements for any system that implements the Transport Services architecture. {{I-D.ietf-taps-interface}} provides further discussion on security and privacy implications of the Transport Services API. This document provides additional guidance on implementation specifics for the Transport Services API and as such the security considerations in both of these documents apply. The next two subsections discuss further considerations that are specific to mechanisms specified in this document.
```
mwelzl, line 1157: JFYI, for others: this was a wording suggestion from @gorryfair - and this text appears in both the interface and the implementation drafts, so this is why I suggest to change it here too.

draft-ietf-taps-interface.md
philsbln: I am thinking the whole time whether a "with FQDN" in addition to "with hostname" would be a better solution here. that makes it totally clear whether a programmer expects canonisation or not

britram (approved):

philsbln (approved):

tfpauly (approved):

PR #1465, opened 2024-01-29
Add property for security protocol/version

tfpauly:


zaheduzzaman (commented):
LGTM.

draft-ietf-taps-interface.md
```diff
@@ -1615,6 +1617,23 @@ SecurityParameters := NewDisabledSecurityParameters()
 SecurityParameters := NewOpportunisticSecurityParameters()
 ~~~
 
+### Allowed security protocols
+
+Name:
+: allowedSecurityProtocols (alternatively, minimumSecurityProtocolVersion and maximumSecurityProtocolVersion)
```
mirjak, line 1623: Does a max really make sense?

And do we want to keep this in there? because I guess a version you need to specify _per_ protocol...

Or maybe we have an interface like add/removeAllowedSecurityProtocol(name, [minimal version]) ?

mwelzl: I'm also surprised by the max., and I also think it should be per protocol.
I think it should be specified to be a "protocol name and version") - in line with the example given below. That should be good enough, why complicate this further.

mwelzl (changes requested):
A small change suggested (see also the comment by @mirjak )

draft-ietf-taps-interface.md
tfpauly:
```suggestion
: allowedSecurityProtocols
```

mwelzl (approved):
Looks great, thanks!

britram (approved):

